Issue #555, opened 2014-05-29
Inter-pixel capacitance


rmjarvis:
Some NIR detectors apparently have an effect called inter-pixel capacitance where charge in a pixel can induce voltage in neighboring pixels, which then spuriously gets interpreted as photons falling into those pixels.

I'm still not completely sure why this isn't just a convolution that can be included as part of the PSF.  A write-up (unpublished) by Jason Rhodes and others seems to imply that this can be modeled by convolving by a 3x3 kernel, which I would think means that it could be included as part of the PSF.  But I was told that it's not that simple.  

So this issue is to investigate this issue in more detail and find a convenient way to include this effect in GalSim.

cf. http://adsabs.harvard.edu/abs/2008SPIE.7021E...3S

rmandelb:
Closing #555.

PR #615, opened 2014-12-05
#555

arunkannawadi:
Request to merge #555 into the master branch.


galsim/detectors.py
```diff
@@ -148,5 +148,133 @@ def addReciprocityFailure(self, exp_time, alpha, base_flux):
...
+    left = pad_array[1:-1,:-2]
+    right = pad_array[1:-1,2:]
+    topleft = pad_array[2:,:-2]
+    bottomright = pad_array[:-2,2:]
+    topright = pad_array[2:,2:]
+    bottomleft = pad_array[:-2,:-2]
+
+    # Ensure that the choice of origin does not matter
+    x0 = IPC_kernel.bounds.xmin # 1 by default
+    y0 = IPC_kernel.bounds.ymin # 1 by default
```
arunkannawadi: The code looks clearner with Image elements being used instead of the array elements since the bottom left pixel on the kernel (as an Image) is multiplied with the bottom left part of the Image and so on. If the array elements are used, then it can potentially be confusing later on.

rmjarvis: I guess.  But at the expense of an unnecessary memory allocation (from the `IPC_kernel.copy()`).  If you want to do it this way, at least change it to use a `view()` call rather than `copy()`, since that won't reallocate space for the array.

arunkannawadi: Ah, I hadn't known of view() earlier. The code uses view() method instead of copy() now.

arunkannawadi:
Sorry for the delay, I was away on vacation. I went through this PR and all the comments seemed to have been addressed. If there are no other concerns, then I think we should merge this in a day or two.

rmjarvis:
Looks fine to me.  OK to merge as far as I'm concerned.

rmandelb:
Ditto.

arunkannawadi:
Merging #555 into the master branch.

arunkannawadi:
Closed #555.

Issue #618, opened 2014-12-09
ChromaticOpticalPSF ?


rmandelb:
Hi all- For #590, I have been thinking of how to represent optical PSFs in a chromatic context.  After a brief e-mail discussion with @jmeyers314 , I would like to propose some new functionality that I think will be useful for anyone who wants to represent chromatic optical PSFs (not particularly tied to WFIRST in any way).  Note for Josh: this has evolved slightly from what we discussed because of my realization that I was thinking of the aberrations issue in an overly-complex way; it's actually quite simple.

Here is the basic problem.  The chromatic effects come into optical PSFs in two ways:
- The obvious way is through the diffraction limit depending on wavelength/diameter.
- The less obvious way is that we require aberrations in units of wavelengths in order to determine the wavefront, but aberrations are typically defined in terms of some physical length, so for a different wavelength, this becomes a different number required by the OpticalPSF.

So the idea would be to have something like this (schematically; I'm sure it needs work on the details to actually be functional):
```
class ChromaticOpticalPSF(ChromaticObject):
    def __init__(self, diam, aberrations, **kwargs):
        self.diam = diam # in m
        self.aberrations = aberrations # in physical units like microns
        self.kwargs = kwargs
    def evaluateAtWavelength(self, wave):
        lam_over_diam = wave / self.diam # and do what's needed to get the right units
        aberrations = self.aberrations / wave # and do what's needed to get the right units
        ret = OpticalPSF(lam_over_diam=lam_over_diam, aberrations=aberrations, **self.kwargs)
        return ret
```

There are potential timing issues since OpticalPSF can be slow.  A few options that Josh suggested:

- Have users thin the bandpasses and SEDs to reflect how many points at which they want to evaluate.  This puts the onus on the user.

- Precompute the PSF at a few points in wavelength as part of the constructor, and then have ```evaluateAtWavelength```  interpolate between those PSF images.  This would require the constructor to have options for the user to control the pre-computation (number of grid points etc.).  For a given combination of telescope diameter + obscuration + struts + aberrations, this could be reused   for different SEDs/bandpasses, since a change would just require reweighting the precomputed quantities appropriately rather than precomputing anything else (as long as the wavelength range of interest hasn't changed).

This is something I was thinking of doing pretty much immediately, to enable some of what I want to do on #590, but comments / questions / suggestions are very welcome.  My tentative plan is to try to get something that works first using something like the above code, which will be slow, and then consider converting to a pre-computation scheme afterwards.

[...]

PR #620, opened 2014-12-26
#618 - interpolated chromatic objects

rmandelb:
Hi all -

Sorry for the obnoxious holiday-season pull request.  I realize this might not get attention for a while.

This pull request is for the new ```InterpolatedChromaticObject``` class and its subclass, ```ChromaticOpticalPSF```.  The former is a way of expediting calculations related to chromatic objects that can be described as GSObjects that have wavelength-dependent keywords or transformations applied to them.  The latter is a specific example for describing the wavelength-dependence of optical PSFs.

This class has a number of keywords that control its speed and accuracy.  It is possible to easily turn off the interpolation and make it do the exact brute force calculation.  For some simple examples that I tried, the startup cost of instantiating one of these ```ChromaticOpticalPSF```s is of order 1 second, and the time to actually draw is 1/10 the time to draw when doing the full calculation.  Often for a single object, the interpolated calculation is not worthwhile due to the overhead, but as soon as this is used for mass production, the interpolated calculation pays off very quickly.  Note that as long as the object was instantiated with sufficient wavelength coverage, it can be reused when drawing through different bandpasses or when convolved with objects that have different SEDs, so the overhead doesn't have to be incurred multiple times.

A few comments and questions:

* When writing the ```InterpolatedChromaticObject``` class, I clearly had ```ChromaticOpticalPSF``` in mind as the main application.  I am open to comments on parts of ```InterpolatedChromaticObject``` that should be generalized for other applications.

* @jmeyers314 , I was wondering if I should use this class to simplify the atmospheric PSF class that you wrote?

* I gave some thought to the issue of flux normalizations since it seems that sometimes ```InterpolatedChromaticObjects``` will be used for PSFs (e.g., ```ChromaticOpticalPSF```) convolving a chromatic galaxy, whereas other times it'll be used as an object with some SED leading to a flux calculation in a given band (if someone makes a galaxy that has wavelength-dependent properties using this class, or if they want to draw a star image corresponding to a ```ChromaticOpticalPSF``` with some SED and SNR).  Right now, my solution is that the ```InterpolatedChromaticObject``` carries around an ```SED``` attribute, which by default is flat, i.e., ```lambda w : 1.0```.  This is what we want if it's to be used as a PSF convolving a chromatic galaxy.  However, it's possible to assign an SED at the time of initialization or make a new instance with a new SED using the ```withSED``` method; in this case, the object has a well-defined flux through each band, as defined by the SED.   For example, if we want to draw an image of what the effective PSF is for a galaxy with a given SED, then we have to assign that SED to the ```ChromaticOpticalPSF``` before drawing.

Note that changing the SED does not incur the overhead of making a completely new object, since the new object just copies over all the data (stored images, etc.).  The documentation of this behavior might not quite be sufficient.  I first wanted to get some feedback on whether this way of dealing with SEDs makes sense, and if people agree then I will add some more information or perhaps examples to the docstring.


jmeyers314:
Hi @rmandelb, I just started going through this.  Looks well done.  A couple quick comments:

1) Is there merit to keeping the interpolation in a subclass instead of integrating it directly into ChromaticObject (to be instigated by the presence of the `waves` keyword)?  I don't have strong feelings about this, and maybe I'm not thinking of some complicating factor, but it seems that this could just have been done without introducing a new subclass.

2) 
> @jmeyers314 , I was wondering if I should use this class to simplify the atmospheric PSF class that you wrote?

Sure!  That sounds like a reasonable thing to do.

3)
>For example, if we want to draw an image of what the effective PSF is for a galaxy with a given SED, then we have to assign that SED to the ChromaticOpticalPSF before drawing.

I'm not sure I understand the purpose of the withSED() method.  To generate effective PSFs, I usually just do:

```
star = galsim.Gaussian(sigma=1.e-8) * sed
conv = galsim.Convolve(psf, star)
conv.drawImage(...)
```

I think this is conceptually cleaner (the sed belongs to the star, not the psf).

galsim/chromatic.py
```diff
@@ -1359,3 +1376,286 @@ def withScaledFlux(self, flux_ratio):
...
+                            results in higher accuracy. [default: 1]
+    """
+    def __init__(self, waves, sed=None, oversample_fac=1.):
+        self.waves = waves
+        self.separable = False
+        if sed is None:
+            self.SED = lambda w: 1.0
+        else:
+            self.SED = sed
+        self.wave_list = np.array([], dtype=float)
```
jmeyers314: I think this should be:
```
if hasattr(self.SED, 'wave_list'):
  self.wave_list = self.SED.wave_list
else:
  self.wave_list = np.array([], dtype=float)
```

galsim/chromatic.py
```diff
@@ -1359,3 +1376,286 @@ def withScaledFlux(self, flux_ratio):
...
+        Make a new InterpolatedChromaticObject with the requested SED.
+
+        @param sed    Either a galsim.SED object to assign to this object as its SED, or a string
+                      'flat' which means a flat SED (equal weight to all wavelengths).  'flat' is
+                      what should be used for PSFs, so they do not modify the flux of the objects
+                      with which they are being convolved. 
+                      [default: 'flat']
+        @return a new InterpolatedChromaticObject
+        """
+        new = self.copy()
```
jmeyers314: similarly, should update self.wave_list in withSED.

galsim/chromatic.py
```diff
@@ -1359,3 +1376,286 @@ def withScaledFlux(self, flux_ratio):
...
+        if wave < min(self.waves) or wave > max(self.waves):
+            raise RuntimeError("Requested wavelength %.1f is outside the allowed range:"
+                               " %.1f to %.1f nm"%(wave, min(self.waves), max(self.waves)))
+
+        # Figure out where the supplied wavelength is compared to the list of wavelengths on which
+        # images were originally tabulated.
+        lower_idx = np.searchsorted(self.waves, wave)-1
+        # There can be edge issues, so watch out for that:
+        if lower_idx < 0: lower_idx = 0
+        if lower_idx == len(self.waves)-1: lower_idx = len(self.waves)-1
```
jmeyers314: Did you mean `if lower_idx > len(self.waves)-1`?

galsim/chromatic.py
```diff
@@ -1359,3 +1376,286 @@ def withScaledFlux(self, flux_ratio):
...
+        frac = (wave-self.waves[lower_idx]) / (self.waves[lower_idx+1]-self.waves[lower_idx])
+
+        # Actually do the interpolation for the image, stepK, and maxK.
+        im = _linearInterp(self.ims, frac, lower_idx)
+        stepk = _linearInterp(self.stepK_vals, frac, lower_idx)
+        maxk = _linearInterp(self.maxK_vals, frac, lower_idx)
+        return self.SED(wave)*im, stepk, maxk
+
+    def drawImage(self, bandpass, force_eval=False, image=None, **kwargs):
```
jmeyers314: I'm not sure if it would make a big timing difference, but I think this could be rewritten to use fewer sums.  For instance, if len(self.waves)=10, then the final profile will be a linear combination of no more than 10 terms, regardless of the value of len(self.wave_list) or len(bandpass.wave_list).  This is a consequence of the linear interpolation; a more generic wavelength-dependence doesn't have this property (although other polynomial interpolation rules like Simpson's rule probably do).  

It might be worth looping through the elements of the combined wave list here to build up the coefficient of each term and then forming the linear combination instead of using DirectSampleIntegrator which doesn't know about the linear interpolation.

Likewise, I don't think you need to check the values of stepk and maxk at each wavelength in the combined wave list; it's probably sufficient to check at the elements of self.waves.

tests/test_chromatic.py
```diff
@@ -1201,6 +1225,269 @@ def test_centroid():
...
+        obj.drawImage(image=im, scale=scale, add_to_image=True)
+    # Flux normalization should not be the same, but the rest should be.  So just renormalize.
+    im *= im_interp.array.sum()/im.array.sum()
+    np.testing.assert_array_almost_equal(
+        im_interp.array, im.array, decimal=4,
+        err_msg='InterpolatedChromaticObject results differ for interpolated vs. brute force')
+
+    # Also check that with a tiny / narrow band, we get exactly what we expect.
+    # Since the band is extremely narrow we can do this test to quite high accuracy.
+    wave_list = [550., 555., 560.]
```
jmeyers314: Since exact_psf doesn't have a wave_list, this object only evaluates at precisely these 3 wavelengths, getting multiplied by zero for 2 of them.  I.e., using `wave_list = [1., 555., 560.]` gives exactly the same results.  I'm not sure if that's what you intended...

tests/test_chromatic.py
```diff
@@ -1201,6 +1225,269 @@ def test_centroid():
...
+    bulge = galsim.DeVaucouleurs(half_light_radius = 1.5*scale)
+    bulge = bulge_SED*bulge
+    tot_gal = gal + bulge
+    obj_exact = galsim.Convolve(exact_psf, tot_gal)
+    obj_interp = galsim.Convolve(interp_psf, tot_gal)
+    im_exact = obj_exact.drawImage(bandpass, scale=scale)
+    im_interp = im_exact.copy()
+    im_interp = obj_interp.drawImage(bandpass, image=im_interp, scale=scale)
+    expected_flux = disk_SED.calculateFlux(bandpass) + bulge_SED.calculateFlux(bandpass)
+    expected_flux *= (1.0 - galsim.GSParams().alias_threshold)
```
jmeyers314: Using alias_threshold here seems like a weird thing to do to me since image corners probe larger radii than that which encloses (1-alias_threshold).  Tests pass on my system with this line commented out anyway.

galsim/chromatic.py
```diff
@@ -1359,3 +1376,286 @@ def withScaledFlux(self, flux_ratio):
...
+        self._last_n_eval = integrator.last_n_eval
+
+        # Apply integral to the initial image appropriately.  This will naturally work properly and
+        # take into account the supplied value of `add_to_image`, which will be included in kwargs.
+        image = int_im.drawImage(image=image, **kwargs)
+        return image
+
+    def withScaledFlux(self, flux_ratio):
+        ret = self.copy()
+        ret.SED *= flux_ratio
```
jmeyers314: I think this only works if `self.SED` is actually an SED, and not `lambda w: 1.0`.

tests/test_interpolatedimage.py
```diff
@@ -880,6 +880,36 @@ def test_conserve_dc():
...
+
+    step_k_val = int_im.stepK()
+    max_k_val = int_im.maxK()
+
+    mult_val = 0.9
+    new_int_im = galsim.InterpolatedImage(im, _force_stepk=mult_val*step_k_val,
+                                          _force_maxk=mult_val*max_k_val)
+    numpy.testing.assert_almost_equal(
+        new_int_im.stepK(), mult_val*step_k_val, decimal=7,
+        err_msg='InterpolatedImage did not adopt forced minimum value for stepK')
```
jmeyers314: Is this really a *minimum* value for stepK and not just *the* value?

and similarly maxK?

rmandelb: You are absolutely correct.  Just pushed a fix.

tests/test_chromatic.py
rmandelb: Fair enough.  There is some missing flux but that is an over-correction for it.  I just removed this, as you suggested.

galsim/chromatic.py
rmandelb: Oh my.  Thanks for the careful reading. :)  You are right, of course.  Fix pushed.

galsim/chromatic.py
rmandelb: I think any polynomial interpolation rule will have this property that there are at most ```len(self.waves)``` distinct terms that are being summed, with well-defined coefficients.

To be honest, the reason I did it this way was that the code seemed straightforward, and I thought there was no way that this was the limiting step.  Instantiation of InterpolatedImages seemed like it should be more expensive.  To test out this gut feeling, I just wrote a script that draws the ChromaticOpticalPSF from the unit tests, but with differently thinned versions of the LSST r bandpass that have a different ```len(bandpass.wave_list)```.  For ```len(bandpass.wave_list)``` = 23, 54, 73, the time to render a PSF image (not convolved with galaxy) was typically 0.5, 0.75, and 1.7 seconds.  That is a kind of weird scaling that I'm not quite sure how to interpret - I'm guessing its the sum of a nearly constant overhead (for the instantiation of the InterpolatedImage that always happens before drawing) plus a term that scales with the length of the combined wave list.  But anyway, it seems you're right - this is more important than I was expecting and I think I should refactor the code to account for it.

The way this would work (I think) is that I would have a subroutine that does the steps in ```_imageAtWavelength``` that find ```lower_idx``` and ```frac``` to interpolate between two stored images.  Using that subroutine here as well, I could figure out (for each wavelength of the combined wave list) which of the stored images contribute to the integral, and with what relative weights.  Then I could just take the total weighted sum, without calling any kind of ```DirectImageIntegrator``` at all.  (I suppose I could leave that new class just in case someone wants to implement a version of the integration that doesn't just use linear/polynomial interpolation between the stored images. But I don't need it if I do what you suggested, I think.)

jmeyers314:
Hi @rmandelb, sorry I'm slow to respond...

>I can't think of any other consequences, so if it were just the renaming of subroutines, I would definitely say we should go ahead.

I guess the code in InterpolatedChromaticObject.__init__ would need to go somewhere too.  It's not totally obvious to me where though.  Now that I look at it, the ChromaticObject constructor expects rather different arguments than the InterpolatedChromaticObject constructor.

I'm curious what @rmjarvis has to say about this too.

>Does that sound correct to you or is there some subtlety that I'm missing?

I think we'll just have to try it and find out.  One potential subtlety is that ChromaticAtmosphere is a currently a function that returns a ChromaticObject and not a subclass of ChromaticObject.  I'm happy to look into this if you like.

>then your "star" has some galaxy-ish SED. 

I guess I'm less bothered by stars with galactic SEDs than I am by chromatic PSFs with SEDs :).  I think it's useful to maintain the notion that the monochromatic PSFs returned by evaluateAtWavelength() are normalized.  The same is not true for monochromatic slices through galaxies, of course.

rmjarvis:
> I'm curious what @rmjarvis has to say about this too.

This is more a question of coding style than anything else.  Putting the functionality into ChromaticObject wouldn't be wrong, but in my opinion, it seems like the more object oriented way to do this is how Rachel did it.

The basic question is probably whether the interpolation paradigm is potentially appropriate for all kinds of Chromatic objects?  Or are there just some Chromatic objects that should be treated as an InterpolatedChromaticObject while others shouldn't?

If it's the former, then the interpolation option should just be an option for all Chromatic classes, and the functionality should be moved to the main ChromaticObject base class.

If it's the latter, then I think it is better to leave the InterpolatedChromaticObject class as a level in the class hierarchy that specific child classes can derive from to get that functionality.

My guess is that it is the latter, since I don't think the Chromatic class (simple object * sed) would benefit from any of the interpolation stuff.  But maybe not.  Maybe even that one should have the interpolation option.

galsim/chromatic.py
rmandelb: I just did this.  The good part:

- The code seems cleaner to me.
- If I run the test script described above, then for the 3 cases I mentioned, the time to render the PSF image with 23, 54, and 73 elements in ```bandpass.wave_list``` is typically 0.5, 0.5, and 0.9 seconds. (Instead of 0.5, 0.75, and 1.7 seconds.)  So the scaling with the number of elements is definitely shallower.  

The weird part:
- I determined that the step that is dominating the times is the step where the InterpolatedImage is drawn (at the very end of drawImage).  For the cases of 54 and 73 elements, that step takes ~0.37 and ~0.7 seconds.  I'm a little confused as to why it's taking longer in one case than in the other - I have no idea why there should be any difference.  The stepk and maxk values are a bit different, but not very.  Any thoughts?

rmandelb:
To play devil's advocate to your devil's advocate :) 

1) "one benefit from letting everything use the interpolation option is that it would provide for a nice unit test." -> I have to be honest, I am a bit loathe to rearrange some pre-existing and smoothly working functionality just to get a nice unit test.  ChromaticObject already works well and has quite a few unit tests!

2) drawImage has different optional arguments for ChromaticObject and InterpolatedChromaticObject right now, and works completely differently (especially now that I just refactored it using Josh's suggestion to avoid needing any integrator from integ.py).  I feel like having a drawImage routine that just says "if BLAH: go to THIS function, else: go to THAT function" is kind of ugly.

I do admit that your ```setupInterpolation``` routine takes care of the issue Josh pointed out (where does the stuff that is in the current InterpolatedChromaticObject constructor go?).

galsim/chromatic.py
rmandelb: Fixed.

galsim/chromatic.py
rmandelb: Fixed.

tests/test_chromatic.py
rmandelb: Well, it is what I intended as a very basic sanity check.  But perhaps it's so silly as to not be useful.  I think this be a more useful/non-trivial test if I were convolving this with a chromatic object that does have a ```wave_list``` - right?

galsim/chromatic.py
rmandelb: I think you're right.

Now that I've gone back to the traditional usage, where if this is a galaxy it'll have a proper SED and it only has ```lambda w:1.0``` if it's a PSF, then does that matter?  (I don't think people would be using ```withScaledFlux``` if it's a PSF, would they?)

rmandelb:
@jmeyers314 - just wanted to say that I think I got rid of all comments about the idea of using an SED for a PSF (and likewise any associated code), but please let me know if you think I missed something.

rmjarvis:
> To play devil's advocate to your devil's advocate :)

That's usually called angel's advocate.

> I am a bit loathe to rearrange some pre-existing and smoothly working functionality just to get a nice unit test. ChromaticObject already works well and has quite a few unit tests!

I meant that as a unit test of the new interpolating functionality.  Not the old ChromaticObject behavior.  Also, I think most of the rearrangement would be in the new stuff.  The code path for things that don't ever call `setupInterpolation` could be made to be essentially unchanged.

> I feel like having a drawImage routine that just says "if BLAH: go to THIS function, else: go to THAT function" is kind of ugly.

Python lets you do something a bit prettier than that, since you can define methods on the fly.  You could have the `setupInterpolation` function swap in the interpolating version of drawImage so any calls to `obj.drawImage(...)` would use that code instead.  

But regardless, I'm more interested in the front end design than the prettiness of the implementation.  We should decide which API makes more sense first, and then implement that, rather than pick an API that has an implementation we happen to prefer.

> drawImage has different optional arguments for ChromaticObject and InterpolatedChromaticObject right now

Only 2 arguments: `force_eval` and `integrator`.  I think `force_eval` would be unnecessary in my proposed paradigm.  And we can just say that `integrator` will be ignored if the user has previously called `setupInterpolation`.

jmeyers314:
I just thought of another issue: applying a transformation (shear, dilate, etc) will wrap the InterpolatedChromaticObject class in a ChromaticObject class, which destroys the interpolation speedup.  

I think it would be valid to apply the transformations to the InterpolatedImage `int_im` in InterpolatedChromaticObject.drawImage() before redraw, though, at least if the transformations are all achromatic.  This might just be a quick change to how _applyMatrix works (and InterpolatedChromaticObject.drawImage).  I'll think about it.

rmandelb:
About SEDs:

> Actually, I think I'd recommend getting rid of the sed keyword in __init__ and the withSED method 
> entirely. The SED of a chromatic profile is really only completely unambiguous for a separable profile 
> (for which it doesn't make any sense to interpolate). For an inseparable profile, the SED by definition 
> varies with position. One could argue I suppose that the SED in this case could refer to the spatially 
> integrated flux at each wavelength, but I can't think of a situation where I would actually want to 
> create different chromatic profiles this way. Let me know if I'm just not being sufficiently creative 
> though.

So let's ignore ```ChromaticOpticalPSF``` for the moment and think about the case where someone has defined an ```InterpolatedChromaticObject``` that they want to correspond to a galaxy light profile.  (Thinking about concretely how/why they might do this: let's consider for example the use of HST data in multiple bands to represent non-parametric galaxy light profiles.  I could imagine that this might involve some ```InterpolatedImage``` representation of the light profile as a function of wavelength, and since those are slow to create and render directly, the ```InterpolatedChromaticObject``` functionality could come in handy.)

In this case, what I mean by ```SED``` is the relative weight to give the images they infer for different values of wavelength.  In other words if you ignore the PSF, then the light profile is essentially the integral of bandpass(w)*SED(w)*image(w).  I guess the ambiguity here is that the stored images for each wavelength might have some non-unity flux.  If they were always constrained to have unit flux, then image(w) for any arbitrary wavelength w would have unity flux given our linear interpolation in wavelength.  In that case, I think that the concept of an SED as a weight function makes sense, and i think it corresponds to exactly what you said - "the SED in this case could refer to the spatially integrated flux at each wavelength".  Do you agree with that statement and that it might be useful in the context I described, or am I off base here?  I do agree with you that it's not the same as an SED in the traditional sense, since the actual SED at each position will be different.

In the case that the object is convolved with an acrhomatic PSF, the light profile should be that PSF convolved with \int bandpass(w)*SED(w)*image(w).  If the PSF is chromatic then I guess it has to be the integral of bandpass(w)*SED(w)*[PSF(w) x image(w)] (where x = convolution) but here we lose the speedup due to interpolation since we have to instantiate an ```InterpolatedImage``` at each wavelength value.

rmandelb:
Message above edited due to formatting issues, please read on github.

rmandelb:
> I just thought of another issue: applying a transformation (shear, dilate, etc) will wrap the 
> InterpolatedChromaticObject class in a ChromaticObject class, which destroys the interpolation 
> speedup.

You are right, and I hadn't thought of this.  For ```InterpolatedChromaticObjects``` that are specifically ```ChromaticOpticalPSF```s, I think that won't be a big deal because I can't think of a context that requires transforming them.  (That is probably why I didn't think this through before now!)  But I can easily see wanting to apply shear or other transformations to other types of ```InterpolatedChromaticObjects```, like those that represent atmospheric PSFs.

So right now, as I understand it, calling ```shear()``` and other transformations of ```ChromaticObjects``` (with potentially wavelength-dependent arguments) just leads to a modification of the ```_A``` attribute.  Then when the profile is being rendered, ```drawImage``` integrates over the profiles from ```evaluateAtWavelength```, which applies those transformations represented by ```_A``` including any wavelength-dependence.  Have I understood this correctly?

So I think that if the transformations are wavelength-independent, then it would be pretty easy to have ```InterpolatedChromaticObject.drawImage()``` call ```evaluateAtWavelength``` without any transformations, and then do the transformations to the final ```InterpolatedImage``` that represents the object after integrating over wavelength.  However, if the transformations are wavelength-dependent, then we would have to use ```ChromaticObject.drawImage()``` and lose all the optimizations.

And I would have noticed this problem if I had had a unit test that transformed one of these things in some way.  I will have to add that once this functionality is included.

rmandelb:
Regarding the API issue and having a separate subclass:

Just to check that I understand what you both have in mind, let's talk specifically about ```ChromaticOpticalPSF```.  In this new model, it would be defined as a subclass of ```ChromaticObject``` with an ```__init__``` that is the same as the current one except without taking ```waves``` as an argument, and it would have an ```evaluateAtWavelength``` method (which contains the contents of its current ```simpleEvaluateAtWavelength```).  When the user instantiates it, the default will be exact evaluation, unless they do ```setupInterpolation```.  Doing so will carry out much of what's currently in ```InterpolatedChromaticObject.__init__```, leading to it having a ```waves``` attribute (plus all the other stuff that goes with it - the stored images etc.).  It would also swap in the version of drawImage that uses the interpolation.  And if someone convolves the ```ChromaticOpticalPSF``` with a galaxy, then it'll have to do the slower evaluation of the light profile for image rendering, but that's already the case.

ChromaticObject itself would have an ```_imageAtWavelength``` method (same as InterpolatedChromaticObject right now) and an ```_evaluateAtWavelength``` that would be used by the interpolated version of ```drawImage```, containing the current contents of the ```InterpolateChromaticObject.evaluateAtWavelength``` method.

I have to think about how that would interact with the shear and other transformation stuff that I just posted about.

rmjarvis:
Yes, I think that summary is basically what I had in mind.

Note that the user can wait until after applying any transformations or convolving by a galaxy before calling `setupInterpolation`, which would then let them use the interpolation machinery on the final object.

In fact, if you don't want to try to handle the transformations of objects that have the interpolation set up, you could just let them revert back to a simple ChromaticObject without interpolation and emit a warning that the previous setupInterpolation call doesn't apply anymore.  (Or maybe only do this if the transformation has a wavelength dependence, since the achromatic ones are handleable.)  Similarly if they convolve by another chromatic object.

rmandelb:
> Note that the user can wait until after applying any transformations or convolving by a galaxy before 
> calling setupInterpolation, which would then let them use the interpolation machinery on the final 
> object.

This requires that subclasses like ```ChromaticOpticalPSF``` or whatever other subclasses have their ```evaluateAtWavelength``` method set up to apply the transformations, right?  This way the stored images will have the transformation already applied. (In fact, this must be a requirement for ChromaticObject in general.)

That sounds neat in principle but I think it is not what we would want in many cases.  For example, imagine we have a ```ChromaticAtmosphere``` that we want to allow to have a mildly varying size and shape for each position on the sky, with the variation being implemented using ```shear``` and ```dilate``` with position-dependent arguments.  In the scenario you describe, the user would have to set up the interpolation machinery separately at each location, whereas if we use some trickery to allow the transformation (at least achromatic ones) on the final InterpolatedImage, the user could set up the interpolation machinery just once for the original object, and then use it for all the slightly different PSFs at each position.

I guess we could just advise the user of these tradeoffs, i.e., that
```
foo = galsim.ChromaticAtmosphere(...)
foo = foo.shear(...)
foo.setupInterpolation()
im = foo.drawImage(...)
```
is doing something different from
```
foo = galsim.ChromaticAtmosphere(...)
foo.setupInterpolation()
foo = foo.shear(...)
im = foo.drawImage(...)
```
and they should think about what makes sense for their use case.

In the first case, we want ```evaluateAtWavelength``` to carry out all the transformations before storing the images, whereas in the second case, we want ```drawImage``` to carry out the transformations at the end on the final InterpolatedImage right before drawing.   Though I guess that we could deal with that like this:

* In the first case, ```evaluateAtWavelength``` could do the transformation and set some flag like self.transformed=True (where it is set to False in the ChromaticObject constructor).  Then after interpolation is set up, it's going to use the custom interpolation version of drawImage which doesn't call ```evaluateAtWavelength```, but rather ```_imageAtWavelength```, where the stored images already have the transformation. The routine will check for self.transformed=True, and that tells it to not use ```self._A``` to transform the final InterpolatedImage.

* In the second case, the custom interpolation version of drawImage will find ```self.transformed=False```, and that is its signal that it should transform using the achromatic ```_A``` matrix after making the InterpolatedImage.  (If ```_A``` is chromatic then it would, as you say, revert to a standard ChromaticObject and emit a warning.)

Does this make sense?
(It's getting complicated...)

rmjarvis:
> In the scenario you describe, the user would have to set up the interpolation machinery separately at each location

I think you misunderstood me.  The user wouldn't have to do so, but rather they could call setupInterpolation after the shear call if that would be more efficient.  If you are doing a bunch of different achromatic shears, then you wouldn't want to do this.  Rather, you would want to setup the interpolation once and then apply the different shears to that.

But if you have a chromatic shear, then I think we still want to allow the user to setup an interpolation scheme after applying the shear to speed up the rendering.  That was really my point.

I think your proposed implementation may be a bit more complicated than necessary.  If there are transformations present when setupInterpolation is called, then the images it makes will include those transformations.  You can then reset `_A` and `_fluxFactor` back to an identity matrix and unity.  Any further (achromatic) transformations should then be applied to the interpolated images.

Applying a chromatic transformation can either be an error or it can remake the images by applying the transformation and redrawing.  I'm not sure which is better, since the latter is probably not the most accurate way to do things, especially if repeated several times, so it might be better to force the user to reorganize the code to do setupInterpolation after the transformations.

rmandelb:
> If there are transformations present when setupInterpolation is called, then the images it makes will 
> include those transformations. You can then reset _A and _fluxFactor back to an identity matrix and 
> unity. 

Good point.

> it might be better to force the user to reorganize the code to do setupInterpolation after the 
> transformations.

Yes, I think that if a chromatic transformation is attempted after setupInterpolation(), it should either raise an exception, or there should be a warning that the stored images are being regenerated.

Josh and Mike: thanks for discussing this change.  Since classes start next week, I will not have time to do this for at least a week, but most likely longer.   I will leave this PR open for a day in case there are any further comments on this API change (or other issues), then I will close it and reopen a clean one whenever I have a chance to make the changes.  The code will look quite different and I think it will be better to basically start with a clean slate.

jmeyers314:
>I think that the concept of an SED as a weight function makes sense, and i think it corresponds to exactly what you said - "the SED in this case could refer to the spatially integrated flux at each wavelength".  Do you agree with that statement and that it might be useful in the context I described, or am I off base here?

Maybe.  Another approach would be to absorb SED(w) into image(w).  In this case it might be useful to have getSED() as an output method.  I think the usefulness of having SED as an input (for this particular use case) then depends on whether or not you could envision changing the SED while keeping all of the monochromatic profiles the same.

>... If the PSF is chromatic then I guess it has to be the integral of bandpass(w) * SED(w) * (PSF(w) x image(w)), where x = convolution, but here we lose the speedup due to interpolation since we have to instantiate an InterpolatedImage at each wavelength value.

It's worth remembering that ChromaticConvolution.drawImage has some magic built in to avoid drawing at each wavelength when possible.  In particular, convolving by a sum of separable profiles (e.g. bulge+disk with different SEDs, or "stars" = tiny Gaussians with an SED) shouldn't destroy the speedup.

rmandelb:
> Another approach would be to absorb SED(w) into image(w). In this case it might be useful to have 
> getSED() as an output method.

I am reluctant to absorb it into the stored images - I think it is safest to enforce the stored images to all have unit flux.

> magic built in to avoid drawing at each wavelength

Yes, good point, I did see that ChromaticOpticalPSF convolved with a chromatic galaxy was faster using interpolation than without interpolation.

Issue #612, opened 2014-11-12
pixel conventions differ between FindAdaptiveMom vs. Image


rmandelb:
As documented in a few places, there are different conventions for what to call the lower left pixel for GalSim images vs. for the FindAdaptiveMom() and EstimateShear() routines.  It might be useful to make them self-consistent.

(Note: issue opened after a discussion elsewhere with @TallJimbo and @PaulPrice )

rmandelb:
Closing #612.

PR #623, opened 2014-12-31
#612 - small improvements and documentation updates in shear estimation routines

rmandelb:
This pull request is for some documentation updates and slight improvements in the EstimateShear() and FindAdaptiveMom() routines.  The changes are as follows:

(1) Documentation fixes.  There were some incorrect statements in the docstrings about centroid conventions differing between GalSim.Images vs. these routines, which are now clarified.  Also, the docstrings now say outright that the image WCS gets completely ignored.  Finally, some example code snippets in the documentation have been updated (many used old syntax that still works but is deprecated).

(2) Propagation of PSF size and shape estimates from EstimateShear() in the C++ layer up to the python.  This means that if someone uses EstimateShear() to get a PSF-corrected galaxy shape, they don't have to do an additional call to FindAdaptiveMom() to get the PSF moments (which wastes time and is an extra line of code).  This involves no change in syntax; it's just extra information in the data structure that gets returned.

(3) Users who want to input a guess for the object centroid, in case the object is not centered in the postage stamp, can now do so by passing in a galsim.PositionD() object as a single keyword argument, instead of passing in two separate keywords.  (If there is any concern about backwards compatibility, I could put the two older args back into the python layer while keeping the C++ layer simplified.)

(4) Some new unit tests related to the functionality in (2) and (3).

(5) Updates to the old unit tests in test_hsm.py to use newer syntax (e.g., shear() instead of applyShear(); drawImage instead of draw()).


PaulPrice:
Thanks for the heads-up.  This is issue [HSC-1136](https://hsc-jira.astro.princeton.edu/jira/browse/HSC-1136).

rmjarvis:
I looked through this, and it all looks fine to me.  If you want, I don't mind merging this now and getting it into 1.2.

rmandelb:
Do you think this is simple / insignificant enough that it's okay to merge with just one person reviewing it?  I was on the fence about that myself.

rmjarvis:
It's probably ok.  But does anyone else want to take a quick look?  @jmeyers314 @barnabytprowe

msimet:
I took a quick peek through and didn't see anything obviously wrong (but I'm not as familiar with this part of the code, so I'm not sure how much help that is).

@rmandelb, if you haven't looked at the diff for this PR, you might get a laugh at the GitHub attempt to figure out what changed when you went to the galaxy-position object instead of two position numbers.

rmandelb:
@rmjarvis , do you think we're okay to merge this?

rmjarvis:
Sure.  Sorry, I had meant to merge this before tagging 1.2, but I forgot about it.  Fine to merge now though.

rmandelb:
Thanks!

rmandelb:
Merging #612.

Issue #616, opened 2014-12-05
Implement Spergel (2010) galaxy profile model


mdschneider:
_Motivation:_ We are interested in speeding up the rendering of galaxy images for use in MCMC algorithms to infer galaxy model parameters given a postage stamp image. An alternative galaxy model parameterization may be helpful. 

The slowest part of the prediction of a galaxy image is the convolution with the PSF. Here's a plan for moving the PSF convolution into a pre-computation step, leaving MCMC steps to a simple linear operation for galaxy image predictions,
* [Spergel (2010)](http://iopscience.iop.org/0067-0049/191/1/58/) demonstrates a Sersic-like profile, but with the important feature that it can be Taylor expanded in ellipticity and relative size in such a way that all model parameters are in coefficients of the expansion. Then, the PSF convolution can be pre-computed with each term in the expansion and subsequent model updates are a simple linear operation.
* If the PSF is represented by a Zernike expansion of the exit pupil wavefront, then it’s possible to derive an expansion of the PSF in terms of the Zernike coefficients. Then we can treat PSF models in pre-computation analogous to the Spergel galaxy profile model.
* A dependency graph for these features is [here](https://www.lucidchart.com/publicSegments/view/54822282-b490-4c3c-b314-06a00a004a68/image.png).
* A first step is to implement the functional form from Spergel (2010) as the new `SBSpergel` class, mimicking `SBSersic`.

While GalSim predictions can be sped up by changing `GSParams` settings, we would like to avoid identifying the combination of accuracy and speed tradeoffs for each application if a re-parameterization can be made to work.

Credit for benchmarks and planning: @jmeyers314

jmeyers314:
I think the first part of this issue (drawing the Spergel profile without any Taylor expansion trickery) is almost ready for a PR.  The last remaining question I have is what to do for photon-shooting when nu<0 (in which case the s.b. profile diverges at the origin).  I think this should be possible in principle, but in practice the rejection sampling near the origin ends up in an infinite loop waiting for a deviate smaller than 1/inf.  Any suggestions?

Some other items of note:

- Drawing a nu=-0.6 Spergel profile is about 3x faster than drawing a n=4 Sersic profile, which has about the same shape near the half_light_radius.  A nu=0.5 Spergel takes the same amount of time as an n=1 Sersic (both exponentials).
[...]

Issue #497, opened 2014-01-08
Weird kernel for the pixel


ghost:
Hi!

I've come across a somewhat bizarre behaviour in the following working spinet which is copied from the example demo1.py script:

``` python
import galsim

pixel_scale = 0.5  # arcsec / pixel

pix = galsim.Pixel(pixel_scale)
image = pix.draw(dx=pixel_scale)
image.write('pix.fits')
```

The resulting fits file looks like i
![pix_fits](https://f.cloud.github.com/assets/4168086/1867683/0bcacbf6-7855-11e3-883e-8c68bbb00122.png)

where black is 0 and white is 0.25. I would expect a 2 by 2 kernel with value 0.25 in all four non zero components. 

This kernel is then used in the convolution by the PSF:

``` python
final = galsim.Convolve([gal, psf, pix])
```

Is that a misinterpretation or a bug?

Cheers
Thibault Kuntzer
EPFL

PR #625, opened 2015-01-04
#616 - Spergel profile

jmeyers314:
This is a pull request to add the Spergel profile (define [here](http://arxiv.org/abs/1007.3248)) as a new GSObject.  The Spergel profile shares some qualitative similarities to the Sersic profile in that it carries an index that varies the central cuspiness and prominence of the wings.  It's the Fourier transform of the Moffat profile, and hence has analytic representations in both real-space and Fourier-space making it relatively quick to draw.

I've implemented truncation in a manner that parallels Sersic profile truncation (note that the FT of a truncated Spergel no longer has a closed-form analytic expression, so a lookup table is used in this case).

This is just the first step for issue #616.  In the long run, we'd like to implement the Spergel profile via its Taylor expansion, which will allow changing ellipticity and size by simply reweighting coefficients of the expansion, and hence also allow a large range of PSF-convolved profiles to be drawn using precomputed convolutions.


galsim/base.py
```diff
@@ -2181,6 +2181,105 @@ def getScaleRadius(self):
...
+        D. N. Spergel, "ANALYTICAL GALAXY PROFILES FOR PHOTOMETRIC AND LENSING ANALYSIS,"
+        ASTROPHYS J SUPPL S 191(1), 58-65 (2010) [doi:10.1088/0067-0049/191/1/58].
+
+    Initialization
+    --------------
+
+    The allowed range of values for the `nu` parameter is -0.85 <= nu <= 0.85.  An exception will be
+    thrown if you provide a value outside that range.  These values are chosen to match the values
+    investigated by the above paper.  At and below nu=0.0, the surface brightness profile diverges
+    at the origin, which currently affects our ability to perform photon-shooting or `real_space`
```
rmandelb: The "which currently affects our ability..." is a bit vague.  It's not clear if this means you just can't do it at all, or can do it less accurately, or what.  I recommend being a little more specific here.

galsim/base.py
```diff
@@ -2181,6 +2181,105 @@ def getScaleRadius(self):
...
+    A Spergel can be initialized using one (and only one) of two possible size parameters:
+    `scale_radius` or `half_light_radius`.  Exactly one of these two is required.
+
+    @param nu               The Spergel index, nu.
+    @param half_light_radius  The half-light radius of the profile.  Typically given in arcsec.
+                            [One of `scale_radius` or `half_light_radius` is required.]
+    @param scale_radius     The scale radius of the profile.  Typically given in arcsec.
+                            [One of `scale_radius` or `half_light_radius` is required.]
+    @param flux             The flux (in photons) of the profile. [default: 1]
+    @param trunc            An optional truncation radius at which the profile is made to drop to
```
rmandelb: This is in the same units as the units used to specify the radius parameter?

galsim/base.py
```diff
@@ -2181,6 +2181,105 @@ def getScaleRadius(self):
...
+    If you are truncating the profile, the optional parameter, `flux_untruncated`, specifies
+    whether the `flux` and `half_light_radius` specifications correspond to the untruncated
+    profile (`True`) or to the truncated profile (`False`, default).  The impact of this parameter
+    is a little subtle; please see the detailed description in the `Sersic` profile documentation
+    for examples.
+
+    Methods
+    -------
+
+    In addition to the usual GSObject methods, Spergel has the following access methods:
```
rmandelb: Probably stupid, but this is weirding me out, what with David Spergel being a person who I know.  How about "the Spergel class"?  (To be fair, if you were to replace "Spergel" with, say, "Exponential" I wouldn't love it - I would rather say "the Exponential class" in that case too.  So I guess it's not just the name thing.)

galsim/base.py
```diff
@@ -2181,6 +2181,105 @@ def getScaleRadius(self):
...
+
+        D. N. Spergel, "ANALYTICAL GALAXY PROFILES FOR PHOTOMETRIC AND LENSING ANALYSIS,"
+        ASTROPHYS J SUPPL S 191(1), 58-65 (2010) [doi:10.1088/0067-0049/191/1/58].
+
+    Initialization
+    --------------
+
+    The allowed range of values for the `nu` parameter is -0.85 <= nu <= 0.85.  An exception will be
+    thrown if you provide a value outside that range.  These values are chosen to match the values
+    investigated by the above paper.  At and below nu=0.0, the surface brightness profile diverges
```
rmandelb: I think it's worth mentioning here the fact that nu<-0.85 is numerically tricky. (especially since the paper does go down to -0.9, not -0.85)

src/SBSpergel.cpp
```diff
@@ -0,0 +1,717 @@
...
+        x0 *= _inv_r0;
+        dx *= _inv_r0;
+        dxy *= _inv_r0;
+        y0 *= _inv_r0;
+        dy *= _inv_r0;
+        dyx *= _inv_r0;
+
+        double x00 = x0; // Preserve the originals for below.
+        double y00 = y0;
+        It valit = val.linearView().begin();
```
rmandelb: What is the purpose of this line?  It seems like you're just about to make a new ```valit``` in the loop right below this, without ever using this one.

rmandelb: (There's another example of this below, in ```fillKValue```.)

tests/SBProfile_comparison_images/spergel_image.py
```diff
@@ -0,0 +1,44 @@
+import os
```
rmandelb: Missing header with license.

tests/galsim_test_helpers.py
```diff
@@ -204,22 +204,10 @@ def do_kvalue(prof, name):
...
-    r = minr
-    sum = 0.
-    while r < maxr:
-        # In this tight loop, it is worth optimizing away the parse_pos_args step.
-        # It makes a rather significant difference in the running time of this function.
-        sum += r * prof.SBProfile.xValue(galsim.PositionD(r,0)) 
-        r += dr
-    sum *= 2. * math.pi * dr
-    return sum
+    f = lambda r: 2 * np.pi * r * prof.SBProfile.xValue(galsim.PositionD(r,0))
```
rmandelb: Nice simplification.
(But is there any reason to jettison the docstring?  I agree the new version is very readable and simple but docstrings are still nice...)

tests/test_base.py
```diff
@@ -1687,6 +1691,340 @@ def test_kolmogorov_flux_scaling():
...
+            do_kvalue(spergel, "Spergel(nu={:1}) ".format(nu))
+
+        # s.b. finite at origin iff nu > 0
+        # Consequently can only test sb sum and photon shooting for nu > 0.
+        if nu > 0.0:
+            np.testing.assert_almost_equal(
+                myImg.array.sum()*dx**2, myImg.added_flux, 5,
+                err_msg="Spergel profile GSObject::draw returned wrong added_flux")
+
+            # Test photon shooting.
```
rmandelb: Sorry, I am a little confused by this restriction.  I thought the photon shooting used some approximations but still works for nu<0.  In this case, why shouldn't it be tested here?

galsim/base.py
```diff
@@ -2181,6 +2181,105 @@ def getScaleRadius(self):
         return self.SBProfile.getScaleRadius()
 
 
+class Spergel(GSObject):
+    """A class describing a Spergel profile.
+
+    The Spergel surface brightness profile is characterized by three properties: its Spergel index
+    `nu`, its `flux`, and either the `half_light_radius` or `scale_radius`.  Given these properties,
```
rmandelb: I thought a little more about the docstring, and i think there are 2 more (small) things I'd like to see:

* I know what a Spergel profile is and why one might want to use it, but I think that someone who hasn't read David's paper won't know that based on this docstring.  I think it would be worth having a sentence in this first paragraph saying something along the lines of the Spergel profile being Sersic-ish but with an analytic Fourier transform, so it can be faster to draw while still looking quite like a realistic galaxy profile.  This is parallel to other docstrings where we've tried to say that such-and-such class is a common form used to represent PSFs or galaxies or whatever.

* A phrase about timing considerations would be worthwhile - even just something simple, like nu=BLAH (something deVauc-ish) is roughly 3x faster to draw than a deVauc, while nu=BLAH2 (exp-ish) is similar in speed to an Exponential.

I realize you have the reference to the paper which could help address my first point, but I think someone looking at the available classes and their docstrings won't have a motivation to look at the paper and figure out what this is without some basic information that it's a profile that can describe galaxies and is sometimes faster than Sersic.

rmandelb:
Hi Josh - I just went over this and I think it looks really good.  As you can see, my comments are quite minor.  Tests pass on both of my systems.

galsim/base.py
jmeyers314: Whoops.  That was from an earlier iteration.  I think photon-shooting and real-space convolution mostly work fine now.  I changed this sentence to:

"These values are chosen to match the values investigated by the above paper, with the exception that our lower limit is slightly above the lower limit in the paper.  This change is due to numerical difficulties integrating the *very* peaky nu < -0.85 profiles."

galsim/base.py
jmeyers314: Yep.  I just went ahead and copied the description from Moffat, which included the line about matching units.  I also copied this to Sersic and DeVaucouleurs, which lacked the units description.

galsim/base.py
jmeyers314: Lol.  Having not met him, I guess "Spergel" is just an abstract concept to me :).  I'm happy to replace Spergel -> Spergel profile where needed.

galsim/base.py
jmeyers314: Addressed in comment above.

src/SBSpergel.cpp
jmeyers314: Good eye.  I copied this from SBSersic.cpp, which also appears to not use it.  It's also in SBKolmogorov and SBMoffat.  I've removed it from all four.

tests/SBProfile_comparison_images/spergel_image.py
jmeyers314: Added the header.

tests/test_base.py
jmeyers314: Whoops again.  I had fixed the code, but forgot to unrestrict the unit test.  I removed the if statement.

src/SBSpergel.cpp
jmeyers314: Ah, ok.  Even better!

galsim/base.py
jmeyers314: I added the following paragraph:

    The Spergel profile is intended as a generic galaxy profile, somewhat like a Sersic profile, but
    with the advantage of being analytic in both real space and Fourier space.  The Spergel index
    `nu` plays a similar role to the Sersic index `n`, in that it adjusts the relative peakiness of
    the profile core and the relative prominence of the profile wings.  At `nu = 0.5`, the Spergel
    profile is equivalent to an Exponential profile (or alternatively an `n = 1.0` Sersic profile).
    At `nu = -0.6` (and in the radial range near the half-light radius), the Spergel profile is
    similar to a DeVaucouleurs profile or `n = 4.0` Sersic profile.  Note that a `nu = -0.6` Spergel
    profile is roughly 3x faster to draw than an `n = 4.0` Sersic profile, though a `nu = 0.5`
    Spergel profile, `n = 1.0` Sersic profile, and an Exponential profile all take about the same
    amount of time to draw.

rmandelb: That's great.  Thank you.

galsim/base.py
rmandelb: Thanks.

galsim/base.py
rmandelb: Sounds good.

rmandelb:
@rmjarvis , I'm getting some compilation errors that seem to show up after your commit that optimized the iterators:
```
src/SBKolmogorov.cpp:161:20: error: no viable conversion from 'iterator' (aka 'VarConjIter<T>') to
      'It' (aka 'VIt<std::complex<double>, 1, tmv::NonConj>')
                It valit = val.col(j).begin();
                   ^       ~~~~~~~~~~~~~~~~~~
/usr/local/include/tmv/TMV_VIt.h:234:9: note: candidate constructor not viable: no known conversion
      from 'iterator' (aka 'VarConjIter<T>') to 'const type &' (aka 'const VIt<std::complex<double>,
      1, (tmv::ConjType)0U> &') for 1st argument
        VIt(const type& rhs) : p(rhs.get()), s(rhs.step()) {}
        ^
/usr/local/include/tmv/TMV_VIt.h:237:9: note: candidate template ignored: could not match 'VIt'
      against 'VarConjIter'
        VIt(const VIt<T,S2,C>& rhs) : p(rhs.get()), s(rhs.step()) {}
        ^
src/SBKolmogorov.cpp:218:12: error: no viable conversion from 'iterator' (aka 'VarConjIter<T>') to
      'It' (aka 'VIt<std::complex<double>, 1, tmv::NonConj>')
        It valit = val.linearView().begin();
           ^       ~~~~~~~~~~~~~~~~~~~~~~~~
/usr/local/include/tmv/TMV_VIt.h:234:9: note: candidate constructor not viable: no known conversion
      from 'iterator' (aka 'VarConjIter<T>') to 'const type &' (aka 'const VIt<std::complex<double>,
      1, (tmv::ConjType)0U> &') for 1st argument
        VIt(const type& rhs) : p(rhs.get()), s(rhs.step()) {}
        ^
/usr/local/include/tmv/TMV_VIt.h:237:9: note: candidate template ignored: could not match 'VIt'
      against 'VarConjIter'
        VIt(const VIt<T,S2,C>& rhs) : p(rhs.get()), s(rhs.step()) {}
        ^
2 errors generated.
scons: *** [src/.obj/SBKolmogorov.os] Error 1
```
(And likewise for the other files that have these iterators.)

rmjarvis:
> @rmjarvis , I'm getting some compilation errors that seem to show up after your commit that optimized the iterators:

You apparently have TMV v0.71.  Upgrade to 0.72.

rmandelb:
Curses.

tests/test_base.py
rmandelb: Thanks.  I reran unit tests and it seems fine on my systems.

jmeyers314:
I've got a small update on performance timings for this PR.  In the doc string for Spergel, there is currently a sentence that reads: 
```
Note that a `nu = -0.6` Spergel profile is roughly 3x faster to draw than an `n = 4.0`
Sersic profile, though a `nu = 0.5` Spergel profile, `n = 1.0` Sersic profile, and an 
Exponential profile all take about the same amount of time to draw.
```
I should note that this is true only when taking advantage of a cached Sersic Hankel transform.  If not taking advantage of the cache, for example if drawing Sersic profiles with n continuously varying near 4.0 and Spergel profiles with nu continuously varying near -0.6, then the Spergel profiles are ~50 times faster to draw.

rmandelb:
Wow.  That's certainly worth mentioning in the docstring, so people can properly compare times including or excluding the overhead of setting up the Hankel transform depending on exactly what they're doing.

galsim/base.py
rmjarvis: Is there a good reason to allow truncation of Spergel profiles?  My understanding was that their main advantage over Sersic is that they have analytic Fourier transforms.  This isn't the case if they are truncated.  So I think maybe we ought not to allow truncation for them.

Unless there is some other advantage of Spergel that would be maintained even if they are truncated.

include/galsim/SBMoffatImpl.h
```diff
@@ -45,26 +45,26 @@ namespace galsim {
         double maxK() const;
         double stepK() const;
 
-        void getXRange(double& xmin, double& xmax, std::vector<double>& ) const 
+        void getXRange(double& xmin, double& xmax, std::vector<double>& ) const
```
rmjarvis: Do you know why your PRs always include all these white space changes?  It would be helpful if you could somehow avoid committing these changes.  It makes it hard to to the review when I have to verify that most of these line changes just involve a single space at the end of a line being removed, which isn't relevant to anything.

You can use `git add -p` to pick through the local changes and select only a subset of them to commit.

Also, if you want to make the white space changes, you can commit them on master and then merge them into the issue branch, so they won't show up on the pull review.  White space changes like this don't have to be reviewed..  :)

src/SBSpergel.cpp
```diff
@@ -0,0 +1,715 @@
...
+        double i0 = (-dy * x00 + dxy * y00) / det;
+        double j0 = (dyx * x00 - dx * y00) / det;
+        dbg<<"i0, j0 = "<<i0<<','<<j0<<std::endl;
+        dbg<<"x0 + dx i + dxy j = "<<x00+dx*i0+dxy*j0<<std::endl;
+        dbg<<"y0 + dyx i + dy j = "<<y00+dyx*i0+dy*j0<<std::endl;
+        int inti0 = int(floor(i0+0.5));
+        int intj0 = int(floor(j0+0.5));
+
+        if ( std::abs(i0 - inti0) < 1.e-12 && std::abs(j0 - intj0) < 1.e-12 &&
+             inti0 >= 0 && inti0 < m && intj0 >= 0 && intj0 < n)  {
```
rmjarvis: My guess is that this isn't likely to be worth doing.  For Sersic, it's important for high n Sersics where the central peak is so sharp that an error of 1.e-18 in the rsq value leads to an error of about 0.5% in the amplitude.  cf. [this comment](https://github.com/GalSim-developers/GalSim/issues/497#issuecomment-31847079).  

For Spergel, the really peaky ones (nu <= 0) actually have an infinite value at the center, so making this "fix" will put an infinity at that point, which isn't likely to be terribly useful.

In fact, this feature of Spergel profiles probably ought to be mentioned in the main doc string.  Namely, that for nu <= 0, the central value is undefined, so there could be problems rendering them directly without convolving by either a PSF or a pixel if the profile center is precisely on a pixel center.

src/SBSpergel.cpp
```diff
@@ -0,0 +1,715 @@
...
+
+        double operator()(double u) const
+        // Return flux integrated up to radius `u` in units of r0, minus `flux_frac`
+        // (i.e., make a residual so this can be used to search for a target flux.
+        {
+            double fnup1 = std::pow(u / 2., _nu+1.)
+                * boost::math::cyl_bessel_k(_nu+1., u)
+                / _gamma_nup2;
+            double f = 1.0 - 2.0 * (1.+_nu)*fnup1;
+            return f - _target;
```
rmjarvis: This is a pretty cool result.  However, I haven't been able to rederive it.  Maple gives me back something with hypergeom functions.  But the difference of what it has and your formula seems to be within numerical errors of zero for some select values of nu, so I assume your answer is correct.

Could you add a comment here with at least a skeleton of your derivation?  Maybe the key Bessel identity you used to get this.

rmjarvis: Ah, actually when integrating from u to infinity, Maple gives back something without the hypergeom, which leads to your result.  Still a comment about this formula in the code would be helpful.

src/SBSpergel.cpp
```diff
@@ -0,0 +1,715 @@
...
+        double _target;
+    };
+
+    double SpergelInfo::calculateFluxRadius(const double& flux_frac) const
+    {
+        // Calcute r such that L(r/r0) / L_tot == flux_frac
+
+        // These seem to bracket pretty much every reasonable possibility
+        // that I checked in Mathematica, though I'm not very happy about
+        // the lower bound....
```
rmjarvis: I'm more concerned about the upper bound.  The input `flux_frac` could be arbitrarily close to 1 if the user puts in a very small `folding_threshold` or `shoot_accuracy`.  Are you sure that 20 is always safe for that?

I think maybe if flux_frac < 0.5, do the call to `bracketLowerWithLimit` as you have here to make sure the lower bound is ok, otherwise call `bracketUpper` to make sure the upper bound is ok.

src/SBSpergel.cpp
```diff
@@ -0,0 +1,715 @@
...
+    {
+        dbg<<"Find HLR for (nu,trunc) = ("<<_nu<<","<<_trunc<<")"<<std::endl;
+        SpergelIntegratedFlux func(_nu, _gamma_nup2, 0.5*getFluxFraction());
+        double b1 = 0.001; // These are pretty conservative...
+        double b2 = 25.0;
+        Solve<SpergelIntegratedFlux> solver(func, b1, b2);
+        xdbg<<"Initial range is "<<b1<<" .. "<<b2<<std::endl;
+        solver.setMethod(Brent);
+        solver.bracketLowerWithLimit(0.); // Expand if needed
+        solver.bracketUpper();
```
rmjarvis: The range you have is actually more than sufficient.  So I don't think you need these bracket calls.

I believe the range of _re values for nu in (-0.85, 0.85) is (0.13, 1.92).  So (0.001, 25) is completely sufficient without any expansion.  You could even shrink it a bit if you want to make this calculation more efficient, but that's not really all that necessary I suspect.  The solver probably zeros in on the right answer pretty quickly as it is.

galsim/base.py
jmeyers314: I was just trying to be as complete as possible.  I'm not aware of any advantage of truncated Spergels over truncated Sersics.  In fact, the truncated Spergels are a bit slower to render than the truncated Sersics.

I'd be fine with removing the truncation.  It would be easy enough to reintroduce if someone finds a use case in the future.

src/SBSpergel.cpp
```diff
@@ -0,0 +1,715 @@
...
+        // so let's take the completely arbitrary value of 10.  (This value was found to be
+        // conservative for Sersic, but I haven't investigated here.)
+        // 10 h^4 <= kvalue_accuracy
+        // h = (kvalue_accuracy/10)^0.25
+        double dlogk = _gsparams->table_spacing * sqrt(sqrt(_gsparams->kvalue_accuracy / 10.));
+        dbg<<"Using dlogk = "<<dlogk<<std::endl;
+        int n_below_thresh = 0;
+
+        // Don't go past k = 500
+        double kmin = dlogk; // have to begin somewhere...
```
rmjarvis: I'm a little concerned about this.  It would be nice if there was a calculation to show that the linear approximation you use for k < kmin has errors less than kvalue_accuracy.  We usually try to make sure of that, and it's not obvious to me that this value for kmin satisfies that criterion.  

However, I've also proposed getting rid of the truncation option, which means this whole function would go away.  In which case the comment is moot.

include/galsim/SBMoffatImpl.h
jmeyers314: Yeah, sorry about that.  I have a line in my .emacs that automatically removes all trailing whitespace upon a save.  My own personal OCD I guess.

I see that there are some emacs packages online though that will apparently only do this for lines that I've actually edited.  I'll look into this.

In the meantime..., supposedly adding a `?w=1` to a github url will ignore whitespace changes.  Also I think this is an option for most diff utils.  `-w` for gnu diff.

rmandelb: > In the meantime..., supposedly adding a ?w=1 to a github url will ignore whitespace changes

That's true, but it does mean you can no longer comment on the code (or at least I wasn't able to when I used that option).

tests/galsim_test_helpers.py
rmjarvis: Similarly, it is worth keeping my comment about why we use `prof.SBProfile.xValue` rather than `prof.xValue`.  In the latter case, the argument checking of the python version slows the function evaluation considerably, which matters when the xValues are being called in a tight loop.  (Both in the original version, and in your improved int1d version.)

galsim/base.py
```diff
@@ -2181,6 +2183,125 @@ def getScaleRadius(self):
...
+
+    For more information, refer to
+
+        D. N. Spergel, "ANALYTICAL GALAXY PROFILES FOR PHOTOMETRIC AND LENSING ANALYSIS,"
+        ASTROPHYS J SUPPL S 191(1), 58-65 (2010) [doi:10.1088/0067-0049/191/1/58].
+
+    Initialization
+    --------------
+
+    The allowed range of values for the `nu` parameter is -0.85 <= nu <= 0.85.  An exception will be
```
rmjarvis: In the limit of nu->infinity, the Spergel profile becomes a Gaussian.  That being the case, I don't think we need to have such a low upper limit for nu.  The practical limit is somewhere around nu = 150 or so, where xnorm start to approach 1.e323 where double precision will overflow.  So perhaps setting nu max to 100 would be reasonable.  I'm sure no one will consider this limit oppressive.  :)

rmjarvis:
OK.  I've finished going through it.  Even checked all the math, and I think it's all correct.  :)  

Tests pass on my systems.

src/SBSpergel.cpp
jmeyers314: Sure thing.  I've removed the (0,0) guard and added the following to the docstring:

```
    Note that for `nu <= 0.0`, the Spergel profile surface brightness diverges at the origin.  This
    may lead to rendering problems if the profile is not convolved by either a PSF or a pixel and
    the profile center is precisely on a pixel center.
```

src/SBSpergel.cpp
jmeyers314: I can't take credit for this result.  It's in Spergel (2010).  Here's the detailed comment I added though:

```
        // Return flux integrated up to radius `u` in units of r0, minus `flux_frac`
        // (i.e., make a residual so this can be used to search for a target flux.
        // This result is derived in Spergel (2010) eqn. 8 by going to Fourier space
        // and integrating by parts.
        // The key Bessel identities:
        // int(r J0(k r), r=0..R) = R J1(k R) / k
        // d[-J0(k R)]/dk = R J1(k R)
        // The definition of the radial surface brightness profile and Fourier transform:
        // Sigma_nu(r) = (r/2)^nu K_nu(r)/Gamma(nu+1)
        //             = int(k J0(k r) / (1+k^2)^(1+nu), k=0..inf)
        // and the main result:
        // F(R) = int(2 pi r Sigma(r), r=0..R)
        //      = int(r int(k J0(k r) / (1+k^2)^(1+nu), k=0..inf), r=0..R) // Do the r-integral
        //      = int(R J1(k R)/(1+k^2)^(1+nu), k=0..inf)
        // Now integrate by parts with
        //      u = 1/(1+k^2)^(1+nu)                 dv = R J1(k R) dk
        // =>  du = -2 k (1+nu)/(1+k^2)^(2+nu) dk     v = -J0(k R)
        // => F(R) = u v | k=0,inf - int(v du, k=0..inf)
        //         = (0 + 1) - 2 (1+nu) int(k J0(k R) / (1+k^2)^2+nu, k=0..inf)
        //         = 1 - 2 (1+nu) (R/2)^(nu+1) K_{nu+1}(R) / Gamma(nu+2)
```

src/SBSpergel.cpp
jmeyers314: >The range you have is actually more than sufficient.

I think I had lower lower limit to handle the truncated cases, but yes, super conservative on the upper limit.  I went ahead and changed these to handle all untruncated -0.85 < nu < 100 cases and removed the bracket calls.

galsim/base.py
jmeyers314: I've now tried extending the upper limit.  I start running into overflow errors in boost::math::cyl_bessel_k though for nu larger than about 4.0, so I went ahead and made this the upper limit.

tests/galsim_test_helpers.py
jmeyers314: Sure.  I put the comment back.

src/SBSpergel.cpp
jmeyers314: Good idea; now implemented.  (And I adjusted the initial bracket to cover just the flux_frac=0.5 cases for -0.85 < nu < 4).

I also added a `solver.setXTolerance(1e-25)` line that seems to have solved my other problem where the root finder was super sensitive to the initial lower limit.  While debugging this, I added two methods to the (somewhat hidden `_galsim`) python layer: `calculateFluxRadius` and `calculateIntegratedFlux`, which I left in place and also use for some c++ vs. Mathematica unit tests.

galsim/base.py
jmeyers314: Removed Spergel truncation.

jmeyers314:
@rmjarvis and @rmandelb -- I think I've hit all your comments now, and even cleaned up the whitespace diff, (@mdschneider, if you have any comments too, let me know).  There's no real hurry for this, but let's say I'll plan to merge by Friday if nothing new comes up.  Thanks!

-Josh

include/galsim/SBSpergel.h
```diff
@@ -0,0 +1,114 @@
...
+ */
+
+#include "SBProfile.h"
+
+namespace galsim {
+
+    namespace sbp {
+
+        // Constrain range of allowed Spergel index nu.  Spergel (2010) Table 1 lists values of nu
+        // from -0.9 to +0.85. I've found that nu = -0.9 is too tricky for the GKP integrator to
```
rmjarvis: 0.85 -> 4.0

rmjarvis: Oh, nvm.  Here, you are talking about what is in Spergel's paper.

galsim/base.py
```diff
@@ -2181,6 +2183,103 @@ def getScaleRadius(self):
...
+
+    For more information, refer to
+
+        D. N. Spergel, "ANALYTICAL GALAXY PROFILES FOR PHOTOMETRIC AND LENSING ANALYSIS,"
+        ASTROPHYS J SUPPL S 191(1), 58-65 (2010) [doi:10.1088/0067-0049/191/1/58].
+
+    Initialization
+    --------------
+
+    The allowed range of values for the `nu` parameter is -0.85 <= nu <= 100.  An exception will be
```
rmjarvis: Ah, but here, this should be 4.0.  Not 100.

galsim/base.py
```diff
@@ -2181,6 +2183,103 @@ def getScaleRadius(self):
...
+    thrown if you provide a value outside that range.  The lower limit is set above the theoretical
+    lower limit of -1 due to numerical difficulties integrating the *very* peaky nu < -0.85
+    profiles.  The upper limit ...
```
rmjarvis: Also, you forgot to finish this sentence.

src/SBSpergel.cpp
```diff
@@ -0,0 +1,524 @@
...
+            double f = 1.0 - 2.0 * (1.+_nu)*fnup1;
+            return f - _target;
+        }
+    private:
+        double _nu;
+        double _gamma_nup2;
+        double _target;
+    };
+
+    class SpergelTaylorIntegratedFlux
```
rmjarvis: I don't think this is used anywhere.  Is it worth keeping?  Maybe.  But if so, should probably add a comment that this used to be used for something, but now is left here as a relic in case anyone finds it useful again in the future.

galsim/base.py
jmeyers314: Yep.  Thanks.

galsim/base.py
jmeyers314: >Also, you forgot to finish this sentence.

So I did.  I added:
```The upper limit is set to avoid numerical difficulties evaluating the modified
    Bessel function of the second kind.
```

src/SBSpergel.cpp
jmeyers314: Right.  I was trying to use this to calculate the small radius enclosed flux before I discovered `solver.setXTolerance()`.

It's probably unlikely that we'd need to revive this, so I think I'll just get rid of it.  It will live on in the git history one way or another anyway.

src/SBSpergel.cpp
```diff
@@ -0,0 +1,524 @@
...
+                _sampler.reset(new OneDimensionalDeviate( *_radial, range, true, _gsparams));
+            } else {
+                // exact s.b. profile diverges at origin, so replace the inner most circle
+                // (defined such that enclosed flux is shoot_acccuracy) with a linear function
+                // that contains the same flux and has the right value at r = rmin.
+                // So need to solve the following for a and b:
+                // int(2 pi r (a + b r) dr, 0..rmin) = shoot_accuracy
+                // a + b rmin = K_nu(rmin) * rmin^nu
+                double flux_target = _gsparams->shoot_accuracy;
+                double shoot_rmin = calculateFluxRadius(flux_target);
```
rmjarvis: I worked ab it on my idea of making this radius be where the integrated *error* in the flux inside r=R is shoot_accuracy rather than the integrated flux, using the small r approximation of K_niu(r).  

I was hoping there was an analytic solution (using the approximate form for K_nu at least), which would obviate the need for calling calculateFluxRadius with the really small argument for the flux_target, but I kind of ran into a wall when I got to figuring out where the error flipped sign.   The zero crossing seems to require a non-linear root solver, so it would probably be a bit complicated to get this idea working.

So I tried out a few concrete values of nu to see if this idea was worth pursuing.  i.e. How conservative is this choice of rmin?

It turns out that for nu only mildly less than 0, this choice of rmin is a bit conservative.  At nu=-0.3, it is about a factor of 5 smaller than it needs to be.  But as nu gets very small, say near -0.8, the difference gets less important.  It comes out almost the same either way.  And since these are the values where the numerical problems were happening, it seems that there isn't really any benefit to this idea.

So good that you figured out the XTolerance thing to make this work as is.  :)

rmjarvis:
Nothing else from me Josh.  Ready to merge as far as I'm concerned.

Issue #626, opened 2015-01-09
Support direct reading PSFEx directly from HDU in galsim.des.DES_PSFEx


joezuntz:
I'm doing some work that involves packaging PSFEx data together with image exposures in different headers of a single FITS file.  For this to work with the galsim facility I need to be able to initialize a galsim.des.DES_PSFEx with an HDU instead of a whole file.

I'm assigning this to myself; I need to make some minor changes to galsim.des.DES_PSFEx to accommodate this.

PR #631, opened 2015-01-19
Enable using an HDU instead of a filename for DES_PSFEx

joezuntz:
(#626) This now seems to work nicely - I've used it in production.


galsim/des/des_psfex.py
```diff
@@ -81,16 +81,16 @@ class DES_PSFEx(object):
...
-    @param image_file_name The name of the fits file of the original image (needed for the
-                           WCS information in the header).  If unavailable, you may omit this
-                           (or use None), but then the returned profiles will be in image
-                           coordinates, not world coordinates.  (Default `image_file_name = None`)
-    @param wcs             Optional way to provide the WCS if you already have it loaded from the
-                           image file. (Default `wcs = None`)
-    @param dir             Optionally a directory name can be provided if the file_name does not 
-                           already include it.  (The image file is assumed to be in the same
-                           directory.) (Default `dir = None`)
+    @param file_name_or_hdu The file name to be read in, or a pyfits HDU.
```
rmjarvis: I don't think we want to change the name of the existing parameter.  If anyone currently uses `file_name` as a named argument, this will break their code.  It also breaks the config stuff that currently uses this parameter name (see `_req_params` below).

There are two options that I think would be ok.  First, you could add a new `hdu` parameter and have either `file_name` or `hdu` be required.  The new parameter should be put last to maintain backward compatibility for people who are relying on the current argument order.

Or second, you could leave the parameter name `file_name`, but just say that the file_name is allowed to be an HDU rather than a string, in which case that HDU is used rather than opening up the file.  This is probably the simpler option, and I think it would be fine.  Also, it would be almost identical to your existing code, just without the change to the parameter name.

joezuntz:
Hi Mike,

Thanks - good point - I've changed the name back to file_name.

rmjarvis:
Great.  Maybe just add a note in the CHANGELOG to mention this change.  But otherwise, all looks fine to me.

joezuntz:
Cool - I've added.

rmandelb:
Looks good to me.

rmjarvis:
Joe, I think this is ready to merge.

Issue #511, opened 2014-03-01
Make GSObjects immutable?


rmjarvis:
This is a proposal.  I welcome comments...

As I've been learning python over the past two years (mostly in the course of working on GalSim), I've come to realize that it works much better with immutable classes.  This is mostly because of the way the python = operator works, being a naming operator rather than assignment.  This is why a number of places in GalSim we have to be careful about not modifying the inputs to functions, since we recognize that this behavior would be surprising to the python user who usually expects functions to not modify the inputs.

I've managed to get a number of classes in GalSim to be immutable.  The new WCS classes are all immutable.  Josh's Bandpass and SED are immutable.  I recently changed Angle and BoundsI/D to be immutable (in the python layer, not in C++).  

The usual syntax when you want to "change" one of these immutable objects is `wcs = wcs.setOrigin(origin)` or `theta = theta.wrap()`.  This reassigns the new value to the original name.  In fact, this is what python already does for the mutating operators like `a += b` and the like.  It is really just shorthand for `a = a + b`.  i.e. the original `a` is preserved if anyone else references it, and locally the quantity `a + b` is given the name `a`.  As such, I have also been moving away from defining these operators in GalSim, letting python use the default definitions, since this will usually behave the way users will expect in python.

However, GSObjects are not currently immutable in GalSim.  In particular, we have functions like `applyShear`, `setFlux`, etc. which do change the object rather than return a new object.  In retrospect, knowing what I know now about how python works, I would have preferred that we had made these functions return new objects that are appropriately modified versions of the original.  i.e. the way the various `create` methods work: `createSheared` and the like.  This would have avoided a number of difficulties we have since had with respect to GSObjects, most notably how hard it is to make them picklable.

So my (somewhat radical) proposal is to switch GSObject and ChromaticObject to an immutable design where we remove or deprecate the mutating methods and switch over to using the form `gal = gal.createSheared(shear)` in all examples and documentation.  In some cases, we'll need to make a new function if the `create` version is not available, such as `setFlux`.

The other half of my proposal is to revisit the names for these functions.  The `create` names are a little stilted sounding, IMO.  It wasn't really a big deal before, since the `apply` methods were available, which I thought read nicely, but I've never particularly liked the `create` function names.  

[...]

PR #632, opened 2015-01-19
#618 - interpolated ChromaticObject functionality

rmandelb:
This is a new pull request for code that enables faster rendering of ChromaticObjects by interpolating between precomputed images.  The code has been substantially modified compared to what was presented in an earlier pull request, thanks to comments from @rmjarvis and @jmeyers314 (and looks different enough now that I think it's worth starting a new PR).   Please check it out and see what you think.

A few final comments and questions:
* @jmeyers314 , one thing I was thinking about was the question of whether it makes sense to ever draw a ChromaticAtmosphere or ChromaticOpticalPSF or any such thing without convolving with something that has an SED.  Currently, this is allowed, but it's not really clear what this corresponds to physically.  Is this something that we should either forbid or at least raise a warning that the physical interpretation is unclear?

* More generally, I was thinking about transformations of certain ChromaticObjects.  Currently, ChromaticAtmosphere is qualitatively different from ChromaticOpticalPSF in that the former is a method that returns a ChromaticObject corresponding to the PSF at some base wavelength along with some chromatic transformations, whereas the latter is actually a class that has an evaluateAtWavelength method.  This means that in practice, if you do
```
foo = galsim.ChromaticAtmosphere(...)
bar = foo.dilate(...)
```
then the ChromaticObject.evaluateAtWavelength() method will take care of the dilation, whereas if you do
```
foo = galsim.ChromaticOpticalPSF(...)
bar = foo.dilate(...)
```
then since evaluateAtWavelength for the ChromaticOpticalPSF class does not explicitly apply transformations, nothing would happen here, so I actually made dilate() and other transformations of ChromaticOpticalPSF invalid.

I could alternatively make ChromaticOpticalPSF have an ```_A``` and ```_fluxFactor``` attribute, and then its ```evaluateAtWavelength``` would have to apply the transformations.  The reason I didn't do that was that I felt that since there is a clearly physical process behind the form of the ChromaticOpticalPSF, enabling arbitrary transformations of them didn't make sense.  However, I am willing to hear arguments to the contrary about this.  And for what it's worth, in principle one can still transform ChromaticOpticalPSF after ```setupInterpolation```.

More generally, this whole experience made me wonder about the idea of people making subclasses of ChromaticObject.  It seems like they have to make evaluateAtWavelength handle transformations explicitly, copying over a lot of code from ChromaticObject.evaluateAtWavelength, which is perhaps something to warn about in the documentation.  But the decision of whether or not to do that seems like it should depend on what the particular subclass is meant to represent.

Anyway, ideas on either of these two points are welcome.


jmeyers314:
Hi @rmandelb .  Sorry so slow to get to this.  I'll try to go through it this weekend so we can potentially discuss next week.

A couple quick comments:

> @jmeyers314 , one thing I was thinking about was the question of whether it makes sense to ever draw a ChromaticAtmosphere or ChromaticOpticalPSF or any such thing without convolving with something that has an SED.

From a physics point-of-view, I think you're right.  From a code point-of-view I'm not sure if we want to explicitly check this.  I guess the original design in my head was for evaluateAtWavelength() to encapsulate all of the wavelength-dependence.  The fact that separable objects (including `GSObject`s) have an `.SED` attribute is more of an exception than a rule.  The SED attribute is there only because its presence allows the separable profile to be drawn more quickly.  Also, none of the existing code attempts to access `.SED` unless the object is also marked as `separable`.

As an example, the `ChromaticSum` class doesn't have an SED attribute, and I'm pretty sure the `setupInterpolation` method will fail for `ChromaticSum` right now since the former assumes that `.SED` exists.

>since evaluateAtWavelength for the ChromaticOpticalPSF class does not explicitly apply transformations, nothing would happen here

I'm not sure I understand.  If `ChromaticOpticalPSF.dilate()` were left undefined (to take one example), then the parent class method `ChromaticObject.dilate()` would take over and wrap the `ChromaticOpticalPSF` inside a `ChromaticObject`.  The `evaluateAtWavelength` of the resulting object (which is now a `ChromaticObject`, similar to how transforming a `Gaussian` say results in a generic `GSObject`) would call `ChromaticOpticalPSF.evaluateAtWavelength()`, transform the result, and return.  This is how `Chromatic`, `ChromaticSum`, etc. all work.  I think this addresses your concern about copying large amounts of code into subclasses of ChromaticObject too.

I agree that it's probably unlikely that anyone will want to transform a `ChromaticOpticalPSF` (or `OpticalPSF`).  However, I don't think we should completely disable a perfectly functional ability.  I'd be more in favor of a warning.

Finally, one thing I should have thought of earlier is mutability, i.e. #511, which probably means that `setupInterpolation` and `noInterpolation` should return new objects instead of operating on themselves.

rmandelb:
Hi Josh -

> I'll try to go through it this weekend so we can potentially discuss next week.

That would be great, if you're not too pressed for time.  I think it'll be more efficient to sort out any remaining issues by talking in person.

> From a physics point-of-view, I think you're right. From a code point-of-view I'm not sure if we want to 
> explicitly check this.

Yes, I think it should be more of a "document this in the docstring" kind of thing.  I just made a note to update the docstrings.

> I'm not sure I understand. If ChromaticOpticalPSF.dilate() were left undefined (to take one example), 
> then the parent class method ChromaticObject.dilate() would take over and wrap the 
> ChromaticOpticalPSF inside a ChromaticObject. 

Maybe I'm doing something wrong in how I've set up this subclass, but that's not how it worked when I tried it.  When I left all those methods defined, then dilate(), shear(), etc. quietly did nothing at all to ChromaticOpticalPSF.  You can see this if you just delete all the lines of code that currently forbid those transformations and do something like
```
import galsim
foo = galsim.ChromaticOpticalPSF(2.4,[0.,0.,0.,0.,0.])
bp = galsim.Bandpass('examples/data/LSST_r.dat').thin()
bar = foo.shear(g2=0.5)
im_bar = bar.drawImage(bp, scale=0.01)
im_bar.write('im_bar.fits')
```

> Finally, one thing I should have thought of earlier is mutability, i.e. #511, which probably means that 
> setupInterpolation and noInterpolation should return new objects instead of operating on themselves.

Well... I would put this interpolation stuff into a different class than say shears and other transformations.  In principle if you set up the interpolation reasonably, you should get essentially the same results, just faster.  But I am open to an argument that it should make a new object.  However, I think names like "setupInterpolation" and "noInterpolation" imply you're acting on the object.  How about "makeInterpolated" for a version that returns a new object, and I think we don't need to have something that gets rid of interpolation in that case (since the old object will still exist in un-interpolated form).

jmeyers314:
>>I'm not sure I understand. If ChromaticOpticalPSF.dilate() were left undefined (to take one example), 
then the parent class method ChromaticObject.dilate() would take over and wrap the 
ChromaticOpticalPSF inside a ChromaticObject.

>Maybe I'm doing something wrong in how I've set up this subclass, but that's not how it worked when I tried it.

Hmm...  I guess it's obvious now that I didn't actually try this, sorry about that!  I should have said that I would have expected the `ChromaticOpticalPSF` object to get wrapped in a `ChromaticObject`.

Looking more carefully at your code, it appears the reason the wrapper didn't work is that you've called 
`super(ChromaticOpticalPSF, self).__init__(self)` inside of `ChromaticOpticalPSF.__init__()`, which seems like a perfectly reasonable thing to do.  The problem is that this then defines the `._A` and `._fluxFactor` attributes which then confuses `._applyMatrix()` into *not* changing the type of `ChromaticOpticalPSF` when you call, for example, `.dilate()`.  I think this boils down to the fact that the class `ChromaticObject` is attempting to do two things: A) it's the base class for all the other chromatic classes, B) it's the object that handles transformations.  This leads to some weirdness, for instance, a directly instantiated `ChromaticObject` always has a `.obj` attribute, but some of its subclasses won't, like `ChromaticSum`.

Personally, I think this behavior is probably okay, and we can work around it (in particular I think this might be simpler than, say, creating a `ChromaticTransform` subclass of `ChromaticObject`.)  One possibility would be to make a `ChromaticObject.wrap(obj)` method that takes care of defining `._A` and `._fluxFactor`, or maybe just making `obj` an optional instead of required keyword to the `ChromaticObject.__init__` method so that `._A` and `._fluxFactor` only get defined if one is explicitly trying to wrap an object.

rmandelb:
...or, I could just remove the ```super(ChromaticOpticalPSF, self).__init__(self)```, which was left over from when the ChromaticOpticalPSF was a subclass of InterpolatedChromaticObject, and do some other changes to make it work.  Please check out the commit that I just pushed, which (a) allows transformations of ChromaticOpticalPSF (I checked using the code snippet above that it works), and (b) doesn't require ```._A``` and ```._fluxFactor``` to always be defined.

I am open to other options like the ```ChromaticObject.wrap()``` but this way does seem to work.

Note that I need to modify the unit tests to test this additional functionality and compare, for example, transformation before vs. after ```setupInterpolation```.  But before I do that, what do you think of this way of handling the problem?

rmandelb:
Hi @jmeyers314 and @rmjarvis - I did the updates that we discussed, so I think this PR is ready for your review.

galsim/chromatic.py
```diff
@@ -130,6 +135,210 @@ def __init__(self, obj):
...
+        transform_done = False
+        for wave in waves:
+            if hasattr(self, '_A') and not self._nullTransformation(wave):
+                transform_done = True
+                break
+        if transform_done and hasattr(self, '_A'):
+            self._A = lambda w: np.matrix(np.identity(3), dtype=float)
+            self._fluxFactor = lambda w: 1.0
+            self._reset = True
+
```
jmeyers314: This works fine, but I think the following is more "pythonic"

```
        if hasattr(self, '_A') and not all([self._nullTransformation(w) for w in waves]):
            self._A = lambda w: np.matrix(np.identity(3), dtype=float)
            self._fluxFactor = lambda w: 1.0
            self._reset = True
```

jmeyers314: Actually, I'm not sure the whole `_reset` thing is necessary.  I think you could just stash the old `_A` (and `_fluxFactor`) in something like `_oldA`, and then retrieve it inside of `noInterpolation`.  You'd probably need to call `applyMatrix` there too to implement any (achromatic) transformations that were done after `setupInterpolation` was called.

galsim/chromatic.py
```diff
@@ -130,6 +135,210 @@ def __init__(self, obj):
...
+            del self.maxK_vals
+            del self.ims
+
+    def hasInterpolation(self):
+        """
+        A routine to check whether an object has interpolation set up or not.
+
+        @returns    True or False
+        """
+        if hasattr(self, 'waves'):
```
jmeyers314: I think this is just `return hasattr(self, 'waves')`.

galsim/chromatic.py
```diff
@@ -1359,3 +1746,74 @@ def withScaledFlux(self, flux_ratio):
...
             return ChromaticObject(self).withScaledFlux(flux_ratio)
+
+class ChromaticOpticalPSF(ChromaticObject):
+    """A subclass of ChromaticObject meant to represent chromatic optical PSFs.
+
+    Chromaticity plays two roles in optical PSFs. First, it determines the diffraction limit, via
+    the wavelength/diameter factor.  Second, aberrations such as defocus, coma, etc. are typically
+    defined in physical distances, but their impact on the PSF depends on their size in units of
+    wavelength.  Other aspects of the optical PSF are achromatic, e.g., the obscuration and struts.
+
```
jmeyers314: This last sentence is a bit confusing.  I think the obscuration and struts are chromatic (since they're diffractive effects), but since they're specified in units of the aperture diameter, and the diameter is specified chromatically, these get correctly implemented without any additional wavelength-dependence.

jmeyers314:
Hi @rmandelb, 
I'm finished looking through this.  Looks good, just a few small points above.  All tests pass.

One small additional question: I noticed while digging around in InterpolatedImage.h there's a  `MultipleImageHelper` class.  It's comment description sounds relevant:
```
     * @brief A Helper class that stores multiple images and their fourier transforms
     *
     * One of the ways to create an SBInterpolatedImage is to build it from a 
     * weighted sum of several component images.  The idea is that the component
     * images would be constant, but the weights might vary across the field of view.
```
Change the last sentence to `... weights might vary with SED or Bandpass.` and I think you get something similar to (a subset of) what you've developed here, but in the c++ layer.  It's not obvious to me that the c++ implementation would actually be faster here, (and I don't want to derail this PR), but I thought it was interesting enough to mention.  Also, as far as I can tell there currently isn't any way to access this in the python layer, is that right?

-Josh

galsim/chromatic.py
rmandelb: To address your first comment on this code, yes, what you wrote is more pythonic (and neater code in general) so I changed to your version.  Thanks.

Regarding your second comment, I see your point that it should be possible to do that, and then ```noInterpolation``` wouldn't have to raise an exception when trying to go back to the non-interpolated state.  The only question I have is whether there is something that could logically break this idea of having ```_oldA``` and ```_oldFluxFactor```?  For example, what if someone has applied a chromatic transformation, then called ```setupInterpolation```, then applied an achromatic transformation?  It seems to me that ```noInterpolation``` would then have to combine the transformations included in ```_A``` and ```_oldA```; is that right?

galsim/chromatic.py
rmandelb: *facepalm* Yes, it is.  Thanks, fixed.

galsim/chromatic.py
rmandelb: You're absolutely right.  I meant that they don't require any additional code/functionality to get the chromaticity included, but I was being sloppy about it.  Pushed a fix.

rmandelb:
Hi Josh - At least some aspect of the MultipleImageHelper is wrapped and accessible in python, for the benefit of the SBInterpolatedImage constructor (you can see this in pysrc/SBInterpolatedImage.cpp or by doing help(galsim._galsim.MultipleImageHelper)).   I think you're right that this potentially has some of what we want, but there's no value at this point in having it in the C++.

Thanks for iterating with me on this again!

galsim/chromatic.py
jmeyers314: >It seems to me that noInterpolation would then have to combine the transformations included in _A and _oldA; is that right?

Yep.  That's what I meant by calling `applyMatrix`.  I think the code in `noInterpolation` could look something like:
```
_tempA=self._A
self._A = self._oldA
self.applyMatrix(_tempA)
```
Or you could just do the matrix multiplication directly, I suppose.

rmandelb: Right of course, sorry.

rmandelb: Josh - I just made two commits to try to do this, but it's not working properly.  I played around with it for quite a while but am stumped.  (Note: there's a failing unit test that demonstrates exactly how it's going wrong.  The short answer is that I tried applying a chromatic transformation without any interpolation, then I did the case where I set up interpolation and applied a chromatic transformation, which is supposed to just unset the interpolation and do the normal way of drawing the transformed object.  These results should therefore be identical, but they don't seem to be.)

Is there something obvious that I'm missing here?  Sorry to dump this on you but I've been staring at the code pointlessly for about an hour since I initially wrote it, and am simply drawing a blank.

jmeyers314: Hi Rachel,
No worries.  It was my request so I think it's perfectly reasonable for you to dump this on me :).  It's entirely possible there's some limiting factor I haven't thought of.  I'll have a chance to take a look at it this evening.

rmandelb: Thanks!  I really think it does make sense to do this, and I suspect I'm just missing some detail.

jmeyers314: Hey Rachel,
I pushed a commit that passes your unit tests (at least on my laptop).  I'm not 100% certain what was causing the problem, but I think it had something to do with `ChromaticConvolution.drawImage` creating copies of the things it was convolving and also creating entirely new objects as part if its strategy to fast track any separable components of the convolution.  I'm guessing these temporary objects did have their interpolation destroyed, but the originals did not.

So instead I tried destroying the interpolation when the transformation is first applied instead of waiting for drawImage to get called.  Seems to work!

-Josh

rmandelb: Thanks, that makes a lot of sense.

I just further complexified the unit test to make sure we were really catching any possible errors.  It now makes a ChromaticAtmosphere, transforms it (chromatically), then makes a copy that has interpolation, then applies chromatic transformations to both the original and the copy.  The only way to get this right is if it really is combining the first and second set of transformations when destroying the interpolation.

rmandelb:
@jmeyers314 , thanks for all the comments (and the fix for my first attempt at tracking transformations through the interpolation process). I think I am done addressing them all.

@rmjarvis , I think this is now ready for you to go over whenever it works for you.

galsim/chromatic.py
jmeyers314: +1

jmeyers314:
Hey Rachel,
All looks good to me.

I wanted to let you know that I just committed a small bugfix rider to this PR.  While I was playing with the `noInterpolation` code I noticed that `drawImage(bandpass, nx=blah, ny=blah)` was raising errors.  I think this was due to not stripping kwargs of these parameters after they were used (so this is my bug, not yours).  I fixed this and slightly modified one of your unit tests so that this functionality gets tested.  Hope you don't mind that I placed this here.

-Josh

rmandelb:
I don't mind at all!  Thanks for the fix.

CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ New Features
   inter-pixel capacitance (#555) and image quantization (#558).
 - Added information about PSF size and shape to the data structure that is
   returned by EstimateShear(). (#612)
+- Added an option to the ChromaticObject class that allows for image rendering
+  via interpolation between stored images.  This option can speed up the image
+  rendering process compared to brute force evaluation for chromatic objects
+  with basic properties that are wavelength-dependent.  New
+  `ChromaticOpticalPSF` class, which allow the diffraction limit and aberrations
```
rmjarvis: Should split this into 2 items I think.

galsim/chromatic.py
```diff
@@ -131,6 +136,216 @@ def __init__(self, obj):
...
+        factor of 10-50.
+
+        Achromatic transformations can be applied either before or after setting up interpolation,
+        with the best option depending on the application.  For example, when rendering many times
+        with the same achromatic transformation applied, it is typically advantageous to apply the
+        transformation before setting up the interpolation.  But there is no value in this when
+        applying different achromatic transformation to each object.  Chromatic transformations must
+        be applied before setting up interpolation, and attempts to render images of
+        ChromaticObjects with interpolation followed by a chromatic transformation will result in an
+        exception being raised.
```
rmjarvis: Is this still the case?  I thought we decided that a chromatic transformation of something with the interpolation set up was valid, but would emit a warning about losing the interpolation.

galsim/chromatic.py
```diff
@@ -131,6 +136,216 @@ def __init__(self, obj):
...
+
+        Because of the clever way that the ChromaticConvolution routine works, convolutions of
+        non-separable chromatic objects that use interpolation with separable objects will still
```
rmjarvis: This sentence is a bit hard to parse.  When I first read it, I connected "with separable objects" to interpolation rather than to convolution.  Recommend reversing the order: "convolution of separable chromatic objects with non-separable ones that use interpolation will still..."

galsim/chromatic.py
```diff
@@ -131,6 +136,216 @@ def __init__(self, obj):
...
+        benefit from these optimizations.  For example, a non-separable chromatic PSF that uses
+        interpolation, when convolved with a sum of two galaxy components each with their own SED
+        will be able to take advantage of this optimization.  In contrast, when convolving two
+        non-separable profiles there is no way to take advantage of the interpolation optimization,
+        so even in the case that interpolation has been set up, it will be ignored and the full
+        calculation will be done.
```
rmjarvis: Perhaps point out that you can wait until you have created the convolution of two non-separable things and _then_ call `setupInterpolation` on that.  This would then interpolate the full convolved profile, which could be helpful if you plan to draw that same profile multiple times.  e.g. if you convolve a ChromaticOpticalPSF with a ChromaticAtmosphericPSF, you can do `setupInterpolation` on that convolution, and then convolve this by many different separable chromatic galaxies and still gain from the interpolation.

galsim/chromatic.py
```diff
@@ -131,6 +136,216 @@ def __init__(self, obj):
...
+
+        For use cases requiring a high level of precision, we recommend a comparison between the
+        interpolated and the more accurate calculation for at least one case, to ensure that the
+        required precision has been reached.  Also note that after calling `setupInterpolation`, it
+        is possible to revert to the exact calculation by calling `noInterpolation`.
```
rmjarvis: I'm not so fond of the name `noInterpolation`.  Functions that do something should generally start with a verb.  Perhaps `removeInterpolation`?

galsim/chromatic.py
```diff
@@ -131,6 +136,216 @@ def __init__(self, obj):
...
+        the parameters of the object depend on wavelength.
+
+        @param waves            The list, tuple, or NumPy array of wavelengths to be used when
+                                building up the grid of images for interpolation.  The wavelengths
+                                should be given in nanometers, and they should span the full range
+                                of wavelengths covered by any bandpass to be used for drawing Images
+                                (i.e., this class will not extrapolate beyond the given range of
+                                wavelengths).  They can be spaced any way the user likes, not
+                                necessarily linearly, though interpolation will be linear in
+                                wavelength.
```
rmjarvis: Recommend adding "... between the specified wavelengths." at the end of this sentence.

galsim/chromatic.py
```diff
@@ -131,6 +136,216 @@ def __init__(self, obj):
...
+                                `oversample_fac`>1 results in higher accuracy but costlier
+                                pre-computations (more memory and time). [default: 1]
+        """
+        # Set up the interpolation (which can be a costly step, depending on the length of
+        # `waves`).
+        self.waves = np.sort(np.array(waves))
+
+        # Make the objects between which we are going to interpolate.  Note that these do not have
+        # to be saved for later, unlike the images.
+        objs = [ self.evaluateAtWavelength(wave) for wave in waves ]
```
rmjarvis: I think here and later where you use waves, you need to use self.waves instead, since self.waves is sorted.  This probably also means that you are missing some unit tests that have waves not already in order.

galsim/chromatic.py
```diff
@@ -131,6 +136,216 @@ def __init__(self, obj):
...
+            self._fluxFactor = lambda w: 1.0
+
+        # Check the fluxes for the objects.  If they are unity (within some tolerance) then that
+        # makes things simple.  If they are not, however, then we have to reset them to unity, and
+        # modify the SED attribute, which now refers to the total flux at a given wavelength after
+        # integrating over the whole light profile.
+        fluxes = np.array([ obj.getFlux() for obj in objs ])
+        if np.any(abs(fluxes - 1.0) > 1000.*np.finfo(fluxes.dtype.type).eps):
+            # Figure out the rescaling factor for the SED.
+            rescale_fac = np.zeros_like(self.waves)
```
rmjarvis: This seems to be a relic from an old implementation of this.  You don't seem to use `rescale_fac` after you build it.  This means the for loop can be more simply written:
```
for obj in objs: obj.setFlux(1.0)
```

rmjarvis: Oh, I see, you do you `rescale_fac` in the SED line.  But there, you can just use `fluxes` instead, rather than copy over `fluxes` to `rescale_fac` here.

galsim/chromatic.py
```diff
@@ -131,6 +136,216 @@ def __init__(self, obj):
...
+                objs[ind].setFlux(1.0)
+            if not hasattr(self, 'SED'):
+                self.SED = lambda w : 1.0
+            self.SED = galsim.LookupTable(x=self.waves, f=self.SED(self.waves)*rescale_fac,
+                                          interpolant='linear')
+
+        # Find the Nyquist scale for each, and to be safe, choose the minimum value to use for the
+        # array of images that is being stored.
+        nyquist_dx_vals = [ obj.nyquistScale() for obj in objs ]
+        dx = min(nyquist_dx_vals) / oversample_fac
```
rmjarvis: Minor point, since it is an internal variable, but we usually use `scale` now rather than `dx`.

galsim/chromatic.py
```diff
@@ -131,6 +136,216 @@ def __init__(self, obj):
...
+
+        # Find the suggested image size for each object given the choice of scale, and use the
+        # maximum just to be safe.
+        possible_im_sizes = [ obj.SBProfile.getGoodImageSize(dx, 1.0) for obj in objs ]
+        n_im = max(possible_im_sizes)
```
rmjarvis: `n_im` sounds like number of images.  Recommend `im_size`, which reads more clearly as being the max of all the possible image sizes here.

galsim/chromatic.py
```diff
@@ -131,6 +136,216 @@ def __init__(self, obj):
...
+
+        # Make the objects between which we are going to interpolate.  Note that these do not have
+        # to be saved for later, unlike the images.
+        objs = [ self.evaluateAtWavelength(wave) for wave in waves ]
+
+        # The evaluation step (above) led to the incorporation of any shears and other
+        # transformations into the stored models between which we interpolation, so reset the
+        # internal attributes that store information about transformations, if indeed there were any
+        # transformations.
+        if hasattr(self, '_A') and not all ([self._nullTransformation(w) for w in waves]):
```
rmjarvis: What if all the transformations are null, but the flux factor is not null?  I noticed later, you do a check for the latter in `noInterpolation`, so I would have thought there should be a check here as well.

galsim/chromatic.py
```diff
@@ -131,6 +136,216 @@ def __init__(self, obj):
...
+        return hasattr(self, 'waves')
+
+    def _evaluateAtWavelength(self, wave):
+        """
+        Evaluate this ChromaticObject at a particular wavelength, either using interpolation or via
+        direct calculation, depending on whether or not interpolation has been set up.
+
+        @param wave     Wavelength in nanometers.
+
+        @returns the monochromatic object at the given wavelength, as a GSObject.
```
rmjarvis: It's not really wrong to have the doc string here, but it's kind of unnecessary.  Python won't display doc strings for private methods (leading underscore) when you say `help(ChromaticObject)` for instance.  It's generally sufficient to just use an in-code comment (hash comment) for private methods and functions.  And in that case, you don't need to include all the @param stuff when you are just documenting things for other developers.

galsim/chromatic.py
```diff
@@ -131,6 +136,216 @@ def __init__(self, obj):
...
+        the originally-stored images.  Also returns values for step_k and max_k, to be used to
+        expedite the instantation of InterpolatedImages.
+
+        @param wave     Wavelength in nanometers.
+
+        @returns an Image of the object at the given wavelength.
+        """
+        # First, some wavelength-related sanity checks.
+        if not hasattr(self, 'waves'):
+            raise RuntimeError("Requested image at some wavelength without setting up "
```
rmjarvis: Shouldn't have these checks in private methods.  Checks like this should always be in the top-level public method, not in private methods, since if the code is written correctly, they should never hit.

Sometimes it is useful to put them in when developing code to help catch mistakes, but at this point, they should be removed as they just slow down the code.

galsim/chromatic.py
```diff
@@ -140,9 +355,14 @@ def drawImage(self, bandpass, image=None, integrator=None, **kwargs):
 
         The task of drawImage() in a chromatic context is to integrate a chromatic surface
         brightness profile multiplied by the throughput of `bandpass`, over the wavelength interval
-        indicated by `bandpass`.
-
-        Several integrators are available in galsim.integ to do this integration.  By default,
+        indicated by `bandpass`.  This integration will take place either via brute-force drawing
+        the image at each wavelength, or via an optimized version of the routine that uses some
```
rmjarvis: "each wavelength" is a bit of a non sequitur at this point in the doc.  You haven't talked about any list of wavelengths yet.  Also, could probably pare this down a bit and just reference `setupInterpolation` for more details.

galsim/chromatic.py
```diff
@@ -164,10 +384,24 @@ def drawImage(self, bandpass, image=None, integrator=None, **kwargs):
                                 for details.)  [default: None]
         @param integrator       One of the image integrators from galsim.integ [default: None,
                                 which will try to select an appropriate integrator automatically.]
+                                This keyword is ignored in the case that optimized interpolation
+                                between precomputed images is to be used to render the image.
```
rmjarvis: Simpler: This keyword is ignored if `setupInterpolation` has been called for this object.

galsim/chromatic.py
```diff
@@ -227,6 +460,152 @@ def drawImage(self, bandpass, image=None, integrator=None, **kwargs):
...
 
+    def _interp_drawImage(self, bandpass, image=None, integrator=None, **kwargs):
+        """Draw method adapted to work for ChromaticImage instances for which interpolation between
+        stored images is being used.  Users should not call this routine directly, and should
+        instead interact with the `drawImage` method.
+        """
+        # Check whether the object has had some chromatic transformation applied to it.  Since we
+        # are interpolating between stored images, this is not a situation that _interp_drawImage
+        # can handle.  We have to lose the interpolation.
+        if hasattr(self, '_A') and self._chromaticTransformation(bandpass):
```
rmjarvis: Are chromatic flux scalings ok here?

galsim/chromatic.py
```diff
@@ -227,6 +460,152 @@ def drawImage(self, bandpass, image=None, integrator=None, **kwargs):
...
+        #     wavelength.
+        # (2) We use the midpoint rule for integration.
+
+        # Figure out the dwave for each of the wavelengths in the bandpass.
+        dw = [bandpass.wave_list[1]-bandpass.wave_list[0]]
+        dw.extend(0.5*(bandpass.wave_list[2:]-bandpass.wave_list[0:-2]))
+        dw.append(bandpass.wave_list[-1]-bandpass.wave_list[-2])
+        # Set up arrays to accumulate the weights for each of the stored images.
+        weight_fac = np.zeros(len(self.waves))
+        for idx, w in enumerate(bandpass.wave_list):
```
rmjarvis: I don't think this will work very well if the band pass is a simple rectangle approximation.  In the normal drawImage, Josh does a union of the bandpass wave_list and the object's wave_list.  I think you want to do the same here with self.waves.

galsim/chromatic.py
```diff
@@ -227,6 +460,152 @@ def drawImage(self, bandpass, image=None, integrator=None, **kwargs):
...
+        wave_list = self._getCombinedWaveList(bandpass)
+
+        # merge self.wave_list into bandpass.wave_list since we are using a sampling integrator
+        bandpass = galsim.Bandpass(galsim.LookupTable(wave_list, bandpass(wave_list),
+                                                      interpolant='linear'))
+
+        # The integration is carried out using the following two basic approaches:
+        # (1) We use linear interpolation between the stored images to get an image at a given
+        #     wavelength.
+        # (2) We use the midpoint rule for integration.
```
rmjarvis: Why midpoint, not trapezoid?  I don't see any reason that trapezoid wouldn't work, but perhaps I'm missing something.

galsim/chromatic.py
rmjarvis: Or, if trapezoid is possible for the interpolated version, perhaps we can change this keyword to be a string, much like how we allow `x_interp` and `k_interp` to be strings rather than actual `Interpolation` objects.  

Then the user could more simply write `integrator='midpoint'` or `integrator='trapezoid'`, which would create the appropriate thing from galsim.integ in the normal case or use the appropriate weights in the interpolation case.  That seems a bit more user-friendly.

galsim/chromatic.py
rmjarvis: Oh, I see.  You did do this with the call to `_getCombinedWaveList` above.  But why not just use that returned `wave_list` here.  Why did you create a new `Bandpass` object?  I think you could have just used `wave_list` here and kept the original `bandpass` in the below call of `bandpass(w)`.  That seems like a more straightforward implementation.  Is there something I'm missing?

galsim/chromatic.py
```diff
@@ -227,6 +460,152 @@ def drawImage(self, bandpass, image=None, integrator=None, **kwargs):
...
+        @param wave     Wavelength in nanometers.
+        @returns True/False (True if the transformation corresponds to no shear, shift dilation,
+        or flux rescaling)
+        """
+        null_A = np.matrix(np.identity(3), dtype=float)
+        A0 = self._A(wave)
+        f0 = self._fluxFactor(wave)
+        # Check whether any transformation was done.
+        return not (np.any(abs(A0-null_A) > 1000.*np.finfo(A0.dtype.type).eps) or \
+                        abs(f0-1.) > 1000.*np.finfo(A0.dtype.type).eps)
```
rmjarvis: 1000 is actually a pretty huge slop here.  10 would be plenty.  (Even 1 would probably work.)  It's not such a big deal if dtype is float, but if it is float32, then eps is only 1.e-7, so 1000 * eps is 1.e-4 which is starting to get into the realm of things the user might actually care about.

galsim/chromatic.py
```diff
@@ -367,6 +746,8 @@ def copy(self):
 
     # Helper function
     def _applyMatrix(self, J):
+        if hasattr(J, '__call__'):
+            self.noInterpolation()
```
rmjarvis: Doesn't this call mean that the check in `_intep_drawImage` won't ever trigger?  If so, I think this is where the warning should be emitted.

Might also need a similar thing when applying a chromatic flux scaling.

galsim/chromatic.py
```diff
@@ -1045,6 +1433,23 @@ def __init__(self, *args, **kwargs):
         else:
             self.separable = False
 
+        # Check quickly whether we are convolving two non-separable things that aren't
+        # ChromaticSums, >1 of each uses interpolation.  If so, emit a warning that the
```
rmjarvis: each -> which

galsim/chromatic.py
```diff
@@ -1359,3 +1770,76 @@ def withScaledFlux(self, flux_ratio):
...
+    the wavelength/diameter factor.  Second, aberrations such as defocus, coma, etc. are typically
+    defined in physical distances, but their impact on the PSF depends on their size in units of
+    wavelength.  Other aspects of the optical PSF do not require explicit specification of their
+    chromaticity, e.g., once the obscuration and struts are specified in units of the aperture
+    diameter, their chromatic dependence gets taken care of automatically.
+
+    When using interpolation to speed up image rendering (see ChromaticObject.setupInterpolation()
+    method for details), including ~10-15 samples across any given bandpass should be sufficient.
+    For moderate accuracy, 5 is often sufficient.  Likewise, for high accuracy, `oversample_fac`
+    should be in the range 1.5-2, whereas for moderate accuracy, 1 is often sufficient.  All of
```
rmjarvis: "high" and "moderate" here are pretty squishy terms.  Is it possible to be more quantitative?  Or is that still an open research topic?

galsim/chromatic.py
```diff
@@ -1359,3 +1770,76 @@ def withScaledFlux(self, flux_ratio):
...
+    these statements assume that aberrations are not very large (typically <~0.25 waves, which is
+    commonly satisfied by space telescopes); if they are larger than that, then more stringent
+    settings are required.
+
+    Also, note that in order to render an image of a real physical object, the ChromaticOpticalPSF
+    should be convolved with some object that has an SED.  Drawing without doing so corresponds to
+    using a flat SED.
```
rmjarvis: This paragraph seems a bit odd.  We don't make the same comment in any other Chromatic class, so why do so here?

galsim/chromatic.py
```diff
@@ -1359,3 +1770,76 @@ def withScaledFlux(self, flux_ratio):
...
+
+    @param   diam          Telescope diameter in meters.
+    @param   aberrations   An array of aberrations, in nanometers.  The size and format of this
+                           array is described in the OpticalPSF docstring.
```
rmjarvis: I'm a bit uncomfortable with this.  The `aberrations` parameter for `OpticalPSF` is given in units of waves.  I think it might be a bit less of a burden on the user to keep the same definition here, but have the user also give a fiducial wavelength at which the aberrations are defined.

In fact, we could add `lambda` and `diam` to `OpticalPSF` (and maybe even deprecate `lam_over_diam` which is a bit unwieldy, imo), in which case we could have the two classes take exactly the same arguments.  

Of course this one would have a proper wavelength dependence given those parameters, and the other would only be correct at a single wavelength.  But I think making the arguments the same would improve the API.

galsim/integ.py
```diff
@@ -154,3 +154,54 @@ def calculateWaves(self, bandpass):
...
+        waves = self.calculateWaves(bandpass)
+        self.last_n_eval = len(waves)
+        for w in waves:
+            ret_val, tmp_stepk, tmp_maxk = imageAtWavelength(w)
+            images.append(bandpass(w)*ret_val)
+            stepk.append(tmp_stepk)
+            maxk.append(tmp_maxk)
+        return self.rule(images, waves), min(stepk), max(maxk)
+
+class DirectSampleIntegrator(DirectImageIntegrator):
```
rmjarvis: You don't seem to use this.  So I'm guessing this is a relic of a past implementation.  Is it worth keeping?

tests/test_chromatic.py
```diff
@@ -490,6 +513,22 @@ def test_chromatic_flux():
...
+    final_int.draw(bandpass, image=image3)
+    int_flux = image3.array.sum()
+    # Be *slight* less stringent in this test given that we did use interpolation.
+    printval(image, image3)
+    np.testing.assert_almost_equal(
+        int_flux/analytic_flux, 1.0, 3,
+        err_msg="Drawn ChromaticConvolve flux (interpolated) doesn't match analytic prediction")
+    # Go back to no interpolation (this will effect the PSFs that are used below).
+    PSF = galsim.ChromaticAtmosphere(mono_PSF, base_wavelength=500,
+                                     zenith_angle=zenith_angle)
```
rmjarvis: Couldn't you just call PSF.noInterpolation here?

tests/test_chromatic.py
```diff
@@ -64,6 +66,27 @@
...
 disk_SED = disk_SED.withFluxDensity(target_flux_density=0.3, wavelength=500.0)
 
+# define the directory containing some reference images
+refdir = os.path.join(".", "chromatic_reference_images") # Directory containing the reference
+
+# Define an ChromaticObject subclass for which interpolation could be helpful, but calculations can
+# be done quickly:
+# It's a Gaussian with a wavelength-dependent sigma [sigma = sigma_0 (wave / 500 nm)]
+# and shear g1 [g1 = 0.1 ((wave)/(500 nm) - 1)]
+class ChromaticGaussian(galsim.ChromaticObject):
```
rmjarvis: This would be better defined within test_interpolated_ChromaticObject where it is used, rather than here at the top of the file.

galsim/chromatic.py
jmeyers314: >In fact, we could add lambda and diam to OpticalPSF (and maybe even deprecate lam_over_diam which is a bit unwieldy, imo)

I always thought the `lam_over_diam` was a bit weird too, since the natural thing to do is to divide the wavelength by the diameter in consistent length units.  This usually does the wrong thing, of course, since (I assume) most users specify `scale` in arcseconds, or maybe pixels, but not radians.  Having separate `lam` (can't use `lambda`, that's a python reserved word) and `diam` keywords probably makes this even more confusing though, since you would need to specify the wavelength in something like "meters arcseconds" and the diameter in "meters".  We could add a third keyword like `radians_in_one_scale_unit`, I suppose, but then we're starting to make the API unwieldy again.  I like the idea of a unified API, but I'm not sure if it's worth replacing one required argument with three for the non-chromatic `OpticalPSF`.

rmjarvis: I was thinking `diam` in meters, `lam` in nm (good point about `lambda`), and then let the code convert the ratio to arcsec for you.  Both numbers are easy for people to get right for a particular optical system, whereas lam_over_diam requires some significant documentation to explain how to calculate it.

The only semi-downside of this option is that is forces people into using arcsec for their world coordinate units if they are using OpticalPSF, but I think that's fine.  It's what we normally use anyway, and it's what Rachel implicitly did here for ChromaticOpticalPSF, which I also think was a perfectly reasonable thing to do.

rmjarvis: Oh and speaking of which, should add that fact about the scale having to be given in arcsec per pixel to the doc string here.

Another idea -- if you want to allow other options for the scale units than just arsec, we could add in an optional kwarg `scale_units` which would default to `galsim.arcsec`.  Probably the only other reasonable alternative would be radians, but at least it would give the user an option.  (They could even define their own AngleUnit if they wanted to use something else.)

jmeyers314: OK.  I like the idea of declaring arcseconds as the de facto standard for `scale` in both OpticalPSF and ChromaticOpticalPSF, with a `scale_units` keyword to optionally override this.

CHANGELOG.md
rmandelb: Agreed.  I just pushed a fix.

galsim/chromatic.py
rmandelb: Docstring was out of date.  Good catch, just pushed a fix.

galsim/chromatic.py
rmandelb: Fixed.

galsim/chromatic.py
rmandelb: I�m fine with that change; done.

galsim/chromatic.py
rmandelb: done

galsim/chromatic.py
rmandelb: Done.

galsim/chromatic.py
rmandelb: Done.

galsim/chromatic.py
rmandelb: Good point, I hadn�t thought of float32.  Fixed.

galsim/chromatic.py
rmandelb: Good point.  Done.

galsim/chromatic.py
rmandelb: Good point, but since it's not actually wrong, I think I will just leave it.  In future I'll avoid doing it this way.

galsim/chromatic.py
rmandelb: Oops.  Fixed.

galsim/chromatic.py
rmandelb: Good point, done.

galsim/chromatic.py
rmandelb: Fixed.

galsim/integ.py
rmandelb: Yes, it's a relic.  I removed it; we can always restore it later if we decide we need it for something.

galsim/chromatic.py
rmandelb: If you look at what ```_nullTransformation``` does, it actually checks both ```_A``` and ```_fluxFactor``` for consistency with a null transformation.  So I think this piece of code is actually okay.

galsim/chromatic.py
rmandelb: I simplified this a bit in https://github.com/GalSim-developers/GalSim/commit/d63eea2637b6a74f2d75048da48f7485d549d400, but please see what you think.

galsim/chromatic.py
rmandelb: It really depends on what you're trying to do.  I gave some numbers based on reproduction of second moments, but someone who really cares about accuracy needs to test this themselves.  I clarified this in https://github.com/GalSim-developers/GalSim/commit/159a692cbc763d1c3eef92c2af52c9922ea5ca0f

galsim/chromatic.py
rmandelb: I think the reason ```lam_over_diam``` was being used, instead of a separate ```lam``` and ```siam``` (my browser keeps auto-correcting this - sorry!), was that this is what determines the diffraction limit.   Given that there are ways of changing both ```lam``` and ```siam``` in ways that preserve the appearance of the PSF, I kind of like the simplicity of having just ```lam_over_diam```.

I'd be more inclined to say that OpticalPSF should stay as-is, and ChromaticOpticalPSF should take ```lam_over_diam``` and ```aberrations``` like OpticalPSF, along with a fiducial wavelength at which the ```lam_over_diam``` and ```aberrations``` are defined.  Sound okay?

This also allows the user to choose whatever arbitrary units they like, I think.

galsim/chromatic.py
rmandelb: Well, most chromatic things have already been explicitly defined as having an SED.  For example, if we define a galaxy using ```gal = galsim.DeVaucouleurs(...)*bulge_sed``` and then convolve it with some PSF, then it explicitly has an SED.  I think ChromaticOpticalPSF and ChromaticAtmosphere are the main cases where one might make a chromatic object that ends up not having an SED, aren't they?

galsim/chromatic.py
rmjarvis: I think it's only simple if you already understand how Airy diffraction works.  I'm not sure everyone using GalSim will remember their physics classes that covered this.

I was looking into where we document how to calculate lam_over_diam and it's really only in Airy that there is a concrete annotated example.  (Also demo3 and demo9 do the calculation.)  So I think a similar explanation/example should be added to OpticalPSF and ChromaticOpticalPSF, since this is kind of a confusing thing.  

However, I've kind of grown fond of the idea of letting GalSim do the `lam * 1.e-9 / diam * 206265` calculation for the user given `lam` and `diam` in their natural units (nm and m).  So could we add that as an option in all these places, but keep the `lam_over_diam` option as well for people who want to use different units or something?  Then in ChromaticOpticalPSF, this would also serve as the fiducial wavelength for the `aberrations`.

galsim/chromatic.py
rmjarvis: Now I'm even more confused about what your point is in this paragraph.  I thought you were using SED loosely in the sense of having some kind of chromatic dependence, not that there was an `SED` attribute.

I thought you were just making the point that if you convolve a chromatic PSF with a non-chromatic galaxy, then the galaxy implicitly has a flat SED.  The same way that if you convolve a `ChromaticAtmosphere` with `Sersic` then the galaxy implicitly has a flat SED.  Likewise if you convolve a `Kolmogorov` with `Sersic * sed` then the PSF implicitly has a flat SED.

So this paragraph seemed like stuff that users of the various chromatic classes should already understand.  But if you think it's a point worth emphasizing, I think this paragraph should be moved to the main `ChromaticObject` doc string.

Unless you are making some other point that I'm not getting right now.  Is there some reason that `ChromaticOpticalPSF` convolved with `Sersic` behaves qualitatively differently than `Kolmogorov * sed` convolved with `Sersic`?  I would think in both cases, you get the right effective PSF over the provided bandpass convolved with a flat-spectrum galaxy.

rmandelb: The point I was making that if you do
```
foo = galsim.ChromaticAtmosphere(...)
foo.drawImage(...)
```
then your image doesn't correspond to anything with an astrophysical SED, it corresponds to something with a totally flat SED.  (Likewise for if you did this with ```ChromaticOpticalPSF```.)   In other words, let's say you have an object with some SED, that you convolve with the ChromaticAtmosphere.  Its effective PSF image cannot be obtained with the above code.

You think this is not worth mentioning for ChromaticAtmosphere / ChromaticOpticalPSF?  While ideally users would have understood this, I think it doesn't hurt to mention.

galsim/chromatic.py
rmandelb: I agree that it would be good to document this better in OpticalPSF and ChromaticOpticalPSF.

I don't want to break old functionality, but I think your compromise is okay.  For OpticalPSF, we can allow either ```lam_over_diam``` or separate diameter and wavelength specification.  For ChromaticOpticalPSF we require both diameter and wavelength, or ```lam_over_diam``` and wavelength, and tell the user to enter aberrations in units of waves at the specified wavelength.   Will do this in a bit.

galsim/chromatic.py
rmjarvis: OK, I think I might see your point here.  You are pointing out that this is not the correct SB profile of a _star_, since star's have an SED that is not included here.  

I think the point you are making is already obvious with respect to galaxies -- that if you use a regular GSObject, then it corresponds to a flat SED.  But maybe it's not as obvious that stars also have SEDs that are not included here in the PSF object.  So here is a proposed paragraph that maybe makes this point more clearly:

Note that a ChromaticOpticalPSF by itself is NOT the correct thing to use to draw an image of a
star.  Stars (and galaxies too, of course) have an SED that is not flat.  To draw a real star, you
should either multiply the ChromaticOpticalPSF object by an SED, or convolve it by a point
source (typically approximated by a very tiny Gaussian) multiplied by an SED:

    >>> psf = galsim.ChromaticOpticalPSF(...)
    >>> star = galsim.Gaussian(sigma = 1.e-6) * psf_sed
    >>> final_star = galsim.Convolve( [psf, star] )
    >>> final_star.drawImage(bandpass = bp, ...)

rmandelb: > I think the point you are making is already obvious with respect to galaxies -- that if you use a 
> regular GSObject, then it corresponds to a flat SED.

Yes, that's fine - if someone doesn't care about chromaticity they can just use GSObjects happily.  My concern was for someone who thought they were properly including chromaticity and making realistic images with the code snippet I gave above (which is wrong).

I am happy to use something like your paragraph + example.  Will swap that in.

galsim/chromatic.py
rmandelb: Well, that's a good point.

I had been thinking that no chromatic transformations are okay, but technically we could accommodate a chromatic flux scaling by just incorporating it as pre-factors multiplying the stored images.  Do you think that makes sense?  If so, then some code refactoring will be necessary to allow this.

galsim/chromatic.py
rmandelb: Fixed this.

galsim/chromatic.py
rmandelb: Laziness.  Midpoint was super easy to code up.  I can set up trapezoidal too, I suppose, but I wasn't sure how necessary it is in this context.

galsim/chromatic.py
rmandelb: You are right, I moved the warning here.
Re: chromatic flux scaling, it depends on what we decide makes sense (see my earlier question).

galsim/chromatic.py
rmjarvis: I was making the opposite point actually, since I didn't notice that `_chromaticTransformation` also checked for a chromatic flux scaling.  So I thought you were neglecting to remove the interpolation in that case, which would be an error.

Since the code you have is correct, I don't see a huge need to make the optimization you mentioned.  The user should normally be able to optimize this themselves by waiting until after they apply their chromatic flux scaling before calling `setupInterpolation`.  And the warning will tell them they ought to.

galsim/chromatic.py
rmjarvis: I think it's probably worthwhile.  Trapezoid really is quite a bit more accurate than midpoint for a given number of samples.  And since the point of this is to be able to cut down the number of samples for the integration, it seems worth it to use trapezoid, where that number can generally be made even smaller.

galsim/chromatic.py
rmandelb: Okay, sounds good.

galsim/chromatic.py
rmandelb: I just made a commit that is meant to address all the concerns raised here:
https://github.com/GalSim-developers/GalSim/commit/b4c8f1f8c35a4fbabdf977c9a0f2e189ea17dbe2
Please take a look and see what you think.

It includes updates to OpticalPSF, ChromaticOpticalPSF, documentation, demo3, and unit tests.  I also made the OpticalPSF updates available in config.  I did not update the config wiki; https://github.com/GalSim-developers/GalSim/wiki/Config-Documentation-v1.1 says it's for v1.1, but should that be v1.2?  Should I make a new page that is for v1.3 and update that one?

rmjarvis:
> I did not update the config wiki; https://github.com/GalSim-developers/GalSim/wiki/Config-Documentation-v1.1 says it's for v1.1, but should that be v1.2? Should I make a new page that is for v1.3 and update that one?

I fixed the link and made a new v1.3 page for you to edit.  (See the side bar that lists all the wiki pages.)

galsim/chromatic.py
rmandelb: Done in https://github.com/GalSim-developers/GalSim/commit/2efdd53a8e52fc41f12023a7d87d1eef57c0af76

But I was being a bit lazy in calling what I was doing midpoint integration; it's now defined a bit better in the comments.  And both methods are now tested.

rmandelb:
Thanks!  I just edited it (and removed "still under development" from the v1.2 page).

I think I have now addressed all your suggestions, though in some cases the changes are not trivial so you should probably go check them out.  I commented on each of those places so they should be easy to find.

galsim/chromatic.py
jmeyers314: Should we add similar keyword arguments to Airy and potentially Kolmogorov (lam_over_r0) too?

rmjarvis: I think Airy, yes.  Kolmogorov maybe not.

Kolmogorov profiles are probably more typically constructed with `fwhm` or `half_light_radius`, so I'm not sure how important it is to improve the `lam_over_r0` usage.

I also thought that we should probably have `ChromaticAiry` class for completeness, but I wasn't going to ask it to be done as part of this PR.  (Unless you feel so inclined...)

I wasn't sure about `ChromaticKolmogorov`.  It's already basically part of the `ChromaticAtmosphere` class.  So we pretty much already have it by another name.

rmandelb: Airy: yes, I can easily add a ```lam``` and ```siam``` option that have similar usage as for the ChromaticOpticalPSF.  That's easy.

ChromaticAiry: Now that I've done ChromaticOpticalPSF, this is so easy that I may as well do it too.

examples/demo3.py
```diff
@@ -169,11 +169,17 @@ def main(argv):
...
+    # calculation:
     lam_over_diam = lam * 1.e-9 / tel_diam # radians
     lam_over_diam *= 206265  # arcsec
+    # Note that we could also have made GalSim do the conversion for us if we did not know the right
+    # factor:
+    # lam_over_diam = lam * 1.e-9 / tel_diam * galsim.radians
+    # lam_over_diam = lam_over_diam / galsim.arcsec
     logger.debug('Calculated lambda over diam = %f arcsec', lam_over_diam)
     # The rest of the values should be given in units of the wavelength of the incident light.
     optics = galsim.OpticalPSF(lam_over_diam,
```
rmjarvis: Perhaps this ought to be changed to use the new lam, diam notation.  Either this or demo9, but one of them probably ought to showcase it.

galsim/optics.py
```diff
@@ -174,8 +200,11 @@ class OpticalPSF(GSObject):
     """
 
     # Initialization parameters of the object, with type information
-    _req_params = { "lam_over_diam" : float }
+    _req_params = { }
     _opt_params = {
+        "lam_over_diam" : float ,
+        "lam" : float ,
```
rmjarvis: This will work, but I think it may give slightly better error messages if `lam_over_diam` and `lam` are considered a "single" item.  One and only one of them is required.  

It's not quite as nice as the other examples of this because of the fact that `diam` has to accompany `lam`.  Also, your runtime checks here do much the same job as the config checks.  But since these parameters are conceptually closer to the meaning of `_single_params` than to `opt_params`, it seems worth moving them:
```
_single_params = [ { "lam_over_diam" : float , "lam" : float } ]
```

galsim/chromatic.py
```diff
@@ -162,12 +381,28 @@ def drawImage(self, bandpass, image=None, integrator=None, **kwargs):
                                 integrate.
         @param image            Optionally, the Image to draw onto.  (See GSObject.drawImage()
                                 for details.)  [default: None]
-        @param integrator       One of the image integrators from galsim.integ [default: None,
+        @param integrator       When doing the exact evaluation of the profile, this argument should
+                                be one of the image integrators from galsim.integ [default: None,
                                 which will try to select an appropriate integrator automatically.]
+                                If setupInterpolation() has been called for this object, then
+                                `integrator` should be a string, either 'midpoint' or
+                                'trapezoidal'. [default: 'trapezoidal']
```
rmjarvis: I think we should enable the string options for the normal integration method as well, so the user doesn't have to wade into the code of `integ.py` to see what the options are for this parameter.

rmandelb: But for the normal integration method, I think it�s more complicated than this.  Users have to choose one of the integration methods like sample integrator or continuous integrator, and  either the midpoint or trapezoidal method.  So I don�t see that using just a string for �midpoint� or �trapezoidal� will do the job.

examples/demo3.py
rmandelb: Since this is a bit simpler, I left demo3 as is, but changed demo9.

rmandelb:
okay, I just pushed a bunch of commits:
- https://github.com/GalSim-developers/GalSim/commit/92e5c6d3a0a21fa6d9e8711ec7c9f59471afbc46 includes the change to how OpticalPSF kwargs are treated by config
- https://github.com/GalSim-developers/GalSim/commit/4ee436618fc32fe07984631925d82871e08cfbc4 is the demo9 update to use the new OpticalPSF functionality
- https://github.com/GalSim-developers/GalSim/commit/8e9a6cbe106fe9f07afa80cab0450cc3ceeb4475 is the update to Airy
- https://github.com/GalSim-developers/GalSim/commit/7358ac5512214c7c14267993f18dc70db13003a3 includes a new ChromaticAiry class.

So other than my question on the specification of integration methods using strings that I posted a little while ago, I think that addresses all the comments I have received so far.

galsim/chromatic.py
rmjarvis: > But for the normal integration method, I think it's more complicated than this.  Users have to choose one of the integration methods like sample integrator or continuous integrator, and  either the midpoint or trapezoidal method. 

In practice, there is really just 'trapezoid' and 'midpoint' there too.  If the bandpass or sed has a wave_list, then it is pretty ill-advised to use the ContinuousIntegrator, which will rebin the wavelength range into 250 (or N) equally spaced partitions.  It's almost always better to use the actual wavelengths where the sed and/or bandpass are sampled.  And if there is no wave_list for either of them, then it's not possible to use SampleIntegrator.  So in both cases, the only real question is whether to use trapezoid or midpoint rule for the integration.  

But also, I'm not suggesting removing the existing option to provide an integrator object if the user wants.  I think that should still be an option for advanced users who want to do something a bit unusual.  I'm just suggesting that the default be switched from `integrator=None` to `integrator='trapezoidal'` and the code that checks the input `integrator` change to:
```
if isinstance(integrator, str):
    if integrator == 'trapezoidal':
        rule = np.trapz
    elif integrator == 'midpoint'
        rule = galsim.integ.midpt
    if len(wave_list) > 0:
        integrator = galsim.integ.SampleIntegrator(rule)
    else:
        integrator = galsim.integ.ContinuousIntegrator(rule)
```

rmandelb: Okay, I see your point.  Done: https://github.com/GalSim-developers/GalSim/commit/37056da85dd4de814ffff0a8317bf4446ea58fa4

rmjarvis:
Looks good to me.  Good to merge as far as I'm concerned.

rmandelb:
Thanks, Mike!  @jmeyers314 , there have been quite a few updates since you last looked at it.  Would you mind checking it over to see if you have any further comments?  I am fine with waiting a few more days to merge if you need some additional time.

galsim/chromatic.py
```diff
@@ -131,7 +136,216 @@ def __init__(self, obj):
...
+        transformation before setting up the interpolation.  But there is no value in this when
+        applying different achromatic transformation to each object.  Chromatic transformations
+        should be applied before setting up interpolation; attempts to render images of
+        ChromaticObjects with interpolation followed by a chromatic transformation will result in
+        the interpolation being unset and the full calculation being done.
+
+        Because of the clever way that the ChromaticConvolution routine works, convolutions of
+        separable chromatic objects with non-separable ones that use interpolation will still
+        benefit from these optimizations.  For example, a non-separable chromatic PSF that uses
+        interpolation, when convolved with a sum of two galaxy components each with their own SED
```
jmeyers314: this should probably say "... when convolved with a sum of two separable galaxy components ..."

galsim/optics.py
```diff
@@ -50,6 +50,15 @@ class OpticalPSF(GSObject):
...
     uses an InterpolatedImage to characterize the profile.
 
+    The diffraction effects are characterized by the diffraction angle, which is a function of the
+    ratio lambda / D, where lambda is the wavelength of the light and D is the diameter of the
+    telescope.  Users can specify this diffraction angle in one of two ways.  The first option is to
+    specify the ratio as `lam_over_diam` in whatever arbitrary units they choose (and then
+    self-consistently specify the image scale in the same units).  The second option is to specify
+    the wavelength as `lam` in units of nanometers and the telescope diameter as `diam` in units of
+    meters.  OpticalPSF will then convert the ratio to arcsec, and image scales must be specified in
+    arcsec as well.
```
jmeyers314: I think you mean convert the ratio to `scale_units` and image scales must be specified in `scale_units` as well.

jmeyers314:
Hi Rachel,
Just two tiny comments from me, and then I think this is good to merge!
-Josh

rmandelb:
Hi Josh - thanks, your comments were spot on, and I just pushed a fix.

I will leave this open until tomorrow morning, in case anyone else wishes to comment.  Then I will merge.

rmandelb:
Merging #618.

Issue #630, opened 2015-01-15
confusion between variance vs. std dev of noise


rmandelb:
I found a bug in UncorrelatedNoise or perhaps something related.  If you do this:
```
import galsim
im = galsim.Image(256,256)
ucn = galsim.UncorrelatedNoise(variance=0.1)
im.addNoise(ucn)
print im.array.std(), im.array.var()
```
then you'll get a standard deviation of 0.1, variance of 0.01.  But I had specified variance=0.1.  So something is getting labeled incorrectly somewhere in UncorrelatedNoise or the routines that it calls.

I will dig into this some more in a bit, just wanted to open the issue.

@emhuff, this isn't all of the problem you're seeing, but it's definitely part of it.

rmjarvis:
Wow.  Sorry. That's probably my mistake.  And I don't seem to have written the appropriate unit tests that would have caught this.

There is one test that would have caught it, but it uses var=1.0, which misses the missed sqrt.  Oops.  I'll fix this.

rmandelb:
Thank you!  I didn't have time to look into it yet, so if you have an idea where it could be happening, it would be great if you can fix it.

So... time for GalSim 1.2.1?  ;)

Issue #501, opened 2014-01-20
Update CorrelatedNoise to handle non-trivial WCS


rmjarvis:
One thing I didn't do on #364 with respect to implementing the new WCS functionality was to make sure CorrelatedNoise worked correctly when applying the noise to an image with something other than a pixel scale WCS.  Also when whitening an image.

It is possible that there is no real work to do here.  It might actually be working correctly.  Both functions use the draw command to realize the noise, which will work with the wcs in the image.  And I think the functions do store the provided wcs into the image correctly.  

However, I did not update the unit tests to make sure that it is all working correctly.  All the unit tests still just use a pixel scale for the WCS.  So mostly this issue is to update the unit tests to test these functions on images with non-trivial WCS's and see if there needs to be any updates to the CorrelatedNoise code.

Issue #589, opened 2014-07-20
Unit tests for non-square image noise generation in the correlated noise module


barnabytprowe:
I think that we support adding correlated noise with a given correlation function to non-square images, but don't have unit tests for this.  This issue will rectify this omission.

PR #634, opened 2015-01-23
#630 Fix error in UncorrelatedNoise variance calculation

rmjarvis:
The profile used for UncorrelatedNoise had a bug in it where the variance was the square of what it should have been.

In addition to fixing this bug, I also added a bunch of unit tests, including ones that involve non-trivial WCS transformations to make sure the profile is set up correctly even if the WCS is not a regular pixel scale.  

I applied these WCS tests to the regular CorrelatedNoise class as well, which required refactoring some of the handling of scale and wcs to correctly deal with non-square pixel options.  I think this addresses what we needed for Issue #501 as well.  I believe the wcs stuff is now working correctly for both CorrelatedNoise and UncorrelatedNoise.


galsim/correlatednoise.py
```diff
@@ -1491,17 +1525,21 @@ def __init__(self, variance, rng=None, scale=None, wcs=None, gsparams=None):
...
 
-        # Need variance == xvalue(0,0)
-        # Pixel has flux of f/scale^2, so use f = variance * scale^2
-        image_pix = galsim.Pixel(scale=1.0, flux=variance * wcs.pixelArea(), gsparams=gsparams)
-        world_pix = wcs.toWorld(image_pix)
-        cf_object = galsim.AutoConvolve(world_pix)
-        _BaseCorrelatedNoise.__init__(self, rng, cf_object)
+        # Need variance == xvalue(0,0) after autoconvolution
+        # So the Pixel needs to have an amplitude of sigma at (0,0)
+        import math
```
rmandelb: I have a question that is probably naive, but I'm curious:

What is the value of importing math and using it for this square root, instead of using the already-imported numpy.sqrt() method?  If you had not imported either of those modules at this point in the code, you would obviously import math since you don't need something that can take the square root of an array here.  Is it that in your experience it is faster to import and use the math.sqrt() method than it is to use numpy.sqrt()?  Or is there some other reason?

rmandelb:
Hi Mike - I just went over this whole thing, and it looks good to me!

Tests pass on both of my systems.

galsim/correlatednoise.py
rmjarvis: For non-arrays, `math.sqrt` is indeed faster, since the `numpy` function has to check if the input is an array or not, whereas the `math` version only works for scalar floats.  I just checked, and the difference is about a factor of 3.

But I'm sure the difference is negligible for uses such as this.  My import statement here is really just because I don't generally import things at the top of the .py file as much as Barney does, so I wasn't thinking that numpy might already be imported.  I generally just import what I need in each function and use it there.  Python is smart enough not to waste any time importing things again if it has already loaded a module into memory.  So there's not really any downside to putting the imports in the functions, and I find it a bit easier to keep track of them there.

rmjarvis:
Thanks Rachel.  @barnabytprowe, I know you're busy of late, but since this module was your baby, do you want to take a look at it as well?  No worries if not.  :)

barnabytprowe:
Hi @rmjarvis ... I can do, and indeed would be delighted to, but it might not be tonight.  Would by Tuesday be OK?

rmjarvis:
Sure.  No rush.  Thanks.

galsim/correlatednoise.py
rmandelb: Out of curiosity I just did a timing test, and found that not only is ```math.sqrt``` faster (which I thought obviously had to the case - though on my system it's a factor of 1.8-2), but importing math AND doing ```math.sqrt``` is faster than ```bumpy.sqrt``` on its own.  That was pretty non-obvious to me.

galsim/correlatednoise.py
```diff
@@ -729,27 +751,20 @@ def _get_update_rootps(self, shape, wcs):
...
-                newcf.wcs = wcs
-            # Then draw this correlation function into an array.  If this is not done at the same
-            # wcs as the original image from which the CF derives, even if the image is rotated,
-            # then this step requires interpolation and the newcf (used to generate the PS below) is
-            # thus approximate at some level
-            newcf = self.drawImage(newcf)
+            # Draw this correlation function into an array.  If this is not done at the same wcs as
+            # the original image from which the CF derives, even if the image is rotated, then this
+            # step requires interpolation and the newcf (used to generate the PS below) is thus
+            # approximate at some level
```
barnabytprowe: I made a slight edit here (5f8c3f2) just to the comment, but actually this got me thinking.  Need we make clear somehwere that this wcs functionality is not necessarily always something that people ought to use, even if they have images of noise that have a wcs?

The reason I pose the question is that typically (but not I guess *totally* exclusively) correlated noise derives from the focal plane detector array, and so a common use case would be:
* Get some images of correlated noise in your camera
* Build a CF
* Use this to generate simulations in GalSim with an approximately similar Gaussian noise pattern
For this, you wouldn't actually want to use the wcs either in the input or output images.  Where the wcs stuff is massively useful is of course when you want to whiten noise in `RealGalaxy` images that have a funky wcs relative to the COSMOS one... But that should be our problem more than the users.

Is this caveat utilitor, or should we mention it?

barnabytprowe: (comment updated in github)

barnabytprowe:
Actually I do have a thought, related to that above: how *does* this play with the output of `getCOSMOSNoise`?  For example, if you wanted to make COSMOS noise but in images that are rotated.  In reality the axis of symmetry in the noise correlation function should also rotate to bear the same relation to the x-y pixel grid (rather than caring about wcs).  Would the correlation function actually also rotate?

barnabytprowe:
Another thought: I suggest deleting the `calculateCovarianceMatrix` method entirely.  I never used it above once, it has a big fat TODO next to it that I won't ever get to and now needs to be gone further into:
```Python
# TODO: Allow this to take a JacobianWCS, rather than just a scale.
```
And in generally is simply enough derived from an image of the correlation function.

barnabytprowe:
(comment above edited on github, sorry)

galsim/correlatednoise.py
rmjarvis: I believe the way it is implemented, if you `addNoise` onto an image that has the same WCS as the one you used when you built the CF, then the effects of the WCS should cancel out.  I.e. it would be the same as if you set the WCS to be `PixelScale(1.)`.  

I _think_ my unit tests for uncorrelated noise at least verify that this is the case if the `CorrelatedNoise` object doesn't actually have any correlations.  I don't know quite how to structure a test to confirm that this is working in the more general case.  If you have ideas for a unit test, I'd be happy to add it.

In any case, the bit that I thought mattered more was the whitening option, where you don't typically have the same WCS in the input and output images.  Currently, we just use a pixel scale difference for RealGalaxy, but I could imagine caring about the slightly non-square pixels in the HST detector or the output image.  Again, if you have an idea for a unit test of this, please let me know.

rmjarvis:
> I suggest deleting the calculateCovarianceMatrix method entirely.

Fine with me.

tests/test_correlatednoise.py
```diff
@@ -36,14 +36,6 @@
...
 smallim_size_odd = 17 # odd-sized version of the above for odd/even relevant tests (e.g. draw)
 largeim_size = 12 * smallim_size # ditto, but when we need a larger image
-xlargeim_size =long(np.ceil(1.41421356 * largeim_size)) # sometimes, for precision tests, we 
-                                                        # need a very large image that will 
-                                                        # fit a large image within it, even if 
-                                                        # rotated
-
-# Decimals for comparison (one for fine detail, another for comparing stochastic quantities)
-decimal_approx = 2
-decimal_precise = 7
```
barnabytprowe: What was the motivation for taking these out of module scope and into the various tests as hard coded numbers?  I found it useful in the old days to be able to modify them, in one place only, to better check for breakages.  Having to search and replace for `2` or `7` seems like a bigger pain.

rmjarvis: Mostly (a) I don't like global variables in general, and (b) I don't really understand your use case of wanting to change the decimal value for lots of unit tests all at once.  Why did you find it useful to be able to do that?

Personally, I find it hard when reading a unit test to understand how precise each test really is when these variables are at module scope.  I can't see the definitions of these variables when I am looking at a test function, so these variables just obfuscate my understanding of how precise the test is.  The hard coded numbers in contrast are self-documenting.  

For instance, some of the times where you used `decimal_precise`, you had a comment saying that it was actually precise to machine precision.  In those cases, I switched to using decimal=12 and removed the comment.  That seems much clearer to me.  

In general, I feel like you should know how accurate a unit test is when you write it.  I generally start with decimal=5 or 6, and only drop it if I understand why the test should not be expected to pass at that level of precision.  Very often there is a good reason to only use decimal=2 or 3, but it is certainly test-specific, so I don't really see the point of having the decimal value for lots of tests set at module scope.  I'd rather have non-default decimal values be set for each test individually to the value appropriate for that specific test.

barnabytprowe: Well it's up to you, as I will not be active from now on, but that's not how I worked.  Some of these tests take a long time to run, and share similar properties, and I found it helpful to be able to ramp up accuracy across the board as much of the *real* output of these tests when building the code was plots, and these changed very informatively as accuracy was increased.  You don't work that way, fine, but you are taking an executive decision to move to your own preferred approach here and I wanted to understand why, and flag it.  I have no idea why global variables aren't a sensible idea in a test situation rather than buried deep in the body of the code.

rmandelb:
Anything else on this PR?  The discussion kind of trailed off a while ago.

rmjarvis:
I still wanted to try to address this comment by Barney:

> Actually I do have a thought, related to that above: how does this play with the output of getCOSMOSNoise? For example, if you wanted to make COSMOS noise but in images that are rotated. In reality the axis of symmetry in the noise correlation function should also rotate to bear the same relation to the x-y pixel grid (rather than caring about wcs). Would the correlation function actually also rotate?

I haven't had time yet, but I was planning to try to work up a unit test for this to at least try to understand what happens in this case and possibly add some documentation about it.

rmandelb:
Oh sorry, I missed that.

barnabytprowe:
As regards my comment as quoted by Mike above (and sorry for going quiet on the matter myself), I have thought a bit more.  If the Correlation function does rotate in the manner I mentioned, I think that for a lot of people this would not be the desired behaviour: in many applications when generating noise to add to final images the orientation of the noise correlations should stay fixed relative to the pixel grid of the instrument that causes them, rather than to celestial north.

*If* this happens (not sure if it does) then a simple `ignore_wcs=True` kwarg would fix it for these noise generation cases.  The under-the-hood stuff in the `RealGalaxy`-type use cases that really needs the WCS functionality for correct whitening etc. could simply set the `ignore_wcs` to true.  It's a bit of a shame to add an extra switch to the interface, but I think it might be hard to satisfy both legitimate uses without it...

rmjarvis:
I'm back to working on this branch now, and I have some results to report about how things work currently.  I pushed an initial version of a test, which doesn't actually have any asserts yet.  It just prints some values so I can see what is going on.  cf. a568dd2

The basic idea is that I take a cosmos noise object, shear it to exaggerate the correlations, and then use it to add noise to images with different WCS functions.  Then I use the CorrelatedNoise class to measure the resulting correlations in the image and compare to the original.

The different wcs functions I use are:
- "same" is the same as the original
- "big" has 2x larger pixels
- "small" has 2x smaller pixels
- "rot" has the same size pixels, but rotated by 90 degrees
- "raw" means I reset the wcs back to the original pixel scale before measuring the correlations.

Then I look at the correlations in the 3x3 grid around the origin in the resulting measured correlation function.  Here is the output:
```
                                orig    same     big    small    rot   big/raw sm/raw  rot/raw 
Covariance at (0.0, 0.0)    :   1.700   1.696   1.699   1.690   1.699   1.699   1.690   1.699 
Covariance at (0.03, 0.0)   :   0.570   0.567   0.873   0.564   0.567   0.047   1.061   0.569 
Covariance at (0.0, 0.03)   :   0.571   0.570   0.871   0.569   0.569   0.044   1.066   0.567 
Covariance at (0.03, -0.03) :   0.111   0.109   0.450   0.115   0.107   0.008   0.603   0.744 
Covariance at (0.03, 0.03)  :   0.744   0.745   0.501   0.742   0.744   0.212   1.167   0.107 
```

So basically, when we measure the noise using the same wcs as was used to draw the noise, we get back to essentially the original values.  The exception is "big", which must be losing too much information by being rendered at effectively wider spacing, so it can't recover the information correctly at the original pixel scale.  I think all that makes sense, although it is possibly not what we want.

I believe Barney's suggestion is that people would instead typically want the raw results to match up with orig, which they currently do not.

I can pretty easily code this up.  The only question is do we care about preserving the current functionality where the code respects the two WCS functions?  Or should we always have `image.addNoise(correlated_noise)` ignore the WCS intrinsic to the noise object and just draw it according to the pixel correlations that were in the original image?  I'm kind of thinking the latter, in which case we won't need the `ignore_wcs` flag, but I'm open to other opinions.

rmjarvis:
Update: I enabled Barney's suggestion locally (but didn't commit).  With that change, here is the output for my new "test":
```
                                orig    same     big    small    rot   big/raw sm/raw  rot/raw 
Covariance at (0.0, 0.0)    :   1.700   1.696   1.697   1.700   1.701   1.697   1.700   1.701 
Covariance at (0.03, 0.0)   :   0.570   0.567   1.134   0.040   0.569   0.570   0.563   0.568 
Covariance at (0.0, 0.03)   :   0.571   0.570   1.134   0.041   0.568   0.571   0.568   0.569 
Covariance at (0.03, -0.03) :   0.111   0.109   0.737   0.009   0.746   0.111   0.106   0.104 
Covariance at (0.03, 0.03)  :   0.744   0.745   0.895   0.215   0.104   0.742   0.747   0.746 
```
So it works. However, it also makes two other tests fail.  `test_output_generation_magnified` and `test_symmetrizing` both changed pixel scales and expected the correlations to change appropriately.  

Which makes me thing maybe we should keep the current functionality where the correlations are both stored and applied in terms of world coordinates, and just document that behavior with some example code for how to get COSMOS noise onto an image with a non-trivial WCS:
```
image = galsim.Image(nx, ny, wcs = complicated_wcs)
cosmos_noise = galsim.getCOSMOSNoise(rng, scale=1.0) # Force it onto a unit pixel scale
image.view(scale=1.0).addNoise(cosmos_noise)
```
What do you think?  It's not that complicated for the user, and I'm starting to think that might be less confusing than having the meaning of a `CorrelatedNoise` object be different when whitening vs when adding noise to an image.

*Full disclosure: The `im.view(scale=1)` syntax doesn't currently exist, but I've been thinking about adding it for some time already, so this would be a good time to do so.*

rmjarvis:
OK, I just pushed a fully baked version of my proposal.  

I now have CorrelatedNoise objects keep track of the original wcs that was used when it was built.  This makes it fairly easy for the user to view an image with that wcs and get the pixel covariances to match the original noise image:
```
image.view(wcs=corr_noise.wcs).addNoise(corr_noise)
```
I added documentation to this effect in the doc strings for both `_BaseCorrelatedNoise` and `getCOSMOSNoise`.  Let me know what you think.

barnabytprowe:
This looks good, but I would like until Sunday to take a look properly (I'm on vacation)... Hope that's OK.  I'm sure it's fine, I'd just like to take a look!

rmjarvis:
Of course.  No rush.  I've also just added more tests that use non-square matrices, so I think this PR can satisfy #589 as well.

barnabytprowe:
@rmjarvis I'm sorry for the delay, longer than I had promised.  I think the proposed solution is *really* good and very elegantly sidesteps (or at least forces the user to explicitly confront in a not-too-onerous way) the potential inconsistencies inherent in applying correlated noise while respecting WCS.

Looks good to me!

rmjarvis:
Great.  Thanks so much Barney!

Issue #639, opened 2015-02-05
Circular top-hat profile


chazshapiro:
Hi - I would like to request a circular top-hat profile, or an ellipse more generally.

This would be useful to the Precision Projector Lab for modelling detector emulations in which the sources are simply illuminated holes.  The profile would always be convolved with a bandlimited PSF.  For now, we use PSFs much larger than the holes (approximate point sources), but this won't always be the case.

Thanks,
Chaz

rmjarvis:
Sure.  This would be easy to add.  I'll just do the circle, since you can always shear it to get any ellipse you want, so a circle is a sufficient implementation.

chazshapiro:
Wonderful - thanks!

chazshapiro:
very cool - thanks

---

PR #649, opened 2015-02-22
#639 TopHat profile

rmjarvis:
This PR is mostly to enable a circular top-hat profile, as requested by @chazshapiro.  That's all pretty straightforward, and the new class is called `TopHat` in python, and `SBTopHat` in C++.

However, along the way, I noticed a bug in our implementation of the `SBBox` class, which I was using as a template for my implementation of `SBTopHat`.  

For all other classes, our implementation of real-space drawing is to render the surface brightness at the pixel center.  However, `SBBox`'s real-space rendering along the edge of the box had been to draw plateau height times the fraction of the pixel area that falls inside the box.  

This isn't consistent with our description of how profiles are supposed to be rendered, so I'm pretty sure this was an incorrect implementation.  It evaded notice, because no one ever really calls this function, so nobody cared.  But I fixed the function to draw either the full plateau height or zero according to whether the pixel center is inside the box or not.  This is also the nature of the implementation of the `TopHat` rendering in real space.

I believe the old implementation is essentially what you get when you convolve a `Box` by a `Pixel`.  In other words the action of `box.drawImage(...)` with default arguments, which is unchanged by the edits on this branch.


rmandelb:
Tests pass on my macbook.

Basic question/comment: I do not love the name for this class.  In my mind, a top hat profile can have any shape - i.e., you can make a square / rectangular top hat function, a circular one, etc.  So I liked the fact that SBBox and Pixel basically tell you something about the shape.  Calling this circular one a TopHat seems a bit too general to me.  Note that I don't actually have a much better idea.  ```Circle``` comes to mind as being slightly better, but I can't claim that I love that either (even though it does seem analogous to ```Box``` / ```Pixel```).

galsim/base.py
```diff
@@ -1864,6 +1850,45 @@ def getHeight(self):
...
+
+    @param radius           The radius of the TopHat, where the surface brightness drops to 0.
+    @param flux             The flux (in photons) of the profile. [default: 1]
+    @param gsparams         An optional GSParams argument.  See the docstring for GSParams for
+                            details. [default: None]
+
+    Methods
+    -------
+
+    In addition to the usual GSObject methods, Box has the following access method:
```
rmandelb: Box -> TopHat

galsim/base.py
```diff
@@ -1864,6 +1850,45 @@ def getHeight(self):
...
+
+        >>> radius = tophat.getGetRadius()
```
rmandelb: getGetRadius -> getRadius

pysrc/SBBox.cpp
```diff
@@ -35,36 +35,40 @@ namespace galsim {
     struct PySBBox 
     {
 
-        static SBBox* construct(const bp::object & width, const bp::object & height, double flux,
```
rmandelb: Why is this going  away?

rmandelb: Oh I see, below is another more compact way to wrap the constructor.  Is there anything different about that functionally?

pysrc/SBBox.cpp
```diff
@@ -35,36 +35,40 @@ namespace galsim {
     struct PySBBox
```
rmandelb: Why does module.cpp in this directory not have to be updated to include the TopHat?

pysrc/SBBox.cpp
rmjarvis: Right.  The `bp::init` syntax is simpler for classes that don't need to do anything special with the arguments (like we do with profiles that can take multiple kinds of sizes).  So this hand-made initializer was basically doing what boost does for us with `bp::init` (i.e. check for the presence of required arguments).

rmjarvis:
> I do not love the name for this class. 

Considering that real top hats *are* circular, I thought the name was rather apt.  

But if you want something else, I think it should be something that evokes a cylinder, rather than a flat circle.  Some options: Cylinder, Slab, Barrel, Drum, Column, Tube, Pedestal, Disk (Yes, I used an online thesaurus).  Do you prefer any of these names?

chazshapiro:
Pinhole?

On Sunday, February 22, 2015, Mike Jarvis <notifications@github.com> wrote:

> I do not love the name for this class.
>
> Considering that real top hats *are* circular, I thought the name was
> rather apt.
>
> But if you want something else, I think it should be something that evokes
> a cylinder, rather than a flat circle. Some options: Cylinder, Slab,
> Barrel, Drum, Column, Tube, Pedestal, Disk (Yes, I used an online
> thesaurus). Do you prefer any of these names?
>
> --
> <https://github.com/GalSim-developers/GalSim/pull/649#issuecomment-75440965>
> .
>


-- 
---
Sent from Gmail

pysrc/SBBox.cpp
rmjarvis: Because I put both of the wrap calls in `pyExportSBBox` rather than make another function to call.  The same way `pyExportSBSersic` also includes the wrap call for `SBDevaucouleurs`.

rmandelb:
> Considering that real top hats are circular, I thought the name was rather apt.

Yeah, I think it would be great if I weren't used to talking about mathematical top-hats defined in Cartesian coordinates.  And I'm willing to buy the argument that nobody but me would be confused by this.  Of the other options, i could deal with Pinhole or Disk.

rmjarvis:
Pinhole is ok with me, since that's the use case for which Chaz requested it.  Although I worry that it might be a bit too specific to that.  If someone else comes up with another use case for this profile, it might be weird to have the circular tophat profile be called a `Pinhole`.  So I guess I still prefer `TopHat` personally, but not so strongly as to insist on it.  :)

Chaz, would you like to make the final call on the name?  You'll be the principal user of this class, at least for the near future.

chazshapiro:
I'd go with TopHat.

---

src/SBBox.cpp
```diff
@@ -283,9 +327,238 @@ namespace galsim {
...
+    {
+        double kr0sq = (k.x*k.x + k.y*k.y) * _r0sq;
+        return kValue2(kr0sq);
+    }
+
+    std::complex<double> SBTopHat::SBTopHatImpl::kValue2(double kr0sq) const
+    {
+        if (kr0sq < 1.e-4) {
+            // Error from omitting next term is about 1.e-16 for kr0sq = 1.e-4
+            return _flux * (1. - kr0sq * ( (1./8.) + (1./192.) * kr0sq ));
```
rmandelb: Is this formula supposed to be so obvious (and the size of next term so obvious) that it doesn't require an explanation?  It didn't seem so obvious to me, so I would appreciate a comment describing it if you don't mind.

rmandelb: I guess it's just the Taylor expansion of the standard J1(x)/x formula?  (Sorry, jet-lagged, my brain is running a bit behind...)

src/SBBox.cpp
```diff
@@ -283,9 +327,238 @@ namespace galsim {
...
+                int i1 = std::max(0, int(std::ceil((-xmax - x0)/dx)));
+                int i2 = std::min(m, int(std::ceil((xmax - x0)/dx)));
+                if (i1 < i2)
+                    val.col(j,i1,i2).setAllTo(_norm);
+            }
+        }
+    }
+
+    void SBTopHat::SBTopHatImpl::fillKValue(tmv::MatrixView<std::complex<double> > val,
+                                            double x0, double dx, int ix_zero,
```
rmandelb: I realize you may be copying syntax from elsewhere that we don't want to change, and this may just be me, but I find it weird to refer to this as x0, dx, etc. instead of kx0, dk, etc. (and likewise throughout the whole function, where x0 and dx and so on are clearly k values).  If you're reading the whole code from start to finish it is kind of obvious what's going on, but if you jump into the code, it seems funny and you have to spend some effort to figure out that the x's and y's are really kx's and ky's.

tests/test_base.py
```diff
@@ -1100,6 +1099,101 @@ def test_box():
...
+        print 'radius = ',radius
+        tophat = galsim.TopHat(radius=radius, flux=test_flux)
+        print 'im.bounds = ',im.bounds
+        print 'im.scale = ',im.scale
+        do_shoot(tophat,im,"TopHat with radius = %f"%radius)
+        do_kvalue(tophat,im,"TopHat with radius = %f"%radius)
+
+    # Check sheared tophat the same way
+    tophat = galsim.TopHat(radius=1.2, flux=test_flux)
+    # Again, the test is very sensitive to the choice of shear here.  Most values fail because
```
rmandelb: Are any of these issues with choices of radius / shear worth mentioning in the docstring for users to watch out for?
(And by users, I mean Chaz. :)

rmandelb:
Okay, as you saw, I just finished going over the code.  Minor issues only.

tests/test_base.py
rmjarvis: I don't think so.  The problem is that this test is checking the routines that render the surface brightness of the profile without convolving by the pixel.  That's where there is a sharp cliff in the profile that can be very close to a pixel center, in which case the FFT method can sometimes end up getting slightly different results at those edge pixels due to numerical rounding.

The normal way you would want to render a `TopHat` is:
```
tophat = galsim.TopHat(radius)
tophat.drawImage(image = image, offset = offset, ...)
```
which implicitly convolves by the `Pixel` profile behind the scenes to correctly render what you would see when shining through a pinhole onto a CCD.  

So I don't think there are any special caveats in this case.  We already spend a while talking about how convolution by a pixel is required to get an accurate rendering and how that's what drawImage does by default (with `method='auto'`).

src/SBBox.cpp
rmjarvis: Yes, it's the Taylor expansion, which I got from Maple.  I'll add a comment to that effect.

src/SBBox.cpp
rmjarvis: Yeah, this was a poor design choice I made ages ago, which I just continued here.  I'll take this opportunity to turn them all into kx, ky instead.

rmjarvis:
I addressed all of Rachel's comments.  Anyone else want to look through this?

include/galsim/SBBox.h
```diff
@@ -69,6 +69,43 @@ namespace galsim {
...
+         */
+        SBTopHat(double radius, double flux, const GSParamsPtr& gsparams);
+
+        /// @brief Copy constructor.
+        SBTopHat(const SBTopHat& rhs);
+
+        /// @brief Destructor.
+        ~SBTopHat();
+
+        /// @brief Returns the radius of the TopHatcar.
```
jmeyers314: What's a TopHatcar?

jmeyers314:
Hi Mike,
  Looks good to me, and all tests pass (even the slow ones).

tests/test_base.py
rmandelb: Fair enough.

include/galsim/SBBox.h
rmjarvis: Fixed.  Thanks.

rmjarvis:
OK.  Thanks everyone.

chazshapiro:
I just felt like saying thank you again for this feature.  I am creating new masks for the JPL scene emulator (projector), which may consist of top-hat apertures etched through 2 material layers.  The layers have different optical transmissions and the etched features may not be perfectly aligned.  It is straightforward to model this and convolve with the projector PSF in GalSim to predict what the NIR detector will see.

rmjarvis:
You're quite welcome of course.  Glad this is still coming in useful.

Issue #645, opened 2015-02-11
TMV error when using transform()


rmandelb:
Hi @rmjarvis - I am trying to use the GSObject.transform() method to do a parity flip, and ran into an error that I don't understand.  I want to remap the image coords from (x, y) to (x', y') where x'=-x and y'=y, so I wrote the code like this:

```
foo = galsim.OpticalPSF(0.1, obscuration=0.2, nstruts=6)
foo_flip = foo.transform(-1.,0.,0.,1.)
im_foo_flip = foo_flip.drawImage(scale=0.01)
```

When I do this, the output is:
```
last col element (2536) must be in 0 -- 768
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/rmandelb/software/lib/python2.7/site-packages/galsim/base.py", line 1249, in drawImage
    image.added_flux = prof.SBProfile.draw(imview.image, gain, wmult)
RuntimeError: TMV Error: Failed Assert statement base::hasSubMatrix(i1,i2,j1,j2,1,1)
```

Do you have any idea what could be going on here?  Is this a case of stupid input (did I misunderstand the args to transform?) or is this a bug somewhere?

rmjarvis:
At the very least the error needs to be changed, since that's not very helpful.  But I don't see anything wrong with what you are doing, so this is probably a bug somewhere.

I was able to reproduce this on my machine, so I'll look into it and see if I can figure out what is going wrong.

rmandelb:
Thanks!

PR #650, opened 2015-02-22
#645 Two small bug fixes

rmjarvis:
1. Rachel discovered a bug where flipping an OpticalPSF profile (with `transform(-1,0,0,1)`) and trying to draw it gave an error.  It turned out that there were a few lines in `SBInterpolatedImage::fillKValue` that assumed that dx and dy were positive, so it failed when the transformation had a flip.  I added a bunch of unit tests to check that this was the only place this assumption was made.  It seems that all the other profiles are fine when flipped.

2. While writing this unit test, I discovered a slight bug in the implementation of drawImage with `nx,ny`.  If either nx or ny was odd, the image was not correctly centered in the center of the image.  So I fixed this as well (and added unit tests for it).


rmandelb:
Very nice - thanks!  I confirm unit tests pass on my Macbook, will look at the code itself when I'm not hideously jet lagged since I definitely won't be useful right now.

rmandelb:
Since I'm back home I decided to try on our Linux cluster, and found a test failure.  So then I reran tests on my Macbook, and the same tests fails.  Yesterday I could have sworn I was on this branch and the tests passed, and you haven't changed the code, so I am confused.  Then I was worried about some screwup when switching branches, so I did some cleanup (scons -c install), re-compiled, and re-installed the code on this branch on both systems, and the same test still fails on both systems:
```
FAIL: Test the various optional parameters to the draw function.
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/rmandelb/git/GalSim/tests/test_draw.py", line 284, in test_draw
    "obj2.draw(nx,ny,scale) (even) did not center in x correctly")
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/site-packages/numpy/testing/utils.py", line 468, in assert_almost_equal
    raise AssertionError(msg)
AssertionError: 
Arrays are not almost equal to 4 decimals
obj2.draw(nx,ny,scale) (even) did not center in x correctly
 ACTUAL: 100.50119583475475
 DESIRED: 100.5
1.79801220597 27.5 27.5 30.0020164042 30.0020164042 0.0
1.79857816637 27.5 27.5 30.0696689965 30.0696689965 0.0
1.79857816997 27.5 27.5 30.069668949 30.069668949 -1.60670452343e-14
1.79857816997 27.5 27.5 30.069668949 30.069668949 -1.60670452343e-14
1.79999895382 109.5 109.5 30.4693839237 30.4693839237 5.18618076832e-13
1.79857633207 27.5 27.5 30.0691940799 30.0691940799 -2.69818346741e-15
1.79841249008 33.5 33.5 45.5833758935 45.5833758935 0.0
1.79999456823 134.5 134.5 46.2173802708 46.2173802708 -1.05994811555e-12
1.80000066927 99.5 99.5 46.2590428105 46.2590428105 -6.87397121661e-14
1.79999855929 99.5 99.5 30.4688838999 30.4688838999 2.39520163952e-14
1.79999855929 99.5 99.5 30.4688838999 30.4688838999 2.39520163952e-14
1.80000066927 99.5 99.5 46.2590428105 46.2590428105 -6.87397121661e-14
1.79999855929 99.5 99.5 30.4688838999 30.4688838999 2.39520163952e-14
1.79999855929 99.5 99.5 30.4688838999 30.4688838999 2.39520163952e-14
      99.501196        49.500601        46.259735        46.234836        7.1857598e-07  

--------------------- >> end captured stdout << ----------------------
```

rmjarvis:
I found a machine that reproduces this error, so I'll look into it.  (On my other machines, that 99.501196 is 99.499998.)

rmjarvis:
It seems to be a numpy calculation inaccuracy for older versions of numpy.  I pushed a fix to always use double precision in the getmoments function, which fixed it.

rmandelb:
That works - thanks!

rmjarvis:
I beefed up the suite of flip tests here, and it actually caught another bug.  If the profile is shifted and then shifted again, and the two shifts couldn't be combined into a single transformation (e.g. if the second shift is inside an Add object, and the first is the shift by 0.5 pixels for even grids), then there was an error in the calculation, which I've now fixed.  So three small bug fixes on this branch now.  :)

rmandelb:
Nice!  I just confirmed that the unit tests (including the longer version when running ```test_transform.py```) pass on both my systems.

rmjarvis:
Did anyone want to look through this code?  It's kind of technical, so I'm not sure how much you want to pore through the actual code changes.  Looking through the additions to the test suite may be sufficient.

rmandelb:
I will take a look at the additional unit tests in the next few days.

tests/test_draw.py
```diff
@@ -272,13 +272,32 @@ def test_draw():
     # Test if we provide nx, ny, and scale.  It should:
     #   - create a new image with the right size
     #   - set the scale
-    im10 = obj2.draw(nx=200, ny=100, scale=dx_nyq)
```
rmandelb: This is a minor point, but I think it would be better to make the various instances of 200, 100, 201, 101, 0.51, etc. set by variables,
```
nx=200
ny=100
scale=0.51
```
and then refer to nx, ny, nx+1, ny+1, scale, etc.  That way if we have some reason to change this code in future, everything necessary will be defined up front.

tests/test_transforms.py
```diff
@@ -647,6 +647,189 @@ def test_integer_shift_photon():
...
+            galsim.OpticalPSF(lam_over_diam=0.17, obscuration=0.2, nstruts=6,
+                              coma1=0.2, coma2=0.5, defocus=-0.1, flux=1.7),
+            # Like with Box, we need to increase the real-space convolution accuracy.
+            # This time lowering both relerr and abserr.
+            galsim.Pixel(0.23, flux=1.7,
+                         gsparams=galsim.GSParams(realspace_relerr=1.e-6,
+                                                  realspace_abserr=1.e-8)),
+            # Note: RealGalaxy should not be rendered directly because of the deconvolution.
+            # Here we convolve it by a Gaussian that is slightly larger than the original PSF.
+            galsim.Convolve([ galsim.RealGalaxy(rgc, index=0, flux=1.7),  # "Real" RealGalaxy
```
rmandelb: Why are these "Real" and "Fake"?  They look like they are just two entries from the catalog, both convolved with a Gaussian as necessary to render an image.

tests/test_transforms.py
```diff
@@ -647,6 +647,189 @@ def test_integer_shift_photon():
...
+            galsim.Convolve([ galsim.RealGalaxy(rgc, index=1, flux=1.7),  # "Fake" RealGalaxy
+                              galsim.Gaussian(sigma=0.08) ]),
+            galsim.Spergel(nu=-0.19, half_light_radius=0.17, flux=1.7),
+            galsim.Sersic(n=2.3, half_light_radius=0.17, flux=1.7),
+            galsim.Sersic(n=2.3, half_light_radius=0.17, flux=1.7, trunc=0.82),
+            # The shifts here caught a bug in how SBTransform handled the recentering.
+            # Two of the shifts (0.125 and 0.375) lead back to 0.0 happening on an integer
+            # index, which now works correctly.
+            galsim.Sum([ galsim.Gaussian(sigma=0.17, flux=1.7).shift(-0.2,0.125),
+                         galsim.Exponential(scale_radius=0.23, flux=3.1).shift(0.375,0.23)]),
```
rmandelb: I see you are now testing flips combined with multiple shifts.  Is there any reason to worry about flips combined with other types of transformations handled by SBTransform?

tests/test_transforms.py
rmjarvis: I believe index = 0 is a real COSMOS image, and index = 1 was a parametric galaxy that you drew for the purposes of the tests in test_real.py.  In that script you use `ind_fake = 1` and `ind_real = 0`.  So I was just following your lead here.

tests/test_transforms.py
rmjarvis: Probably not, but I'll add some to be sure.  And the problem with the double shift had nothing to do with the flip.  It used to raise an exception just drawing it in the first place.

tests/test_transforms.py
rmandelb: Sorry, for some reason i had in my head you were using one of the example catalogs, not the test catalog.  Never mind.

rmjarvis:
Rachel, I addressed the comments you made.  Did you want to look through the main code at all, or are you happy to let this merge?

rmandelb:
I am fine with your going ahead to merge it.

Issue #643, opened 2015-02-10
Deprecate things we nominally got rid of in v1.1


rmjarvis:
In version 1.1, we semi-deprecated a lot of old syntax in preference to our new API.  However, we didn't officially deprecate any (or at least much) of the old API.  I think it is time to do so.

For the next release, v1.3, I'd like to have all the old syntax give deprecation warnings (not errors yet) indicating what the preferred function to use is.

rmandelb:
+1

PR #652, opened 2015-03-13
#643 Actually deprecate all the stuff we've been describing as discouraged or obsolete or even deprecated

rmjarvis:
I made a `GalSimDeprecationWarning` class that gets raised in a `warning.warn` call whenever someone uses a deprecated function or parameter.  It tells the user what function to use instead of what they did use, along with which version of GalSim this feature started being deprecated.  I think I got all the versions correct.  Most were version 1.1.  

Along the way, I noticed a few things that we forgot to do in previous edits.
- Position and GSParams had not been made immutable as we had done for most other classes (pretty much everything except Image). So I did that here.
- We had officially discouraged `setRNG()` for Noise types.  But there were times when we wanted to make a new noise type with a different `rng`.  So I made the `copy` function optionally take an `rng` parameter to use for the new copy of the Noise object.
- The unit tests still used the old (now deprecated) syntax.  This was mostly intentional to make sure that legacy code would still work correctly.  But bad habits leaked in and some newer unit tests also used the old syntax.  Anyway, I switched everything to use the new syntax now.

This is a big PR in terms of number of lines, but there's not actually a lot to look at in detail.  So if possible, I'd like to get a quick turnaround on this one.  I have another issue almost ready to submit that I'd like to get into v1.3 as well, but it was branched off of this branch, so it will be cleaner to wait until this one is merged before doing the PR on it.


rmandelb:
I should be able to take a look over the weekend or early next week, so hopefully I won't hold things up too long.  (I have to focus on non-GalSim things today.)

galsim/deprecated/__init__.py
```diff
@@ -15,3 +15,29 @@
...
+#       for python system deprecations.  I think GalSim deprecations are more appropriately
+#       considered UserWarnings, which are not ignored by default.
+class GalSimDeprecationWarning(UserWarning):
+    def __init__(self, s):
+        super(GalSimDeprecationWarning, self).__init__(self, s)
+
+def depr(f, v, s1, s2=None):
+    import warnings
+    s = str(f)+' has been deprecated since GalSim version '+str(v)+'.'
+    if s1:
```
rmandelb: Isn't s1 required?  Or is the idea that if there is no suggestion to make for the user, it'll be False?

rmandelb: Oh I see, you use an empty string for functionality that has no replacement.  Never mind.

galsim/deprecated/shear.py
```diff
@@ -0,0 +1,52 @@
...
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions, and the disclaimer given in the documentation
+#    and/or other materials provided with the distribution.
+#
+
+import galsim
+from galsim.deprecated import depr
+
+def Shear_setE1E2(self, e1=0.0, e2=0.0):
+    """Deprecated method that will be deprecated eventually."""
```
rmandelb: Deprecated method that will be deprecated eventually?  That sounds kind of weird and redundant to me.  How about "Deprecated method to set the Shear to some (e1, e2)"?  (Same comment applies to the other things in shear.py)

galsim/noise.py
```diff
@@ -204,13 +189,18 @@ def GaussianNoise_applyTo(self, image):
...
-Discouraged; will be deprecated.
-"""
 
-def GaussianNoise_copy(self):
-    return _galsim.GaussianNoise(self.getRNG(),self.getSigma())
+def GaussianNoise_copy(self, rng=None):
+    """Returns a copy of the Gaussian noise model.
+
+    By default, the copy will share the BaseDeviate random number generator with the parent
+    instance.  However, you can provide a new rng to use in the copy if want with
```
rmandelb: "if want" -> "if you want"

rmandelb: (same comment applies below)

galsim/deprecated/__init__.py
rmjarvis: Right.  I did that for calculateCovarianceMatrix, which we are just removing, not replacing.  I should add a commend to that effect here so developers will know how to call this in that kind of case.

galsim/deprecated/shear.py
rmjarvis: Yes.  I think it used to say something like "Obsolete method that will be deprecated."  I replace obsolete, but forgot to check that the rest of the string was sensible.  I'll change these.

rmandelb:
I looked at the stuff in galsim/ and its subdirectories carefully, and it seemed fine... but I have to admit my eyes glazed over once I was partway through the stuff in tests/.  I suspect that mistakes in that code would simply result in failed tests, so you would've caught them already.

rmjarvis:
> I have to admit my eyes glazed over once I was partway through the stuff in tests/.

lol.  Thanks, Rachel.  I started by doing all the things in galsim/deprecated, which pulled up a huge number of warnings (not failures) in the unit tests.  Then I edited those until all the warnings went away.  So I think I got them all.  I suppose I could have missed one or two within a `if __name__ == '__main__'` block, but if so, someone will notice eventually and fix it then.

rmandelb:
Yes, sounds good to me.

rmjarvis:
I addressed all of Rachel's comments.  Did anyone else want to take a look?  I'll plan to merge Wednesday if I don't hear otherwise before then.

Issue #218, opened 2012-07-19
make simple Boost.Python-wrapped objects copyable/picklable


TallJimbo:
We should define `__reduce__` methods for whatever Boost.Python wrapped classes we can, and `__copy__` and `__deepcopy__` methods for immutable objects.  This will allow them to work with the the built-in `copy` and `pickle` modules (note that pure-Python classes should for the most part already work with these modules, unless they contain Boost.Python objects as attributes).

Pickle may be particularly important if we want to use the `multiprocessing` module, as I think it transfers objects between processes by pickling them.

Issue #653, opened 2015-03-19
MultipleImageHelper


jmeyers314:
I came across the MultipleImageHelper code while looking through the InterpolatedImage files to see what might be required to add InterpolatedKImage (ala #642), and also while thinking of profiles that are built up as series expansions (ala #628).  I can't actually figure out how to use it though.

This much seems to work:
```
import galsim
aimg = galsim.Gaussian(fwhm=0.5).drawImage()
mih = galsim._galsim.MultipleImageHelper(image=aimg.image, pad_factor=4.0)
```

but then I can't figure out how to pass this to SBInterpolatedImage:
```
# This fails.
obj = galsim._galsim.SBInterpolatedImage(multi=mih, weights=[1.0],
                                         xInterp=galsim.InterpolantXY(galsim.Quintic(tol=1e-4)),
                                         kInterp=galsim.InterpolantXY(galsim.Quintic(tol=1e-4)))
```

Similarly, I can't figure out how to create a MultipleImageHelper that actually handles multiple images:
```
# This fails.
bimg = galsim.Gaussian(sigma=0.6).drawImage()
mih = galsim._galsim.MultipleImageHelper(images=[aimg.image, bimg.image], pad_factor=4.0)
```

In both cases, I suspect it has something to do with me providing a python list where the c++ is expecting a std::vector, which apparently isn't compatible.  My boost-fu ends about there though.

I tried grepping around for all the instances of `_galsim.SBInterpolatedImage`, and as far as I can tell the multi image machinery is never used or tested in any .py files.  Did this get abandoned?

rmjarvis:
No, it never really got used.  I think Gary's idea about this was that you might want to have an object modeled as a weighted sum of several different images, and then reweight them on the fly, say to derive best fit parameters of a decomposition for instance.  @gbernstein please correct me if I'm wrong here.

But we never really used it that way, and honestly, I think that kind of functionality (e.g. #628) is better implemented in the python layer by making sums of profiles (not all of which need to be InterpolatedImages).  Our code for rendering sums with arbitrary scalings is I think just as efficient as what you would get by using the MultiplImageHelper, and more flexible as well.

So my inclination would be to just remove the MultipleImageHelper class and only have SBInterpolatedImage be constructible from a single image.  

[...]

Issue #644, opened 2015-02-11
it's 2015 - update license?


rmandelb:
Should the license be updated to say 2012-2015?

Mike, I think you had a script that updated the license in all files when we changed from GPL to BSD.  Can your script be used to update this too?

rmjarvis:
Yes.  I can do this pretty easily.   

The script is in GalSim/devel/updatecr, btw.  A perl script no less!  :)

rmandelb:
Thanks!  A perl script - so retro.

rmjarvis:
Done on master

rmandelb:
Thanks!

Issue #590, opened 2014-08-01
WFIRST module


rmandelb:
This issue is similar in flavor to #556, but I'm interested in working on this one first.  The idea is to make a WFIRST module.  Comments on any of the below are welcome:

(1) Bandpasses: I would like there to be routines to get Bandpass objects for the planned WFIRST bandpasses, i.e.
```
my_bandpass = galsim.wfirst.getBandpass(...)
```
where the argument might be a string that is the name of the bandpass.  An alternative would be to make a routine that returns all bandpass info, e.g., 
```
bandpass_list = galsim.wfirst.getBandpasses()
```
where it returns a list of tuples, each containing the name of the bandpass and the associated Bandpass object.

(2) WCS information for the chips, similar to what's described on #556.

(3) PSF information.  I need to dig a bit more into what information I have before deciding how to put this in.  Will probably post some thoughts once I've done that.

(4) Information about various detector effects such as those discussed in #552, #553, #554, #555.  My assumption is that the code to apply those effects will not be in a WFIRST module, but rather in a GalSim "detectors" (or something) module, but presumably those routines will require inputs like the function defining nonlinearity, the type of kernel to apply to simulate inter pixel capacitance, etc.  So the WFIRST module needs to know those things about the WFIRST detectors, to provide inputs for anyone who wants to use the routines in the "detectors" module to make WFIRST-y detector effects.

Any comments or suggestions?

rmjarvis:
I made the changes to FitsHeader for you.  It should be working.  Let me know if that also makes AstropyWCS work now.  It should be able to handle TAN-SIP just fine, so you should not need to muck around with `fits_wcs_types`.  If that still doesn't work, let's figure out why and fix it rather than override the current `fits_wcs_types` list.

rmandelb:
Cool, thanks.  I’ve made a note to try the AstropyWCS stuff the next time I do work on this branch.

rmandelb:
@arunkannawadi - please take a look at the latest version of demo13, and see what you think.  It's a much more realistic galaxy population.

rmandelb:
@arunkannawadi , we have to think about unit tests before we can make a PR (which in any case has to wait until #618 is merged, but it would be good to be ready as soon as that is done).

Here are the routines that currently lack unit tests:
1) WFIRST background routines.
2) WFIRST PSFs.
3) WFIRST WCS.
4) WFIRST bandpass.
5) COSMOS catalog stuff that I just put in catalogs.py.
[...]

PR #655, opened 2015-03-21
#218 Make everything picklable

rmjarvis:
This is a really old issue that's been kind of irksome that we haven't gotten around to yet.  So a few weeks ago, I decided to tackle it.  I'm not actually 100% done yet, but it's close enough that I think it's worth having people take a look.

Almost all of our classes now have the following features:
- Picklable.  i.e. `cPickle.loads(cPickle.dumps(obj))` gets you back an exact copy of the original
- Copyable.  i.e. `copy.copy(obj)` makes a copy of the object.  Anything with an `rng` attribute won't be identical, since the RNG is just copied, not duplicated.
- Deep copyable.  i.e. `copy.deepcopy(obj)` makes an exact copy of the original.  Things with an `rng` attribute _are_ identical in this case.
- Repr-able.  i.e. `eval(repr(obj))` makes an exact copy of the original.  The usual [recommendation](http://stackoverflow.com/questions/1436703/difference-between-str-and-repr-in-python) for repr is that it be "unambiguous".  A common way to make sure of that is to have the repr string be `eval`-able, at least in some context (e.g. after `import galsim`).  However, in our case, this means the repr string in some cases is rather long, so I'm quite willing to reconsider whether this is a feature we really want.
- String-able.  i.e. `str(obj)` returns a reasonable looking string that describes the object.

Probably most people won't want to read through the code changes I made to make all this work, but I would encourage people to play around with printing `str(obj)` and `repr(obj)` to see how things look.


rmjarvis:
Some points worth noting:
- The reprs that are currently a bit unwieldy are things like `InterpolatedImage` that includes a full image as part of the repr, and objects with `rng` attributes, since the serialization of a `BaseDeviate` is kind of long.  So we might choose to add some ambiguity in those cases in return for brevity.
- I added a Transformation class as a type of GSObject.  This is now the return type from all transformations (shear, expand, withFlux, etc.)
- I haven't done the repr, str stuff for ChromaticObjects yet.  I think I'll want to add a ChromaticTransformation class there in parallel with the new structure for GSObject.  I think that will simplify a lot of the way transformations are currently handled there.  But I'll wait until #590 is merged, so I can incorporate the new Interpolation stuff properly.
- Other classes I haven't gotten to yet: FitsHeader, Catalog, Dict, PowerSpectrum, PowerSpectrumRealizer, PowerSpectrumEstimator, NFWHalo, Cosmology.  I think these are all easy.  I just haven't gotten to them yet.
- I completely rewrote the Shear class in the python layer.  We don't really ever need the Shear class in the C++ layer except to pass information back and forth, and there is nothing computationally intensive about it, so rather than do the boost.python pickling stuff, I just rewrote the class in python.  Also, I changed the fundamental data about the shear from (e1,e2) to a single complex (g1 + 1i g2), stored in an attribute called `_g`.  Most of the math about shears is a lot easier in terms of complex g than (e1,e2) (shear addition for instance!).  So no more `CppShear`.
- I got rid of the 2d interpolant stuff from the python layer.  We always just wrap 1d interpolants in an `InterpolantXY`.  We mostly insulated the user from this, but there were lots of places in the docs that referred to the 2d stuff that I think is just simpler to remove.  If we ever decide to have an interpolant that is not the same in x and y, we can bring it back, but until we have a use case for that, I think this is simpler.
- I got rid of the wrappings of the MultipleImageHandler stuff.  As Josh noted in #653, we don't ever use it, and there are no unit tests of it.  Again, I left it there in the C++ layer in case we decide we want it, but I don't think there is currently any need for it in the python layer.
- I added a zeropoint option to the Bandpass constructor.  Only a float value, so if you want to do something complicated, you still need to do `withZeropoint`.  But if you know the number, you can set it on construction.
- I added a redshift option to the SED constructor, so you can set that right away rather than doing `atRedshift`.

rmjarvis:
One more comment about the implementation on this PR.  I think it's worth describing a bit how I got pickling to work for SBProfiles.

Most boost.python-wrapped classes just need a couple hooks in the python layer to get pickling to work.  There are two options.
1. Define `__getinitargs__` with the arguments you would need to pass to the constructor to recreate the object.  This is a simple way to do it for things with very light constructors, so it is usually the way to do our various SBProfile sub-classes, since they don't do much work in the constructor.
2. Define `__getstate__` to get some complete description of the current state of the object (usually a dict) and `__setstate__` to take that information and bring an object back to that state.  This is what is required for things with file access and such, like RealGalaxyCatalog.  We can't pickle the locks or the open file handles, so we close them in the dict of the current state and recreate what we need on the other side.

Most SBProfile classes were easy to implement with the `__getinitargs__` method.  The tricky ones were the compound objects, since they store bare SBProfile objects in them that lose their original construction type.  e.g. a sheared SBGaussian is an SBTransform which has a `adaptee` object in it that is just an SBProfile.  i.e. not an SBGaussian anymore.  

To make pickling on them work, I made the repr function for SBProfile objects a virtual function in C++ that returns the python repr string.  A base object can call `repr` and virtual inheritance means the derived class produces the correct string.  Then in the python layer, this string can be used with `eval` to recreate the original object with the correct type, even when they've lost their original type information from being inside a compound object.

So the pickling for the base class `SBProfile` doesn't use `__getinitargs__`.  Rather it has `__getstate__` return the repr string, and then `__setstate__` evals the string to build the correct kind of object.  

This means that we cannot relax the repr for any of the SBProfile classes.  These need to be complete, even if they are a bit unwieldy.  SBInterpolatedImage in particular is long.  However, the GSObject classes don't rely on this trick for pickling.  They use much more efficient methods for serialization, so if we want to make their reprs more readable, that would be ok.

rmandelb:
@rmjarvis - I need to set aside a decent chunk of time to go over this (not today but hopefully this week).  But for now I wanted to mention some issues I'm having compiling the code and running unit tests:

1) On the linux cluster here, I am getting some compiling errors:
```
In file included from include/galsim/SBProfile.h:44,
                 from include/galsim/OneDimensionalDeviate.h:29,
                 from src/OneDimensionalDeviate.cpp:22:
include/galsim/Angle.h: In member function ‘void galsim::Angle::sincos(double&, double&) const’:
include/galsim/Angle.h:206: error: no matching function for call to ‘galsim::Angle::sincos(const double&, double*, double*) const’
include/galsim/Angle.h:204: note: candidates are: void galsim::Angle::sincos(double&, double&) const
```

2) On my Macbook, it crashed and burned when I tried to run unit tests:
```
Starting python tests...
E...........................................................................................................................................................python(13674,0x7fff79962310) malloc: *** error for object 0x7fff7aca7330: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
```
so then I tried ```scons tests -j1``` in case it was a multiprocessing issue, but the same thing happened.

rmandelb:
Sorry, forgot to say, on the Linux cluster the compiler is g++ version 4.4.7.

rmjarvis:
Sorry, I fixed the g++ error.  Forgot to run tests on linux before doing the PR, so that should be fixed now.

The Mac problem sounds like you need to do `rm -rf .scon*` and rebuild.  But let me know if that doesn't work for you.

rmandelb:
Thanks.  The problem on the linux cluster is now fixed.  On my Mac, I tried what you suggested but the error persists.

rmjarvis:
> On my Mac, I tried what you suggested but the error persists.

Could you try going into the tests directory and typing `run_all_tests`.  Then tell me the last bit that gets written before it crashes.  That should at least tell me what it's working on when it has a problem.

rmandelb:
Whew, that took a long time. CppShapeData is the cause of this problem:
```
time for test_shearest_shape = 0.55
Try pickling  galsim.hsm.HSMParams(3.0,3.6,25.0,1,2,400,-1,0.25,8000.0,15.0,4,-1000.0)
Try pickling  galsim.hsm.HSMParams(1.0,1.6,2.0,0,0,100,4,0.05,80.0,5.0,2,99.0)
Try pickling  galsim.hsm.ShapeData(image_bounds=galsim.BoundsI(xmin=1, xmax=94, ymin=1, ymax=94), moments_status=0, observed_shape=galsim.Shear((0.02971327146432768-0.0445738748080978j)), moments_sigma=2.390927314758301, moments_amp=1.601151466369629, moments_centroid=galsim.PositionD(x=47.4999999926344, y=47.50000001050785), moments_rho4=2.2458355239761225, moments_n_iter=26)
Try pickling  galsim.hsm.ShapeData(image_bounds=galsim.BoundsI(xmin=1, xmax=94, ymin=1, ymax=94), moments_status=0, observed_shape=galsim.Shear((0.02971327146432768-0.0445738748080978j)), moments_sigma=2.3011386394500732, moments_amp=1.7979614734649658, moments_rho4=2.2458355239761225, moments_n_iter=26, correction_status=0, corrected_e1=0.12574654817581177, corrected_e2=-0.18868330121040344, meas_type='e', corrected_shape_err=0.0, correction_method='REGAUSS', resolution_factor=0.5634469985961914, psf_sigma=1.5248305797576904, psf_shape=galsim.Shear((-2.7641240407660916e-09-5.120215362808267e-09j)))
Try pickling  galsim._galsim.CppShapeData(galsim.BoundsI(xmin=0, xmax=0, ymin=0, ymax=0),-1,0.0,0.0,-1.0,-1.0,galsim.PositionD(x=0.0, y=0.0),-1.0,0,-1,-10.0,-10.0,-10.0,-10.0,'None',-1.0,'None',-1.0,-1.0,0.0,0.0,'')
python(19757,0x7fff79962310) malloc: *** error for object 0x7fff7aca7330: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
./run_all_tests: line 3: 19757 Abort trap: 6           $python $test
```

jmeyers314:
Hi Mike,

I haven't looked through this in detail yet, but will plan to.

Tests pass on my mac, with one warning:
```
............................................................................/Users/josh/src/GalSim/galsim/base.py:1062: UserWarning: You called drawImage with no `method` parameter for an object that includes convolution by a Pixel.  This is probably an error.  Normally, you should let GalSim handle the Pixel convolution for you.  If you want to handle the Pixel convolution yourself, you can use method=no_pixel.  Or if you really meant for your profile to include the Pixel and also have GalSim convolve by an _additional_ Pixel, you can suppress this warning by using method=fft.
  "You called drawImage with no `method` parameter "
```
I haven't tracked this down yet.

I have a question related to this branch: would it be reasonable to use `__repr__` to implement `__eq__` by value instead of python's default which is to implement `__eq__` by testing if the two vars are the same object?  (Maybe `__hash__` too?)

For example, currently this is the situation:
```
import galsim
a = galsim.Gaussian(fwhm=0.5)
b = galsim.Gaussian(fwhm=0.5)
print a == b # returns False
print a is b # returns False
print repr(a) == repr(b) # returns True
```

We could override `__eq__` to turn the first comparison into a True, and still let the `is` operator handle the case where you really do want to check if the vars are the same object.  I've wanted this feature a few times in the past for caching results.

rmjarvis:
I thought about implementing `__eq__` for everything on this branch too, but I didn't get around to it.

Using `repr` would be slightly inefficient, but quick to implement.  Probably in most cases, it would be fine, since who really cares if == is efficient.  :)  But most of the time it would be easy to look at the repr function and see what needs to go into the eq implementation.  e.g. For Gaussian it would be
```
def __eq__(self, other):
    return self.sigma == other.sigma and self.flux == other.flux and self.gsparams == other.gsparams
```
since those are the three attributes that enter into the repr.

I'll work on adding these.

jmeyers314:
>since who really cares if == is efficient.

As long as it's more efficient than drawing images (especially chromatically), I'll be happy :)

rmjarvis:
> Try pickling  galsim._galsim.CppShapeData(galsim.BoundsI(xmin=0, xmax=0, ymin=0, ymax=0),-1,0.0,0.0,-1.0,-1.0,galsim.PositionD(x=0.0, y=0.0),-1.0,0,-1,-10.0,-10.0,-10.0,-10.0,'None',-1.0,'None',-1.0,-1.0,0.0,0.0,'')
python(19757,0x7fff79962310) malloc: *** error for object 0x7fff7aca7330: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug

I'm going to take a stab here, and guess that this might be related to the two different std libraries on Macs, one of which has dynamically allocated strings, and the other doesn't.  So maybe these are getting mixed up between the compilers that were used for python, boost, and/or galsim.  

I pushed a simple edit to check this.  If this works, I'll work more on it, since it's not really a satisfactory solution, but if it doesn't, then I'll have to think more about what might be the problem.

rmandelb:
No, same error unfortunately.

One possibility is simply that I should clean up my build of GalSim.  I originally went with Enthought python but it is build with a different compiler than the one used when compiling GalSim, and that's occasionally been a source of trouble.  If this problem is something that is only going to come up because of the oddities of how I happened to build GalSim on this system, then perhaps the answer is that I should fix my build to do something a bit less weird.  (It's not the first time that this has been the problem.)

rmandelb:
@rmjarvis , I started playing around with this.  Just a few comments:

> But I'll wait until #590 is merged, so I can incorporate the new Interpolation stuff properly.

The interpolation stuff is already in place (it's not from #590, it's from another branch that is already merged).  So there's no need to wait for that one.

I'm having trouble with reconstructing InterpolatedImage objects using ```eval(repr(obj))```.  In particular, if I do the following:
```
import galsim
g = galsim.Gaussian(sigma=0.5)
im=g.drawImage(scale=0.1)
int_im=galsim.InterpolatedImage(im)
x = eval(repr(int_im))
```
then I get the following error:
```
NameError                                 Traceback (most recent call last)
/Users/rmandelb/git/GalSim/<ipython-input-5-e5154c13d31f> in <module>()
----> 1 x = eval(repr(int_im))

/Users/rmandelb/git/GalSim/<string> in <module>()

NameError: name 'array' is not defined
```

As far as I can tell, this happens because of the following part of ```repr(int_im)```: ```array=array(...)```.  The only way this would work is if I had done something like ```from numpy import array``` (and if I do that, then I get an error from another part of the ```repr``` which uses ```dtype=float32```, which also hasn't been defined).  I'm not sure how to fix this; even if you were to write these as ```numpy.array``` and ```numpy.float32```, it still requires the user to have imported numpy as itself (not ```as np```).  Any ideas?  At minimum, this error seems to suggest that some more unit tests of this object reconstruction would be helpful.

rmandelb:
The problem above also is an issue for reconstructing ```Image```s.

rmandelb:
Question about the ```str``` for transformations.  If i do something like
```
import galsim
g=galsim.Gaussian(sigma=0.5)
g=g.shear(e1=0.5)
str(g)
```
then the result is ```'galsim.Gaussian(sigma=0.5, flux=1.0).transform(1.31607401295,0.0,0.0,0.759835685652)'```.  This is not terribly illuminating.  Would it be possible to have the ```str``` for a ```Transformation``` object say what is the equivalent shear / magnification / shift / flux change?  I see that the ```repr``` does give a jacobian, an offset, and a ```flux_ratio```, which is at least a bit more information even if not as good as an explicit division into shear etc.

rmjarvis:
> The only way this would work is if I had done something like from numpy import array (and if I do that, then I get an error from another part of the repr which uses dtype=float32, which also hasn't been defined).

Yes, that is exactly what you need to do.  Also typically `from numpy import float32, ...`.  The point is not that `eval(repr(obj))` will always work in all situations, but rather that there is at least some state in which it will work.  That means that it is a complete, unambiguous description of the object with no missing details.

We would never advertise `eval(repr(obj))` as being a good way to reconstruct GalSim objects.

rmjarvis:
> Would it be possible to have the str for a Transformation object say what is the equivalent shear / magnification / shift / flux change? 

I guess so.  We have that as a function in the AffineTransform WCS class.  Although, if the decomposition involves a flip, we don't have a non-`transform` version of that.  

But I'm not sure I see the point of doing so.  The main reason to have a sensible str output, as I see it, is to tell whether two objects are equivalent or how they differ.  Not as precisely as repr, but usually close enough.  I've wanted this often in the past when trying to get the example scripts to match up with the yaml version.  Printing out `str(final)` is (now) a great way to see where two objects differ.  

I don't know how much this would improve by decomposing it into `shear(...).rotate(...).dilate(...).transform(0,1,1,0)`.  (The last being what we would need if there was a flip involved.)

rmandelb:
> I don't know how much this would improve by decomposing it

My thinking was that if there was a mistake in some aspect of the transformation (e.g., a shear hadn't been applied that should've been applied) then that would show up more easily in ```shear(...).rotate(...).dilate(...).transform(0,1,1,0)``` than in the current form.  But YMMV.

rmjarvis:
OK.  I implemented this.  

If someone only does a single shear or rotate for instance, it's probably clearer.  Once they start stringing together several, I'm not sure that it is, since the decomposition may not be in the same order they applied their operations.  But play around with it and see what you think.

I could have it revert back to transform if it is not purely one of the more basic operations perhaps.

rmandelb:
> Once they start stringing together several, I'm not sure that it is, since the decomposition may not be 
> in the same order they applied their operations.

So perhaps the answer is that if it appear to be a combination of two non-commuting operations, then revert back to transform?

rmandelb:
Hi Mike - I dug into the failure a bit more.  Here's what I found:

There's a ShapeData object ```res``` that is successfully pickled.  I find that if I make an analogous ```CppShapeData``` object (all quantities the same as in ```res```) then it can't pickle that ```CppShapeData```.  But if I modify just its error message to be something that has nonzero length, then it can pickle that particular object just fine.

If I understood your previous commit correctly, you had been trying making the ```ShapeData``` error message have nonzero length, but the ```CppShapeData``` error message would still have had zero length, and that was the cause of the original crash so that's why that commit didn't help.  Changing the CppShapeData defaults requires a modification to include/galsim/hsm/PSFCorr.h, and I confirmed that if I make that change then nothing crashes any more.  I'm guessing this isn't necessarily the solution we want to adopt (seems kludgy) but is that enough information to help sort out what's happening?

rmjarvis:
> is that enough information to help sort out what's happening?

Yes, that helped.  See if the commit I just pushed works for you.  I think I was on the right track before, but as you pointed out, my proposed fix didn't actually do what I wanted it to do.  I think this one does...

rmandelb:
Yes, fixed.  Thanks.

jmeyers314:
Hi Mike, 

I tracked down the warning which was ruining my perfect little string of '..... ...' (how does one indicate an ellipsis for a train of periods I wonder...?).  It came from one of the lambdas to a do_pickle inside test_compound.py.  I went ahead and added a commit that added method keywords to each of these so my unit tests come out pretty again.

galsim/bandpass.py
```diff
@@ -83,85 +83,116 @@ class Bandpass(object):
...
+            # Manual override!  Be careful!
+            self.wave_list = _wave_list
+            # This also means that red_limit and blue_limit are already set correctly.
+            # Don't change them.
+            assert self.blue_limit is not None
+            assert self.red_limit is not None
+            return
+
+        # Account for wave_factor in wavelength limits
+        if self.wave_factor != 1.9:
```
jmeyers314: I think this should be `1.0` not `1.9`.  Though I guess this makes the if statement unnecessary since if `self.wave_factor` *is* 1.0, then dividing by it is harmless.

galsim/bandpass.py
```diff
@@ -211,28 +229,17 @@ def __div__(self, other):
...
-        return ret
+            tp = lambda w: self.func(w) / other
+
+        return Bandpass(tp, blue_limit, red_limit, _wave_list=wave_list)
 
     # Doesn't check for divide by zero, so be careful.
     def __rdiv__(self, other):
+        # MJ: I can't think what the use case is here.  Why would we want to divide 
+        #     something (scalar or funct) by a bandpass?
         blue_limit = self.blue_limit
```
jmeyers314: My thought here was that one might have a bandpass describing the total system throughput, and another describing the atmosphere throughput.  One could then divide them to get the "above-the-atmosphere" transmission function, which might be a useful photometric calibration tool.  Probably the opposite (multiplying to obtain the total system throughput) is more likely in practice, but I didn't think it hurt to have this just in case.

galsim/base.py
```diff
@@ -208,15 +208,19 @@ def __init__(self, obj):
...
             self.SBProfile = obj
         else:
             raise TypeError("GSObject must be initialized with an SBProfile or another GSObject!")
-        # a couple of definitions for using GSObjects as duck-typed ChromaticObjects
-        self.separable = True
-        self.SED = lambda w: 1.0 # flat spectrum in photons/nanometer
-        self.wave_list = np.array([], dtype=float)
+
+    # a couple of definitions for using GSObjects as duck-typed ChromaticObjects
+    @property
```
jmeyers314: What's the benefit of property vs. attribute here?

jmeyers314:
Thanks for adding `__eq__`.  I think `__hash__` is required for dictionary key use and my future caching plans though:

```
import galsim
a = galsim.Gaussian(fwhm=1)
b = galsim.Gaussian(fwhm=1)
a == b # True!  Awesome.
a is b # False, that's good.
d = {a:3}
d[a] # 3
d[b] # Gives KeyError
```

Maybe `__hash__ = lambda self: hash(repr(self))` is sufficiently quick and dirty?

galsim/bandpass.py
rmjarvis: Doesn't the regular `__div__` work for this use case?  `__rdiv__` is only necessary for expressions like `1.0 / bandpass` where the lhs is not a `Bandpass` instance.

galsim/compound.py
```diff
@@ -614,13 +727,200 @@ def __init__(self, obj, real_space=None, gsparams=None):
...
+
+_galsim.SBAutoCorrelate.__getinitargs__ = lambda self: (
+        self.getObj(), self.isRealSpace(), self.getGSParams())
+_galsim.SBAutoCorrelate.__getstate__ = lambda self: None
+_galsim.SBAutoCorrelate.__setstate__ = lambda self, state: 1
+_galsim.SBAutoCorrelate.__repr__ = lambda self: \
+        'galsim._galsim.SBAutoCorrelate(%r, %r, %r)'%self.__getinitargs__()
+
+
+def Transform(obj, jac=(1.,0.,0.,1.), offset=galsim.PositionD(0.,0.), flux_ratio=1.,
```
jmeyers314: Why is Transform in compound.py?  The other objects in compound all consist of combinations of two or more GSObjects (admittedly, in some cases these are the same object like in AutoCorrelate).  Deconvolve is  an exception but probably belongs in compound since that's where Convolve lives.

Transform seems like a different kind of beast though.  I think I'd be happier putting it in base.py or even creating transform.py.

galsim/base.py
rmjarvis: In this case, the main reason for doing this is that lambda expressions are not picklable.  So making it a property means it doesn't actually appear in the `__dict__`, so it doesn't need to be pickled.  Doing this with all three was just for consistency to make them all look the same, but it doesn't really matter much for the other two.

The other two benefits, which are inconsequential in this case, but have been a reason for me to use `@property` in other cases are:
1. Efficiency.  The code only creates the thing to be returned when it needs it rather than on every construction.  If the attribute is rarely needed and non-trivial to make, then this can be a good reason to use a property.  I don't think either the array or the lambda are terribly time consuming to make, so it doesn't matter much here, although maybe it's a slight efficiency gain.
2. Read-only.  I don't think we are worried about users writing other values to these attributes, but in other cases, I've used properties to help enforce the immutability of some of our objects.  You can write setters for properties, but the simple thing I did here means that these are all read-only.

galsim/bandpass.py
jmeyers314: Ah right.  In that case I agree that this is probably not useful.  Feel free to excise.

galsim/fitswcs.py
```diff
@@ -104,7 +104,15 @@ def __init__(self, file_name=None, dir=None, hdu=None, header=None, compression=
...
-            self._tag = file_name
+            if dir is not None:
+                import os
+                self._tag = repr(os.path.join(dir,file_name))
+            else:
+                self._tag = repr(file_name)
+            if hdu is not None:
+                self._tag += ', hdu=%r'%hdu
+            if compression is not 'auto':
+                self._tag += ', compression=%r'%copression
```
jmeyers314: typo

jmeyers314: same typo occurs a few times in this file actually.

galsim/sed.py
```diff
@@ -198,42 +228,46 @@ def __call__(self, wave):
...
+                   _wave_list=self.wave_list,
+                   _blue_limit=self.blue_limit, _red_limit=self.red_limit)
+
+    def __truediv__(self, other):
+        return self.__div__(other)
 
     def __rdiv__(self, other):
-        # SEDs can be divided by scalars or functions (callables)
-        ret = self.copy()
+        # MJ: Does this make sense?  Is there a use case for 1/sed or f(wave) / sed?
```
jmeyers314: I suppose this one is also excisable.

galsim/bandpass.py
rmjarvis: I deprecated them.

rmjarvis:
> Maybe __hash__ = lambda self: hash(repr(self)) is sufficiently quick and dirty?

Done.

galsim/fitswcs.py
rmjarvis: Thanks.  Fixed.

galsim/compound.py
rmjarvis: Done.

galsim/base.py
jmeyers314: Cool.  I especially like the read-only.  Thanks for the info.

galsim/bandpass.py
```diff
@@ -463,12 +430,49 @@ def thin(self, rel_err=1.e-4, preserve_range=False):
...
+                self._orig_tp == other._orig_tp and
+                self.red_limit == other.red_limit and
+                self.blue_limit == other.blue_limit and
+                self.wave_factor == other.wave_factor and
+                self.zeropoint == other.zeropoint and
+                (self.wave_list == other.wave_list).all())
+
+    def __getstate__(self):
+        d = self.__dict__.copy()
+        del d['_tp']
```
jmeyers314: By deleting `_tp` and `func`, I guess it's not possible for someone to pickle a `Bandpass` (or `SED`) initialized from an ascii file, send it to a friend, and then have them be able to unpickle it (unless they have the same ascii file at the same path).  Maybe this is worth mentioning somewhere?

Alternatively, is there are reason not to pickle the LookupTable if that's how the throughput/flambda is being represented (other than being unwieldy)?

jmeyers314:
@rmjarvis , I'm finished looking through what you've got here so far.  Once you get the remaining classes done, it should be short work to get this merged I think.

galsim/bandpass.py
rmjarvis: Good point.  `func` needs to be deleted, since that's always a lambda expression.  But I changed it so `_tp` (and `_spec' in SED) is only deleted if it is not a LookupTable.  As you point out, these are picklable directly, and it's probably more efficient to do so rather than read them in from the file again.

rmjarvis:
OK.  I think I'm basically done here.  I managed to get a lot more of the Chromatic stuff picklable than I expected.  I was able to avoid some of the key lambda expressions, so now most of the ways to mess up the picklability is when the user directly supplies a function or lambda expression themselves.  

Of course, in many cases, there isn't really another option to do the kinds of things you would want to do with Chromatic objects, but so it goes.

Some items worth pointing out:
- I made a ChromaticTransformation that works the same way as the regular Transformation, which let me pull out the `_A` and `_fluxFactor` out of the main ChromaticObject class.  It also simplified a lot of the logic for setupInterpolation.
- I changed setupInterpolation to return an object of type InterpolatedChromaticObject, which simplified the logic even further.  I know we iterated on that design choice a fair bit, so if you don't like that change, we can revert it.  It's commit 93b1217.
- I changed ChromaticAtmosphere to a class rather than a function, so I could delay the application of the functional transformation until it is actually used.

So take a look.  I think each commit preserves the working state of the code and unit tests, so you should be able to look at the last few days of commits individually to get a sense of each basic change to the code.

galsim/chromatic.py
```diff
@@ -107,34 +106,28 @@ class ChromaticObject(object):
...
+        return 'galsim.ChromaticObject(%r)'%self.obj
+
+    def __str__(self):
+        return 'galsim.ChromaticObject(%s)'%self.obj
+
+    def __eq__(self, other): return repr(self) == repr(other)
+    def __ne__(self, other): return not self.__eq__(other)
+    def __hash__(self): return hash(repr(self))
 
     def setupInterpolation(self, waves, oversample_fac=1.):
```
jmeyers314: I'm wondering about the name of this method in the refactored code.  I think `setupInterpolation` sounds more like you're manipulating the current object (which used to actually be the case) instead of returning a new one.  Maybe just `interpolate` to go along with `dilate`, `shear`, etc.?

jmeyers314:
Hi Mike, 
I've looked through the rest of your changes.  I'm happy with the new class structure, and the code in general.  Looks ready to merge to me.

galsim/chromatic.py
rmjarvis: Fine with me.  Rachel?  This was your functionality.  Are you ok with changing the name of this function to `interpolate`?

rmandelb:
Mike, sorry I�ve been ignoring this.  Between the T/JP workshop Monday and Tuesday plus having invited some people to stay at CMU on Wednesday to work on projects, I have not had any time to take a good look at this.  Will do so today/tomorrow.

rmjarvis:
No rush.  :)

rmandelb:
> I changed setupInterpolation to return an object of type InterpolatedChromaticObject,

While I know I initially did things this way, I became quite convinced in our discussion that it was best to leave it as a ChromaticObject.  To me, the convincing point is that in principle if you do the interpolation well enough (i.e., at high enough resolution), then you should exactly recover the uninterpolated result when you call drawImage().  This is intrinsically different from applying some kind of transformation (which means the object should look different when you draw it, and where I think there's a much better argument for returning a different class).

rmandelb:
Sorry, I didn't finish my thought:
if you do not like my argument and prefer to go ahead with the new scheme of having setupInterpolation() return a new object, then I agree with Josh that the name of the method should be changed to make it clear that it returns a new object instead of doing something to the existing one.

galsim/bandpass.py
```diff
@@ -83,90 +83,122 @@ class Bandpass(object):
...
-                    test_wave = 700
-                try:
-                    tp = eval('lambda wave : ' + tp)
-                    tp(test_wave)
-                except:
-                    raise ValueError(
-                        "String throughput must either be a valid filename or something that "+
-                        "can eval to a function of wave. Input provided: {0}".format(throughput))
+        self._orig_tp = throughput  # Save this for pickling.
+        self._tp = _tp              # This is orig_tp turned into an actual function
```
rmandelb: Probably I am just being stupid, but I found this comment confusing.  ```self._tp``` is will hold the throughput turned into an actual function later on (after you call ```_initialize_tp```), but right now it's just None, isn't it?

This doesn't matter to users but I think this comment could be clarified to make the code easier to read/parse by developers.

galsim/bandpass.py
```diff
@@ -83,90 +83,122 @@ class Bandpass(object):
...
+
+        if blue_limit > red_limit:
```
rmandelb: Should check for ```>=``` instead of just ```>```?

galsim/bandpass.py
```diff
@@ -83,90 +83,122 @@ class Bandpass(object):
...
 
         # Assign blue and red limits of bandpass
-        if isinstance(tp, galsim.LookupTable):
-            if blue_limit is None:
-                blue_limit = tp.x_min
-            if red_limit is None:
-                red_limit = tp.x_max
+        if isinstance(self._tp, galsim.LookupTable):
+            if self.blue_limit is None:
+                self.blue_limit = float(self._tp.x_min)/self.wave_factor
```
rmandelb: Oh, is this the bug fix that you mentioned?  (originally it wasn't divided by ```self.wave_factor```)
Do we have unit tests that would catch this kind of thing if it comes up in future?

rmandelb: Oh never mind, I see that originally it would have been divided by ```self.wave_factor``` later on.  Ignore that question.

rmjarvis: Yep.  I added more units tests in test_bandpass, which is how I discovered the bug.  So yes, we have such tests now.

rmjarvis:
> While I know I initially did things this way, I became quite convinced in our discussion that it was best to leave it as a ChromaticObject. To me, the convincing point is that in principle if you do the interpolation well enough (i.e., at high enough resolution), then you should exactly recover the uninterpolated result when you call drawImage(). This is intrinsically different from applying some kind of transformation (which means the object should look different when you draw it, and where I think there's a much better argument for returning a different class).

I'm open to putting it back, but the counter-argument to this is that in fact, the interpolated version is _not_ exactly the same as the uninterpolated version.  So if we are being serious about having immutable types, then it should really return a different object.  Otherwise people may be surprised that their object works slightly differently after sending it off to some function, which they may not have written, that internally calls `setupInterpolation` on the object.

galsim/bandpass.py
rmjarvis: (If this wasn't the bug exactly, it was something like this.  One of the unit tests failed when it used Angstroms, and it was something about how the wave_factor was applied that didn't quite work correctly in that case.)

galsim/bandpass.py
```diff
@@ -83,90 +83,122 @@ class Bandpass(object):
...
+                    # If neither `blue_limit` nor `red_limit` is defined, then the Bandpass should
+                    # be able to be evaluated at any wavelength, so check.
+                    test_wave = 700
+                try:
+                    self._tp = eval('lambda wave : ' + self._orig_tp)
+                    self._tp(test_wave)
+                except:
+                    raise ValueError(
+                        "String throughput must either be a valid filename or something that "+
+                        "can eval to a function of wave. Input provided: {0}".format(throughput))
```
rmandelb: I think this should be ```self._orig_tp```, not ```throughput``` (not accessible in this routine).

galsim/bandpass.py
```diff
@@ -83,90 +83,122 @@ class Bandpass(object):
...
+        else:
+            self._tp = self._orig_tp
```
rmandelb: Should we not do some sanity checks of this, just like when there is string input?

galsim/base.py
```diff
@@ -269,18 +273,9 @@ def __truediv__(self, other):
     # Make a copy of an object
     def copy(self):
         """Returns a copy of an object.
```
rmandelb: Correct me if I'm wrong, but I think that if the GSObject has an RNG then the RNG is being copied, not duplicated, so we can't expect identical images (e.g., for a RealGalaxy).  This seems like a caveat that belongs in the docstring.

galsim/base.py
```diff
@@ -1561,6 +1650,35 @@ def getLamOverD(self):
         """
         return self.SBProfile.getLamOverD()
 
+    def getObscuration(self):
+        """Return the `lam_over_diam` parameter of this Airy profile.
```
rmandelb: copy/paste error in docstring?

galsim/chromatic.py
```diff
@@ -1132,52 +1004,121 @@ def ChromaticAtmosphere(base_obj, base_wavelength, **kwargs):
...
+        @returns the monochromatic object at the given wavelength.
+        """
+        return self.build_obj().evaluateAtWavelength(wave)
+
+    def drawImage(self, bandpass, image=None, integrator='trapezoidal', **kwargs):
+        """
+        See ChromaticObject.drawImage for a full description.
+
+        This version usually just calls that one, but if the transformed object has had
+        setupInterpolation called on it, then there is an optimization we can apply here.
```
rmandelb: Technically the optimization doesn't happen here, but in other routines, right?  It's in the choice of which drawImage() actually gets called.

galsim/chromatic.py
```diff
@@ -1252,6 +1189,237 @@ def evaluateAtWavelength(self, wave):
...
+                if hasattr(obj._frat, '__call__'):
+                    self._frat = lambda w: obj._frat(w) * flux_ratio
+                else:
+                    self._frat = obj._frat * flux_ratio
+
+        else:
+            self.original = obj
+            self._jac = jac
+            self._off = offset
+            self._frat = flux_ratio
```
rmandelb: Any reason not to call this ```self._flux_ratio```?  While skimming the code I didn't notice this line and only saw ```_frat``` later on, and wondered what it was (seems non-obvious to me).

galsim/nfw_halo.py
```diff
@@ -42,7 +42,15 @@ def __init__(self, omega_m=0.3, omega_lam=0.7):
         self.omega_lam = omega_lam
         self.omega_c = (1. - omega_m - omega_lam)
         #self.omega_r = 0
```
rmandelb: This has nothing to do with this PR,  but if we are ignoring radiation in most of the code, is there any reason to keep the two commented-out lines that involve ```omega_r```?

rmandelb:
A random thought that occurred to me as I was going through this (not done yet):
It would be nice to show the use of ```__str__``` and ```__eq__``` for some of the objects for which this functionality is newly enabled in a demo or two.

galsim/shear.py
```diff
@@ -76,169 +77,279 @@ class Shear(object):
...
+        return self._g.real * self._g2e(abs(self._g)**2)
+
+    def getE2(self):
+        """Return the e2 component of the distortion.
+        Note: s.getE2() is equivalent to s.e2
+        """
+        return self._g.imag * self._g2e(abs(self._g)**2)
+
+    def getE(self):
+        """Return the magnitude of the distortion |e1 + i e2| = sqrt(e1**2 + e2**2)
```
rmandelb: We seem to go back and forth in these docstrings between i and j for sqrt(-1).  Should probably choose just one and stick with it.

galsim/chromatic.py
rmjarvis: Just brevity.  For private (implementation-detail) attributes, I don't feel as obliged to be verbose.  But I can change it.

galsim/chromatic.py
rmjarvis: Technically it happens in the ChromaticTransformation.drawImage call.  But the user doesn't need to know that.  Maybe the user doesn't even need to know that there is an optimization at all.  So I could probably just remove this while para.

galsim/nfw_halo.py
rmjarvis: I assume they were put in as a starter implementation in case we eventually wanted to enable radiation.  I don't think it hurts to leave them here.

rmandelb: There are other places that would also have to change, but don't have comments.  So it seemed a bit inconsistent to me.  But I guess it doesn't matter, we can cross that bridge when/if we come to it.  :)

galsim/chromatic.py
rmandelb: That makes sense to me.  We talk about the optimization in other places that make more sense to me.

rmandelb:
> So if we are being serious about having immutable types, then it should really return a different object. 

OK.

galsim/shear.py
rmjarvis: I think the reason is that for getShear() I thought it made more sense to use the python version of i, which is j.  So I wanted to point out that getShear() was g1 + 1j * g2, which is how you would build the complex object in python.  Then that idea carried over into some others like getG and getBeta.

But for these, it's probably better to just use i (like I did here), since they are all just describing the math, not python, and i is the more normal thing to use for that.

rmjarvis:
> A random thought that occurred to me as I was going through this (not done yet):
It would be nice to show the use of __str__ and __eq__ for some of the objects for which this functionality is newly enabled in a demo or two.

I couldn't think of any place to use `__eq__`, but demo7 seems like a good place to use the str functionality.  See what you think about my latest commit: 9248b0d.

rmandelb:
> See what you think about my latest commit: 

Looks good to me.

Also, in case it wasn't clear, I did finish going over the diff yesterday and am happy for this to be merged once you're done with my comments (which I think may be the case already, based on the commits above that I glanced through).

rmjarvis:
Great.  I did address all the other comments you had yesterday.  Thanks for those!

I'll plan on merging tonight, and then do #644 on master.  Then we can work on releasing v1.3.

rmandelb:
Sounds good to me.

Issue #642, opened 2015-02-10
InterpolatedKImage


rmjarvis:
In scoping out what will be required for #640, Josh Realized it would be helpful to have a way to specify a profile in terms of its fourier image.  i.e. an InterpolatedKImage object.

I think this could share most of the implementation that is currently in SBInterpolatedImage.  Really, it would just need to change how xtab and ktab get built, using the input image for the latter rather than the former.  Then once they are built, most of the rest of the functions would work the same way.

So I think SBInterpolatedKImage could either be a subclass of SBInterpolatedImage or we could make a new base class that both could derive from.

jmeyers314:
Hi all,
I think I have a handle on the c++ part of this issue now.  I've added a `SBInterpolated` parent class for both `SBInterpolatedImage` and `SBInterpolatedKImage`, and roughly laid out which methods/variables are in common and which probably need to be in the derived classes.  I also completely excised the `MultipleImageHelper` class, which helped me see how `SBInterpolatedImage` worked.

What's left is the Fourier logic part.  I have a few basic questions on how to go about filling in _ktab and _xtab in `SBInterpolatedKImage`:

1. Do the `pad_factor`, `stepk`, and `maxk` parameters make sense for `SBInterpolatedKImage`?  It's not obvious to me what one would do with any of these if provided the Fourier-space image directly instead of calculating it from either a real-space image or some analytic expression.
2. Does dk (=kimage.scale from the python layer) need to be directly passed to the `SBInterpolatedKImage` constructor?  I didn't see anything like scale or dx in `BaseImage`, but maybe I missed it.
3. What's a reasonable size for _xtab?  I guess the simplest thing is to get a `goodFFTSize` on _ktab (or _ktab + _kinterp), pad _ktab to this size with zeros, and then transform.  I'm not sure if one needs to consider anything like `pad_factor` here or not.

Does that sound right?

rmjarvis:
I'm not sure I know all the answers here, but hopefully @gbernstein can correct me if I'm wrong about any of this.

First, I think we probably wouldn't need `pad_factor`.  Or any of the padding options that we have in the regular InterpolatedImage class.  

I think stepk can probably be taken as the scale of the Fourier image.  I guess we'd want to restrict the image to not have any complicated wcs.  Just a pixel scale, which people should be able to override I guess with a scale parameter to the InterpolatedKImage constructor.

[...]

Issue #297, opened 2012-10-16
Generating noise with a general physical correlation function


barnabytprowe:
We want to be able to do this for...
- Padding the `RealGalaxy`-progenitor HST training data images with noise similar to that which they already contain
- Whitening the noise in the final sheared, PSF-convolved images derived from such `RealGalaxy` objects

The maths behind doing this is pretty straightforward, see e.g. http://en.wikipedia.org/wiki/White_noise#Simulating_a_random_vector

I think that we would want to progress in two stages:
1. Generate a covariance matrix for the all the desired positions `(i, j)` in an output `Image` by reading off values from a tabulated (in general 2D, although with 2-fold symmetry) correlation function.
2. Use the covariance matrix to simulate a noise vector as described in the Wiki page above, then populate the output `Image` with this noise.

This is the issue for these changes.  I think this should all be written in C++, and I will use Mike's TMV routines for the linear algebraic heavy lifting.

PR #662, opened 2015-04-27
#642 InterpolatedKImage

jmeyers314:
Hi all,

This PR adds the InterpolatedKImage object, which can be used to create a GSObject by specifying samples of the Fourier-transform of a surface brightness profile.  

A few notes:
- In the c++ layer, I created the class `SBInterpolated`, which serves as the parent class for both `SBInterpolatedImage` and `SBInterpolatedKImage`.  In hindsight, there is actually quite a bit less in common between these two classes than I had initially anticipated, so it might make sense to take out this added layer of abstraction, though I didn't bother for the moment.
- I completely removed the MultipleImageHelper class from the c++ layer, which I thought made the SBInterpolatedImage code easier to understand.  If desired, I could put this back now that I've figured out what's going on.
- Since I don't currently have a use-case for real-space convolution or photon-shooting with InterpolatedKImage, I followed Mike's suggestion and didn't implement these or the SBInterpolatedKImage.xValue() methods.
- I computed (and then cached) the centroid by computing the DFT.  If there's an easier way, I'd love to hear it!
- There's nothing analogous in InterpolatedKImage to noise_pad in InterpolatedImage.

-Josh


gbernstein:
On Apr 27, 2015, at 6:54 PM, Josh Meyers <notifications@github.com> wrote:

> 	• I computed (and then cached) the centroid by computing the DFT. If there's an easier way, I'd love to hear it!
> 
You can calculate the centroid as a sum over your Fourier components using this (which needs some 2\pi factors to be correct…):


The \ast means convolution here.  FT(f) is what you already have and is evaluated on a grid.  The subtlety is that the Fourier transform of the function x is not convergent.  But this actually makes sense because your function is formally periodic once you’ve put it on a grid in k space.  So what you really want (and are doing with the DFT) is taking the centroid of f(x) times a boxcar function spanning the range +-L/2 where L is the total length of your transform.

Hence what you should plug into FT(x) is the Fourier transform of the function
g(x) = x (|x|<L/2), 0 otherwise.  This transform will be purely imaginary.

jmeyers314:
>You can calculate the centroid as a sum over your Fourier components

Thanks @gbernstein!  I figured there must be a more clever way to get the centroid.  I gave up after trying to convert the x into d/dk; clearly going down the wrong path.  It should be straightforward to put in your formula and figure out the necessary factors of pi.

jmeyers314:
> It should be straightforward to put in your formula and figure out the necessary factors of pi.

I was wrong.  The x-centroid integrand works out to something like 

2 i ( kx L cos(kx L/2) - 2 sin(kx L/2) sin(ky L/2) / (kx^2 * ky) * val(kx, ky)

and something similar for the y-centroid integrand.  The problem is that since these need to know about L, I can't implement them using a global function pointer passed to KTable::integrate.  I thought about overloading KTable::integrate to accept a member function pointer, but templated member function pointers aren't allowed in c++ and overloading specifically with a SBInterpolatedKImage::SBInterpolatedKImageImpl member function pointer seems rather inelegant.  Any suggestions from the c++ gurus out there?

gbernstein:
Does it help that kx*L/2 is always a multiple of pi, so the cos are just (-1)^n and sin=0?

jmeyers314:
>Does it help that kx*L/2 is always a multiple of pi, so the cos are just (-1)^n and sin=0?

Cool, I didn't notice that!  I'm not sure it's enough though, at least for using KTable::integrate, since I'd still need to figure out n using only kx, and no knowledge of L or dk.  

Maybe the right approach is to compute the centroids in the constructor using the Image objects passed there, instead of relying on KTable.

jmeyers314:
Hi all,

I got @gbernstein's magic Fourier centroiding algorithm to work.  I haven't explicitly timed it, but in principle it's now O(n) (since only the Fourier components with kx=0 or ky=0 contribute) instead of O(n^2 log n) for the 2D DFT.

There's no explicit hurry for this to be reviewed, I suppose, but I want to make sure it isn't forgotten!  Any takers?

rmjarvis:
I'll try to make time for this next week.  I've been swamped lately leading up to the DES meeting this week, which is now over.

rmandelb:
At the risk of sounding like I'm just saying "ditto", I was swamped preparing for the HSC weak lensing meeting this W-F, but hope to catch up on this and other things next week.

jmeyers314:
@rmandelb, @rmjarvis , sounds good.  Looking forward to your comments.

-Josh

galsim/interpolatedimage.py
```diff
@@ -550,6 +546,178 @@ def __setstate__(self, d):
...
+
+        # set up k_interpolant if none was provided by user, or check that the user-provided one
+        # is of a valid type
+        if k_interpolant is None:
+            self.k_interpolant = galsim.Quintic(tol=1e-4)
+        else:
+            self.k_interpolant = galsim.utilities.convert_interpolant(k_interpolant)
+
+        scale = self._real_kimage.scale
+        if self._stepk < scale:
```
rmandelb: Should perhaps emit a warning if we are ignoring the user-input ```stepk```.  For example, this could indicate that they are accidentally giving the routine input images that are not what they meant to give.

include/galsim/SBInterpolatedImage.h
```diff
@@ -248,20 +146,75 @@ namespace galsim {
...
+
+    class SBInterpolatedKImage : public SBInterpolated
+    {
+    public:
+        /**
+         * @brief Initialize internal quantities and allocate data tables based on a supplied 2D
+         * image.
+         *
+         * @param[in] realKImage  Real part of input Fourier-space Image
+         *                        (any of ImageF, ImageD, ImageS, ImageI).
```
rmandelb: I thought that in the python layer, these had to be floats, not ints?

include/galsim/SBInterpolatedImage.h
```diff
@@ -248,20 +146,75 @@ namespace galsim {
...
-        boost::shared_ptr<Interpolant> getKInterp() const;
 
     protected:
 
         class SBInterpolatedImageImpl;
 
-        // Regular SBProfile pimpl constructor so as to be available to derived classes
-        SBInterpolatedImage(SBProfileImpl* pimpl) : SBProfile(pimpl) {}
+        // Wow, three classes here... I'm not even sure how this works,
+        // but the tests seem to pass...
```
rmandelb: (1) This comment does not exactly inspire confidence.  :)
(2) Honestly, I agree.  I find this confusing.  If anyone wants to educate me more on this, I'm happy to learn.

src/SBInterpolatedImage.cpp
```diff
@@ -635,6 +444,82 @@ namespace galsim {
         }
     }
 
+    // JM: Does this ever get used?  _galsim.SBInterpolatedImage.__repr__ is defined
```
rmandelb: I think you're right, it doesn't get used.  Perhaps worth ripping out?

src/SBInterpolatedImage.cpp
```diff
@@ -900,7 +785,204 @@ namespace galsim {
...
+        ImageAlloc<double> im(N, N);
+        tmv::MatrixView<double> m(im.getData(), N, N, 1, im.getStride(), tmv::NonConj);
+        m += val.realPart();
+        return im;
+    }
+
     // instantiate template functions for expected image types
+
+    // JM: Do the integer templates ever actually get used?  InterpolatedImage raises an
+    //     error on non-floating point input.
```
rmandelb: Ah, I see this relates to my comment above about float/int in C++ vs. float in python.

rmandelb:
Hi Josh - I went over this just now.  It looks pretty good to me.  The unit tests pass on my Mac; I wasn't able to try on the linux cluster here yet but can do so tomorrow.

rmandelb:
Tests pass on the linux cluster too.

galsim/interpolatedimage.py
jmeyers314: Sounds reasonable.  I added warning if stepk is smaller than kimage.scale.

include/galsim/SBInterpolatedImage.h
jmeyers314: Ha!  Yeah, this was my way of calling attention to this line to see if I got it right!  Let's see if @rmjarvis has any comment here...

src/SBInterpolatedImage.cpp
jmeyers314: It's probably okay I suppose to just prohibit and excise all integer interpolated objects.  These wouldn't be very accurate in general, can be mimicked with floats and round(), and can always be put back in if someone has a good use case for them.  I'll go ahead and remove these if there are no objections.

jmeyers314: Ok.  These are now removed.

src/SBInterpolatedImage.cpp
jmeyers314: Just tried ripping this out, but then the compiler let me know that repr is a pure virtual method of SBProfileImpl, and so needs to be defined for all subclasses.  Perhaps it is callable somewhere via polymorphism of an SBProfile object?

I went ahead and removed the comment, but left the method.

include/galsim/SBInterpolatedImage.h
rmandelb: I think this line has to be updated now that you ripped out some stuff.

galsim/interpolatedimage.py
rmandelb: Thanks.

include/galsim/SBInterpolatedImage.h
jmeyers314: Yep, thanks.  Done.

galsim/interpolatedimage.py
```diff
@@ -25,6 +25,7 @@
 from . import _galsim
 from ._galsim import Interpolant
 from ._galsim import Nearest, Linear, Cubic, Quintic, Lanczos, SincInterpolant, Delta
+import numpy as np
```
rmjarvis: I know we do this in other files too, but I'd actually prefer to move in the other direction.  The more things we import at file scope the longer it takes to do `import galsim`.  Already it takes about a second.  So my preference would be to keep these imports inside the classes and functions that use them.

galsim/interpolatedimage.py
```diff
@@ -550,6 +546,182 @@ def __setstate__(self, d):
...
+    The InterpolatedKImage class is useful if you have a non-parametric description of the Fourier
+    transform of an object as a pair of Images -- one for the real part and one for the imaginary
+    part -- of the complex-valued profile that you wish to manipulate / transform using GSObject
+    methods such as shear(), magnify(), shift(), etc.  Note that neither real-space convolution nor
+    photon-shooting of InterpolatedKImages is currently implemented.  Please submit an issue at
+    http://github.com/GalSim-developers/GalSim/issues if you require either of these use cases.
+
+    The primary intended use case for InterpolatedKImage is to create ChromaticObjects from
+    PSF-deconvolved multi-band HST images, a functionality that will be added in a future
+    pull-request.
```
rmjarvis: This para is probably not necessary.  It doesn't really add anything to the understanding of InterpolatedKImage itself.

galsim/interpolatedimage.py
```diff
@@ -550,6 +546,182 @@ def __setstate__(self, d):
...
+
+    The images required for creating an InterpolatedKImage are precisely those returned by the
+    GSObject `.drawKImage()` method.  The `a` and `b` objects in the following command will produce
+    identical images when drawn with the `.drawImage()` method:
+
+    >>> a = returns_a_GSObject()
+    >>> b = galsim.InterpolatedKImage(*a.drawKImage())
```
rmjarvis: This isn't really true.  You might say "...will produce essentially equivalent images", but identical is definitely overstating the case here.
```
>>> import galsim
>>> a = galsim.Gaussian(flux=1.7, sigma=2.3)
>>> b = galsim.InterpolatedKImage(*a.drawKImage())
galsim/interpolatedimage.py:674: UserWarning: Provided Real and Imag kimages do not meet GSParams.maxk_threshold criterion.
  "Provided Real and Imag kimages do not meet GSParams.maxk_threshold criterion.")
>>> im1 = a.drawImage(scale=0.25)
>>> im2 = b.drawImage(scale=0.25)
>>> im1.array
array([[  1.84585705e-18,   3.53282801e-18,   6.64980917e-18, ...,
          6.69098084e-18,   3.57130845e-18,   1.88142065e-18],
       [  3.57958274e-18,   6.77390280e-18,   1.26729025e-17, ...,
          1.26795224e-17,   6.77823682e-18,   3.58197536e-18],
       [  6.70457721e-18,   1.26956036e-17,   2.37427471e-17, ...,
          2.37967852e-17,   1.27480775e-17,   6.75501538e-18],
       ..., 
       [  6.71199826e-18,   1.26954282e-17,   2.37347450e-17, ...,
          2.37287711e-17,   1.26900094e-17,   6.70712369e-18],
       [  3.63048496e-18,   6.84009835e-18,   1.27541324e-17, ...,
          1.26999736e-17,   6.78783087e-18,   3.58060141e-18],
       [  1.89215476e-18,   3.57907113e-18,   6.69621442e-18, ...,
          6.68293279e-18,   3.56788310e-18,   1.88293336e-18]], dtype=float32)
>>> im2.array
array([[ -5.54597435e-08,  -5.58401254e-08,  -5.62082931e-08, ...,
         -6.89198032e-08,  -6.68243842e-08,  -6.48945928e-08],
       [ -3.36760735e-08,  -3.32897372e-08,  -3.28244560e-08, ...,
         -5.21154071e-08,  -4.99594144e-08,  -4.79943765e-08],
       [ -6.82893431e-09,  -5.73644288e-09,  -4.51138993e-09, ...,
         -2.74646297e-08,  -2.55707686e-08,  -2.38654927e-08],
       ..., 
       [ -2.38654927e-08,  -2.55707686e-08,  -2.74646297e-08, ...,
         -4.51138993e-09,  -5.73644288e-09,  -6.82893431e-09],
       [ -4.79943765e-08,  -4.99594144e-08,  -5.21154071e-08, ...,
         -3.28244560e-08,  -3.32897372e-08,  -3.36760735e-08],
       [ -6.48945928e-08,  -6.68243842e-08,  -6.89198032e-08, ...,
         -5.62082931e-08,  -5.58401254e-08,  -5.54597435e-08]], dtype=float32)
>>> numpy.max(im1.array)
0.0031840832
>>> numpy.max(im2.array)
0.0031777034
>>> im3 = a.drawImage(scale=0.25, method='fft')
>>> im3.array
array([[  1.84585705e-18,   3.53282801e-18,   6.64980917e-18, ...,
          6.69098084e-18,   3.57130845e-18,   1.88142065e-18],
       [  3.57958274e-18,   6.77390280e-18,   1.26729025e-17, ...,
          1.26795224e-17,   6.77823682e-18,   3.58197536e-18],
       [  6.70457721e-18,   1.26956036e-17,   2.37427471e-17, ...,
          2.37967852e-17,   1.27480775e-17,   6.75501538e-18],
       ..., 
       [  6.71199826e-18,   1.26954282e-17,   2.37347450e-17, ...,
          2.37287711e-17,   1.26900094e-17,   6.70712369e-18],
       [  3.63048496e-18,   6.84009835e-18,   1.27541324e-17, ...,
          1.26999736e-17,   6.78783087e-18,   3.58060141e-18],
       [  1.89215476e-18,   3.57907113e-18,   6.69621442e-18, ...,
          6.68293279e-18,   3.56788310e-18,   1.88293336e-18]], dtype=float32)
>>> numpy.max(im3.array)
0.0031840832
>>> numpy.max(im3.array) - numpy.max(im2.array)
6.3797925e-06
```
So with `method='fft'` the max is closer to equal, so that's better, but still numerically different.  So "identical" isn't really accurate here.

galsim/interpolatedimage.py
```diff
@@ -550,6 +546,182 @@ def __setstate__(self, d):
...
+
+    The real- and imaginary-part Images must have the same data type, same bounds, and same scale.
+    The only wcs permitted is a simple PixelScale.  Furthermore, the complex-valued Fourier profile
+    formed by `real_kimage` and `imag_kimage` must be Hermitian, since it represents a real-valued
+    real-space profile.  The outputs from `drawKImage` automatically meet all of these criteria.
```
rmjarvis: I'm not sure how useful it is to mention that the output of drawKImage meets these criteria.  It reads to me like you are implying that this is the recommended way for the user to generate these images.  

But presumably if you had an object with which you could call `drawKImage`, then you would not build an InterpolatedKImage with it.  You already have it as a GSObject.  So I'd recommend removing this last sentence.  The previous items are what the user needs to know.

galsim/interpolatedimage.py
```diff
@@ -550,6 +546,182 @@ def __setstate__(self, d):
...
+
+    def __init__(self, real_kimage, imag_kimage, k_interpolant=None, stepk=None,
+                 gsparams=None):
+
+        # make sure real_kimage, imag_kimage are really `Image`s, are floats, and are congruent.
+        if not isinstance(real_kimage, galsim.Image) or not isinstance(imag_kimage, galsim.Image):
+            raise ValueError("Supplied kimage is not an Image instance")
+        if ((real_kimage.dtype != np.float32 and real_kimage.dtype != np.float64)
+            or (imag_kimage.dtype != np.float32 and imag_kimage.dtype != np.float64)):
+            raise ValueError("Supplied image is not an image of floats or doubles!")
```
rmjarvis: "double" is a C++ term.  Just float is fine here for the error text, even though we also allow np.float64.  Or you can write out `float32 or float64`.  These are the terms we use in the Image class doc string.  Or perhaps "floating point Image" would be similarly descriptive.

galsim/interpolatedimage.py
```diff
@@ -550,6 +546,182 @@ def __setstate__(self, d):
...
+        if real_kimage.bounds != imag_kimage.bounds:
+            raise ValueError("Real and Imag kimages must have same bounds.")
+        if real_kimage.scale != imag_kimage.scale:
+            raise ValueError("Real and Imag kimages must have same scale.")
+
+        # Make sure any `wcs`s are `PixelScale`s.
+        if ((real_kimage.wcs is not None
+             and not isinstance(real_kimage.wcs, galsim.PixelScale))
```
rmjarvis: You can also use the shorthand `wcs.isPixelScale()`.

galsim/interpolatedimage.py
```diff
@@ -550,6 +546,182 @@ def __setstate__(self, d):
...
+        else:
+            maxk_threshold = gsparams.maxk_threshold
+        edge = np.zeros_like(real_kimage.array, dtype=bool)
+        edge[[0,-1],:] = True # top and bottom
+        edge[:,[0,-1]] = True # left and right
+        edge_max = np.max(np.abs(real_kimage.array[edge] + 1j*imag_kimage.array[edge]))
+        if (edge_max / real_kimage.array.max()) > maxk_threshold:
+            import warnings
+            warnings.warn(
+                "Provided Real and Imag kimages do not meet GSParams.maxk_threshold criterion.")
```
rmjarvis: Considering that this got triggered in my example above with `galsim.Gaussian(flux=1.7, sigma=2.3)`, is there a bug in `drawKImage`?  Or in this test?  Seems like that shouldn't have triggered a warning, right?

I did a quick look, and the max edge value is 0.01, compared to the maximum of 1.7, which gives a ration of 0.006, quite a bit more than maxk_threshold = 0.001.  So I'm thinking this is pointing to a bug in drawKImage.

galsim/interpolatedimage.py
```diff
@@ -550,6 +546,182 @@ def __setstate__(self, d):
...
+        # an image to be pickled, but at least it will be through the normal pickling rules,
+        # rather than the repr.
+        d = self.__dict__.copy()
+        del d['SBProfile']
+        return d
+
+    def __setstate__(self, d):
+        self.__dict__ = d
+        self.__init__(self._real_kimage, self._imag_kimage, k_interpolant=self.k_interpolant,
+                      stepk=self._stepk, gsparams=self._gsparams)
```
rmjarvis: I don't think you want to repeat all the tests that are in the `__init__` function.  Just need to rebuild the SBProfile:
```
self.__dict__ = d
self.SBProfile = galsim._galsim.SBInterpolatedKImage(
            self._real_kimage.image, self._imag_kimage.image,
            self._real_kimage.scale, self._stepk, self.k_interpolant, self._gsparams))
```

include/galsim/SBInterpolatedImage.h
```diff
@@ -29,109 +29,29 @@
...
-     * weighted sum of several component images.  The idea is that the component
-     * images would be constant, but the weights might vary across the field of view.
-     * (E.g. they could be principal components of the PSF).
-     *
-     * This class stores those images along with helpful derived information
-     * (most notably, the fourier transforms), so that each SBInterpolatedImage
-     * doesn't have to recalculate everything from scratch.
-     */
-    class MultipleImageHelper
+    class SBInterpolated : public SBProfile
```
rmjarvis: I don't see the point of this extra level of indirection.  The `SBInterpolated` class doesn't seem to provide any significant code reuse.  It and its Impl class both have trivial constructors.  And the only function that it defines is `getKInterp`, which is also trivial.  So I think it would be better to just have both `SBInterpolatedImage` and `SBInterpolatedKImage` derive directly from `SBProfile`.

include/galsim/SBInterpolatedImage.h
rmjarvis: This wasn't actually used at all, so you could have just removed it.  If you comment out this line on master, everything still compiles and passes.  So I suspect it was a remnant from something old that we don't actually do anymore.

The corresponding line in SBProfile is the important one.  That's the constructor that builds the base SBProfile object with whatever the appropriate pimpl type is.  Something derived from SBProfileImpl, such as SBInterpolatedImageImpl.  This is where all the magic is that makes SBProfiles act like python objects, being trivially copyable and such.  

Of course, all that is probably obsolete now that we've made GSObjects immutable.  It will probably go away in my grand redesign of the back end that I've been thinking about for some time now.  (The goal being to get rid of our dependence on boost.python.)

include/galsim/SBInterpolatedImageImpl.h
```diff
@@ -26,9 +26,58 @@
...
+
+        bool isAxisymmetric() const { return false; }
+
+        // We'll use false here, but really, there's not an easy way to tell.
+        // Certainly an Image _could_ have hard edges.
+        bool hasHardEdges() const { return false; }
+
+        // This class will be set up so that both x and k domain values
+        // are found by interpolation of a table:
+        bool isAnalyticX() const { return true; }
```
rmjarvis: This should be false for `SBInterpolatedKImage`.  At least until we implement `xValue`.  That's really what this (poorly-named) function is about.  If this is true, then `xValue` is implemented.  A better name for these would be something like `hasDirectlyCalculableXValues`.

src/SBInterpolatedImage.cpp
```diff
@@ -900,6 +782,198 @@ namespace galsim {
...
+        dbg<<"kimage bounds = "<<realKImage.getBounds()<<std::endl;
+        assert(_kInterp.get());
+
+        Ninitial = std::max(realKImage.getXMax()-realKImage.getXMin()+1,
+                            realKImage.getYMax()-realKImage.getYMin()+1);
+        dbg<<"Ninitial = "<<Ninitial<<std::endl;
+        Nk = goodFFTSize(int(Ninitial));
+        dbg<<"Nk = "<<Nk<<std::endl;
+
+        _ktab = boost::shared_ptr<KTable>(new KTable(Nk, _dk));
```
rmjarvis: Probably worth adding a comment in here somewhere about the difference between dk and stepk, since they are both nominally the same thing.  So it's worth clarifying for future developers who may need to understand this code.

dk is the scale of the input image, and the KTable is constructed with the same scale.  stepk on the other hand is the maximum scale that a *rendered* Fourier image will have.  So stepk is only used in the calculations done by drawImage to figure out how to render an image, not in any of the stuff here.

src/SBInterpolatedImage.cpp
```diff
@@ -900,6 +782,198 @@ namespace galsim {
...
+                 _ktab->kSet(kx, ky, kvalue);
+                 xxdbg<<"ikx,iky,kx,ky = "<<ikx<<','<<iky<<','<<kx<<','<<ky<<std::endl;
+                 xxdbg<<"kvalue = "<<kvalue<<std::endl;
+             }
+        }
+        _flux = kValue(Position<double>(0.,0.)).real();
+        dbg<<"flux = "<<_flux<<std::endl;
+
+        xcentroid = NAN;
+        ycentroid = NAN;
```
rmjarvis: Let's not use NAN as a flag that these haven't been set yet.  The support for NAN, and particularly isnan, in C and C++ has kind of a sordid history.  I'm not confident that `std::isnan` will work correctly on all compilers and OS's.  

Better to just set a bool variable `centroidIsDefined = false` here, and check for that later.

src/SBInterpolatedImage.cpp
```diff
@@ -900,6 +782,198 @@ namespace galsim {
...
+                which works out to
+                2 i ( kx L cos(kx L/2) - 2 sin(kx L/2)) sin (ky L/2) / kx^2 / ky.
+                Noting that kx L/2 = ikx pi, the cosines are -1^ikx and the sines are 0.
+                Of course, lim kx->0 sin(kx)/kx is 1 though, so that term survives.  Algebra
+                eventually reduces the above expression to what's in the code below.
+             */
+            double xsum(0.0), ysum(0.0);
+            for (int iky = -Ninitial/2; iky < Ninitial/2; iky++) {
+                if (iky == 0) continue;
+                ysum += std::pow(-1.0, iky) / iky * _ktab->kval(0, iky).imag();
```
rmjarvis: A minor efficiency point here.  Rather than `pow(-1.0, iky)`, you can just keep a variable that keeps track of this and update it each time through the loop.
```
int iky = -Ninitial/2
double sign = (iky % 2 == 0) ? 1.0 : -1.0;
for (; iky < Ninitial/2; iky++) {
    sign = -sign;
    ...
```

src/SBInterpolatedImage.cpp
```diff
@@ -900,6 +782,198 @@ namespace galsim {
...
+    {
+        int N = _ktab->getN();
+        tmv::Matrix<std::complex<double> > val(N,N);
+        int Nko2 = Nk/2;
+        fillKValue(val.view(), -Nko2, 1, Nko2, -Nko2, 1, Nko2);
+        ImageAlloc<double> im(N, N);
+        tmv::MatrixView<double> m(im.getData(), N, N, 1, im.getStride(), tmv::NonConj);
+        m += val.realPart();
+        return im;
+    }
```
rmjarvis: This isn't what we want here.  The only use of these two functions is to be able to serialize the SBInterpolatedKImage more efficiently than the version we get by using `repr`.  Calling `fillKValue` is not very efficient.  

For SBInterpolatedImage, it was fine to take the view of the XTable, since that has the right shape and it doesn't actually require copying any values to make the ImageView, so it is efficient to pass that back and reasonably efficient to remake the SBInterpolatedImage from that.

Here, KTable stores things in a different order than the input images, so getting a view of the real and imag parts wouldn't be the right thing to input back as a the real and imag images in the constructor. Which is why you implemented these functions this way.  But if we're going to do all this, it's probably better to just keep the originals around and use them again.

However, I think a better option would be to make an alternate version of the constructor that would accept the raw KValue data as input and efficiently recreate the KTable.  I'd suggest adding the centroid values to the arg list so if they have been calculated, then it would avoid recalculating those as well.  (You'd also need the bool flag of whether they are calculated or not yet.)

Then instead of these two functions, we would have one function that returns the KValue data packaged up as an Image, but not really being any kind of useful image.

```
template <typename T>
    SBInterpolatedKImage::SBInterpolatedKImageImpl::SBInterpolatedKImageImpl(
        const BaseImage<double>& data,
        double dk, double stepk, boost::shared_ptr<Interpolant2d> kInterp,
        double xcen, double ycen, bool cenIsSet, const GSParamsPtr& gsparams) :
    ...,
    _xcentroid(xcen), _ycentroid(ycen), _cenIsSet(cenIsSet)
{
    ...
    
    Nk = data.getYMax()-data.getYMin()+1)
    _ktab = boost::shared_ptr<KTable>(new KTable(Nk, _dk));
    std::complex<double>* kptr = _ktab.getArray();
    const double* ptr = data.getData();
    for(int i=Nk*Nk; i>0; --i, ptr+=2)
        *kptr++ = std::complex<double>(*ptr, *ptr+1);
}

ConstImageView<double> SBInterpolatedKImage::SBInterpolatedKImageImpl::getKData() const
{
    int Nk = _ktab->getN();
    data = reinterpret_cast<const double*>(_ktab->getArray())
    return ConstImageView<double>(data, boost::shared_ptr<double>(),
                                  2*N, Bounds<int>(0,2*N-1,0,N-1));
}
```
I didn't try this out, so there could be bugs in the above, but this should be close.

include/galsim/SBInterpolatedImageImpl.h
```diff
@@ -175,6 +194,56 @@ namespace galsim {
...
+        { throw SBError("SBInterpolatedKImage::shoot() is not implemented"); }
+
+    protected:  // Made protected so that these can be used in the derived CorrelationFunction class
+
+        std::string repr() const;
+        double _dk;
+        int Ninitial;
+        int Nk;
+        mutable double xcentroid;
+        mutable double ycentroid;
```
rmjarvis: These should have leading underscores.  Our style in the C++ layer has been that member variables get a leading underscore to mark them as such.  Kind of like the `self.` prefix in the python layer for member variables.  But shorter.

tests/test_interpolatedimage.py
```diff
@@ -1036,6 +1036,72 @@ def test_stepk_maxk():
...
+def test_kround_trip():
+    a = final
+    real_a, imag_a = a.drawKImage()
+    b = galsim.InterpolatedKImage(real_a, imag_a)
+
+    # Check picklability
+    do_pickle(b)
+    do_pickle(b, lambda x: x.drawImage())
+    # do_pickle(b.SBProfile)
+    # do_pickle(b.SBProfile, lambda x: repr(x))
```
rmjarvis: These should work too.  Are the commented out because they failed?  Or too slow?

tests/test_interpolatedimage.py
```diff
@@ -1036,6 +1036,72 @@ def test_stepk_maxk():
...
+        np.testing.assert_array_almost_equal(img_a.array, img_b.array, 5)
+
+    # Try some additional transformations:
+    a = a.shear(g1=0.2, g2=-0.2).shift(1.1, -0.2).dilate(0.7)
+    b = b.shear(g1=0.2, g2=-0.2).shift(1.1, -0.2).dilate(0.7)
+    img_a = a.drawImage()
+    img_b = b.drawImage(img_a.copy())
+    # Fails at 6th decimal
+    np.testing.assert_array_almost_equal(img_a.array, img_b.array, 5,
+                                         "Transformed InterpolatedKImage image drawn incorrectly.")
```
rmjarvis: Maybe worth adding a test of an InterpolatedKImage convolved by something else?  Just to be more complete in the ways people will use the class.

galsim/interpolatedimage.py
jmeyers314: Sure.  Removed.

galsim/interpolatedimage.py
jmeyers314: Fair enough.  Changed "identical" -> "essentially equivalent" as suggested.

Out of curiosity though, I'm not sure what you mean by "with `method='fft'` the max is closer to equal".  Aren't im1 and im3 identical above?  (And doesn't Gaussian * Pixel default to method=fft?)

galsim/interpolatedimage.py
jmeyers314: >It reads to me like you are implying that this is the recommended way for the user to generate these images.

I think my point was that you could use drawKImage if you wanted an example of valid input.  I changed the sentence to "(To see an example of valid input to `InterpolatedKImage`, you can look at the output of `drawKImage`)."

galsim/interpolatedimage.py
jmeyers314: Changed message (here and in `InterpolatedImage`) to:

"Supplied image does not have dtype of float32 or float64!"

galsim/interpolatedimage.py
jmeyers314: Done.

galsim/interpolatedimage.py
jmeyers314: Yep.  Changed.

include/galsim/SBInterpolatedImage.h
jmeyers314: Sounds like the least confusing thing is to just delete the line then, which is what I've now done.

include/galsim/SBInterpolatedImageImpl.h
jmeyers314: >Our style in the C++ layer has been that member variables get a leading underscore to mark them as such.

Ah, okay.  I didn't realize the full significance here.  Changed.

src/SBInterpolatedImage.cpp
jmeyers314: OK.  Good to know.  Switched to `centroidIsDefined`.

src/SBInterpolatedImage.cpp
jmeyers314: Cool.  Done.

galsim/interpolatedimage.py
rmjarvis: You're right.  Sorry.  For some reason, I thought the first one was doing real-space rendering, but of course it isn't, as you pointed out.  So I thought I was doing something with the 'fft' specification.

galsim/interpolatedimage.py
jmeyers314: Would this slow down the classes/functions that have to do the importing?  I think I'd rather have the package be slow to initially import than slow down any of the individual functions.  Probably worth benchmarking both options.

rmjarvis: I don't believe so.  I think python only actually reads in the files the first time a module is imported.  After that it's just about letting that namespace be visible in the current scope.

rmjarvis: In any case, import numpy is pretty quick, so this probably isn't too important.  I suspect the one that is slowing us down is the `try: import astropy.io.fits; except: import pyfits` bit in our `__init__.py` file.  That one takes a bit longer.  I should see if I can find a way to delay that until we need it.

src/SBInterpolatedImage.cpp
jmeyers314: Okay.  I think most of that makes sense.  Let me check: the next steps would be something like: expose the alternative c++ constructor in python in pysrc/SBInterpolatedImage.cpp, and then write `SBIKI.__getinitargs__()` assuming that this alternative constructor will get called upon unpickling?  (I assume you're not suggesting I replace the current constructor...)  Is that right?

rmjarvis: Yes.  This alternate constructor would only be used for the serialization.  The current constructor would still be the one that InterpolatedKImage would call.  And boost python should handle the distinction for you via the different argument signatures.

jmeyers314: Alright.  I think I got this point working.  A couple questions though:

- I got a little confused deciding if there needed to be both `Interpolant` and `Interpolant2d` versions of the alternative constructor.  I landed on only `Interpolant`, since the object returned by `getKInterp` is always the 1d version.  This raises the question though that maybe the original 2d constructor should be coded as `InterpolantXY` instead of `Interpolant2d`, since `getKInterp` assumes that the `get1d` method exists.  It's quite possible I've missed something subtle involving the static_cast in getKInterp though.
- I wrote `repr` to use `getKData()` instead of `getRealKImage` and `getImagKImage`, which, as pointed out, is more efficient but does potentially mean that the `repr` string is less human-understandable.  Since the real/imag version is also pretty arduous though, maybe this isn't so important.

tests/test_interpolatedimage.py
jmeyers314: These used to fail, but now work after rewriting the pickability stuff above.

src/SBInterpolatedImage.cpp
rmjarvis: I got rid of all the XY and 2d interpolants in the python layer, since we don't ever use any kind of Interpolant2d other than via an XY version of a 1d interpolant.  So now everything is just Interpolant, which means the 1d interpolant used in both directions.

I didn't try to go through the C++ layer to clean up all the places where we assume that a 2d interpolant is really an XY interpolant, but I'm hard pressed to imagine a reason to ever have one that isn't.  I'm kind of inclined to just make it official and remove the Interpolant2d and InterpolantXY classes entirely and just always use the 1d classes, but I didn't try to do so yet.

TL;DR: what you did is fine.  :)

jmeyers314: >I'm kind of inclined to just make it official and remove the Interpolant2d and InterpolantXY classes entirely and just always use the 1d classes, but I didn't try to do so yet.

Sounds reasonable to me.  It would be a pretty weird interpolation kernel indeed that isn't at the very least separable (although maybe this comes up if you're interpolating in curved space like on the surface of a sphere?).  I'll leave the c++ layer as is to await more targeted cleanup.

tests/test_interpolatedimage.py
jmeyers314: Added a simple test for InterpolatedKImage convolved by a Kolmogorov.

galsim/interpolatedimage.py
jmeyers314: >I'm thinking this is pointing to a bug in drawKImage.

Hmm...  Looks like it missed by one pixel:
```
a = galsim.Gaussian(flux=1.7, sigma=2.3)
re, im = a.drawKImage()
print re.array.shape # (14, 14)
print re.array[7,7] # 1.7
print re.array[7, 7+6] # 0.0100987
print a.kValue(0, 6*a.stepK()) # 0.0100987
print a.kValue(0, 6*a.stepK())/1.7 # 0.005940
print a.kValue(0, 7*a.stepK())/1.7 # 0.000933
```

I think it makes sense to truncate images such that they include the nearest pixel beyond where the threshold gets crossed, rather than at the last pixel inside of the threshold.  Perhaps increment by 2 rows/columns in SBProfile::getGoodImageSize()?

rmjarvis: I don't know.  Based on this, I think drawKImage got it right.  The pixels that were omitted were ok to omit according to the official definition of what maxk_threshold means.

I think I'd rather just remove this warning.  It's similar to the case of someone providing an image when there is still flux at larger scales than they included in the image.  InterpolatedImage just implicitly assumes the surface brightness is zero outside the bounds of the provided image.  We don't warn them if there is still significant flux at the edge of the image.

I think it probably makes sense to do the same here.  Any kvalues off the edge of the provided images are just taken to be zero.  I'm not sure we need a warning if the kvalues are still marginally significant near the boundary.

jmeyers314: >We don't warn them if there is still significant flux at the edge of the image.

OK. Good point.  I also checked to see if adding the extra row/col made any difference in how tightly I could set the unit test tolerances.  There's no improvement, so I'll just remove the warning.

src/SBInterpolatedImage.cpp
jmeyers314: Added the following comment (and assert) to SBInterpolatedKImageImpl constructor:
```
        // Note that _dk is the pitch of realKImage and imagKImage.  In contrast, _stepk indicates
        // the maximum pitch for drawImage() to use when rendering an image, which may be set
        // larger than _dk for efficiency.
        assert(_stepk >= _dk);
```

include/galsim/SBInterpolatedImageImpl.h
jmeyers314: Done.  (In newly de-abstracted SBInterpolatedKImage)

include/galsim/SBInterpolatedImage.h
jmeyers314: Yep.  I initially assumed there would be more overlap which is why this exists.  But I agree with you it's not very useful in hindsight.  I've removed this layer.

galsim/interpolatedimage.py
jmeyers314: Oh, and one last comment on this comment.  The drawKImage docstring is lying about the default scale being the nyquistScale right?  It's actually stepK() (which I guess is kind of like the Nyquist scale of the Fourier transform, but I still found this confusing).

galsim/interpolatedimage.py
jmeyers314: In the spirit of PEP 8 then, I think I'll leave this as is.

jmeyers314:
@rmjarvis,  I think I've hit all your comments now.  Let me know if you see anything new or something I missed.

rmjarvis:
All looks good to me.  

I removed some more leftovers from PR #297, which is where that [extra constructor](https://github.com/GalSim-developers/GalSim/commit/6591e2d915f80cda4750d88d1c83d422d0be057f) was introduced to enable the CorrelationFunction class, which we then abandoned.

I also fixed a minor error in an err_msg format string, which my python2.6 system complained about.

jmeyers314:
>All looks good to me.

Great, thanks for getting through this @rmjarvis and @rmandelb (also with tips from @gbernstein).  I'll leave this open another day or two before merging, just in case anyone else wants to take a look.

>I removed some more leftovers from PR #297, which is where that extra constructor was introduced to enable the CorrelationFunction class, which we then abandoned.

Ah, okay.  I looked around a bit for this can couldn't find it.  This makes more sense now that I know it was abandoned.

Issue #663, opened 2015-05-01
latest version of GalSim broke WFIRST module getPSF()


rmandelb:
The GalSim WFIRST module getPSF routine uses setupInterpolation(), which was removed in the latest version of GalSim.  Need to fix this so that the routine is usable again.  Also, should have a unit test that would have caught this.

I'll take care of it within a few days.

rmjarvis:
Sorry. I thought I grepped for all instances of setupInterpolation when I was making that switch.  I must have missed going one directory further down in the directory structure.

rmandelb:
No worries, it's easily fixed.  :)  And now that I play around with it, I realize that there were unit tests in ```test_wfirst.py``` that would have caught this if we ran the script directly (but not as part of ```scons tests```).  I just moved the fastest of those out of the ```if``` statement so it gets run with ```scons tests```.

However - the problem I'm having is that this test is now failing, even though it's one that should pass at very high precision.  The test is as follows:
- We had previously made an achromatic WFIRST PSF at a particular wavelength (```use_lam=900 nm```) and drew it into an image.
- We then make an interpolated WFIRST PSF, for which one of the wavelengths used for interpolation is 900 nm.
- We evaluate the interpolated WFIRST PSF at 900 nm and draw it into an image.

Previously the images from the original achromatic PSF and the interpolated one agreed to 8 decimal places, which I think should trivially be the case.  Now, the test fails unless I restrict it to 4 decimal places.  You can see this is if you go to branch #663 and run ```test_wfirst.py```.  This change in precision suggests to me that something important might have actually changed.  Can you think offhand of anything from your own code rearrangements in chromatic.py that might cause this?  If not, I will start digging into this further.

rmandelb:
@rmjarvis - do you have any comments or suggestions on what I was finding with the unit tests? (https://github.com/GalSim-developers/GalSim/issues/663#issuecomment-98729298)

jmeyers314:
[...]

PR #665, opened 2015-05-22
#663

rmandelb:
This short and simple PR is fixing a place in the WFIRST module that relied on an older version of some of the chromatic object code, and also modifying the unit tests to: (a) account for some of the recent chromatic changes and (b) run some more of the tests when ```scons tests``` is called instead of reserving them for when ```test_wfirst.py``` is run directly.

### DIFF
--- a/galsim/wfirst/wfirst_psfs.py
+++ b/galsim/wfirst/wfirst_psfs.py
@@ -279,7 +279,8 @@ def storePSFImages(PSF_dict, filename, bandpass_list=None, clobber=False):
     SCA_index_list = []
     for SCA in PSF_dict.keys():
         PSF = PSF_dict[SCA]
-        if not isinstance(PSF, galsim.ChromaticOpticalPSF):
+        if not isinstance(PSF, galsim.ChromaticOpticalPSF) and \
+                not isinstance(PSF, galsim.InterpolatedChromaticObject):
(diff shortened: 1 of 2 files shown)


jmeyers314:
Looks good to me.  

I have one comment, which actually applies to code written many commits ago but occurs near the current commits: I think test_wfirst.py:396 makes more sense as
```
  obj_chrom = psf_chrom.evaluateAtWavelength(use_lam)
```
than as currently written:
```
  obj_chrom = psf_achrom.evaluateAtWavelength(use_lam)
```

rmandelb:
Whew.  Yes, Josh, you are right.  As written, ```psf_chrom``` was never actually used for anything!  Thanks for your careful attention to detail; I've fixed it now (the test still passes, fortunately).

rmandelb:
@rmjarvis , are you interested in taking a look at this?  This is a pretty minor change (but important for the demo to be functional) so I'd like to merge by the end of this week.  If you don't have time before then, I think this one should be okay with just one person going over it.

rmjarvis:
I just did a cursory check, and it seems fine to me.  I'll try to find time to catch up on the other PRs that are waiting for me.  Been super hectic with DES lately.

rmandelb:
Thanks for looking it over!  Since you did check it out, I'll plan to merge tonight unless I hear objections from anybody else.

Issue #666, opened 2015-05-22
Interleaving images


arunkannawadi:
GalSim's build-in methods to generate surface brightness profiles, for e.g. ```galsim.Gaussian```, ```galsim.OpticalPSF``` etc., can be explicitly convolved with the native pixel response and then drawn using a smaller pixel scale to give images that are finely sampled. However, when pixel level effects, like non-linearity and IPC have to be incorporated, then the images have to be drawn at the native pixel scale, which are typically sampled below Nyquist rate. Highly sampled images including these pixel level effects can be obtained from a dither sequence with uniform offsets by interleaving them, without having to invoke external image combination softwares.

arunkannawadi:
The implementation of  ```galsim.utilities.interleave``` as it is now, relies on the user giving the list of images in the right order of offsets. The default ordering is a little unnatural and can be confusing for the user. Moreover, such errors from the user's side cannot be caught through unit tests. Thus, it is recommended that the user provide a list of offsets corresponding to each of the images in the list.

arunkannawadi:
Re-opening this issue.

As mentioned in the first comment, simulating the effect of detector effects like IPC requires generation of individual low-resolution exposures. The `interleaveImages` routine combines a dither sequence (obtained in some specific way, suitable only for simulations) of multiple exposures to obtain a well resolved image. In GalSim, it is much faster to draw a high resolution image of a profile than to draw multiple low-resolution exposures, even if the number of sampling points is the same. Therefore, it is only natural to have an `deInterleaveImages` routine that does the opposite of `interleaveImages` routine, i.e., given a high resolution image, it generates a dither sequence by slicing the image. I am re-opening this issue to address this aspect.

PR #667, opened 2015-05-27
Interleaving Images (#666)

arunkannawadi:
This is a PR for the new routine ``interleaveImages`` in galsim/utilities.py. This routine can combine multiple coarsely sampled images and return a finely sampled image if the original set of images differed by equally spaced offsets. This routine is mainly of interest since it can increase the resolution of the image after incorporating pixel-level effects, like nonlinearity, interpixel capacitance and brighter-fatter effect.

### DIFF
--- a/galsim/utilities.py
+++ b/galsim/utilities.py
@@ -498,9 +498,14 @@ def interleaveImages(im_list, N, offsets, add_flux=True, suppress_warnings=False
 
     if isinstance(im_list[0],galsim.Image):
         y_size, x_size = im_list[0].array.shape
-        scale = im_list[0].scale
+        wcs = im_list[0].wcs
     else:
         raise TypeError("'im_list' must be a list of galsim.Image instances")
+
+    if wcs is None:
+        scale = im_list[0].scale
+    else:
+        scale = wcs.PixelScale
 
     for im in im_list[1:]:
         if not isinstance(im,galsim.Image):
@@ -542,10 +547,12 @@ def interleaveImages(im_list, N, offsets, add_flux=True, suppress_warnings=False
         # Fix the flux normalization
         img /= 1.0*len(im_list)
 
-    if (n1==n2):
-        if scale is not None:
-            img.scale = im_list[0].scale*(1./n1)
+    if wcs is not None:
+        img.wcs = galsim.Jacobian(1.*scale/n1, 0., 0., 1.scale/n2)
+    elif scale is not None:
+        if (n1==n2):
+            img.scale = 1.scale/n1
     elif suppress_warnings is False:
         import warnings
-        warnings.warn("Interleaved image could not be assigned a pixel scale automatically")
+        warnings.warn("Interleaved image could not be assigned a PixelScale automatically.")
     return img
--- a/galsim/utilities.py
+++ b/galsim/utilities.py
@@ -421,16 +421,33 @@ _gammafn._a = ( 1.00000000000000000000, 0.57721566490153286061, -0.6558780715202
 def interleaveImages(im_list, N, offsets, add_flux=True, suppress_warnings=False,
     catch_offset_errors=True):
     """
-    Interleaves two or more images and outputs a larger image.
+    Interleaves the pixel values from two or more images and into a single larger image.
 
-    The sampling length of simulated images can be set arbitrarily using the `pixel_scale' argument
-    in drawImage() routine appropriately. However, pixel level detector effects can be included
-    only on images drawn at the native pixel scale, which are typically undersampled. Nyquist
-    sampled images that also include the effects of detector non-idealities can be obtained by
-    drawing multiple undersampled images (with the detector effects included) that are offset from
-    each other by a fraction of a pixel. This is equivalent to obtaining a finer sampled image from
-    a dither sequence, except that this routine handles only equispaced offsets, centered at 0. The
-    dither sequence  must be a list of galsim.Image instances supplied through 'im_list'.
+    This routine can be used to build a Nyquist-sampled image from a set of images that were
+    observed with pixels larger than the Nyquist scale. This routine converts a list of images taken
+    at a series of (uniform) dither offsets into a single higher resolution image, where the value
+    in each final pixel is the observed pixel value from exactly one of the original images.
+
+    In the original observed images, the integration of the surface brightness over the pixels is
+    equivalent to convolution by the pixel profile and then sampling at the centers of the pixels.
+    This procedure simulates an observation sampled at a higher resolution than the original images,
+    while retaining the original pixel convolution. Such an image can be obtained in a fairly
+    simple manner in simulations of surface brightness profiles by convolving them explicitly with
+    the native pixel response and setting a lower sampling scale (or higher sampling rate) using the
+    `pixel_scale' argument in drawImage() routine and setting the `method' parameter to `no_pixel'.
+    However, pixel level detector effects can be included only on images drawn at the native pixel
+    scale, which happen to be undersampled in most cases. Nyquist-sampled images that also include
+    the effects of detector non-idealities can be obtained by drawing multiple undersampled images
+    (with the detector effects included) that are offset from each other by a fraction of a pixel.
+
+    This is equivalent, in principle, to other procedures that build a higher resolution image from
+    a set of low resolution images, such as MultiDrizzle and IMCOM. A minor disadvantage of this
+    routine compared to other external generic image combination softwares is that the images must
+    be offset in equal steps along the horizontal and vertical directions respectively. This is
+    difficult to acheive with real observations but can be precisely acheived in a series of
+    simulated images. A major advantage is that the noise in the final image is not correlated as
+    the pixel values are from just a single input image. Thus, this routine preserves the noise
+    properties of the pixels.
 
     Here's an example script using this routine:
 
--- a/galsim/utilities.py
+++ b/galsim/utilities.py
@@ -432,8 +432,10 @@ def interleaveImages(im_list,N,offsets=None,add_flux=True,suppress_warnings=Fals
                              directions respectively.
     @param offsets           A list containing the offsets as galsim.PositionD instances
                              corresponding to every image in `im_list'. The offsets must be equally
-                             spaced and must add to zero in both the directions. Usage is highly
-                             recommended. [default:None]
+                             spaced and must add to zero in both the directions. The default offset
+                             ordering is to vary the offset in x from positive to negative for every
+                             offset in y which should go from positive to negative. Providing
+                             `offsets' is highly recommended. [default:None]
     @param add_flux          Should the routine add the fluxes of all the images (True) or average
                              them (False)?
     @param suppress_warnings Suppresses the warnings about the pixel scale of the output, if True.
@@ -485,8 +487,8 @@ def interleaveImages(im_list,N,offsets=None,add_flux=True,suppress_warnings=Fals
             for i in xrange(n1):
                 img_array[j::n2,i::n1] = im_list[n1*j+i].array[:,:]
     else:
-        # dx[i] = i/n
-        # default_offset[i]/n = -(i+0.5)/n+0.5 = -i/n + 0.5*(n-1)/n
+        # DX[i'] = -(i+0.5)/n+0.5 = -i/n + 0.5*(n-1)/n
+        #    i  = -n DX[i'] + 0.5*(n-1)
         for k in xrange(len(offsets)):
             dx, dy = offsets[k].x, offsets[k].y
             i = int(round((n1-1)*0.5-n1*dx))
--- a/galsim/utilities.py
+++ b/galsim/utilities.py
@@ -498,14 +498,14 @@ def interleaveImages(im_list, N, offsets, add_flux=True, suppress_warnings=False
 
     if isinstance(im_list[0],galsim.Image):
         y_size, x_size = im_list[0].array.shape
-        wcs = im_list[0].wcs
     else:
         raise TypeError("'im_list' must be a list of galsim.Image instances")
 
-    if wcs is None:
-        scale = im_list[0].scale
+    wcs = im_list[0].wcs
+    if wcs.isPixelScale():
+        scale = wcs.scale
     else:
-        scale = wcs.PixelScale
+        scale = None
 
     for im in im_list[1:]:
         if not isinstance(im,galsim.Image):
@@ -514,9 +514,9 @@ def interleaveImages(im_list, N, offsets, add_flux=True, suppress_warnings=False
         if im.array.shape != (y_size,x_size):
             raise ValueError("All galsim.Image instances in 'im_list' must be of the same size")
  
-        if im.scale != scale:
+        if im.wcs != wcs:
             raise ValueError(
-                "All galsim.Image instances in 'im_list' must have the same pixel scale")
+                "All galsim.Image instances in 'im_list' must have the same WCS")
 
     img_array = np.zeros((n2*y_size,n1*x_size))
     # The tricky part - going from (x,y) Image coordinates to array indices
@@ -547,12 +547,14 @@ def interleaveImages(im_list, N, offsets, add_flux=True, suppress_warnings=False
         # Fix the flux normalization
         img /= 1.0*len(im_list)
 
-    if wcs is not None:
-        img.wcs = galsim.Jacobian(1.*scale/n1, 0., 0., 1.scale/n2)
-    elif scale is not None:
-        if (n1==n2):
-            img.scale = 1.scale/n1
+    if scale is not None:
+        if n1==n2:
+            img.wcs = galsim.PixelScale(1.*scale/n1)
+        else:
+            img.wcs = galsim.JacobianWCS(1.*scale/n1, 0., 0., 1.*scale/n2)
+    elif isinstance(wcs,galsim.JacobianWCS):
+        img.wcs = galsim.JacobianWCS(1.*wcs.dudx/n1,1.*wcs.dudy/n2,1.*wcs.dvdx/n1,1.*wcs.dvdy/n2)
     elif suppress_warnings is False:
         import warnings
-        warnings.warn("Interleaved image could not be assigned a PixelScale automatically.")
+        warnings.warn("Interleaved image could not be assigned a WCS automatically.")
     return img
--- a/galsim/utilities.py
+++ b/galsim/utilities.py
@@ -537,10 +537,10 @@ def interleaveImages(im_list, N, offsets, add_flux=True, suppress_warnings=False
 
         img_array[j::n2,i::n1] = im_list[k].array[:,:]
 
-    if add_flux is True:
-        img = galsim.Image(img_array)
-    else:
-        img = galsim.Image((1.0/(len(im_list)))*img_array)
+    img = galsim.Image(img_array)
+    if not add_flux:
+        # Fix the flux normalization
+        img /= 1.0*len(im_list)
 
     if (n1==n2):
         if scale is not None:
--- a/galsim/utilities.py
+++ b/galsim/utilities.py
@@ -582,3 +582,99 @@ def interleaveImages(im_list, N, offsets, add_flux=True, suppress_warnings=False
         import warnings
         warnings.warn("Interleaved image could not be assigned a WCS automatically.")
     return img
+
+class LRU_Cache:
+    """ Simplified Least Recently Used Cache.
+    Mostly stolen from http://code.activestate.com/recipes/577970-simplified-lru-cache/,
+    but added a method for dynamic resizing.  The least recently used cached item is
+    overwritten on a cache miss.
+
+    @param user_function   A python function to cache.
+    @param maxsize         Maximum number of inputs to cache.  [Default: 1024]
+
+    Usage
+    -----
+    >>> def slow_function(*args) # A slow-to-evaluate python function
+    >>>    ...
+    >>>
+    >>> v1 = slow_function(*k1)  # Calling function is slow
+    >>> v1 = slow_function(*k1)  # Calling again with same args is still slow
+    >>> cache = galsim.utilities.LRU_Cache(slow_function)
+    >>> v1 = cache(*k1)  # Returns slow_function(*k1), slowly the first time
+    >>> v1 = cache(*k1)  # Returns slow_function(*k1) again, but fast this time.
+
+    Methods
+    -------
+    >>> cache.resize(maxsize) # Resize the cache, either upwards or downwards.  Upwards resizing
+                              # is non-destructive.  Downwards resizing will remove the least
+                              # recently used items first.
+    """
+    def __init__(self, user_function, maxsize=1024):
+        # Link layout:     [PREV, NEXT, KEY, RESULT]
+        self.root = root = [None, None, None, None]
+        self.user_function = user_function
+        self.cache = cache = {}
+
+        last = root
+        for i in range(maxsize):
+            key = object()
+            cache[key] = last[1] = last = [last, root, key, None]
+        root[0] = last
+
+    def __call__(self, *key):
+        cache = self.cache
+        root = self.root
+        link = cache.get(key)
+        if link is not None:
+            # Cache hit: move link to last position
+            link_prev, link_next, _, result = link
+            link_prev[1] = link_next
+            link_next[0] = link_prev
+            last = root[0]
+            last[1] = root[0] = link
+            link[0] = last
+            link[1] = root
+            return result
+        # Cache miss: evaluate and insert new key/value at root, then increment root
+        #             so that just-evaluated value is in last position.
+        result = self.user_function(*key)
+        root[2] = key
+        root[3] = result
+        oldroot = root
+        root = self.root = root[1]
+        root[2], oldkey = None, root[2]
+        root[3], oldvalue = None, root[3]
+        del cache[oldkey]
+        cache[key] = oldroot
+        return result
+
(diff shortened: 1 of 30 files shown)


rmandelb:
Hi Arun - before I review the code, I have a request for some documentation updates.  I think the docstring is missing the big picture: if you are not including effects like IPC, then you can trivially recover an image sampled however you like by using drawImage() to draw into an image with whatever pixel scale you want.  However, once pixel-level detector effects are included, it is not obvious how to correctly draw an oversampled image that includes those detector effects properly, and that's what this routine is for.  So I suggest documentation updates in a few places:

* The docstring for the interleaving routine itself should better explain its purpose for existence, including a description of how one would normally get a high-resolution image and why that is no longer the right procedure once certain detector effects are included.
* The docstring for the WFIRST module in ```galsim/wfirst/__init__.py``` should mention somewhere appropriate that the images will be undersampled, but there is a way to use GalSim to reconstruct Nyquist sampled images, pointing to this routine.
* The docstring for the WFIRST getPSF() module should also briefly mention and point to this routine.

arunkannawadi:
Hi Rachel,

I included the updates you suggested in the docstring of the
``interleaveImages`` routine and in the init file for WFIRST. However, I am
not sure if this needs to go into the getPSF() routine as well. Maybe, the
routines like applyNonlinearity() and applyIPC() can specify that it is
only sensible to apply these to images drawn at native pixel scale and that
the resolution can be increased using interleaveImages() ?

On Wed, May 27, 2015 at 8:50 AM, Rachel Mandelbaum <notifications@github.com
> wrote:

> Hi Arun - before I review the code, I have a request for some
> documentation updates. I think the docstring is missing the big picture: if
> you are not including effects like IPC, then you can trivially recover an
> image sampled however you like by using drawImage() to draw into an image
> with whatever pixel scale you want. However, once pixel-level detector
> effects are included, it is not obvious how to correctly draw an
> oversampled image that includes those detector effects properly, and that's
> what this routine is for. So I suggest documentation updates in a few
> places:
>
>    - The docstring for the interleaving routine itself should better
>    explain its purpose for existence, including a description of how one would
>    normally get a high-resolution image and why that is no longer the right
>    procedure once certain detector effects are included.
>    - The docstring for the WFIRST module in galsim/wfirst/__init__.py
>    should mention somewhere appropriate that the images will be undersampled,
>    but there is a way to use GalSim to reconstruct Nyquist sampled images,
>    pointing to this routine.
>    - The docstring for the WFIRST getPSF() module should also briefly
>    mention and point to this routine.
>
> —
> <https://github.com/GalSim-developers/GalSim/pull/667#issuecomment-105894597>
> .
>



-- 
Regards

Arun Kannawadi Jayaraman
Carnegie Mellon University,
Pittsburgh, PA 15213

arunkannawadi:
Before any of you point out, the default ordering of generating offsets can
be potentially very confusing for most of the users. I thought of making
offsets a mandatory input. Let me know if you concur and I'll make that
change immediately.

On Wed, May 27, 2015 at 6:03 PM, Arun Kannawadi <arunkannawadi@cmu.edu>
wrote:

> Hi Rachel,
>
> I included the updates you suggested in the docstring of the
> ``interleaveImages`` routine and in the init file for WFIRST. However, I am
> not sure if this needs to go into the getPSF() routine as well. Maybe, the
> routines like applyNonlinearity() and applyIPC() can specify that it is
> only sensible to apply these to images drawn at native pixel scale and that
> the resolution can be increased using interleaveImages() ?
>
> On Wed, May 27, 2015 at 8:50 AM, Rachel Mandelbaum <
> notifications@github.com> wrote:
>
>> Hi Arun - before I review the code, I have a request for some
>> documentation updates. I think the docstring is missing the big picture: if
>> you are not including effects like IPC, then you can trivially recover an
>> image sampled however you like by using drawImage() to draw into an image
>> with whatever pixel scale you want. However, once pixel-level detector
>> effects are included, it is not obvious how to correctly draw an
>> oversampled image that includes those detector effects properly, and that's
>> what this routine is for. So I suggest documentation updates in a few
>> places:
>>
>>    - The docstring for the interleaving routine itself should better
>>    explain its purpose for existence, including a description of how one would
>>    normally get a high-resolution image and why that is no longer the right
>>    procedure once certain detector effects are included.
>>    - The docstring for the WFIRST module in galsim/wfirst/__init__.py
>>    should mention somewhere appropriate that the images will be undersampled,
>>    but there is a way to use GalSim to reconstruct Nyquist sampled images,
>>    pointing to this routine.
>>    - The docstring for the WFIRST getPSF() module should also briefly
>>    mention and point to this routine.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/GalSim-developers/GalSim/pull/667#issuecomment-105894597>
>> .
>>
>
>
>
> --
> Regards
>
> Arun Kannawadi Jayaraman
> PhD candidate, Dept. of Physics,
> Carnegie Mellon University,
> Pittsburgh, PA 15213
>



-- 
Regards

Arun Kannawadi Jayaraman
Carnegie Mellon University,
Pittsburgh, PA 15213

rmandelb:
I'd just make that mandatory.

galsim/utilities.py
```diff
@@ -418,3 +418,101 @@ def _gammafn(x):
...
+                             them (False)?
+    @param suppress_warnings Suppresses the warnings about the pixel scale of the output, if True.
+
+    @returns the interleaved image
+    """
+
+    if isinstance(N,int):
+        n1,n2 = N,N
+    elif isinstance(N,tuple):
+        n1,n2 = N
```
jmeyers314: You should probably also check that len(N) == 2 here.

jmeyers314: Actually, permitting only tuples, and not, for example, lists or numpy arrays seems a bit draconian.  Maybe try something like the `canindex` function inside of parse_pos_args.

tests/test_utilities.py
```diff
@@ -156,6 +156,132 @@ def test_check_all_contiguous():
...
+def test_interleaveImages():
+    import time
+    t1 = time.time()
+    # 1) Interleave arrays as Images
+    x_size, y_size = 16,10
+    n1, n2 = 2, 5
+    im_list = []
+    for i in xrange(n1):
+        for j in xrange(n2):
+            x = np.arange(-x_size+2.*i/n1,x_size+2.*i/n1,2)
```
jmeyers314: Any reason why all the '2's in this test?  It took me a while to realize that it would still work if these were 1's.

galsim/utilities.py
```diff
@@ -418,3 +418,102 @@ def _gammafn(x):
...
+    @param im_list           A list containing the galsim.Image instances to be interleaved.
+    @param N                 Number of images to interleave in either direction. It can be of type
+                             `int' if equal numbers of images are interleaved in both directions or
+                             a tuple of two integers, containing the number of images in x and y
+                             directions respectively.
+    @param offsets           A list containing the offsets as galsim.PositionD instances
+                             corresponding to every image in `im_list'. The offsets must be equally
+                             spaced and must span an entire pixel area. The offset values must
+                             be symmetric around zero, hence taking positive and negative values.
+                             The default offset ordering is to vary the offset in x from positive to
```
jmeyers314: I would have naively expected the offsets to go from negative to positive.  I agree with @rmandelb , probably best to just require the offsets explicitly.

tests/test_utilities.py
```diff
@@ -156,6 +156,132 @@ def test_check_all_contiguous():
...
+    img_randperm = galsim.utilities.interleaveImages(im_list_randperm,(n,n),offsets=offset_list_randperm)
+
+    np.testing.assert_array_equal(img_randperm.array,img.array,\
+        err_msg="Interleaved images do not match when 'offsets' is supplied")
+    assert img_randperm.scale == img.scale
+
+    # 3a) Increase resolution along one direction - square to rectangular images
+    n = 2
+    g = galsim.Gaussian(sigma=3.7,flux=100.)
+    g1 = g.shear(g=1.*(n**2-1)/(n**2+1),beta=0.0*galsim.radians)
```
jmeyers314: Clever.  +1

jmeyers314:
Hi Arun,

That's all from me.  All minor stuff I think.

galsim/utilities.py
rmandelb: Or ```hasattr(N, '__iter__')```?

arunkannawadi: I included Rachel's suggestion since that seemed simple enough. Though ``N`` can be a NumPy array, it just doesn't seem very natural to make it a NumPy array. So the docstring just says ``N`` has to be a list or a tuple of two integers.

arunkannawadi: ``len(N)==2`` is checked now.

galsim/utilities.py
arunkannawadi: Done. The only (very minor) disadvantage with having to specify ``offsets`` compulsorily is that you can no longer use this routine to interleave NumPy arrays. But that's a very rare thing to do and one is prone to make errors in getting the right offsets.

tests/test_utilities.py
arunkannawadi: No particular reason. This test is now removed completely since ``offsets`` is made mandatory now.

arunkannawadi:
Thanks for your comments @jmeyers314 :)

tests/test_utilities.py
arunkannawadi: Thanks! I had to justify why interleaving only along one direction is even a thing one would want to do.

arunkannawadi:
Unit tests can capture any bugs in the routine but cannot catch the mistakes that the user is likely to make. So I added some consistency checks in the routine to check for the correctness of ``offsets``. If one is absolutely sure of the ``offsets`` provided, then there is an option to turn the checks off to save some time. @jmeyers314 can you have a look as well?

galsim/utilities.py
```diff
@@ -418,3 +418,113 @@ def _gammafn(x):
...
+
+        i = int(round((n1-1)*0.5-n1*dx))
+        j = int(round((n2-1)*0.5-n2*dy))
+
+        if catch_offset_errors is True:
+            err_i = (n1-1)*0.5-n1*dx - round((n1-1)*0.5-n1*dx)
+            err_j = (n2-1)*0.5-n2*dy - round((n2-1)*0.5-n2*dy)
+            tol = 1.e-6
+            if abs(err_i)>tol or abs(err_j)>tol:
+                raise ValueError("'offsets' must be a list of galsim.PositionD instances with x "
```
jmeyers314: I think I'd just report "... values spaced by 1/n1 ..." so it's clear that the constraint depends on n1, and isn't a magic number.

galsim/utilities.py
```diff
@@ -418,3 +418,113 @@ def _gammafn(x):
...
+    each other by a fraction of a pixel. This is equivalent to obtaining a finer sampled image from
+    a dither sequence, except that this routine handles only equispaced offsets. The dither sequence
+    must be a list of galsim.Image instances supplied through 'im_list'.
+
+    @param im_list             A list containing the galsim.Image instances to be interleaved.
+    @param N                   Number of images to interleave in either directions. It can be of
+                               type `int' if equal number of images are interleaved in both
+                               directions or a list or tuple of two integers, containing the number
+                               of images in x and y directions respectively.
+    @param offsets             A list containing the offsets as galsim.PositionD instances
```
jmeyers314: It might help to explicitly mention that the lower and upper limits for the offset values are +/- 0.5.  You could also give an example like, for N = 2, valid values for offsets must be some permutation of [galsim.PositionD(-0.25, -0.25), galsim.PositionD(-0.25, 0.25), galsim.PositionD(0.25, -0.25), galsim.PositionD(0.25, 0.25)].

galsim/utilities.py
arunkannawadi: ``n1`` and ``n2`` are internal variables. I edited the message to output the ``N`` that user provided so the user can easily infer what the offsets should look like.

galsim/utilities.py
arunkannawadi: I have added a full example script to use the routine.

jmeyers314: Excellent.  Thanks.

galsim/utilities.py
jmeyers314: Oh right.  Your solution looks good.  Thanks.

galsim/utilities.py
```diff
@@ -418,3 +418,133 @@ def _gammafn(x):
                0.00000000000000000141, -0.00000000000000000023, 0.00000000000000000002
              )
 
+def interleaveImages(im_list,N,offsets,add_flux=True,suppress_warnings=False,catch_offset_errors=True):
+    """
+    Interleaves two or more images and outputs a larger image.
```
rmjarvis: This needs a lot more explanation at the start of this docstring.  "Interleaving" images is not a standard thing to do to astronomical images, so you need explain what it is doing right at the start.  Here is a straw man for you to edit:

Interleaves the pixel values from two or more images into a single larger image.  This routine is used to convert a list of images taken at a series of dither offsets into a single higher resolution image, where the value in each final pixel is the observed pixel value from exactly one of the original images.  It can be used to build a Nyquist-sampled image from images that were observed with pixels larger than the Nyquist scale.

In the original observed images, the integration of the surface brightness over the pixels is equivalent to convolution by the pixel profile and then sampling at the centers of the pixels.  This procedure thus simulates an observation sampled at a higher resolution than the original images, but still with the same pixel convolution.

One of the advantages of this procedure over other procedures that build a higher resolution image from a set of low resolution images, such as drizzle, is that the noise in the final image is not correlated.  Each pixel value comes from just a single input image.  The noise properties of the pixels are thus preserved.

A significant disadvantage is that the input images must be uniformly spaced in both x and y, which is difficult to achieve with real observations, but which can be precisely achieved with a series of simulated images.

galsim/utilities.py
```diff
@@ -418,3 +418,133 @@ def _gammafn(x):
                0.00000000000000000141, -0.00000000000000000023, 0.00000000000000000002
              )
 
+def interleaveImages(im_list,N,offsets,add_flux=True,suppress_warnings=False,catch_offset_errors=True):
```
rmjarvis: Need spaces after the commas.

galsim/utilities.py
```diff
@@ -418,3 +418,133 @@ def _gammafn(x):
...
+            if i<0 or j<0 or i>=x_size or j>=y_size:
+                raise ValueError("'offsets' must be a list of galsim.PositionD instances with x "
+                            +"values spaced by 1/{0} and y values by 1/{1} around 0 for N = ".format(n1,n2)+str(N))
+
+        img_array[j::n2,i::n1] = im_list[k].array[:,:]
+
+    if add_flux is True:
+        img = galsim.Image(img_array)
+    else:
+        img = galsim.Image((1.0/(len(im_list)))*img_array)
```
rmjarvis: This might be slightly clearer as:
```
img = galim.Image(img_array)
if not add_flux:
    # Fix the flux normalization.
    img /= len(im_list)
```

galsim/utilities.py
```diff
@@ -418,3 +418,133 @@ def _gammafn(x):
...
+    for im in im_list[1:]:
+        if not isinstance(im,galsim.Image):
+            raise TypeError("'im_list' must be a list of galsim.Image instances")
+
+        if im.array.shape != (y_size,x_size):
+            raise ValueError("All galsim.Image instances in 'im_list' must be of the same size")
+ 
+        if im.scale != scale:
+            raise ValueError(
+                "All galsim.Image instances in 'im_list' must have the same pixel scale")
```
rmjarvis: Are you requiring them to have PixelScale WCS's?  Perhaps you want to take `wcs` from the first and check that `im.wcs != wcs` here.

galsim/utilities.py
```diff
@@ -418,3 +418,133 @@ def _gammafn(x):
...
+        img = galsim.Image(img_array)
+    else:
+        img = galsim.Image((1.0/(len(im_list)))*img_array)
+
+    if (n1==n2):
+        if scale is not None:
+            img.scale = im_list[0].scale*(1./n1)
+    elif suppress_warnings is False:
+        import warnings
+        warnings.warn("Interleaved image could not be assigned a pixel scale automatically")
```
rmjarvis: Au contraire.  If the originals did have PixelScale WCS's, then the appropriate final WCS is
```
img.wcs = galsim.Jacobian(scale/n1, 0., 0., scale/n2)
```
Also, I'd put the check for `if wcs is not None` at the start, since the warning doesn't make much sense in that case, as the scale couldn't be assigned automatically even if n1 == n2 then.

tests/test_utilities.py
```diff
@@ -156,6 +156,159 @@ def test_check_all_contiguous():
...
+    gal = galsim.Convolve([g,galsim.Pixel(1.0)])
+    im_list = []
+    offset_list = []
+    n = 2
+    for j in xrange(n):
+        for i in xrange(n):
+            im = galsim.Image(16*n,16*n)
+            offset = galsim.PositionD(-(i+0.5)/n+0.5,-(j+0.5)/n+0.5)
+            offset_list.append(offset)
+            gal.drawImage(image=im,scale=1.0,method='no_pixel',offset=offset)
```
rmjarvis: Don't use 1.0 for the scale here.  It's too easy for bugs to be missed when using 1.0 as a value in a test suite.

tests/test_utilities.py
```diff
@@ -156,6 +156,159 @@ def test_check_all_contiguous():
...
+
+    VE = ValueError('No error')
+    try:
+        N = (n,n)
+        galsim.utilities.interleaveImages(im_list,N=N,offsets=offset_list,catch_offset_errors=True)
+    except ValueError as VE:
+        pass
+    message =  "'offsets' must be a list of galsim.PositionD instances with x values"\
+                          +" spaced by 1/{0} and y values by 1/{1} around 0 for N = ".format(n,n)+str(N)
+    assert VE.message == message
```
rmjarvis: I don't think we need to regression test the specific error message here.  Just test that an error was raised.

tests/test_utilities.py
rmjarvis: Also, rather than convolve by the Pixel above, and then using 'no_pixel' here, you should just draw this the normal way with `g.drawImage(image=im, scale=scale, offset=offset)` and let galsim do the pixel convolution for you.

Then below, you will need to do the explicit convolution by `Pixel(scale)` when rendering the comparison image at `scale=scale/n`.  So still use 'no_pixel' there.

tests/test_utilities.py
```diff
@@ -156,6 +156,159 @@ def test_check_all_contiguous():
...
+        galsim.utilities.interleaveImages(im_list,N=N,offsets=offset_list,catch_offset_errors=True)
+    except ValueError as VE:
+        pass
+    assert VE.message == message
+
+    # 2a) Increase resolution along one direction - square to rectangular images
+    n = 2
+    g = galsim.Gaussian(sigma=3.7,flux=100.)
+    g1 = g.shear(g=1.*(n**2-1)/(n**2+1),beta=0.0*galsim.radians)
+    gal1 = g1 #galsim.Convolve([g1,galsim.Pixel(1.0)])
```
rmjarvis: Leaving these comments here is a bit confusing.  Either add an explicit comment that you are now interleaving surface brightness images rather than the normal pixel-convolved images that one would normally make.  Or do the normal pixel convolution here (by letting GalSim do it for you as I described above) and then later convolving by `Box(scale, scale/n/n)` and rendering with `wcs=galsim.Jacobian(scale, 0, 0, scale/n/n)`.  (At least I think that's what should work...)

tests/test_utilities.py
```diff
@@ -156,6 +156,159 @@ def test_check_all_contiguous():
...
+    DY = np.arange(0.0,1.0,1.0/(n*n))
+    DY -= DY.mean()
+    for dy in DY:
+        im = galsim.Image(16,16)
+        offset = galsim.PositionD(0.0,dy)
+        offset_list.append(offset)
+        gal1.drawImage(im,offset=offset,scale=1.0,method='no_pixel')
+        im_list.append(im)
+
+    img = galsim.utilities.interleaveImages(im_list,N=[1,n**2],offsets=offset_list,add_flux=False,suppress_warnings=True)
```
rmjarvis: Should add a comment highlighting your test of `add_flux=False` here.  The flux in the test profile below has the same flux as the input profiles.  Unlike the flux for the other tests where you scaled it up.

galsim/utilities.py
arunkannawadi: I don't quite understand this. Why can't the scale be assigned even if n1==n2? The image can have no WCS but can still have scale right?

rmjarvis: A scale is just a special kind of WCS.  The simplest one, in fact.  Behind the scenes, if you set `im.scale=0.2` say for an image, this is really equivalent to `im.wcs = galsim.PixelScale(0.2)`.

However, if an image actually has a non-PixelScale wcs then trying to access it with `im.scale` will raise an exception.  So in our routines, we always need to check for wcs, not scale.

tests/test_utilities.py
arunkannawadi: Certain tests seem to fail if I don't specify a scale. Is the issue only with 1.0? I can set some non-trivial scale in such cases?

galsim/utilities.py
arunkannawadi: Got it. I think I have made the WCS to be handled in a generic way.

tests/test_utilities.py
arunkannawadi: I added a comment now.

tests/test_utilities.py
arunkannawadi: Sorry, I meant to clean it up for pushing, which I clearly forgot about.

tests/test_utilities.py
rmjarvis: Yes.  Use a value other than 1.0.  Just to make sure that there isn't anything that forgot to multiply or divide by scale at some point.  That kind of thing.

tests/test_utilities.py
arunkannawadi: I just didn't want to mistake it for some other error that could have risen. Now I just check for ValueError and if it isn't raised, I raise a RuntimeError.

galsim/utilities.py
arunkannawadi: Done.

galsim/utilities.py
arunkannawadi: Done.

tests/test_utilities.py
rmjarvis: We normally use the `assert_raises` function for this.  cf. test_wfirst.py (among many others).  You do need to wrap it in a try/except block to catch an ImportError on systems that don't have nose installed.  But it's a bit cleaner than do this manually.

rmandelb:
@arunkannawadi , I reiterate my previous concern:

>I think the docstring is missing the big picture: if you are not including effects like IPC, then you can 
>trivially recover an image sampled however you like by using drawImage() to draw into an image with 
> whatever pixel scale you want. However, once pixel-level detector effects are included, it is not 
> obvious how to correctly draw an oversampled image that includes those detector effects properly, 
> and that's what this routine is for. So I suggest documentation updates in a few places:
>
> The docstring for the interleaving routine itself should better explain its purpose for existence, 
> including a description of how one would normally get a high-resolution image and why that is no 
> longer the right procedure once certain detector effects are included.

The docstring still doesn't do this, as I think Mike commented also.  One of the first jobs of the docstring should be to explain what exactly the routine is doing and why.  He gave you a nice suggestion in https://github.com/GalSim-developers/GalSim/pull/667#discussion_r31881386, why not use that (or a lightly edited version)?

rmandelb:
To clarify: I saw you updated the docstring a little, but I still don't think it's understandable to somebody who hasn't been discussing this with you already.

galsim/utilities.py
```diff
@@ -417,3 +417,151 @@ def _gammafn(x):
...
+                               type `int' if equal number of images are interleaved in both
+                               directions or a list or tuple of two integers, containing the number
+                               of images in x and y directions respectively.
+    @param offsets             A list containing the offsets as galsim.PositionD instances
+                               corresponding to every image in `im_list'. The offsets must be spaced
+                               equally and must span an entire pixel area. The offset values must
+                               be symmetric around zero, hence taking positive and negative values,
+                               with upper and lower limits of +0.5 and -0.5 (limit values excluded).
+    @param add_flux            Should the routine add the fluxes of all the images (True) or average
+                               them (False)? [default:True]
```
rmandelb: our convention is that the colons have spaces after them when declaring defaults (comment applies to this and other places where you have a default statement in this docstring)

arunkannawadi: Done.

arunkannawadi:
@rmandelb, I have now included an edited version of @rmjarvis 's suggestion for the docstring.

galsim/utilities.py
arunkannawadi: Done.

rmjarvis:
I edited the docstring a bit further.  I'm happy with it now.  Happy to merge this PR whenever you're ready.

arunkannawadi:
Thanks! I am happy with it too now. I will merge it by tomorrow evening if
there are no objections.

On Fri, Jul 3, 2015 at 11:43 AM, Mike Jarvis <notifications@github.com>
wrote:

> I edited the docstring a bit further. I'm happy with it now. Happy to
> merge this PR whenever you're ready.
>
> —
> <https://github.com/GalSim-developers/GalSim/pull/667#issuecomment-118378192>
> .
>



-- 
Regards

Arun Kannawadi Jayaraman
Carnegie Mellon University,
Pittsburgh, PA 15213

Issue #661, opened 2015-04-20
WFIRST module accuracy parameters


rmandelb:
This is an issue that should be addressed before we tag v1.3:

It seems that there are some numerical inaccuracies in the rendering of PSFs in the WFIRST module, apparently related to the default Fourier-space sampling.  As I understand it, the Fourier-space sampling for optical PSFs is chosen based on the obscured Airy, but if there are struts then one might need better sampling.  Some tests seem to confirm that the sampling is adequate if we remove the struts, and problems only arise with struts.  The fix for this problem differs depending on whether approximate struts are used or the full pupil plane image is used.

I will put some more plots and comments soon, but I'm mainly opening this not terribly informative issue now to flag the fact that there is one more thing to do before tagging v1.3.

rmandelb:
@arunkannawadi - I was trying to put together a simple script that demonstrates this issue without directly using the WFIRST module (since it should be possible to show using OpticalPSfs, where we have more control over the choice of adding aberrations / struts / whatever) - but my simple script does not seem to demonstrate it well.  Do you have a simple script that demonstrates this?  Is it in a repository that I can access?

arunkannawadi:
@rmandelb, I don't have a script as I was just playing around with them in
the terminal directly. I will consolidate them all and put in a script.

On Wed, Apr 29, 2015 at 10:33 PM, Rachel Mandelbaum <
notifications@github.com> wrote:

> @arunkannawadi <https://github.com/arunkannawadi> - I was trying to put
> together a simple script that demonstrates this issue without directly
> using the WFIRST module (since it should be possible to show using
> OpticalPSfs, where we have more control over the choice of adding
> aberrations / struts / whatever) - but my simple script does not seem to
> demonstrate it well. Do you have a simple script that demonstrates this? Is
> it in a repository that I can access?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/GalSim-developers/GalSim/issues/661#issuecomment-97636929>
> .
>



-- 
Regards

Arun Kannawadi Jayaraman
PhD candidate, Dept. of Physics,
Carnegie Mellon University,
Pittsburgh, PA 15213

rmandelb:
Thanks, please do.

arunkannawadi:
@rmandelb, I have shared the script with you. The script generates the MTF
for many different simpler PSFs for comparison. You can turn-off those by
setting ```draw_simple_profiles=True```. Further, it goes on to generate
[...]

PR #668, opened 2015-06-03
#661 - WFIRST module accuracy parameters

rmandelb:
This is a short pull request that includes an option to have a higher-fidelity representation of the WFIRST PSFs in Fourier space compared to what you get with the default settings.  See the plots on the issue page for details of how this setting changes the MTF.

I did not put this into the CHANGELOG because this PR will go into the same release as the initial PR for the WFIRST module, so it's basically a slight improvement of one of the new features in this version of GalSim rather than an improvement on a previous version of GalSim.

There is a debatable philosophical point here related to how the WFIRST module is set up:
Unlike OpticalPSF or ChromaticOpticalPSF, the user cannot adjust all the parameters related to the WFIRST optical PSF freely.  Instead, the WFIRST module getPSF() method just has default settings and the new 'high accuracy' settings.  The reason for this choice was that once we are working from the WFIRST pupil plane image, the OpticalPSF kwargs cannot be adjusted freely (i.e., many possible changes to the oversampling or pad factor would result in the routine throwing an exception because the supplied pupil plane image is inadequate in some way).  I think it would take a lot of work to make it possible to adjust the OpticalPSF kwargs completely freely, and I did not feel that it was necessary to do this for an initial version of the WFIRST module.  However, I wanted to flag this design decision, particularly now that we have a 'high accuracy' setting that also isn't adjustable (i.e., it's just a binary choice for default or high accuracy with nothing in between).  Any comments or concerns about this choice?

### DIFF
--- a/galsim/wfirst/wfirst_psfs.py
+++ b/galsim/wfirst/wfirst_psfs.py
@@ -122,9 +122,11 @@ def getPSF(SCAs=None, approximate_struts=False, n_waves=None, extra_aberrations=
                                    bandpass.
                                    [default: False]
     @param    high_accuracy        If True, make higher-fidelity representations of the PSF in
-                                   Fourier space, to minimize aliasing.  This is more expensive in
-                                   terms of time and RAM, and may not be necessary for many
-                                   applications. [default: False]
+                                   Fourier space, to minimize aliasing (see plots on
+                                   https://github.com/GalSim-developers/GalSim/issues/661 for more
+                                   details).  This setting is more expensive in terms of time and
+                                   RAM, and may not be necessary for many applications.
+                                   [default: False]
     @returns  A dict of ChromaticOpticalPSF or OpticalPSF objects for each SCA.
     """
     # Check which SCAs are to be done using a helper routine in this module.
--- a/galsim/wfirst/wfirst_psfs.py
+++ b/galsim/wfirst/wfirst_psfs.py
@@ -65,6 +65,13 @@ def getPSF(SCAs=None, approximate_struts=False, n_waves=None, extra_aberrations=
     simple, understood wavelength-dependence.  The resulting object can be used to draw into any of
     the WFIRST bandpasses.
 
+    For applications that require very high accuracy in the modeling of the PSF, with very limited
+    aliasing, the `high_accuracy` option can be set to True.  When using this option, the MTF has a
+    value below 1e-4 for all wavenumbers above the band limit when using `approximate_struts=True`,
+    or below 3e-4 when using `approximate_struts=False`.  In contrast, when `high_accuracy=False`
+    (the default), there are some bumps in the MTF above the band limit that reach an amplitude of
+    ~1e-2.
+
     By default, no additional aberrations are included above the basic design.  However, users can
     provide an optional keyword `extra_aberrations` that will be included on top of those that are
     part of the design.  This should be in the same format as for the ChromaticOpticalPSF class,
--- a/galsim/wfirst/wfirst_psfs.py
+++ b/galsim/wfirst/wfirst_psfs.py
@@ -134,7 +134,7 @@ def getPSF(SCAs=None, approximate_struts=False, n_waves=None, extra_aberrations=
     # Deal with some accuracy settings.
     if high_accuracy:
         if approximate_struts:
-            oversampling = 3.0
+            oversampling = 3.5
         else:
             oversampling = 2.0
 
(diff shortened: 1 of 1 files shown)


rmjarvis:
This looks fine to me.

rmandelb:
Anyone else?  @jmeyers314 ?  @arunkannawadi ?

jmeyers314:
Hey @rmandelb , I thought maybe @barnabytprowe was planning to look at this, but all looks good to me.

barnabytprowe:
Hey guys, sorry for the long non-response on this one... I have been playing a little in spare moments on vacation, quick update.

## Pupil plane image results
Wow, I had no idea this would use so much RAM in high accuracy mode!  Basically way more than my laptop has, and it never completed in reasonable time (e.g. overnight, even.)  Sorry to not be much use here!

## Approximate struts results
I could get the `approximate_struts` image to be better quality and sub 1.e-4 in terms of higher frequency (i.e. ultra-bandpass) gunk without unreasonable memory costs, simply by upping the `oversampling` to 3.5 from 3.0 when building the `OpticalPSF` within `wfirst_psfs.py` (around line 138):
fig2_6_after

The `pad_factor` keyword here has little effect on the amplitude of the bumps, so so much for that idea.  So in terms of this PR I think that I'm happy as is, but might suggest bumping the 3.0 up to 3.5 for `oversampling`.

What I did notice and which I found interesting was that increasing the oversampling helps smooth those bumps, but it does not impact the "continuum" i.e. "green cross" pattern stuff.  Here is an example with oversampling set to 5:
fig2_6_after_os5

Note there is also a slight tail, at the edge of the sharp drop in the MTF at the bandlimit, that is also not reduced by increasing the oversampling.  I don't know the source of this artifact for certain, but my hypothesis is that it is caused by the representation of the curved edges of the pupil using a grid with straight, and sharp, vertical and horizontal boundaries. The reason (I'm guessing) this doesn't "go away" if you increase oversampling or the `pad_factor` is that you aren't in fact reducing the amount of vertical or horizontal edge in the image, you are just switching from vertical to horizontal more often.

I don't know if this is right, and I certainly shouldn't hold up this PR more while I try out the `float` pupil plane idea to try and see if this reduces the effect.  (I'm betting it will.)  I might make an issue for this if there is interest.

Moreover, this idea doesn't really help with the pupil plane image version of the WFIRST PSF.  Does anyone know if there is a vector-graphical representation of the same pupil plane image, in some decipherable polygonal format? (I'm guessing the raster image used here must have come ultimately from some sort of CAD model...)  Because if so it would be really good to try and use that somehow - either by making a smoothed boundary version of what we already have, or using it directly.  But that's an even bigger project.  Probably best to work out if what we have in this PR is good enough.  My preference for sub 1.e-4 artifacts is definitely conservative.

barnabytprowe:
Pithy summary: OK for this PR but might consider oversampling 3.5 for approx_struts.  Harder for the pupil plane version.  Seem to be hitting a bit of a wall that I'd like to understand better.  Will consider opening an issue out of pure personal intrigue!  For WFIRST purposes thinking about how to test what is "good enough" is probably necessary.

rmandelb:
Hi @barnabytprowe - thanks very much for playing around with this!  Now that I'm back, I took your suggestion of increasing the oversampling to 3.5 for ```approximate_struts```.

If nobody raises any objection, I propose to merge this on the morning of Tuesday July 7 (giving time for people who might be off already for the July 4 weekend to comment once they are back).

To address the more detailed points above:
- I agree that it's likely the horizontal/vertical sharp boundaries that are likely causing some of the low-level issues, and this may be worth exploring further at some point.
- For the pupil plane image, I can inquire what formats are available.
- The definition of "good enough" depends on exactly what you're trying to do.  There are many possible uses of these sims and I suspect that the requirements for some are more stringent than for others...

rmandelb:
On another note: once this PR and the other open PR are merged, shall we tag v1.3 sometime next week?

rmjarvis:
Fine with me.  (On both questions.)

arunkannawadi:
Looks good to me.

As far as we know for now, such high accuracy is certainly needed to give these images as inputs to IMCOM. Do we know of any other case where these highly accurate PSFs are a necessity? Currently, the docstring only says that such high accuracies may not be needed for many applications. It may be worthwhile to add a line in the docstring saying explicitly that ``high_accuracy = True`` is needed if a PSF dither sequence is to be used by IMCOM or something of that sort?

rmandelb:
@arunkannawadi - I think it depends a lot on the application, so it's hard to say anything general.  But I agree that it's worth adding a line about applications that require a strictly band-limited PSF with aliasing below a certain limit (such as IMCOM).  Will do that now, and then merge.

rmandelb:
Okay, this is done: https://github.com/GalSim-developers/GalSim/commit/4c3c0c8ac12ee7165bd28137fb019e17b6ce3934

Further docstring tweaks (if necessary) can be done on master before we tag v1.3.

Issue #670, opened 2015-06-11
Cache chromatic drawImage intermediate results


jmeyers314:
This issue is to implement caches of intermediate products in the drawImage() and drawKImage() methods of `ChromaticObject` and subclasses where possible.

There are several instances where one may wish to call the `drawImage()` (or `drawKImage()`) method of a `ChromaticObject` with only minor changes between invocations, and in particular with no changes to `SED`s, `Bandpass`es, and chromatic PSFs.  Examples include forward-model fitting/sampling and ring tests.  In these cases, it would be advantageous to avoid recomputing certain intermediate results of `ChromaticObject.drawImage()` and `.drawKImage()`, such as the product of an SED and a Bandpass in the case of separable profiles, or the effective (wavelength-integrated) profile in the case of convolution of separable and inseparable profiles.

Previously, creating such a cache was problematic since it was difficult to recognize when particular GalSim objects had been seen before.  Now that (most) GalSim objects are immutable, though, and carry `__eq__` and `__hash__` methods, caching is feasible.  

My current plan is to steal the code for the cache from here: http://code.activestate.com/recipes/577970-simplified-lru-cache/ , but I'm happy to hear alternatives.

jmeyers314:
This still needs to be cleaned up a bit, but I wanted to report that some of the ring test results in chroma are sped up by a factor of ~18 from 107s to 6s with this update!

@mdschneider, @wadawson I think this may speed up some of the chromatic drawings you've been playing with too.

rmandelb:
Nice!

PR #671, opened 2015-06-12
#670 chromatic caches

jmeyers314:
This PR implements two caches in order to avoid recomputing intermediate products that can recur when drawing chromatic objects.  Specifically, these are the integral of the product of SED and bandpass for separable objects, and the effective profiles used in ChromaticConvolution.drawImage.  

The former cache only leads to a speedup of ~10% for the simple test I tried, but this probably depends somewhat on how finely the SED*bandpass integrand is sampled so could be more important in other cases.

The latter cache improves the timing of some ring test code in chroma by a factor of ~18.

The only tricky part I think worth mentioning is that since numpy ndarrays and lists are not hashable, in a few places I had to convert these into tuples to get the cache algorithm to work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,12 +28,12 @@ API Changes
   issue, and we can add them back. (#218)
 - Made all returned matrices consistently use numpy.array, rather than
   numpy.matrix.  We had been inconsistent with which type different functions
-  returned, so now all matrices are numpy.arrays.  If you rely on the 
+  returned, so now all matrices are numpy.arrays.  If you rely on the
   numpy.matrix behavior, you should recast with numpy.asmatrix(M). (#218)
-- Deprecated CorrelatedNoise.calculateCovarianceMatrix, since it is not used 
+- Deprecated CorrelatedNoise.calculateCovarianceMatrix, since it is not used
   anywhere, and we think no one really has a need for it. (#630)
 - Officially deprecated the methods and functions that had been described as
-  having been removed or changed to a different name.  In fact, many of them 
+  having been removed or changed to a different name.  In fact, many of them
   had been still valid, but no longer documented.  This was intentional to
   allow people time to change their code.  Now these methods are officially
   deprecated and will emit a warning message if used. (#643)
@@ -49,14 +49,14 @@ New Features
 
 - Made all GalSim objects picklable unless they use fundamentally unpicklable
   things such as lambda expressions.  Also gave objects better str and repr
-  representations (str(obj) should be descriptive, but relatively succinct, 
+  representations (str(obj) should be descriptive, but relatively succinct,
   and repr(obj) should be unambiguous).  Also made __eq__, __ne__, and __hash__
   work better. (#218)
 - Added ability to set the zeropoint of a bandpass on construction.  (Only
   a numeric value; more complicated calculations still need to use the method
   `bandpass.withZeropoint()`.) (#218)
 - Added ability to set the redshift of an SED on construction. (#218)
-- Updated CorrelatedNoise to work with images that have a non-trivial WCS. 
+- Updated CorrelatedNoise to work with images that have a non-trivial WCS.
   (#501)
 - Added new methods of the image class to simulate detector effects:
   inter-pixel capacitance (#555) and image quantization (#558).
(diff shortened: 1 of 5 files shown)


galsim/chromatic.py
```diff
@@ -265,11 +319,7 @@ def drawImage(self, bandpass, image=None, integrator='trapezoidal', **kwargs):
         wave_list = self._getCombinedWaveList(bandpass)
 
         if self.separable:
-            if len(wave_list) > 0:
-                multiplier = np.trapz(self.SED(wave_list) * bandpass(wave_list), wave_list)
-            else:
-                multiplier = galsim.integ.int1d(lambda w: self.SED(w) * bandpass(w),
-                                                bandpass.blue_limit, bandpass.red_limit)
+            multiplier = _multiplier_cache(self.SED, bandpass, tuple(wave_list))
```
rmjarvis: Why do you call tuple here?  Does a numpy array not work as a key for the cache for some reason?

galsim/chromatic.py
```diff
@@ -1692,6 +1742,13 @@ def drawImage(self, bandpass, image=None, integrator='trapezoidal', iimult=None,
         Works by finding sums of profiles which include separable portions, which can then be
         integrated before doing any convolutions, which are pushed to the end.
 
+        This method uses a cache to avoid recomputing 'effective' profiles, which are the 
+        wavelength-integrated products of inseparable profiles, the spectral components of 
+        separable profiles, and the bandpass.  Because the cache size is finite, users may find 
+        that it is more efficient when drawing many images to group images using the same 
+        SEDs, bandpasses, and inseparable profiles (generally PSFs) together in order to hit the 
+        cache more often.
```
rmjarvis: You should mention that the maxsize is 10 to help the user plan this out.

Also, it might be nice to provide a function where the user could change the maxsize.  The modification to LRU_Cache wouldn't be too hard I think to enable this.

tests/test_chromatic.py
```diff
@@ -1313,6 +1313,9 @@ def evaluateAtWavelength(self, wave):
             ret = ret.shear(g1 = this_shear)
             return ret
 
+        def __repr__(self):
+            return 'galsim.ChromaticGaussian(%r)'%self.sigma
+
```
rmjarvis: Are there any substantive tests we should add for this?  Something that intentionally draws the same thing multiple times (with different PSFs maybe) to make sure we hit the cache, then don't, then do again. Or something like that?

galsim/chromatic.py
jmeyers314: Yeah, ndarrays are not immutable, and therefore are not hashable.  Tupling gets around this, though I agree it's a bit hacky.

galsim/chromatic.py
jmeyers314: I added a resize method to LRU_Cache and resize methods for each of the specific chromatic caches.  As far as I could tell, this change required either marking the caches with `global` or moving them from the module level to the class level.  I opted for the latter, though this did require me to attach each cache to its class outside of the class definition.

jmeyers314: Oh, and I added that the maxsize is 10 to each docstring.

tests/test_chromatic.py
jmeyers314: I wasn't sure that we needed to test the chromatic caches specifically, but I did add a test for the LRU_Cache itself.  I think this probably sufficient coverage, but let me know if you think I'm omitting a particular failure mode.

CHANGELOG.md
```diff
@@ -108,7 +108,8 @@ Bug Fixes and Improvements
 - Fixed a bug in rendering profiles that involve two separate shifts. (#645)
 - Fixed a bug if drawImage was given odd nx, ny parameters, the drawn profile
   was not correctly centered in the image. (#645)
-
+- Added intermediate results cache to `ChromaticObject.drawImage` and
```
rmandelb: I'm not sure this is enough to say what the purpose of the change was; include the fact that this can speed up rendering by such-and-such factor?

jmeyers314: Sure.  Here's my new cut at the CHANGELOG.md entry:

```
- Added intermediate results cache to `ChromaticObject.drawImage` and
  `ChromaticConvolution.drawImage`, which, for some applications, can
  significantly speed up (anywhere from 10% to 2000%) the rendering of groups
  of similar (same SED, same Bandpass, same PSF) chromatic profiles.
```

jmeyers314:
Anything more @rmandelb and @rmjarvis ?

I think @rmandelb has seen the latest commit here, but @rmjarvis , you may wish to take another look since I reorganized when adding the cache resize option.

rmjarvis:
Looks good.  Thanks.

jmeyers314:
Hey @rmandelb , were you okay with merging this now?

rmandelb:
yes

PR #672, opened 2015-06-14
Pyfits 2.4 compatibility.

rmjarvis:
This is a fairly small PR to address some failures in the test suite for older versions of pyfits.  I have a system with pyfits 2.4 that was failing many of the FitsHeader tests we introduced in #590.  So I fixed those.

The two changes that matter for people who don't care about that are:
1. I now allow a FitsHeader to be constructed from a list of (key, value) pairs in addition to a pyfits.header or a dict.  And I made that the way the repr works when there isn't a more succinct string to use, rather than the string version of the pyfits.Header object.  (The latter didn't work for my old pyfits, but also I think the list of (key, value) pairs looks cleaner in those cases anyway.)
2. I also took the opportunity to move our pyfits import out of `__init__.py` and into `_pyfits.py`.  So now `import galsim` is much faster, which is nice for an interactive session.  And you now get the galsim pyfits by writing `from galsim._pyfits import pyfits`.  The first time you do so, it will do the try/except bit to import either astropy.io.fits or pyfits.

BTW, I didn't bother to create an issue for this, so the branch is pyfits_2.4.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,9 +60,11 @@ New Features
   (#501)
 - Added new methods of the image class to simulate detector effects:
   inter-pixel capacitance (#555) and image quantization (#558).
-- Enable constructing a FitsHeader object from a dict.  This had been a hidden
+- Enabled constructing a FitsHeader object from a dict.  This had been a hidden
   feature that only half worked.  It should now work correctly.  Also, allow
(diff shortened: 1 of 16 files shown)


rmandelb:
This looks good to me, and tests pass on both my systems.

jmeyers314:
Looks good to me too.  Tests all pass.

Issue #675, opened 2015-07-17
update some aspects of WFIRST module


rmandelb:
Some parts of the WFIRST PSF should be updated to match what is posted at
http://wfirst.gsfc.nasa.gov/science/sdt_public/wps/references/instrument/

This issue (assigned to me) is for that work.  Three possible updates:
- updating Zernikes at field centers (done, see PR #676)
- add variation of Zernikes across SCAs
- pupil plane updates

rmandelb:
@arunkannawadi , @chazshapiro:
I took care of the first item on the list today, updating the Zernikes at the centers of each SCA to the new values.  This seems to have largely fixed the pesky chromatic trends that we were seeing.  For example, if I take this script and run it on the master branch:
```
import galsim
import galsim.wfirst as wf

SCA=11
blue_wavelength=850.
red_wavelength=1500.
draw_scale = 0.01 # arcsec

# First check WFIRST effective PSF using only obscured Airy, and compare the adaptive size using 
# the two wavelengths.
pix = galsim.Pixel(wf.pixel_scale)
blue_psf = galsim.Convolve(
    galsim.OpticalPSF(lam=blue_wavelength, diam=wf.diameter, obscuration=wf.obscuration), pix)
red_psf = galsim.Convolve(
    galsim.OpticalPSF(lam=red_wavelength, diam=wf.diameter, obscuration=wf.obscuration), pix)
blue_im = blue_psf.drawImage(method='no_pixel', scale=draw_scale)
red_im = red_psf.drawImage(method='no_pixel', scale=draw_scale)
blue_res = blue_im.FindAdaptiveMom()
red_res = red_im.FindAdaptiveMom()
print 'Obscured Airy adaptive sizes: ',blue_res.moments_sigma,red_res.moments_sigma
print '  Ratio (blue/red): ',blue_res.moments_sigma/red_res.moments_sigma

# Then, do the same trick, but using the aberrated PSFs.
pix = galsim.Pixel(wf.pixel_scale)
blue_psf = galsim.Convolve(
    wf.getPSF(SCAs=SCA, approximate_struts=True, wavelength=blue_wavelength)[SCA], pix)
red_psf = galsim.Convolve(
    wf.getPSF(SCAs=SCA, approximate_struts=True, wavelength=red_wavelength)[SCA], pix)
blue_im = blue_psf.drawImage(method='no_pixel', scale=draw_scale)
[...]

PR #676, opened 2015-07-21
#675 - small WFIRST module update

rmandelb:
This PR is for a small WFIRST module update that I would like to merge to master before tagging v1.3.  The aberrations in an earlier optical design were too large, and incorporation of the newer design gives a PSF that behaves more nicely (e.g., it's larger in the red bands, like a diffraction-limited PSF, instead of having such large aberrations that it becomes larger in the blue bands).  I also updated the docstring to indicate the relationship between these PSFs and the files on the WFIRST project webpage.

Two other possible updates mentioned in #675 were not done, since they are a bit more work.  They will be left for a later version of GalSim.


galsim/wfirst/wfirst_psfs.py
```diff
@@ -448,3 +437,37 @@ def _find_limits(bandpasses, bandpass_dict):
         if bp.blue_limit < min_wave: min_wave = bp.blue_limit
         if bp.red_limit > max_wave: max_wave = bp.red_limit
     return min_wave, max_wave
+
+def _parse_new_zernikes(infile, out_dir):
```
jmeyers314: Is there a danger that this function name will become dated?  Maybe just `_parse_zernikes()` or `_parse_zernikes_0715()` would work better?

jmeyers314:
Looks good to me @rmandelb .  All tests pass.

galsim/wfirst/wfirst_psfs.py
rmandelb: That's a good point.  The new stand-alone routine in devel/external/ has a more specific name according to your suggestion.

rmandelb:
@rmjarvis , did you have any other comments besides that one, or are we good to merge this?

rmjarvis:
Fine to merge as far as I'm concerned.

rmandelb:
Merging #675.

Issue #678, opened 2015-08-24
build test failure


beckermr:
So it appears that when using the fits reader from astropy, the test suite for galsim has a failure.

```======================================================================
ERROR: Test the PyAstWCS class
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/beckermr_admin/Desktop/galsim_install/src/GalSim-1.3.0/tests/test_wcs.py", line 1613, in test_pyastwcs
    wcs = galsim.PyAstWCS(file_name, dir=dir)
  File "/usr/local/lib/python2.7/site-packages/galsim/fitswcs.py", line 434, in __init__
    wcsinfo = self._load_from_header(header, hdu)
  File "/usr/local/lib/python2.7/site-packages/galsim/fitswcs.py", line 476, in _load_from_header
    fc = starlink.Ast.FitsChan(starlink.Atl.PyFITSAdapter(hdu))
  File "/usr/local/lib/python2.7/site-packages/starlink/Atl.py", line 106, in __init__
    self.pyfits_3_1_0 = ( LooseVersion(pyfits.__version__) >=
AttributeError: 'module' object has no attribute '__version__'
```

It turns out this is not a failure of galsim, but actually of starlink, since this library is asking for the `__version__` attribute which is not set for the latest pyfits version in astropy. I don't know if galsim needs to do anything about this. Should this be marked as a known failure?

rmjarvis:
Thanks for posting this Matt.  This wasn't a problem with previous versions of starlink.  I've posted an issue on their site [here](https://github.com/timj/starlink-pyast/issues/7).

beckermr:
OK. With the latest starlink-pyast that was just pushed to pypi (3.8.0) the test suite passes just fine. I am closing this issue.

PR #681, opened 2015-09-04
A few updates to WCS classes

rmjarvis:
I made a few updates to the WCS classes, precipitated by Matt's bug report related to PyAst #678.  This spawned some an [issue](https://github.com/timj/starlink-pyast/issues/8) on the pyast GitHub page.  (Feel free to read through that issue page if you're interested.)

The upshot is that I realized that there was something I wasn't quite doing right in using PyAst.  It would only come into play for FITS files that have the two celestial axes listed as Dec, RA.  It's pretty unusual, but technically allowed by the FITS standard.  I know this was true, but I hadn't realized that the way I was using PyAst would fail for these files.  So I added such a file to the test suite to make sure all our WCS classes work correctly in that case.

* PyAst now does work correctly for these files
* Astropy v1.0.1 worked, but 0.2.4 and 0.3.1 failed, so I added a workaround for older astropy versions, so they should now work correctly.
* Astropy v1.0.1 also seems to have fixed another bug that I had had a workaround for, so I put back the native astropy call for versions >= '1.0.1'.  This was their issue [1977](https://github.com/astropy/astropy/issues/1977)
* WcsTools worked fine.
* GSFitsWCS failed with a RuntimeError, since I knew about this shortcoming.  But on this branch, I went ahead and implemented the fix to make it work.  So it works correctly now.

If you want to test this (rather than just look at the code), the branch name is `pyast_tpv` (since it started with a bug in how PyAst handled TPV files, although that's not very descriptive of what the branch turned into in the end), and you'll want to run `python test_wcs.py` directly from the command line to get the tests of TAN-FLIP, which aren't enabled for the nosetests run you get with `scons tests`.


rmandelb:
Will go over the code later, but tests pass on both my machines.

galsim/fitswcs.py
```diff
@@ -964,9 +991,12 @@ def _read_header(self, header):
...
         ctype2 = header['CTYPE2']
-        if not (ctype1.startswith('RA---') and ctype2.startswith('DEC--')):
-            raise NotImplementedError("GSFitsWCS can only handle cases where CTYPE1 is RA " +
-                                      "and CTYPE2 is DEC")
+        if ctype1.startswith('DEC--') and ctype2.startswith('RA---'):
+            flip = True
+        elif ctype1.startswith('RA---') and ctype2.startswith('DEC--'):
+            flip = False
+        else:
+            raise RuntimeError("The WCS read in does not define a pair of celestial axes" )
```
rmandelb: Might be slightly more informative to also spit out the ctype1 and ctype2 values (since they are used to determine whether this error should be raised).

rmandelb:
Sorry for forgetting this for an embarrassingly long time.  Looks good to me (just one minor, optional point that I noted on the code).

jmeyers314:
Hi @rmjarvis, all looks good to me, and all tests pass.

galsim/fitswcs.py
rmjarvis: Done: 3823df2

Sample error message when I changed DEC to DEG:
```
>>> wcs = galsim.GSFitsWCS(header=h)
  File "/Library/Python/2.7/site-packages/galsim/fitswcs.py", line 976, in __init__
    self._read_header(header)
  File "/Library/Python/2.7/site-packages/galsim/fitswcs.py", line 1001, in _read_header
    ctype1, ctype2))
RuntimeError: The WCS read in does not define a pair of celestial axes. Expecting CTYPE1,2 to start with RA--- and DEC--.  Got RA---TAN, DEG--TAN
```

rmjarvis:
Thanks both.  I'll go ahead and merge now.

galsim/fitswcs.py
rmandelb: Cool, thanks.

Issue #683, opened 2015-10-19
inseparable chromatic drawImage(nx=..., ny=...) fails


jmeyers314:
Some of the image setup logic for chromatic drawImage fails when trying to set the output image properties of the convolution of two inseparable profiles with the nx=, ny=, scale= keywords.  For example:

```
import galsim
import numpy as np
waves = np.arange(500, 1000)
psf = galsim.ChromaticAiry(lam=700, diam=0.1)
bp = galsim.Bandpass(lambda x: 1, blue_limit=500, red_limit=1000)
sed1 = galsim.SED(galsim.LookupTable(waves, waves**0, interpolant='linear'),
                  flux_type='fphotons').withFlux(1.0, bp)
sed2 = galsim.SED(galsim.LookupTable(waves, waves**1, interpolant='linear'),
                  flux_type='fphotons').withFlux(1.0, bp)
g1 = galsim.Sersic(n=1, half_light_radius=1.0) * sed1
g2 = galsim.Sersic(n=2, half_light_radius=0.5) * sed2

img = galsim.Convolve(g1+g2, psf).drawImage(bp) # works
img = galsim.Convolve(g1, psf).drawImage(bp, nx=32, ny=32, scale=0.2) # works
img = galsim.Convolve(g1+g2, psf).drawImage(bp, nx=32, ny=32, scale=0.2) # fails
```

I think I've traced this to some missing `kwargs.pop()` cleanup lines in chromatic.py.  PR coming soon...

jmeyers314:
Merged.  Closing.

PR #684, opened 2015-10-19
#683

jmeyers314:
These are just a few minor changes to the way chromatic profiles' `drawImage` methods setup output `Image`s.  In general, the following `kwargs` are not needed once an image is setup, and therefore it's a good idea to just remove them from `kwargs` once an image exists: `dtype`, `scale`, `wcs`, `nx`, `ny`, `bounds`.  We were doing this inconsistently before, so this PR just introduces consistency and a simple unit test to verify that previous problems are gone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,4 +6,5 @@ Bug Fixes
 
 - Fixed a bug in some of the WCS classes if the RA/Dec axes in the FITS header
   are reversed (which is allowed by the FITS standard). (#681)
-
+- Fixed a bug in the way Images are instantiated for certain combinations of
+  ChromaticObjects and image-setup keyword arguments (#683)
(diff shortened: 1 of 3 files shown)


rmandelb:
Looks good to me.  Tests pass on both my machines.  Need to update CHANGELOG, otherwise I'm fine with merging.

jmeyers314:
Thanks @rmandelb .

Added to the CHANGELOG:
```
- Fixed a bug in the way Images are instantiated for certain combinations of
  ChromaticObjects and image-setup keyword arguments (#683)
```

rmandelb:
Great, thanks.

rmjarvis:
Finally getting around to looking at these PRs now.  

My only comment on this one is that the same list of pop commands appears 5 times within the same file.  To me this begs to be a helper function.  Maybe `_remove_setup_kwargs(kwargs)` or something along those lines.

jmeyers314:
>To me this begs to be a helper function. Maybe _remove_setup_kwargs(kwargs) or something along those lines.

Sounds good to me.  Done.

rmjarvis:
Good.  Looks fine to merger I think.

Issue #680, opened 2015-09-03
Separable ChromaticObject with null SED at bandpass effective wavelength is undrawable.


jmeyers314:
Drawing separable `ChromaticObject`s currently works by first drawing the achromatic object at the effective wavelength of the filter in question, and then scaling this image by the wavelength integral of bandpass * SED.  This fails, however, if the value of the SED is zero at the effective wavelength of the filter, a situation that probably doesn't occur too often with realistic spectra/filters but can occur (and just did for my office-mate!) for contrived tophat filters and SEDs.

One possible solution would be to search for a fiducial wavelength (probably still starting at the filter effective wavelength, or maybe at the midpoint between the filter blue/red edges) for which the SED has non-zero flux.

rmandelb:
Tell your office-mate not be so contrived!

Just kidding.  I think your solution is reasonable.  I would imagine it could be fixed by taking this
```
        # setup output image (semi-arbitrarily using the bandpass effective wavelength)
        prof0 = self.evaluateAtWavelength(bandpass.effective_wavelength)
```
and doing something slightly more complicated to ensure that a reasonable wavelength with nonzero flux is chosen, so that later on you can do the rescaling by ```multiplier``` without a problem.

rmandelb:
Can close this now?

jmeyers314:
Right.  Yes, thanks.

PR #685, opened 2015-10-20
#680

jmeyers314:
This PR adds a method `_fiducial_profile` to chromatic objects that returns an achromatic profile by evaluating the chromatic profile at a fiducial wavelength.  Previously, this fiducial wavelength had just been the effective wavelength of a `Bandpass`.  The new method adds fallback wavelengths to try to ensure that the fiducial achromatic profile has non-zero flux, which is necessary to rapidly draw separable profiles.

### DIFF
--- a/galsim/chromatic.py
+++ b/galsim/chromatic.py
@@ -147,10 +147,12 @@ class ChromaticObject(object):
         Return a monochromatic profile of a separable chromatic object that can be scaled to give
         the monochromatic profile at any wavelength.
         """
-        candidate_waves = [bandpass.effective_wavelength,
-                           0.5 * (bandpass.blue_limit + bandpass.red_limit),
-                           bandpass.blue_limit,
-                           bandpass.red_limit]
+        bpwave = bandpass.effective_wavelength
+        candidate_waves = [bpwave, 0.5 * (bandpass.blue_limit + bandpass.red_limit)]
+        candidate_waves = np.union1d(candidate_waves, bandpass.wave_list)
+        candidate_waves = np.union1d(candidate_waves, self.wave_list)
+        candidate_waves = candidate_waves[np.argsort(np.abs(candidate_waves - bpwave))]
+
         for w in candidate_waves:
             prof0 = self.evaluateAtWavelength(w)
             if prof0.flux != 0:
--- a/galsim/chromatic.py
+++ b/galsim/chromatic.py
@@ -142,6 +142,21 @@ class ChromaticObject(object):
         """
         ChromaticObject._multiplier_cache.resize(maxsize)
 
+    def _fiducial_profile(self, bandpass):
+        """
+        Return a monochromatic profile of a separable chromatic object that can be scaled to give
+        the monochromatic profile at any wavelength.
+        """
+        candidate_waves = [bandpass.effective_wavelength,
+                           0.5 * (bandpass.blue_limit + bandpass.red_limit),
+                           bandpass.blue_limit,
+                           bandpass.red_limit]
+        for w in candidate_waves:
+            prof0 = self.evaluateAtWavelength(w)
+            if prof0.flux != 0:
+                return w, self.evaluateAtWavelength(w)
+        raise ValueError("Could not fiducial wavelength where SED * Bandpass is nonzero.")
+
     def __repr__(self):
         return 'galsim.ChromaticObject(%r)'%self.obj
 
@@ -281,8 +296,8 @@ class ChromaticObject(object):
         if self.separable: assert hasattr(self, 'SED')
         assert hasattr(self, 'wave_list')
 
-        # setup output image (semi-arbitrarily using the bandpass effective wavelength)
-        prof0 = self.evaluateAtWavelength(bandpass.effective_wavelength)
+        # setup output image using fiducial profile
+        wave0, prof0 = self._fiducial_profile(bandpass)
         image = prof0.drawImage(image=image, setup_only=True, **kwargs)
         # Remove from kwargs anything that is only used for setting up image:
         kwargs.pop('dtype', None)
@@ -296,7 +311,7 @@ class ChromaticObject(object):
 
         if self.separable:
             multiplier = ChromaticObject._multiplier_cache(self.SED, bandpass, tuple(wave_list))
-            prof0 *= multiplier/self.SED(bandpass.effective_wavelength)
+            prof0 *= multiplier/self.SED(wave0)
             image = prof0.drawImage(image=image, **kwargs)
             return image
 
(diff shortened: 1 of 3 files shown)


galsim/chromatic.py
```diff
@@ -142,6 +142,25 @@ def resize_multiplier_cache(maxsize):
...
+        Return a fiducial achromatic profile of a chromatic object that can be used to estimate
+        default output image characteristics, or in the case of separable profiles, can be scaled to
+        give the monochromatic profile at any wavelength or the wavelength-integrated profile.
+        """
+        bpwave = bandpass.effective_wavelength
+        candidate_waves = [bpwave, 0.5 * (bandpass.blue_limit + bandpass.red_limit)]
+        candidate_waves = np.union1d(candidate_waves, bandpass.wave_list)
+        candidate_waves = np.union1d(candidate_waves, self.wave_list)
+        # Prioritize wavelengths near the bandpass effective wavelength.
+        candidate_waves = candidate_waves[np.argsort(np.abs(candidate_waves - bpwave))]
```
rmandelb: OK, so the idea here is that in most cases it will start at the effective wavelength, get nonzero flux, and return immediately, so the fact that this list is kind of long doesn't normally make the function expensive?  Nifty.

tests/test_chromatic.py
```diff
@@ -1717,6 +1717,31 @@ def test_ChromaticAiry():
...
+    red_flux = waves > 550.0
+    bp = galsim.Bandpass(galsim.LookupTable(waves, waves**0, interpolant='linear'))
+    blue_sed = galsim.SED(galsim.LookupTable(waves, blue_flux, interpolant='linear'))
+    red_sed = galsim.SED(galsim.LookupTable(waves, red_flux, interpolant='linear'))
+
+    gal1 = galsim.Gaussian(fwhm=1) * blue_sed
+    gal2 = galsim.Gaussian(fwhm=1) * red_sed
+    img1 = gal1.drawImage(bp)
+    img2 = gal2.drawImage(bp)
+    # Didn't see an obvious np.testing method for checking finiteness, so just use assert.
```
rmandelb: I think you're right that there isn't one, so if you want to use a np.testing routine it would have to be something like this:
```
np.testing.assert_equal(np.isfinite(img1.array.sum()).astype(int), 1, "whatever")
```

rmandelb:
Looks good to me.  Should update CHANGELOG.

galsim/chromatic.py
jmeyers314: Exactly.  One could, I suppose, try to short-circuit this method before the sort too.  I didn't bother since probably the sort is still subdominant to actually drawing images (though I didn't actually check this).

tests/test_chromatic.py
jmeyers314: True.  Are there any side benefits then to using numpy.testing methods over plain asserts?  I think the assert is a bit easier to read.

rmandelb: I find the code easier to read with assert, but the error message when there's a problem is a bit more informative / easier to read with np.testing (just in my opinion).  For example, consider

```
x = 1
y = 3
assert x==y,'badness happened'
```
which gives the following:
```
AssertionError                            Traceback (most recent call last)
/Users/rmandelb/<ipython-input-6-1ea81ded12f9> in <module>()
----> 1 assert x==y,'badness happened'

AssertionError: badness happened
```

This tells me the code that failed.  But if I had done
```
np.testing.assert_equal(x,y,err_msg='badness happened')
```
I would have gotten
AssertionError: 
```
Items are not equal: badness happened
 ACTUAL: 1
 DESIRED: 3
```
which tells me how wrong things were, instead of just that they were wrong.

jmeyers314:
Added to CHANGELOG:

```
- Improved ability of ChromaticObjects to find fiducial achromatic profiles
  and wavelengths with non-zero flux. (#680)
```

tests/test_chromatic.py
jmeyers314: Sure, but for this particular case, I think the text string after the assert is probably sufficient.  The additional information the numpy.testing would reveal (`Items are not equal.  ACTUAL: 0, DESIRED: 1`) isn't all that helpful.

galsim/chromatic.py
rmjarvis: I guess I would have preferred only doing all this if the bpwave had zero flux, since that is by far the dominant case:
```
prof0 = self.evaluateAtWavelength(bandpass.effective_wavelength)
if prof0.flux != 0:
    return w, prof0
else:
    candidate_waves = ...
    ...
```
Even if this is subdominant, the unioning and sorting steps just seems wasteful to do every time.

galsim/chromatic.py
```diff
@@ -142,6 +142,25 @@ def resize_multiplier_cache(maxsize):
...
+
+        for w in candidate_waves:
+            prof0 = self.evaluateAtWavelength(w)
+            if prof0.flux != 0:
+                return w, self.evaluateAtWavelength(w)
```
rmjarvis: Also, either way, you have an extra call to `evaluateAtWavelength` here.  The return should just be `w, prof0`.

galsim/chromatic.py
```diff
@@ -142,6 +142,25 @@ def resize_multiplier_cache(maxsize):
...
+    def _fiducial_profile(self, bandpass):
+        """
+        Return a fiducial achromatic profile of a chromatic object that can be used to estimate
+        default output image characteristics, or in the case of separable profiles, can be scaled to
+        give the monochromatic profile at any wavelength or the wavelength-integrated profile.
+        """
+        bpwave = bandpass.effective_wavelength
+        candidate_waves = [bpwave, 0.5 * (bandpass.blue_limit + bandpass.red_limit)]
+        candidate_waves = np.union1d(candidate_waves, bandpass.wave_list)
+        candidate_waves = np.union1d(candidate_waves, self.wave_list)
```
rmjarvis: And another efficiency point.  `np.union1d` includes a sort, which you don't care about, since you're resorting later.  Better to just do `candidate_waves += wave_list` here as a concatenation.

galsim/chromatic.py
jmeyers314: >I would have preferred only doing all this if the bpwave had zero flux, since that is by far the dominant case:

Done.

galsim/chromatic.py
jmeyers314: >Also, either way, you have an extra call to evaluateAtWavelength here. The return should just be w, prof0.

Yep.  Thanks.  Fixed.

galsim/chromatic.py
jmeyers314: >And another efficiency point. np.union1d includes a sort, which you don't care about, since you're resorting later. Better to just do candidate_waves += wave_list here as a concatenation.

It turned out to be easier to work with numpy.concatenate than +=, but point taken.  Fixed.

rmandelb:
Is this done?  It looks like Mike and I both went over it...

jmeyers314:
>Is this done? It looks like Mike and I both went over it...

I was just waiting for Mike and you to approve my corrections.  I think it's basically done though.

rmjarvis:
Looks good to me.

rmandelb:
Looks good to me, too.  ( Also @jmeyers314 , I can't speak for Mike, but I really don't mind if you ping me after a while if I seem to be sitting on stuff in a way that is holding you back...  if I'm too busy I'll just say so, but it could be that it's actually fallen off my stack and I would be happy to put it back on!  Sorry about this.)

jmeyers314:
>I really don't mind if you ping me after a while if I seem to be sitting on stuff in a way that is holding you back...

@rmandelb , don't worry, this wasn't holding me back from anything. :)  But you're right, it's good to double check that PRs haven't simply been forgotten after a while.

PR #686, opened 2015-10-20
KSB width

jmeyers314:
A quick issue-less PR (in branch ksb_width).  For treating galaxy color-gradients, [Semboloni++13](http://adsabs.harvard.edu/abs/2013MNRAS.432.2385S) show that one can trade off SNR and shear bias for moments-based algorithms by manipulating the width of the moments-measuring weight function.  @sowmyakth (grad student at Stanford) and I would like to investigate this further.

After a brief email exchange with @rmandelb , I learned that flexible weight functions could be pretty quickly added to the galsim.hsm package for the KSB method.  This PR adds two keywords to HSMParams that affect the weight function used by the KSB method:

- ksb_sig_weight: assert the width of the weight function (in pixels) directly
- ksb_sig_factor: multiply the auto-generated width by this factor

(Sorry about the whitespace.  Anyone know what happens if I clean up the whitespace in master now?   will it then show up in other existing PRs?, or are PR diffs only relative to the branch at the time of the PR?)

### DIFF
--- a/include/galsim/hsm/PSFCorr.h
+++ b/include/galsim/hsm/PSFCorr.h
@@ -21,10 +21,10 @@
 #define GalSim_PsfCorr_H
 
 /**
- * @file PSFCorr.h 
+ * @file PSFCorr.h
  *
  * @brief Contains functions for the hsm shape measurement code, which has functions to carry out
- *        PSF correction and measurement of adaptive moments. 
+ *        PSF correction and measurement of adaptive moments.
  *
  * All functions are in the hsm namespace.
  */
@@ -137,6 +137,11 @@ namespace hsm {
      *                           exception.
      * @param ksb_moments_max    Use moments up to ksb_moments_max order for KSB method of PSF
      *                           correction.
+     * @param ksb_sig_weight     The width of the weight function for the KSB method.  Normally,
+     *                           this is computed from the measured moments of the galaxy image.
+     *                           This keyword overrides this calculation.
+     * @param ksb_sig_factor     Factor by which to multiply the weight function width for the KSB
+     *                           method.  Default is 1.0.
      * @param failed_moments     Value to report for ellipticities and resolution factor if shape
      *                           measurement fails.
      */
@@ -151,6 +156,8 @@ namespace hsm {
                   double _max_amoment,
                   double _max_ashift,
                   int _ksb_moments_max,
+                  double _ksb_sig_weight,
+                  double _ksb_sig_factor,
                   double _failed_moments) :
             nsig_rg(_nsig_rg),
             nsig_rg2(_nsig_rg2),
@@ -163,6 +170,8 @@ namespace hsm {
             max_amoment(_max_amoment),
             max_ashift(_max_ashift),
             ksb_moments_max(_ksb_moments_max),
+            ksb_sig_weight(_ksb_sig_weight),
+            ksb_sig_factor(_ksb_sig_factor),
             failed_moments(_failed_moments)
         {}
 
@@ -181,6 +190,8 @@ namespace hsm {
             max_amoment(8000.),
             max_ashift(15.),
             ksb_moments_max(4),
+            ksb_sig_weight(0.0),
+            ksb_sig_factor(1.0),
             failed_moments(-1000.)
             {}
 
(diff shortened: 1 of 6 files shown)


rmjarvis:
> Anyone know what happens if I clean up the whitespace in master now? will it then show up in other existing PRs?, or are PR diffs only relative to the branch at the time of the PR?

The diff is relative to the last point of contact with the master branch.  Typically the last time you merged master into the branch.  Or if you haven't, then the point where you made the new branch.

jmeyers314:
>The diff is relative to the last point of contact with the master branch. Typically the last time you merged master into the branch. Or if you haven't, then the point where you made the new branch.

Ah okay.  Just to be sure though, you're saying if I merge in a WS-cleaned master now, this will be immediately reflected in this diff but no others?

rmjarvis:
Yes.  You can also merge the cleaned master into the other PR branches, and they'll stay ok.

jmeyers314:
Thanks @rmjarvis .  Whitespace is solved.

include/galsim/hsm/PSFCorr.h
```diff
@@ -137,6 +137,11 @@ namespace hsm {
      *                           exception.
      * @param ksb_moments_max    Use moments up to ksb_moments_max order for KSB method of PSF
      *                           correction.
+     * @param ksb_sig_weight     The width of the weight function for the KSB method.  Normally,
+     *                           this is computed from the measured moments of the galaxy image.
+     *                           This keyword overrides this calculation.
+     * @param ksb_sig_factor     Factor by which to multiply the weight function width for the KSB
```
rmandelb: One might wonder if you are supposed to use only one of these (but not both).  the code is set up so that using both is an option; might mention this explicitly?

include/galsim/hsm/PSFCorr.h
```diff
@@ -137,6 +137,11 @@ namespace hsm {
      *                           exception.
      * @param ksb_moments_max    Use moments up to ksb_moments_max order for KSB method of PSF
      *                           correction.
+     * @param ksb_sig_weight     The width of the weight function for the KSB method.  Normally,
```
rmandelb: Should say whether units = pixels or units = whatever was used for the image ```scale``` parameter.

tests/test_hsm.py
```diff
@@ -781,6 +781,30 @@ def test_bounds_centroid():
     t2 = time.time()
     print 'time for %s = %.2f'%(funcname(),t2-t1)
 
+def test_ksb_sig():
```
rmandelb: The test doesn't seem all that helpful, in the sense that (for example) if the code was accidentally disabled by some bug, you would still get the right answer.  I recommend doing a test that would make sure the new sig parameters are actually being used.  For example: make a two-component galaxy such that there is a clear ellipticity gradient.  Measure shear using two different effective radii, and require that difference in the shears has the expected sign and is nonzero.

rmandelb:
Tests pass on both my machines.  Other than the 3 small comments I put on the code, my only other comment is that you should mention this in CHANGELOG.

include/galsim/hsm/PSFCorr.h
jmeyers314: Done.

include/galsim/hsm/PSFCorr.h
jmeyers314: Yep.  Here's the new docstring snippet:

```
     * @param ksb_sig_weight     The width of the weight function (in pixels) to use for the KSB
     *                           method.  Normally, this is derived from the measured moments of the
     *                           galaxy image; this keyword overrides this calculation.  Can be
     *                           combined with ksb_sig_factor.
     * @param ksb_sig_factor     Factor by which to multiply the weight function width for the KSB
     *                           method (default: 1.0).  Can be combined with ksb_sig_weight.
```

rmandelb: Great, thanks.

tests/test_hsm.py
jmeyers314: Good idea.  I added your suggested test to the testing suite (just for `ksb_sig_factor`).  I think this plus what I had before then covers both `ksb_sig_factor` and `ksb_sig_weight`.

jmeyers314:
added to CHANGELOG:
```
- Added ability to manipulate the width of the moment-measuring weight function
  for the KSB shear estimation method of the galsim.hsm package. (#686)
```

tests/test_hsm.py
rmandelb: Looks good, thanks!

rmjarvis:
No comments.  Looks fine to me.

Issue #640, opened 2015-02-06
Chromatic Real Galaxy


jmeyers314:
GalSim has both `RealGalaxy` objects, and `ChromaticObject`s, but doesn't have `ChromaticRealGalaxy` (yet!).

@rmjarvis, @rmandelb, and I talked a bit about how one might take high-res HST images of a galaxy through 2+ filters, deconvolve the known (chromatic) HST PSF, and then reconvolve by a different chromatic PSF (to presumably less resolution than HST, c.f. `RealGalaxy`).  This would be especially helpful for studies of galaxies with color-gradients for LSST/Euclid/etc.

[Semboloni++13](http://adsabs.harvard.edu/abs/2013MNRAS.432.2385S) section 5.2 equation 25 briefly lays this out for the case where the SED at each point in the galaxy is linear in wavelength.  We think we can generalize this to the case of generic SEDs, for instance, typical bulge and disk SEDs.

This issue is to develop the framework for a `ChromaticRealGalaxy` object.  For the moment, I think it's reasonable to limit the scope to just the framework.  In particular, this issue will not attempt to actually assemble a catalog of HST images and PSFs; we'll do that later, and in the meantime we can test with simulated GalSim images.

jmeyers314:
Hey @rmjarvis, @rmandelb, @gbernstein, @barnabytprowe.

I'm in the middle of figuring out how to propagate correlated noise through `RealChromaticGalaxy`, which is under construction in this branch.  See some notes on this [here](https://github.com/jmeyers314/CGNotes).  I believe the right way to handle the noise propagation is to work Fourier mode by Fourier mode.  However, I'm finding that I'm unsure about the right way to consistently compute the required Fourier coefficients for 3 different objects: the input noise power spectrum (given a `CorrelatedNoise` instance), the input HST images, and the (SED*bandpass)-weighted input PSFs.

In particular, I see that the Fourier transforms in `CorrelatedNoise` are done via np.fft, rather than using the stored profile's .drawKImage() method.  What's the reason for this?  I created the script pasted below to investigate the difference between drawK and np.fft, and realized quickly that for the amplitudes at least, the difference is precisely the left panel of Bernstein&Gruen2014 Figure 1.    (So why not use drawKImage with x_interpolant='sinc'?)  I don't understand the phase differences, though I suppose these don't matter for the noise generation in `CorrelatedNoise` since new noise gets generated with uniformly random phase anyway.

[...]

PR #687, opened 2015-10-22
#640

jmeyers314:
Hi all, this PR isn’t ready for a merge yet, but I’d like some feedback on what’s currently in place.  The two main additions are a ChromaticRealGalaxy class that currently only implements the “Special (undocumented) way to build a ChromaticRealGalaxy” (analogous to what’s in RealGalaxy), and the CovarianceSpectrum class which is a generalization of a CorrelatedNoise object (the name is meant to evoke a generalization of a power spectrum).

I’ve placed two test scripts in devel/external/test_chromatic_real_galaxy/ that show off some of the new capabilities.

The test_CRG.py script will add together several parametric profiles to create a galaxy with color gradients, draw this gal through mock HST-like PSFs and a bandpasses, and then use those images to try and reconstruct what would have been seen through a mock Euclid-like PSF and wider bandpass.  You can change things like the number of SED components, number of input bandpasses, pixel-sizes and image sizes from the command line.  Here’s an example for the case of 3 SED components and 3 input images:  you can clearly make out the color gradient from the input images.
`>python test_CRG.py --NSED 3 --Nim 3 --SNR 300 --plot`

￼figure_1

The test_CRG_noise.py script empirically verifies that noise correlation functions get properly propagated.  It generates a series of input “HST” image sets of pure noise from prescribed input correlation functions (one for each filter), pushes these image sets through the ChromaticRealGalaxy analysis, and then compares the observed correlation function of the simulated output images to the prediction made in the ChromaticRealGalaxy code.  Again, you can manipulate a number of parameters from the command line.  Here’s an example:
`>python test_CRG_noise.py --Ntrial 5000 --plot`
￼
figure_2

The 2d images are the log of the pixel-pixel covariance, and the 1d plots are linear- and log-scaled slices through the center of the 2d images.  As you can see, the prediction is a bit high in variance (the value at Delta x = 0), about 15%.  Depending on settings, the difference seems to vary between about 10% too low and 15% too high.  I’m not sure what causes this.

A few points worth mentioning about the current implementation:
- The input images are assumed to be square, as these produce square Fourier transform “pixels” which is required by InterpolatedKImage.  I think this can fairly easily be made more general though.
- The input images are assumed to all have the same wcs; i.e., there’s no attempt to handle dithering or rotation.  This is probably okay for now, since multi-band HST data is typically drizzled onto a common pixel grid, though we may want to extend this in the future.
- The code is written to take ChromaticObjects as input PSFs for each image.  I have recently realized, however, that it still runs fine if one just asserts GSObjects as the input PSFs, i.e., if one ignores the fact that blue and red stars in the same filter have different effective PSFs in the HST images.  It’s possible that capturing the inter-band chromatic PSF differences while ignoring the intra-band chromatic PSF differences is sufficient (and it’s definitely much simpler), albeit somewhat less rigorous.
- Currently, the noise propagation stuff is computed for every ChromaticRealGalaxy.  If there are use cases where one doesn’t care about whitening the noise, then this is an unnecessary time sink.
- I couldn’t think of a good way to write something analogous to CorrelatedNoise.convolveWith(), or at least something that automatically gets called when a ChromaticRealGalaxy is convolved with another profile.  This is because turning the CovarianceSpectrum into a CorrelatedNoise object requires that the output filter be specified, which isn’t done until drawImage is called.  So instead, there’s a toNoise() method of CovarianceSpectrum with arguments for both a PSF and a Bandpass.  I.e., you have to explicitly convolve the noise in parallel with the ChromaticRealGalaxy; it doesn’t happen automatically like for RealGalaxy.

Finally, I took a look at the fits format for RealGalaxyCatalog.  I think this can be trivially extended for at least the simplest version of ChromaticRealGalaxyCatalog where one ignores the intraband chromaticity of the PSF.  One would just have two rows for each galaxy instead of one (joined by a common IDENT).  An open question is whether or not we want to pre-deconvolve the HST PSFs and just store that.  My rough estimate of the time/storage tradeoff is that predeconvolution saves a factor of ~10 in speed at a cost of a factor of ~few in storage for the case of 2 images and 2 SEDs per galaxy.  Maybe the right approach is to ship the original input images, but then make available a routine to convert this catalog into a deconvolved catalog.


galsim/correlatednoise.py
```diff
@@ -667,6 +667,55 @@ def drawImage(self, image=None, scale=None, wcs=None, dtype=None, wmult=1., add_
...
+        @param wmult        A multiplicative factor by which to enlarge (in each direction) the size
+                            of the image, if you are having drawKImage() automatically construct the
+                            images for you.  [default: 1]
+        @param add_to_image Whether to add to the existing images rather than clear out anything in
+                            the image before drawing.  Note: This requires that `re` and `im` be
+                            provided and that they have defined bounds. [default: False]
+
+        @returns the tuple of Image instances, `(re, im)` (created if necessary)
+        """
+        # Check for obsolete dk parameter
```
rmjarvis: No need to include the check for the deprecated dk parameter here.  It's a new function. :)

rmandelb:
A few comments on your note, without reading the code:

> The input images are assumed to be square, as these produce square Fourier transform “pixels”
> which is required by InterpolatedKImage. I think this can fairly easily be made more general though.

Right.  I don't think this is a horrible limitation for a first implementation though.

> The input images are assumed to all have the same wcs; i.e., there’s no attempt to handle dithering 
> or rotation. This is probably okay for now, since multi-band HST data is typically drizzled onto a 
> common pixel grid, though we may want to extend this in the future.

Ditto.  This seems like something to document but perhaps leave as-is for the initial implementation.  (As we've seen before, basically everything we do is found to require some re-designing once people are using it, so I don't feel it's too bad to have some minor limitations in the code given that we basically expect to have to fiddle with it more.)

> The code is written to take ChromaticObjects as input PSFs for each image. I have recently realized, 
> however, that it still runs fine if one just asserts GSObjects as the input PSFs, i.e., if one ignores the 
> fact that blue and red stars in the same filter have different effective PSFs in the HST images. It’s 
> possible that capturing the inter-band chromatic PSF differences while ignoring the intra-band 
> chromatic PSF differences is sufficient (and it’s definitely much simpler), albeit somewhat less 
> rigorous.

Sorry if I am missing something, but it seems to me that this would be easy to test using some realistic stellar SEDs?  (i.e., test the difference between proper treatment of chromatic PSFs vs. achromatic GSObject PSFs)  I feel that this is more important to nail down in an initial implementation, since in practice for HST we generally have an achromatic PSF estimate.

> Currently, the noise propagation stuff is computed for every ChromaticRealGalaxy. If there are use 
> cases where one doesn’t care about whitening the noise, then this is an unnecessary time sink.

Could have a kwarg to shut this off?  (but default is on)

> I couldn’t think of a good way to write something analogous to CorrelatedNoise.convolveWith(), or at 
> least something that automatically gets called when a ChromaticRealGalaxy is convolved with 
> another profile. This is because turning the CovarianceSpectrum into a CorrelatedNoise object 
> requires that the output filter be specified, which isn’t done until drawImage is called. So instead, 
> there’s a toNoise() method of CovarianceSpectrum with arguments for both a PSF and a Bandpass. 
> I.e., you have to explicitly convolve the noise in parallel with the ChromaticRealGalaxy; it doesn’t 
> happen automatically like for RealGalaxy.

Hmmmm.  It would be nice to have that as an option.  What if the Chromatic convolver had a kwarg for the output filter (optional, but if you know it, it could let ChromaticRealGalaxy handle this behind the scenes)?

Finally - I am not loving the ```CovarianceSpectrum``` name.  I don't know why but I'm not finding it obvious/intuitive.  What about ```ChromaticCorrelatedNoise```? (just like we have ```OpticalPSF``` and ```ChromaticOpticalPSF```, ```RealGalaxy``` and ```ChromaticRealGalaxy```, etc.)

jmeyers314:
Working on this branch again.

Quick question:  I'm considering changing `RealGalaxyCatalog.getGal()` to `.getGalImage()` and `.getPSF()` to `.getPSFImage()`, since these are a bit more accurate, and also leave open the possibility of using the names `.getPSF()` and `.getGal()` to return GSObjects or ChromaticObjects, which matches the use of `.getPSF()` for `PhaseScreenList` and even `DES_PSFEx`.  (I ran into this since `ChromaticRealGalaxy` needs to know the input PSF as a `ChromaticObject`, not an `Image`).  My quick grep through the galsim/ directory shows only real.py, scene.py, and wfirst would be affected, but I wanted to check if anyone could think of a reason not to make this change.

rmjarvis:
> I wanted to check if anyone could think of a reason not to make this change.

No objection from me.

jmeyers314:
Out of curiosity, is there a good reason there's no `galsim.ImageC` for complex arrays?  (or maybe `galsim.KImageC`?)  Seems like this would be more natural output for `drawKImage` and input for `InterpolatedKImage` than the real and imaginary parts separately.

rmjarvis:
> Out of curiosity, is there a good reason there's no galsim.ImageC for complex arrays? (or maybe galsim.KImageC?) Seems like this would be more natural output for drawKImage and input for InterpolatedKImage than the real and imaginary parts separately.

In my refactoring for #799, I found this to be a useful change, so this is done now on that branch.

jmeyers314:
Wow.  Tough merge.  345 commits....

Ready to add the last feature though: proper flux normalization.

jmeyers314:
I've been looking into managing flux normalization of `ChromaticRealGalaxy`  (which I think is the last thing I want to finish before rerequesting a review on this very old PR :) ).  While working on this, I noticed the following behavior, which surprised me a bit:

```python
icat = galsim.RealGalaxyCatalog(AEGIS_I_catalog_file)
vcat = galsim.RealGalaxyCatalog(AEGIS_V_catalog_file)
crg = galsim.ChromaticRealGalaxy([vcat, icat], index=0)
print(crg.calculateMagnitude(icat.getBandpass())) # prints 24.90
print(icat.mag[0]) # prints 25.17
```

I.e., there's a ~30% difference between the `ChromaticObject` magnitude and the catalog magnitude.  I don't think this is a bug in `ChromaticRealGalaxy`, as there's an analogous discrepancy also in `RealGalaxy` (it's just harder to see since there's no real analogy to `calculateMagnitude` for achromatic objects).  Here's a scatter plot showing the difference between `-2.5 log10(gal.getFlux())` and the real galaxy catalog magnitude (i.e., `cat.mag[index]`) vs. the magnitude for the COSMOS 25.2 catalog:

cosmos_flux_v_mag

The ridgeline of that plot is basically the zeropoint ~26.

My guess for what's going on is that the magnitude in the catalog (measured by SExtractor MAG_AUTO) only takes pixels into account inside the segmentation map, and probably weights them too.  But `real_galaxy.getFlux()` is using the pixel sum of the wrapped `InterpolatedImage`, so is affected by noisy pixels not really belonging to the galaxy.  I think the impact of this is that the "flux" of the real galaxy, as you might try to specify with the `flux` keyword argument to `RealGalaxy()` or `simReal()` for instance, rescales based on the pixel sum instead of what's probably a better estimate of the flux from SExtractor.  The difference is generally in the 5-25% range, but can be as large as a couple of magnitudes at the faint end (which admittedly at these magnitudes isn't actually a lot of flux).

I'm not sure if this calls for any code changes or additions (`.withCatalogFlux()`?), but I do think we could at least improve the documentation of the `RealGalaxy` normalization (unless this is documented somewhere already?  If so, I didn't find it.)

rmjarvis:
I think there are probably places in the InterpolatedImage code that relies on the flux being equal to np.sum(image.array).  So allowing these objects to have a different nominal flux might lead to problems.  So I guess I'm inclined to agree about pushing this onto the documentation, rather than change any code for it.

jmeyers314:
>I think there are probably places in the InterpolatedImage code that relies on the flux being equal to np.sum(image.array).

I assumed as much as well.

The idea of adding `.withCatalogFlux()` and `.withCatalogMagnitude` methods are growing on me though.  I could imagine `.catalogFlux` and `.catalogMagnitude` attributes for `RealGalaxy`s and maybe `calculateCatalogFlux` and `calculateCatalogMagnitude` methods for `ChromaticRealGalaxy`s (that take `Bandpass` arguments).

rmandelb:
I tthink your explanation of the effect is definitely correct.  Indeed, my preferred way to normalize the real galaxy image fluxes is to rescale based on known zeropoint offsets between the dataset I'm simulating and COSMOS (so I'm multiplying all postage stamps by a fixed factor), rather than by doing anything that includes rescaling based on the RealGalaxy flux.

jmeyers314:
I think this is finally ready for review!

Summary of additions/changes:

- Added `ChromaticRealGalaxy`, the chromatic version of `RealGalaxy`.   This object uses multi-band HST images of galaxies to produce realistic `ChromaticObject`s, including realistic SEDs and color gradients.
- Added `CovarianceSpectrum`, which is a generalization of a power spectrum (or generalization of a `CorrelatedNoise` object in galsim-speak).  Mostly an implementation detail, as it gets turned into a `CorrelatedNoise` object when an image of a `ChromaticRealGalaxy` gets drawn.
- Added `drawKImage` to `ChromaticObject` and `BaseCorrelatedNoise`.
- Added `wcs` keyword to `GSObject.drawKImage`.  Similarly, added support for non-`PixelScale` wcses to `InterpolatedKImage`.  This was mostly to handle Fourier transforms of non-square real space images.
- Reorganized the packaged bandpasses and SEDs.  They now live in the $galsim_meta_data/bandpasses and $galsim_meta_data/SEDs.  Also updated the bandpass and SED download scripts, and added script for HST AEGIS and CANDELS filters.  Pointed LSST bandpass download script to modern repository.
- Added test_CRG.py and test_CRG_noise.py to devel/modules/chromatic_real_galaxy/.  These are useful test/demo scripts, but too slow to use as unit tests.
- Added devel/modules/CGNotes.tex which goes through the math behind `ChromaticRealGalaxy` and `CovarianceSpectrum`.
- Added a small test catalog (5 objects) for testing `ChromaticRealGalaxy`.
- Added documentation to `RealGalaxy` and `ChromaticRealGalaxy` clarifying that methods like `withFlux` set the postage stamp flux, not the galaxy flux, and that the two can be quite different if the signal-to-noise ratio of the source image(s) is low.
- changed return type of `RealGalaxyCatalog.getPSF` and `RealGalaxyCatalog.getGal` to be `GSObject`s instead of `Image`s.  Use `getPSFImage` and `getGalImage` instead for getting `Image`

rmjarvis:
Awesome.  Long time coming...  :) 

I'll probably take a look this week as a procrastination measure to avoid doing some writing for the Y1 DES paper.

rmjarvis:
@jmeyers314 I haven't done a careful look yet, but I wanted to push back a bit on the new wcs option for drawKImage and InterpolatedKImage.  Could you explain why you decided that was necessary?

AFAICT, it looks like the only place you use it is to deal with non-square input images in ChromaticRealGalaxy, which lead to dkx != dky.  But in the regular RealGalaxy, there are non-square input images, which we simply pass to InterpolatedImage as they are.  This class (in the C++ layer) in turn pads out the image with zeros to make it square.

Why isn't this acceptable for ChromaticRealGalaxy too?

jmeyers314:
>Could you explain why you decided that was necessary?

Sure.  I think it was mostly because I forgot that that's what RealGalaxy did.  That said, I believe the slow part of CRG is solving the weighted linear least squares problem for each Fourier mode.  So reducing the number of modes will, I think, be a non-negligible speed up for what's a pretty slow bit of code.  If you'd like to hide the wcs stuff from the user (`_wcs` kwarg?) and just make it an implementation detail of CRG, though, that's fine with me.

rmjarvis:
> If you'd like to hide the wcs stuff from the user (_wcs kwarg?) and just make it an implementation detail of CRG, though, that's fine with me.

It's more that I think adding wcs as an option to drawKImage is confusing.  The meaning of "WCS" in that case is not really very related to what WCS means everywhere else.  It's not from pixels to sky.  It's from pixels to k-space.

I'll try a version locally that doesn't use a non-pixelscale wcs in CRG and see what breaks and/or gets slower and report back.

rmjarvis:
I just changed the wcs in CRG to use the minimum of the two stepk values, and the only thing that broke was the round trip test that is dominated by ringing.  The ringing seems to have increased a bit, so it only passes if 5% -> 8% in one of the tests (6% in the other).  I'm not very able to evaluate whether this is actually a problem or not.

cf. ebaf707

The running time for all `test_crg` functions in test_real.py went from 1:35 -> 1:39, which seems relatively negligible to me.  

Profiling indicates that `np.lstsq` is indeed the tall pole at 40 sec (from 38 previously).  I believe I can pretty massively speed that up with a call to the C++ layer to use tmv instead of numpy.  Plus, probably SVD is unnecessary, and we can use QRP which is much faster.

TL;DR I'd prefer to switch back to having drawKImage and InterpolatedKImage only taking a scale, not a more general wcs.

jmeyers314:
I think you have a typo: `nyx` -> `nky`.  Also, the test images that aren't from the test catalog are all square (should probably change this!), so these already have nkx = nky.  Of the images from the test catalog, the most non-square one is 46x70 pixels.  I time that galaxy in particular as taking ~6.5s to construct when padding out to a square image, and ~4.5s to construct when using the rectangular image.

I haven't looked at the rectangularness of the entire catalog though.  Maybe this situation is relatively uncommon.

rmjarvis:
Sorry, yes.  And actually, fixing that also fixed the test that was failing.

galsim/real.py
```diff
@@ -839,3 +905,426 @@ def _parse_files_dirs(file_name, image_dir, dir, noise_dir, sample):
...
+                coef[iy, ix] = x
+                Sigma[iy, ix] = dx
+                # Save work by filling in conjugates.
+                coef[-iy, -ix] = np.conj(x)
+                Sigma[-iy, -ix] = np.conj(dx)
+        else:
+          galsim._galsim.ComputeCRGCoefficients(
+            coef.ctypes.data, Sigma.ctypes.data,
+            w.ctypes.data, kimgs.ctypes.data, PSF_eff_kimgs.ctypes.data,
+            NSED, Nim, nkx, nky)
```
rmjarvis: @jmeyers314 FYI, I re-wrote this section in C++, which sped it up by an order of magnitude.  So np.lstsq is no longer the tall pole for running the unit tests at least.  It's mostly now drawImage and drawKImage.

jmeyers314: That's awesome!  Thanks!

If non-square wcs's no longer yield a speedup, then I'm fine with axing them.  I'll try profiling with the test_CRG and test_CRG_noise scripts and see if it makes any difference.

tests/test_real.py
```diff
@@ -257,8 +428,149 @@ def test_noise():
...
+    np.testing.assert_almost_equal(
+            obj1.noise.getVariance(),
+            obj2.noise.getVariance() * galsim.real.HST_area**2)
+
+
+@timer
+def test_crg_noise_draw_transform_commutativity():
+    """Test commutativity of ChromaticRealGalaxy correlated noise under operations of drawImage and
+    applying transformations.
+    """
```
rmjarvis: Are these sufficient tests of the CRG noise calculations?  I was expecting to see something where a CRG was made from a few pure (correlated) noise fields and then drawn, and compare the propagated noise to the measured noise from the drawn image.  In other words, some specific version of what is in your `test_CRG_noise.py` script.

I know it's not the fastest calculation in the world, but can we find some version of it that tests things fairly roughly that doesn't take minutes to run?  Perhaps something that only achieves 10% agreement rather than 1% agreement using smaller and/or fewer images?

examples/demo12.py
```diff
@@ -78,7 +83,7 @@ def main(argv):
     SED_names = ['CWW_E_ext', 'CWW_Sbc_ext', 'CWW_Scd_ext', 'CWW_Im_ext']
     SEDs = {}
     for SED_name in SED_names:
-        SED_filename = os.path.join(galsim.meta_data.share_dir, '{0}.sed'.format(SED_name))
+        SED_filename = os.path.join(datapath, 'SEDs/{0}.sed'.format(SED_name))
```
jmeyers314: >I like the better organization of putting the SEDs and bandpasses each in their own subdirectory of the main `share_dir`.

Thanks!  me too.

>we should at the very least declare this as an API change

added to CHANGELOG

>it might be worth either checking if the path looks like "..../share/galsim/*.sed" and adding in the "SEDs" bit with a deprecation warning

Done.

>Second, I think it would be nice if the Bandpass and SED constructors could look in share_dir/bandpasses and share_dir/SEDs automatically

Done.

>This should really be join(datapath, 'SEDs', '{0}.sed'.format(SED_name).

Also done.  I fixed the Bandpass and SED occurrences of this.  There were a few more that only live in the tests dir that I didn't bother with.

galsim/chromatic.py
```diff
@@ -1567,9 +1661,26 @@ def drawImage(self, bandpass, image=None, integrator='trapezoidal', **kwargs):
...
+            self._last_wcs = image.wcs
             return image
         else:
-            return ChromaticObject.drawImage(self, bandpass, image, integrator, **kwargs)
+            image = ChromaticObject.drawImage(self, bandpass, image, integrator, **kwargs)
+            self._last_wcs = image.wcs
+            return image
+
+    @property
+    def noise(self):
```
jmeyers314: >I think we should go through and move all the noise propagation to noise property functions

Sounds good to me.  Looking into it...

>GSObject and Chromatic object can have default implementations that return None,

I think this may be problematic, since `GSObject.__init__` can have another `GSObject` passed in, and we want to propagate the noise of that object.  So I think we might need to have the base object noise still be a regular attribute so it can be overwritten and vary instance to instance.  We can certainly make `Transform.noise` a property though (and `Sum`, `Convolve`, ...), and delay noise computations for them.

galsim/chromatic.py
```diff
@@ -1567,9 +1661,26 @@ def drawImage(self, bandpass, image=None, integrator='trapezoidal', **kwargs):
...
+        # 1) All transformations are achromatic.
+        # 2) This ChromaticTransformation wraps a ChromaticConvolution with a valid noise property.
+        if (hasattr(self._jac, '__call__') or
+            hasattr(self._offset, '__call__') or
+            not self._flux_ratio._const):
+            raise TypeError("Cannot propagate noise through chromatic transformation")
+        noise = self.original.noise
+        jac = self._jac
+        flux_ratio = self._flux_ratio(42.) # const, so use any wavelength
+        return noise.transform(jac[0,0], jac[0,1], jac[1,0], jac[1,1])*flux_ratio**2
```
jmeyers314: Not completely sure I grokked this one.  Is this 2c608bf6bc what you had in mind?

rmjarvis: Yes, although I'd forgotten that this, being a BaseCorrelatedNoise, would need the extra bits for the rng and wcs.  Up to you whether you want to keep the longer (but slightly more efficient) version or keep what you originally had.

galsim/correlatednoise.py
```diff
@@ -1465,3 +1506,86 @@ def __repr__(self):
...
+class CovarianceSpectrum(object):
+    """Class to hold a `ChromaticSum` noise covariance spectrum (which is a generalization of a
+    power spectrum or equivalently a correlation function).
+
+    Analogous to how a galsim.CorrelatedNoise object stores the variance and covariance of a
+    galsim.Image object, a galsim.CovarianceSpectrum stores the variance and covariance of the
+    Fourier mode amplitudes in different components of a `ChromaticSum`.  Note that the covariance
+    in question exists between different SED-components of the `ChromaticSum`, and not between
+    different Fourier modes, which are assumed to be uncorrelated.  This structure arises naturally
+    for `RealChromaticGalaxy`s (see devel/modules/CGNotes.pdf for more details).
```
jmeyers314: tbh I can never remember which one of these I landed on.  Fixed here and elsewhere.

galsim/chromatic.py
jmeyers314: Actually, what's the use case for making a `GSObject` out of a subclass?  I can only find a couple cases of this in the repo (that I think could be eliminated), and I feel like the fact that it's a base class but can also wrap subclasses has gotten in the way before...

rmjarvis: I think that was just an implementation convenience (and barely much of one at that), so I'd be very happy to remove it if it's no longer convenient.  :)

jmeyers314: I took the noise bit out of `GSObject.__init__` entirely, and also removed the ability to pass in a different `GSObject`.  (I'm not sure we really need the bit of `GSObject.__init__` that remains either (passing in an `SBProfile`), but that at least wasn't in the way so I just left it).

I decided not to go with a standard python `@property` for `GSObject.noise`, since that would have required a `noise.setter` method and `_noise` instance variable too (to enable users to attach their favorite `_BaseCorrelatedNoise` object to, e.g., `gaussian_galaxy.noise`).  I did find, however, that a [`lazy_property`](http://stackoverflow.com/a/6849299) would work.  It turns out that not only does this delay and cache the computation of an "attribute", due to minutiae of the python descriptor protocol, the cached value is even faster to access than a regular `@property` - I think just as fast as a regular attribute.

With this in place, I removed all the noise propagation bits from the transformation methods and compound constructors, and find that, for example, `sheared = real_galaxy.shear(g1=0.1)` goes down from ~40 microseconds to ~20 microseconds.  Fun :).

Finally, to squash a bug, I decided to have `RandomWalk` inherit from `GSObject` instead of `Sum`.  Despite being a sum of a bunch of Gaussians, `RandomWalk` doesn't really behave like a `Sum`.  In particular, there's no `self._obj_list` attribute, which some of the `Sum` methods depend on.  I think it's simpler to change the parent class than to keep overriding methods that work for `Sum` but not `RandomWalk`.

next up, chromatic noise...

rmjarvis: > I'm not sure we really need the bit of GSObject.__init__ that remains either (passing in an SBProfile

I think the logic there was to make sure that GSObjects always have an `SBProfile` attribute.  But given the way that Python works (unlike C++) it's just as easy to forget to call `GSObject.__init__(self, sbp)` as it is to set `self.SBProfile = sbp`.  So it's probably an unPythonic legacy of the fact that our initial python layer code was written by C++ coders.  :)

rmjarvis: BTW, I like that `@lazy_property` implementation you stole from SO.  Very neat!

galsim/gsobject.py
```diff
@@ -1884,6 +1881,7 @@ def drawKImage(self, image=None, nx=None, ny=None, bounds=None, scale=None,
                             If `scale` is None and `image` is None, then use the Nyquist scale.
                             If `scale <= 0` (regardless of `image`), then use the Nyquist scale.
                             [default: None]
+        @param wcs          The wcs. [default: None]
```
rmjarvis: I took the liberty of reverting this (and the InterpotedKImage wcs option) to make my life easier merging in the new stuff that just got merged to master.  The commit is 44da24d823d401966e182688d54e86141c1c70af if you want to look through and make sure I didn't remove anything you think is important.  

Nothing broke in the unit tests when I did that (other than the unit tests specifically designed to test these features, which I removed).  I also ran your test_CRG and test_CRG_noise scripts, which also seem to be ok.

galsim/interpolatedimage.py
```diff
@@ -803,18 +818,12 @@ def __setstate__(self, d):
...
-def _SBIKI_getinitargs(self):
-    if self._cenIsSet():
-        return (self._getKData(), self.dK(), self.stepK(), self.maxK(), self.getKInterp(),
-                self.centroid().x, self.centroid().y, True, self.getGSParams())
-    else:
-        return (self._getKData(), self.dK(), self.stepK(), self.maxK(), self.getKInterp(),
-                0.0, 0.0, False, self.getGSParams())
-_galsim.SBInterpolatedKImage.__getinitargs__ = _SBIKI_getinitargs
+_galsim.SBInterpolatedKImage.__getinitargs__ = lambda self: (
+        self._getKData(), self.stepK(), self.maxK(), self.getKInterp(), self.getGSParams())
```
rmjarvis: This isn't covered.  Should be easy to add a `do_pickle` call to cover this.

galsim/chromatic.py
```diff
@@ -421,6 +421,97 @@ def drawImage(self, bandpass, image=None, integrator='trapezoidal', **kwargs):
...
+        # drawn one profile at a time, and hence _last_n_eval will only represent the final
+        # component drawn.
+        self._last_n_eval = integrator.last_n_eval
+
+        # Apply integral to the initial image appropriately.
+        # Note: Don't do image = integral and return that for add_to_image==False.
+        #       Remember that python doesn't actually do assignments, so this won't update the
+        #       original image if the user provided one.  The following procedure does work.
+        if add_to_image:
+            image += image_int
```
rmjarvis: This branch isn't covered.  Probably should add a test of `drawKImage(..., add_to_image=True)`.

galsim/chromatic.py
```diff
@@ -1143,11 +1234,14 @@ def drawImage(self, bandpass, image=None, integrator='trapezoidal', **kwargs):
...
         @returns the drawn Image.
         """
+        # Store the last bandpass used.
+        self._last_bp = bandpass
         if self.SED.dimensionless:
             raise ValueError("Can only draw ChromaticObjects with spectral SEDs.")
 
         int_im = self._get_interp_image(bandpass, image=image, integrator=integrator, **kwargs)
         image = int_im.drawImage(image=image, **kwargs)
+        self._last_wcs = image.wcs
```
rmjarvis: These two lines aren't covered, which means this whole function is never being called in the tests.  Not directly related to this PR I guess, but probably worth doing.

galsim/real.py
```diff
@@ -509,6 +559,7 @@ def close(self):
         self.loaded_files = {}
 
     def getNObjects(self) : return self.nobjects
+    def __len__(self): return self.nobjects
```
rmjarvis: This line is not covered.  It makes sense as a thing for a RealGalaxyCatalog to have, but I guess if you used it for something, you ended up removing that use.  Could probably add a simple test in the main rgc tests that `len(rgc) == rgc.getNobjects() == rgc.nobjects`.

galsim/real.py
```diff
@@ -571,14 +620,26 @@ def _getFile(self, file_name):
...
-    def getGal(self, i):
+    def getBandpass(self):
+        """Returns a Bandpass object for the catalog.
+        """
+        try:
+            bp = real_galaxy_bandpasses[self.band[0].upper()]
+        except KeyError:
+            raise ValueError("Bandpass not found.  To use bandpass '{0}', please add an entry to "
+                             "the galsim.real.real_galaxy_bandpasses "
+                             "dictionary.".format(self.band[0]))
```
rmjarvis: Probably worth adding a test of this.  i.e check that it raises a ValueError when some custom bandpass isn't in the dict.  Then add it in by hand in the test and check again, and it should work.

galsim/real.py
```diff
@@ -839,3 +908,420 @@ def _parse_files_dirs(file_name, image_dir, dir, noise_dir, sample):
...
+            if logger:
+                logger.debug('ChromaticRealGalaxy %d: Got psf', use_index)
+
+            bands = [rgc.getBandpass() for rgc in real_galaxy_catalogs]
+
+            xis = []
+            for rgc in real_galaxy_catalogs:
+                noise_image, pixel_scale, var = rgc.getNoiseProperties(use_index)
+                # Make sure xi image is odd-sized.
+                if noise_image.array.shape[0] % 2 == 0:
```
rmjarvis: The False branch here never gets run.  I guess all the noise images in the RGCs have even dimensions?  If so, maybe just put a `# pragma: no branch` here with a comment that all of ours are odd, but we don't require that for user-provided RGCs.

galsim/utilities.py
```diff
@@ -1331,3 +1331,46 @@ def rand_with_replacement(n, n_choices, rng, weight=None, _n_rng_calls=False):
...
+    meant to be used for lazy evaluation of an object attribute.
+    property should represent non-mutable data, as it replaces itself.
+    """
+    def __init__(self, fget):
+        self.fget = fget
+        self.func_name = fget.__name__
+
+    def __get__(self, obj, cls):
+        if obj is None:
+            return None
```
rmjarvis: This branch never happens.  And actually, it seems a weird thing to check.  How could `obj` be `None`?  Isn't `obj` the instance of the class that you're calling `obj.some_attribute` on?  So it doesn't seem like `None` is a viable possibility.

galsim/gsobject.py
```diff
@@ -869,7 +840,7 @@ def _shift(self, offset):
         return new_obj
 
     # Make sure the image is defined with the right size and wcs for drawImage()
-    def _setup_image(self, image, nx, ny, bounds, add_to_image, dtype, odd=False, wmult=1.):
+    def _setup_image(self, image, nx, ny, bounds, add_to_image, dtype, odd=False, N=None, wmult=1.):
```
rmjarvis: After removing the wcs stuff from drawKImage, the `N` option here is never used.  I guess I should have removed it as part of that commit, but I didn't remember to.

galsim/correlatednoise.py
```diff
@@ -1465,3 +1504,86 @@ def __repr__(self):
...
+        pk = galsim.ImageCD(pkout + 0j, scale=stepk)  # imag part should be zero
+        iki = galsim.InterpolatedKImage(pk)
+        iki *= wcs.pixelArea()**2  # determined this empirically
+        return galsim.correlatednoise._BaseCorrelatedNoise(rng, iki, wcs)
+
+    def __eq__(self, other):
+        return (isinstance(other, galsim.CovarianceSpectrum) and
+                self.SEDs == other.SEDs and
+                self.Sigma == other.Sigma)
+    def __ne__(self, other): return not self.__eq__(other)
```
rmjarvis: This isn't tested.  Should probably add a quick test that `covspec1 != covspec2` for two different CovarianceSpectra.

galsim/interpolatedimage.py
jmeyers314: Done.  Even caught a bug.

galsim/chromatic.py
jmeyers314: There's a bug here too, which is related to `gsobject.drawKImage()` not having a `setup_only` kwarg, so I'm implementing that.  One question though: should we raise an exception in `gsobject.drawKImage()` if both `add_to_image` and `recenter` are True?  Seems like it might be surprising behavior to do both.

rmjarvis: > should we raise an exception in gsobject.drawKImage() if both add_to_image and recenter are True? Seems like it might be surprising behavior to do both.

Why would it be surprising?  `recenter=True` is the default, and I suspect `recenter=False` would be a rare use case.  I mostly added that option because `_drawKImage` doesn't recenter, so I wanted to have a place to describe what that meant, and then the doc string for `_drawKImage` can say it is equivalent to `drawKImage(..., recenter=False)`.

But it seems to me that the normal use case for `add_to_image=True` would be something like:
```
obj1.drawKImage(im)
obj2.drawKImage(im, add_to_image=True)
```
That should probably behave sensibly without a warning.

jmeyers314: Oh right.  I meant it would be surprising to actually do a recenter on a not-to-be-zeroed image.  (If it already exists, then we shouldn't be messing with its wcs, right?)  If the image is already centered, then `recenter=True` should always be fine, regardless of the state of `add_to_image`.

So maybe the condition I want to raise on is actually:
```python
if recenter and image.center() != galsim.PositionI(0,0) and add_to_image:
    raise ValueError("Cannot recenter a non-centered image when add_to_image=True")
```

rmjarvis: That sounds ok to me.

galsim/utilities.py
jmeyers314: I think `obj=None` occurs if you try to access the property from the class instead of from an instance of the class.  (See https://docs.python.org/3/reference/datamodel.html#invoking-descriptors).  I kept it mostly b/c it was in the SO answer and in the pure python description of a `@property`: https://docs.python.org/3/howto/descriptor.html#properties.

So I guess we need to decide what ought to happen if a user tries to access, e.g., `galsim.Gaussian.noise`...  I don't have a strong opinion on this, but returning `None` seems reasonable (though this would also be the return value without this branch I think).

rmjarvis: Looking into this a bit, I think None is the wrong thing to return:
```
>>> galsim.Gaussian.sigma
<property object at 0x10578f940>
>>> print galsim.Gaussian.noise
None
```
In the Python docs, they don't return `None` for this branch. they return self, which IMO looks a bit more consistent:
```
>>> galsim.Gaussian.noise
<galsim.utilities.lazy_property object at 0x10f6d7490>
```
Without the branch entirely, it raises an error
```
>>> galsim.Gaussian.noise
  File "/Users/Mike/lib/python2.7/site-packages/galsim/utilities.py", line 1373, in __get__
    setattr(obj, self.func_name, value)
AttributeError: 'NoneType' object has no attribute 'noise'
```

jmeyers314: Yep.  `return self` is better.  Will use that.

galsim/real.py
```diff
@@ -259,10 +279,14 @@ def __init__(self, real_galaxy_catalog, index=None, id=None, random=False,
...
-        # If flux is None, leave flux as given by original image
-        if flux is not None:
-            flux_rescale = flux / self.original_gal.getFlux()
-        if flux_rescale is not None:
+        # Only alter normalization if a change is requested
+        if flux is not None or flux_rescale is not None or normalize_area:
+            if flux_rescale is None:
+                flux_rescale = 1
+            if normalize_area:
+                flux_rescale /= HST_area
```
jmeyers314: Done.  Also, renamed `normalize_area` -> `area_norm` to better match new behavior.

galsim/real.py
```diff
@@ -839,3 +905,426 @@ def _parse_files_dirs(file_name, image_dir, dir, noise_dir, sample):
...
+
+        if rng is None:
+            self.rng = galsim.BaseDeviate()
+        elif not isinstance(rng, galsim.BaseDeviate):
+            raise TypeError("The rng provided to ChromaticRealGalaxy constructor "
+                            "is not a BaseDeviate")
+        else:
+            self.rng = rng
+        self._rng = self.rng.duplicate()  # This is only needed if we want to make sure eval(repr)
+                                          # results in the same object.
```
jmeyers314: This, along with the `__hash__` implementation, is something I appear to have copied from `RealGalaxy`.  Are these questionable there too?

src/SBInterpolatedImage.cpp
```diff
@@ -895,31 +881,61 @@ namespace galsim {
         }
         _flux = kValue(Position<double>(0.,0.)).real();
         dbg<<"flux = "<<_flux<<std::endl;
+        setCentroid();
+    }
+
+    void SBInterpolatedKImage::SBInterpolatedKImageImpl::setCentroid() const {
```
jmeyers314: There wasn't anything inappropriate.  I just thought this was more elegant than using the old `_cenIsSet` boolean member variable, more aligned with how SBII works, and simplifies `SBIKI.__getinitargs__`.  Calculating the centroid for an SBIKI is (surprisingly) only O(N), so less painful than computing the centroid of an SBII (O(N^2)), which is done in its constructor.

share/SEDs/getSEDs.py
```diff
@@ -16,23 +16,27 @@
 #    and/or other materials provided with the distribution.
 #
 """@file getSEDs.py
-Grab example SEDs from the web, clip them at 22050 Angstroms, and then thin with rel_err = 1.e-5 or
-1.e-3 (two versions).
-Note that the outputs of this script, which are the files GALSIM_DIR/share/CWW_*.sed, are
+Grab example SEDs from the web, clip them and then thin.
+Note that the outputs of this script, which are the files GALSIM_DIR/share/SEDs/CWW_*.sed, are
```
jmeyers314: These scripts are mostly just for provenance; I don't expect users to run them.  (I had already added the downloaded data to the repo).  I thought they'd be more discoverable if they lived next to the data, though, which is why they're in share/.

tests/test_real.py
jmeyers314: I more-or-less copied `test_CRG_noise.py` into an actual test, and used an `if __name__ == '__main__'` block to isolate a fastish version (~6 sec on my laptop) from some more interesting versions (~60 sec).

jmeyers314:
Hey Mike,  I think I've either attempted to address or asked for further feedback on each of your comments now.  Let me know what you think of the changes.

galsim/real.py
rmjarvis: Not sure.  I was proposing to remove this line and see if anything breaks.  :)  So maybe try the same experiment with RealGalaxy.  I *think* we don't need it in either case, but I didn't verify.

src/SBInterpolatedImage.cpp
rmjarvis: OK.  I guess this might be premature optimization.  I don't know that the centroid calculation is a tall pole of optimization for any use case with this class.  If it turns out that this is a slow step, it's not too hard to bring back the delayed calculation.

share/SEDs/getSEDs.py
rmjarvis: I guess I'd rather include a README.txt instead of these scripts in the share directory and just point to the scripts in GalSim/devel/ (or some subdirectory).  The README should probably just say the URL where we got each of the kinds of files and what thinning parameters we used on them.  That seems more useful to an end user trying to understand provenance than parsing the scripts (simple though they are).

src/SBInterpolatedImage.cpp
jmeyers314: I had some fun with a profiler visualization tool (https://github.com/jrfonseca/gprof2dot) I saw someone using on slack.  Here's the output for the test_CRG_noise.py devel script:

output

The InterpolatedKImage initializer (the box on the left labeled `interpolatedimage.py:677:__init__`) takes 2.5% of the total run time, of which 1.7% is the Hermitian conjugacy sanity check (the `numeric:2310:allclose` box).  So I think the centroid calculation is probably sufficiently fast.

rmjarvis: Nice.  :)  Looks like a more useful optimization would be to let the user skip that sanity check if they know they are Hermitian.  But no need to do that here....

jmeyers314:
Just ran into a recursion limit in an extended test of `test_CRG_noise.py` caused b/c there's a Sum(Sum(Sum(...))) 1000x deep, and the `Sum._prepareDraw` recurses through this hierarchy.  I can solve this by consolidating Sums of Sums in `Sum.__init__` (i.e., Sum(Sum(a, b), c) -> Sum(a, b, c), but that makes Sum associative, which we currently explicitly test against.  Any reason to not eliminate this test and just make Sum associative?

rmjarvis:
> Any reason to not eliminate this test and just make Sum associative?

No problem with me.  `git blame` tells me that you added that test.  :) 

In fact, the C++ layer does make such things equivalent anyway.  If an argument is an `SBAdd` object, then it just adds its components to the list, not the compound object.  So I don't see any problem with doing this in the python layer as well.

jmeyers314:
> git blame tells me that you added that test. :)

Sounds like something I would do :).   I didn't realize this already was occurring in the C++ layer.  Good to know; will add to python.

src/SBInterpolatedImage.cpp
jmeyers314: I went ahead and made a `_InterpolatedKImage` function that creates a `InterpolatedKImage` object without the sanity checks.  Using it in `ChromaticRealGalaxy` brings that 2.5% above down to 0.6%.

galsim/real.py
jmeyers314: The `rng` attribute for `RealGalaxy` is used in (up to) three places: to optionally pick a random index from the catalog, to pad the input galaxy image when creating the `InterpolatedImage`, and to initialize the returned object's `noise`(`covspec`) attribute.  For the first two of these, I'd argue that once the index and noise-padded images are created, the `rng` isn't needed any more for equality testing, hashing, or eval(repr())-ing.  For the third case, there may be an argument that two `RealGalaxy`s that are the same except for the `rng` of their noise attributes should be treated as unequal, but I'd note that this isn't how we test for equality / hash / eval(repr()) any other `GSObject`s, which can also, in principle, have `noise` attributes.  

So I think my proposal would be to not add `rng` (or `_rng`) as an attribute to either `RealGalaxy` or `ChromaticRealGalaxy`, not include `noise` (or `covspec`) in the equality testing or hashing, and have eval(repr()) create an object with the same `noise` or `covspec` variance/covariance, but with different `rng`s from the original.  We'd probably want to add some documentation too clarifying that two `RG`s or `CRG`s (really all `GSObject`s and `ChromaticObject`s) can compare equal despite different noise/covspec attribute rngs.

rmjarvis: I think we should probably keep `rng` as an attribute, since it's technically part of the public API of RealGalaxy, and it does provide an easy way for the user to access the BaseDeviate object that was used for the various random numbers.  It also makes it easier to skip the `eval(repr(obj))` test in `do_pickle` since we automatically skip that test for object with an `rng` attribute, so we wouldn't need any special handling there.  (Indeed I just checked, and removing all the lines with `self._rng` worked without any edits to the tests.)

But I agree with your other suggestions about hash and eq.

rmjarvis:
Josh, as you probably saw, I made a few cleanup edits, but I think this is basically ready to go if you're happy with them.  Let's say plan to merge on Monday to give anyone else who wants to take a look one last chance to comment.

jmeyers314:
Sure.  I still had plans to deal with the `rng` attributes and move the `SED`/`Bandpass` provenance scripts into devel, replaced by a readme in the share dir.  I also realized that `CRG` probably deserves a `noise_pad` kwarg similar to `RG`.  I think all of those are straightforward though.  Monday's probably fine.

PR #690, opened 2015-11-10
#556 New LSST module with WCS using LSST stack for implementation

rmjarvis:
(Text originally by @danielsf)

I think this implements all of the functionality you need to create a WCS for an LSST detector, provided you already have the LSST stack installed.

There are two classes defined:

LsstCamera wraps the whole afwCameraGeom model for the LSST camera. This class performs the exact (according to the LSST stack) transformations between RA, Dec and pixel coordinates.

LsstWCS uses LsstCamera to create a WCS for a single detector on the LSST Camera.

I have added a test_lsst.py script to the tests/ directory to test all of the relevant functionality.

Let me know what needs fixing.


rmandelb:
A few random thoughts before digging into this very deeply:

- Is there any chance to get a subset of this functionality without requiring the LSST stack?  I just wonder if we could give even some minimal opportunity to use an LSST WCS e.g. by pre-computing and storing a few WCS for some pre-selected locations on the sky and orientation angles, so that if people want to use the LSST module with a proper chip geometry for a single pointing or a few pointings, they don't have to install the whole LSST stack (=big overhead).

- Some aspects of the coding style, and the documentation format/style, differ from GalSim standards.  Should we be enforcing uniformity of the coding and documentation styles on these survey-specific modules?  My inclination would be yes, but I wanted to open that up for discussion.  (To be a little more specific: the docstrings don't use the doxygen-style specification of inputs and outputs; there's no top-level docstring for the lsst module; some of the code doesn't have the level of commenting that I'm used to, explaining the algorithm - e.g. ```_nativeLonLatFromRaDec```.  There's more, but these were just the obvious things that jumped out at me when I looked at it quickly.)

rmandelb:
@rmjarvis - I was wondering if you have any thoughts on the suggestions I made above?

rmjarvis:
I admit I haven't looked at this PR much yet.  I still need to install the DM stack on one of my machines so I can test it out.  I'll try to get on that in the next week or so.

I agree with your comment about the coding (and especially documentation) style.  I can go through and clean up some of that.  No need for Scott to conform to our preferences, but for the stuff that a user sees, I'll try to make it match the rest of GalSim style.

I think the no-stack option will have to be a separate project from this PR.  I think it's still a long term goal, but not what Scott signed on for.  :)  So we can open another issue about trying to get these features available without the LSST stack.

rmandelb:
> I agree with your comment about the coding (and especially documentation) style. I can go through 
> and clean up some of that. No need for Scott to conform to our preferences, but for the stuff that a 
> user sees, I'll try to make it match the rest of GalSim style.

> I think the no-stack option will have to be a separate project from this PR. I think it's still a long term 
> goal, but not what Scott signed on for. :)

Fair enough!

rmjarvis:
I'm trying to find a machine on which I can install the LSST stack.  So far, I've tried the binary installation on my Mac, which failed (cf. https://github.com/airnandez/lsst-cvmfs/issues/5) and on a linux box, I thought I would try from source, since I thought that might be easier, but that both failed and kicked me off my terminal session:
```
$ source newinstall.sh 

LSST Software Stack Builder
=======================================================================

diff: missing operand after `-'
diff: Try `diff --help' for more information.
Detected git version 2.8.1.51.g6a269e5. OK.


LSST stack requires Python 2.7; you seem to have Python 2.7.3 on your
path (alias python='/data3/DECAM/global/Python-2.7.3/bin/python'
	/data3/DECAM/global/Python-2.7.3/bin/python).	 Please set up a compatible python interpreter,
prepend it to your PATH, and rerun this script.	 Alternatively, we can set
up the Miniconda2 Python distribution for you.

In addition to Python 2.7, some LSST packages depend on recent versions of numpy,
matplotlib, and scipy. If you don't have all of these, the installation may fail.
Using the Miniconda2 Python distribution will ensure all these are set up.

Miniconda2 Python installed by this installer will be managed by LSST's EUPS
package manager, and will not replace or modify your system python.

Would you like us to install the Miniconda2 3.19.0.lsst4 Python distribution (if unsure, say yes)? y 

Installing EUPS (v2.0.1)...  done.
-bash: /home/mjarvis/lsst/eups/bin/setups.sh: No such file or directory
Installing Miniconda2 Python Distribution ... 
-bash: eups: command not found
Connection to folio.physics.upenn.edu closed.
```
On logging back in, indeed that file does not exist:
```
$ ls /home/mjarvis/lsst/eups/bin/
$
```
Any thought about what I should try next?  Should I try from source on my Mac?  (I thought that was supposed to be a bit hairy, so I was hoping to avoid it.)  Or do you have a guess about what went wrong with either of the two ways I tried so far?

danielsf:
This is going to sound like a cop-out, but can you try again.

GitHub was down for an indeterminate amount of time this afternoon.  I wonder if you were running this script (and thus trying to clone the EUPS git repo) while that was happening (in which case I have no idea what the failure would look like).

Apologies in advance if that was not the problem.

rmjarvis:
Nope.  Good idea, but I tried again (from scratch on a fresh directory) and got exactly the same behavior.

timj:
The install script is meant to be executed and not sourced: `bash newinstall.sh`. That explains why you were kicked out of your shell. Can you try it again using that technique? You should also be fine on Mac is you use the "w_2016_15" tag.

rmjarvis:
Thanks.  I did manage to get it installed from source on my Mac, but the linux box is still giving me trouble.  If I use `bash newinstall.sh` instead (as the instruction say!) I still get the following:
```
Installing EUPS (v2.0.1)...  done.
newinstall.sh: line 259: /home/mjarvis/lsst/eups/bin/setups.sh: No such file or directory
Installing Miniconda2 Python Distribution ... 
newinstall.sh: line 267: eups: command not found
```
And the `eups/bin/` directory is empty.  No `setups.sh` (or any other) file there.

danielsf:
One thing to try:

newinstall.sh is supposed to be run like
```
bash newinstall.sh
```
not
```
source newinstall.sh
```
We think that would at least explain why your shell exited.  Try that....?

timj:
The most recent version of `newinstall.sh` is meant to check for failure in eups install. Do you have an `eupsbuild.log` file left over?

```bash
(
                mkdir _build && cd _build
                if [[ -z $EUPS_GITREV ]]; then
                        # Download tarball from github
                        $cmd curl -L $EUPS_TARURL | tar xzvf -
                        $cmd cd eups-$EUPS_VERSION
                else
                        # Clone from git repository
                        $cmd git clone "$EUPS_GITREPO"
                        $cmd cd eups
                        $cmd git checkout $EUPS_GITREV
                fi

                $cmd ./configure --prefix="$LSST_HOME"/eups --with-eups="$LSST_HOME" --with-python="$PYTHON"
                $cmd make install

        ) > eupsbuild.log 2>&1 && echo " done." || { echo " FAILED."; echo "See log in eupsbuild.log"; exit -1; }
```

rmjarvis:
I'm willing to pass on the linux box for now unless someone thinks of something obvious to try, since I have an apparently working version on my Mac.

So after the regular installation I followed the instructions [here]:
```
eups distrib install GalSim 1.2
cd /Users/Mike/lsst/DarwinX86/boost/1.55.0.1.lsst2/lib
install_name_tool -id $PWD/libboost_python.dylib libboost_python.dylib
cd /Users/Mike/GalSim
scons TMV_DIR=$LSST_PKGS/tmv/0.72/ FFTW_DIR=$LSST_PKGS/fftw/3.3.3/ BOOST_DIR=$LSST_PKGS/boost/1.55.0.1.lsst2/
```
This seemed to work in that GalSim ran through the intro without failure (a feat in its own right) and built everything.

Now if I `scons install PREFIX=~` this installs the currently built GalSim into /Users/Mike/lsst/DarwinX86/miniconda2/3.19.0.lsst4/lib/python2.7/site-packages, but importing galsim still gives me version 1.2.

So I think I need to do some setup command now to tell eups to use this version rather than 1.2.  Could you help me out with that?  I'm not sure what that command is.

danielsf:
The command to get a list of the packages eups knows about is
```
eups list
```
This will give you a list formatted as
```
version package_name list_of_eups_tags
```
the way to setup a package is with
```
setup package_name -t eups_tag
```
so you will probably want something like
```
setup GalSim -t lsst.1.2
```
or whatever the eups tag is for the eups-installed version is

danielsf:
In a telecon until 11am.  I will return to this thread once that is done.

rmjarvis:
I have this line:
```
GalSim                1.2        	b666 current b659 setup
```
But I don't think any of these are the new development version I'm trying to install.  How do I get it to add the development installation to the list of allowed versions?  I thought `setup -r` something...  But I'm sure I don't really know what I'm doing.

rmjarvis:
BTW, @timj I tried to send you the eupsbuild.log file to your cornell account.  But it bounced, so maybe I don't actually have your correct email address.  Anyway, you can grab it from here: 
http://www.sas.upenn.edu/~mjarvis/eupsbuild.log

timj:
@rmjarvis you might want to try `@lsst.org` instead of `@cornell.edu`...

Your eups build log looks for all the world like you are trying to use python 2.4 or something (something that is at least older than 2.6) since `print_function` is not working. EUPS now requires >= 2.6.

This might be a consequence of https://jira.lsstcorp.org/browse/DM-5636 (and lsst/lsst#19) where you get errors indicating that you are using a version that is not the one being tested. What python do you have in `/usr/bin/python` on that machine? I am guessing that setting `$PYTHON` to the python you want to use will fix things.

jmeyers314:
I'm certainly not an expert, but I think the command you're looking for might be something like:

`eups declare -r DEVEL_VERSION_PATH -t $USER`

and then you can setup via `setup galsim $USER`

danielsf:
Sorry.  Misunderstood.

Josh is right.  In order to declare a new package to eups, the syntax is
```
eups declare -r path/to/package/root package_name package_version
eups declare -r path/to/package/root package_name package_version -t eups_tag
```
(ordinarily, you could do this in one line, but a but was introduced in eups version 2.0.1, which is the version I assume you have). So you would want to do something like
```
eups declare -r /path/to/your/galsim/ GalSim $USER
eups declare -r /path/to/your/galsim/ GalSim $USER -t $USER
```
and then you can set it up via
```
setup GalSim -t $USER
```

rmjarvis:
> eups declare -r /path/to/your/galsim/ GalSim $USER
> eups declare -r /path/to/your/galsim/ GalSim $USER -t $USER
> setup GalSim -t $USER

OK.  This is progress, I guess.  Now it seems to be trying to use the development version:
```
$ python
Python 2.7.11 |Continuum Analytics, Inc.| (default, Dec  6 2015, 18:57:58) 
[GCC 4.2.1 (Apple Inc. build 5577)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
>>> import galsim
Segmentation fault: 11
```
Seems to be some kind of linking problem I guess.  I'll see if I can track that down.

rmjarvis:
> What python do you have in /usr/bin/python on that machine?

Yes, the one in /usr/bin/python is old.  2.4 I think.  I have `python` aliased to a 2.7 version.  I guess I'll have to muck with the PATH to make sure that's the one eups finds.  

BTW, it's kind of a pain to reset this after each failure, since the eups installation is setting all its directories to have unwritable permissions, so `rm -rf` doesn't work.  I first need to `chmod +w -R eups` in order to remove it.  Seems kind of an odd "feature".

danielsf:
We know.  There's a race condition.  This was the best way to avoid it.

timj:
@rmjarvis setting PATH won't fix it. You need to set `$PYTHON` or else specify the python you want to use on the command line.

rmjarvis:
I figured out the seg fault.  It was linking the wrong libpython2.7.dylib file.  The fix was
```
scons EXTRA_LIB_PATH=$LSST_PKGS/miniconda2/3.19.0.lsst4/lib
```
So I'm pretty much rolling on my laptop now.  

The one thing that isn't working yet is nosetests.  Is there a version of that in the stack that will use the miniconda python?  (The nosetests executable I have is trying to use my system python and failing.)

timj:
I think you may just want to do `conda install nose` or something similar.

rmjarvis:
> conda install nose 

Thanks that worked.

@danielsf I'm now finally at the point where I am running your code.  :)  And I'm getting an error:
```
$ python test_lsst.py
  File "test_lsst.py", line 7, in <module>
    from galsim.lsst import LsstCamera, LsstWCS, _nativeLonLatFromRaDec
  File "/Users/Mike/GalSim/galsim/lsst/__init__.py", line 19, in <module>
    from lsst_wcs import *
  File "/Users/Mike/GalSim/galsim/lsst/lsst_wcs.py", line 16, in <module>
    raise ImportError("You cannot use the LSST module.\n"
ImportError: You cannot use the LSST module.
You either do not have the LSST stack installed,
or you have it installed, but have not set it up.
------------
To install the LSST stack, follow the instructions at:

https://confluence.lsstcorp.org/display/SIM/Catalogs+and+MAF

NOTE: you must build the stack with the python you are using to
run GalSim.  This means that when the stack asks you if you want
it to install Anaconda for you , you MUST SAY NO.

------------
If you have installed the stack, run

source $LSST_HOME/loadLSST.bash
setup obs_lsstSim -t sims
```
But the last line doesn't work for me:
```
$ setup obs_lsstSim -t sims
Unable to find an acceptable version of obs_lsstSim
$ eups list obs_lsstSim
eups list: Unable to find product obs_lsstSim
$ 
```

danielsf:
Probably because you did not install it.  You'll want to run
```
eups distrib install obs_lsstSim -t sims
```

rmjarvis:
OK.  Success.  Thanks all for the help!

galsim/lsst/lsst_wcs.py
```diff
@@ -0,0 +1,1153 @@
...
+except ImportError:
+    raise ImportError("You cannot use the LSST module.\n"
+                      "You either do not have the LSST stack installed,\n"
+                      "or you have it installed, but have not set it up.\n"
+                      "------------\n"
+                      "To install the LSST stack, follow the instructions at:\n\n"
+                      "https://confluence.lsstcorp.org/display/SIM/Catalogs+and+MAF\n\n"
+                      "NOTE: you must build the stack with the python you are using to\n"
+                      "run GalSim.  This means that when the stack asks you if you want\n"
+                      "it to install Anaconda for you , you MUST SAY NO.\n\n"
```
rmjarvis: I didn't say no, and I didn't have any trouble in that respect.  I'd suggest removing this sentence.  It just means that the python you need to use for using galsim.lsst is the miniconda python that LSST installs.  That seems fine and probably the easiest option for most users.

Also, I updated the [GalSim wiki page](https://github.com/GalSim-developers/GalSim/wiki/Building-GalSim-with-the-LSST-stack) about building with the stack, so perhaps that would make a better link for this error message.

galsim/lsst/lsst_wcs.py
```diff
@@ -0,0 +1,1153 @@
...
+        dra = ra - self._pointing.ra/galsim.radians
+        cradif = np.cos(dra)
+        sradif = np.sin(dra)
+        sdec = np.sin(dec)
+        cdec = np.cos(dec)
+        denom = sdec * self._sin_dec + cdec * self._cos_dec * cradif
+        xx = cdec * sradif/denom
+        yy = (sdec * self._cos_dec - cdec * self._sin_dec * cradif)/denom
+        xx *= -1.0
+        return xx*self._cos_rot - yy*self._sin_rot, xx*self._sin_rot + yy*self._cos_rot
```
rmjarvis: This seems like a reimplementation of the above function `pupilCoordsFromPoint`.  Seems like both should be simple wrappers around something that does the actual work and would be a single implementation.  Also, I suspect you'll want to use the `project_rad` function, which acts like `project` (used above) but on numpy arrays of ra, dec.

galsim/lsst/lsst_wcs.py
```diff
@@ -0,0 +1,1153 @@
...
+
+        x_g = x*self._cos_rot + y*self._sin_rot
+        y_g = -1.0*x*self._sin_rot + y*self._cos_rot
+
+        x_g *= -1.0
+
+        denom = self._cos_dec - y_g * self._sin_dec
+        d = np.arctan2(x_g, denom) + self._pointing.ra/galsim.radians
+        ra = d%(2.0*np.pi)
+        dec = np.arctan2(self._sin_dec + y_g * self._cos_dec, np.sqrt(x_g*x_g + denom*denom))
```
rmjarvis: Likewise, can this use the `deproject` function?  Maybe cleaner than reimplementing this functionality here.

tests/test_lsst.py
```diff
@@ -0,0 +1,928 @@
...
+
+        with warnings.catch_warnings(record=True) as ww:
+            wcs1 = LsstWCS(self.pointing, 49.0*galsim.degrees,
+                           chip_name='R:0,1 S:1,1', camera=self.wcs.camera)
+
+        self.assertEqual(str(ww[0].message),
+                         "The camera you passed to LsstWCS does not have the same\n"
+                         "pointing and rotation angle as you asked for for this WCS.\n"
+                         "LsstWCS is creating a new camera with the pointing and\n"
+                         "rotation angle you specified in the constructor for LsstWCS.")
```
rmjarvis: This needs some way to run the tests from the command line.  Normally this looks like:
```
if __name__ == '__main__':
    test_something()
    test_something_else()
    ...
```
I'm less familiar with the `unittest` module, but you need to add something here to make it possible to run the tests by hand from the command line.

tests/test_lsst.py
```diff
@@ -0,0 +1,928 @@
...
+
+
+    def test_radec(self):
+        """
+        Test that the conversion from pixel coordinates to RA, Dec orks
+        """
+
+        # test one-at-a-time use case
+        for ra_control, dec_control, xx, yy in \
+            zip(np.radians(self.wcs_data['ra']), np.radians(self.wcs_data['dec']), self.wcs_data['xpix'], self.wcs_data['ypix']):
```
rmjarvis: Some of these lines are a bit long.  We try to keep to 100 char in GalSim.

galsim/lsst/lsst_wcs.py
```diff
@@ -0,0 +1,1153 @@
...
+try:
+    import lsst.pex.logging as pexLog
+    import lsst.daf.base as dafBase
+    import lsst.afw.geom as afwGeom
+    import lsst.afw.cameraGeom as cameraGeom
+    import lsst.afw.image as afwImage
+    import lsst.afw.image.utils as afwImageUtils
+    import lsst.meas.astrom as measAstrom
+    from lsst.afw.cameraGeom import PUPIL, PIXELS, TAN_PIXELS, FOCAL_PLANE
+    from lsst.obs.lsstSim import LsstSimMapper
```
rmjarvis: One of these is importing matplotlib (which seems like an error, but ok).  On my machine this gives a warning:
```
>>> import galsim.lsst
/Users/Mike/lsst/DarwinX86/miniconda2/3.19.0.lsst4/lib/python2.7/site-packages/matplotlib/font_manager.py:273: UserWarning: Matplotlib is building the font cache using fc-list. This may take a moment.
  warnings.warn('Matplotlib is building the font cache using fc-list. This may take a moment.')
```
Can you please wrap these in a context where warnings are ignored so the user doesn't need to see this.  Even better would be to fix whichever package in the stack is doing this, since there's no reason for that to be a top level import in any of these modules.

galsim/lsst/lsst_wcs.py
```diff
@@ -0,0 +1,1153 @@
...
+        ysq = np.power(delta_yList,2).sum()
+
+        aMatrix = np.array([
+                           [xsq, offDiag, 0.0, 0.0],
+                           [offDiag, ysq, 0.0, 0.0],
+                           [0.0, 0.0, xsq, offDiag],
+                           [0.0, 0.0, offDiag, ysq]
+                           ])
+
+        coeffs = np.linalg.solve(aMatrix, bVector)
```
rmjarvis: This all seems like a pretty convoluted way to figure out this 2x2 matrix.  It would be helpful for anyone coming along later trying to debug something if you could explain the steps involved here.  Maybe some mathy comments with what the matrix equation is that you are setting up and solving here.

tests/test_lsst.py
rmjarvis: Also, we have a tradition of timing each test and giving the time to run it at the end of the test.  This is helpful for finding especially long-running tests that might merit being trimmed down a bit.  Also helpful as a progress log while the tests are running.  See any of the other test*.py files for examples.

rmjarvis: I'm not sure that you need to unit test the exact wording of the warning message.  Normally we just check that the appropriate class of warning is raised.  However, I'm getting an error in this test when I run it:
```
ERROR: Test that you can pass a camera from one WCS to another
  File "/Users/Mike/GalSim/tests/test_lsst.py", line 924, in test_passing_camera_by_hand
    self.assertEqual(str(ww[0].message),
IndexError: list index out of range

```
So it seems that maybe the warning didn't actually get emitted...

tests/test_lsst.py
```diff
@@ -0,0 +1,928 @@
+from __future__ import with_statement
+import unittest
+import numpy as np
+import warnings
+import os
+import galsim
+from galsim.lsst import LsstCamera, LsstWCS, _nativeLonLatFromRaDec
```
rmjarvis: This can't be at the top level.  We can't require the LSST stack for regular runs of nosetests (via `scons tests`).  So there needs to be some graceful way to ignore the ImportErrors if someone is running `scons tests` and doesn't have the stack installed.  OTOH, we probably want to let those errors through when someone is running this from the command line.  So probably something like
```
try:
    from galsim.lsst import LsstCamera, LsstWCS, _nativeLonLatFromRaDec
except:
    if __name__ != '__main__':
        print "No LSST stack detected.  Skipping tests in test_lsst.py"
        no_stack=True
    else:
        raise
else:
    no_stack=False
```
then in each test something like
```
if no_stack: return
```
so they gracefully ignore the lack of an installed LSST stack.

timj: I think these should use the standard unittest skipping infrastructure. Set a flag as you have in the modification above and then use the `skipif` decorator for the tests.

tests/test_lsst.py
```diff
@@ -0,0 +1,928 @@
...
+
+    outputs
+    ------------
+    the angular separation between points 1 and 2 in radians
+
+    From http://en.wikipedia.org/wiki/Haversine_formula
+    """
+    t1 = np.sin(lat2/2.-lat1/2.)**2
+    t2 = np.cos(lat1)*np.cos(lat2)*np.sin(long2/2. - long1/2.)**2
+    return 2*np.arcsin(np.sqrt(t1 + t2))
```
rmjarvis: This is fine, although the tests would probably be a bit more self-documenting if they used the CelestialCoord `distance` function.  The term `haversine` is a bit jargonny compared to `distance`

tests/test_lsst.py
```diff
@@ -0,0 +1,928 @@
...
+                sa = np.sin(alpha)
+                rotX = np.array([[1.0, 0.0, 0.0],
+                                    [0.0, ca, sa],
+                                    [0.0, -sa, ca]])
+
+                cb = np.cos(raPointing)
+                sb = np.sin(raPointing)
+                rotZ = np.array([[cb, -sb, 0.0],
+                                    [sb, cb, 0.0],
+                                    [0.0, 0.0, 1.0]])
```
rmjarvis: Totally minor point, but these matrices could be lined up a bit nicer.

tests/test_lsst.py
```diff
@@ -0,0 +1,928 @@
...
+            self.assertAlmostEqual(lat, latc, 10)
+
+
+    def testNativeLongLatComplicated(self):
+        """
+        Test that nativeLongLatFromRaDec works by considering stars and pointings
+        at non-intuitive locations.
+        """
+
+        np.random.seed(42)
```
rmjarvis: This is probably not a huge problem here, but the `numpy.random` PRNG uses a global state variable, so it's not really thread-safe.  I'm guessing these tests aren't in danger of failing for particularly unlucky choices of random numbers, but to be safe you'd be better off using a well-defined state variable:
```
rng = np.random.RandomState(42)
raPointingList = rng.random_sample(nPointings)*2.0*np.pi
decPointingList = rng.random_sample(nPointings)*np.pi - 0.5*np.pi
```

rmjarvis:
@danielsf I'm not sure what your availability for working on this is, but we're planning to release the next GalSim version in a couple/few weeks probably.  Did you want to try to get this included in that release?  Do you have time to work on responding to the comments I made?  It's ok to push back on some of them if you disagree, or to ask me to do things that you don't have the time or inclination to work on.

danielsf:
Sorry.  I've been distracted by other work.  I did see your comments.  I am going to be at a meeting next week (April 25-30).  I can try to work on this during the downtime.  However, it is probably safer to say that I won't get this done until May 6th.  Is that reasonable?

I know for a fact there are some bugs in what I did (i.e. the WCSs that are being created are not correct).  These have been fixed in the LSST stack implementation of this code.  I just need to port the fixes over.

If you need to release without this, I understand and apologize.

rmjarvis:
Thanks, Scott.  I think May 6 would be fine for making it into 1.4, so let's plan on trying to get this included.

danielsf:
To what level of detail does this WCS need to go?

Since first writing the LSST stack implementation of this code, my understanding of what "RA, Dec" means has been refined somewhat.

Previously, the code in this PR used spherical geometry to convert RA, Dec into "native longitude and latitude" (what RA, Dec would be if the celestial pole was directly overhead), fit a Tan WCS to that native longitude and latitude, then refined that Tan WCS based on the actual geometry of the LSST camera.

The correct thing to do is to transform RA, Dec to add in the motion of the Earth (precession, nutation, aberration), then convert to native longitude and latitude and proceed as before.

Adding the motion of the Earth involves re-implementing a bunch of stack code that ultimately depends on PALPY.  Is that functionality you want in GalSim (or is it functionality you already have in GalSim, and I just can't find it)?  Or should we just proceed as before, negelcting the motion of the Earth?  I suspect that is a level of detail that is really only required for people who are trying to simulate an actual LSST pointing, as opposed to just simulating a possible LSST image.

danielsf:
This point may actually be moot.

I just checked, and obs_lsstSim (which is the LSST stack package that actually contains the camera model) already depends on afw (the 'applications framework') and boost, which are the biggest offenders in the 'it takes too long to compile' category.  Since we are already bringing in those dependencies, we don't lose much agility just depending on the LSST stack-implemented coordinate transformation routines.  I will re-factor the code to do so as much as possible.

rmjarvis:
>  I suspect that is a level of detail that is really only required for people who are trying to simulate an actual LSST pointing, as opposed to just simulating a possible LSST image.

I think that's right.  Nowhere else in GalSim do we try to get any of these kinds of details right.  We pretty much assume that there is some RA, Dec system that you are using and the WCS just translates between that and chip coords.  We don't do precession or distinguish FK5 from ICRS, etc.

Furthermore, most of what we care about is the complexity of the WCS relative to the center of the field of view.  Not getting the overall pointing perfectly oriented.  So I'd suggest just ignoring these details and start at whatever the most convenient RA/Dec system is for constructing the WCS.

danielsf:
I originally had ```_neworigin``` modify ```self```.  I believe that was incorrect.  ```_newOrigin``` is supposed to return a new WCS.  Is my new thinking correct?

rmjarvis:
>  _newOrigin is supposed to return a new WCS. 

Yes.  Just about everything* in GalSim is designed to be immutable, so methods that change something return a new object with the applied change, leaving the original as it was.

*I believe the only exceptions are Image, since modifying images is kind of the point of GalSim, and the various random deviates whose internal states change every time they get used.

galsim/lsst/lsst_wcs.py
danielsf: The code that they share in common is so short (really just the last line) that I couldn't think of a way to make them wrappers that wasn't more complicated than what is already there.  Also, it wasn't clear to me what I should do about the fact that pupilCoordsFromFloat returns unitless floats while pupilCoordsFromPoint returns a CelestialPosition.

I did get rid of the PALPY re-implementation and use project_rad() as you suggested.

Note: I am implementing all of these changes on another branch on my local machine.  I will merge that branch into this PR and push it up to github when I am ready for you to take another look at it.

galsim/lsst/lsst_wcs.py
danielsf: Done

galsim/lsst/lsst_wcs.py
rmjarvis: Perhaps something like this:
```
def pupilCoordsFromPoint(self, point):
    if hasattr(point, '__len__'):
        ra = [ p.ra for p in point ]
        dec = [ p.dec for p in point ]
        return self.pupilCoordsFromFloat(ra, dec)
    else:
        return self.pupilCoordsFromFloat(point.ra, point.dec)
```

danielsf: But won't we still need to make sure that FromPoint returns a unit-ful quantity while FromFloat returns a unit-less quantity?  Or did I make that convention up?

rmjarvis: The doc strings claim they both output "The x and y coordinates on the pupil in radians".  But if you want to add units, that would be a simple addition.  My point is mostly that the meat of the calculation be in one place.

danielsf: I see.  Do you have a preference whether these methods return unit-less scalars versus galsim.angles?

rmjarvis: Doesn't matter to me.  As long as it's well documented I'm fine with either.  This is mostly an internal function I guess, not something most users would use, so whatever is more convenient for how you use it.  Although it's maybe more intuitive to have them return the same thing, whichever you choose.

tests/test_lsst.py
danielsf: We were running afoul of python's natural tendency to filter warnings such that the same warning is not raised more than once (the test in question tries to provoke the warning in two different ways).  I have added a warning filter to allow the warning to be raised multiple times (since, presumably, we want the user to be alerted every time they try to create a WCS with a camera that is not configured to the same world origin as the WCS).

danielsf: Regarding testing the exact wording of the warning: I like to do that to make sure that the warning (or exception) being raised is the one I thought I was provoking in the unit test (rather than some other warning from earlier in the traceback).  I have been told this is excessively paranoid and fragile.  If you would prefer I not do that, I can just test that a warning is raised.

rmjarvis: That's fine.  The fragility is just that if you decide to change the warning at some point, then the test also needs to change, but that's not so onerous. So you can leave it this way if you prefer.

rmandelb: In that case, perhaps a good idea to put a comment in the code where the warnings are emitted, saying something like "if you change this warning, you have to go change such-and-such unit test"?

rmjarvis: I don't think that's necessary.  It should be clear what needs to be done, since the test will fail in a way that says the warning message is wrong.

danielsf:
I am under the impression that I have implemented all of the fixes requested in this review, if you want to have another look at the code.  Sorry it took so long.

galsim/lsst/lsst_wcs.py
```diff
@@ -0,0 +1,1164 @@
...
+        yTanPixMin, yTanPixMax = self._getTanPixelBounds()
+
+
+        xPixList = []
+        yPixList = []
+        nameList = []
+
+        #dx and dy are set somewhat heuristically
+        #setting them eqal to 0.1(max-min) lead to errors
+        #on the order of 0.7 arcsec in the WCS
```
rmjarvis: This comment confuses me.  The next two lines use 0.5(max-min) for dx and dy, not 0.1.  So is that more accurate than using 0.1, or less?  If less (or the same), then I think we want something better, since 0.7 arcsec is a huge error!  That's wrong by of order the full width of a typical PSF.

danielsf: It is more accurate.  Setting dx, dy to 0.1(max-min) caused an overfitting of the WCS.  test_tan_wcs() and test_tan_sip_wcs() in test_lsst.py verify that the WCS is accurate to within a milli-arcsecond.

danielsf:
Just checking: are y'all waiting on me for something?

rmjarvis:
> Just checking: are y'all waiting on me for something?

No.  I'm planning to look through this again, but I haven't had a chance yet.  I was also going to go through and edit the doc strings to match our normal style, but I wasn't going to make you do that.

@jmeyers314 or @rmandelb  Would either of you like to take a look?

rmandelb:
I would like to take a look, but likely won�t have time for ~a week.

rmjarvis:
I updated the doc strings to match our normal style, and I made a few other minor edits.  So I'm mostly happy with this now.  I just have two concerns.

1. The unit tests are pretty slow.  We usually aim to have each test function take < 1s ideally.  This isn't always possible, but a lot of these tests are many seconds, and it seems the culprit is the call to `LsstCamera`.  This takes ~1.3 seconds on my machine, so that's completely dominating the time for these tests.  Any clue why this is so slow?

The slow line is
```
self._camera = LsstSimMapper().camera
```
Is this always the same object?  Can we get away with having a singleton pattern for this and only call `LsstSimMapper()` the first time we need it, and reuse it after that?

2. The other concern I have is when I run `scons tests`, it freezes up.  `nosetests` runs fine (with no arguments) in the tests directory, as does `python test_lsst.py`.  However, a multiprocessing nosetests run (e.g. `nosetests --processes=2`), which is what `scons tests` runs, freezes, presumably on the new lsst tests.  I checked and `scons tests -j1`, which runs in single-process mode, does run through successfully.

Do you know if this is expected?  Is the lsst stack not multi-process-safe?  Should we put in some kind of warning that if you have the LSST stack installed, and thus are running the tests in test_lsst.py, that you need to run `scons tests -j1`?  Or if the race conditions are in `LsstSimMapper`, maybe the singleton idea I had above will fix this too...

danielsf:
```LsstSimMapper().camera``` is always the same, so, yes, we should be able to limit it so that it is only called once for all of the unit tests.

As far as multi-processing, I don't know of any reason why this portion of the stack should not be multi-process-safe.  Maybe it is multiple instances of ```LsstSimMapper()``` all trying to access the same data at once.

I will see what happens when I eliminate the multiple calls to ```LsstSimMapper()```.

**Comment by timj** [comment]
We definitely want to fix problems with freezing. It's not at all expected (and we tend to run tests in parallel ourselves [albeit not with nose])

**Comment by rmjarvis** [comment]
> I will see what happens when I eliminate the multiple calls to LsstSimMapper().

I think the easiest way to implement this would be to have `LsstCamer._camera` be a class variable initialized to None at class scope.  And then in `__init__`, do
```
if self._camera is None:
    self._camera = LsstSimMapper().camera
```
I think this will then only call LsstSimMapper() once in any particular python session that uses this.  Modulo multiprocessing race conditions, that is.

**Comment by danielsf** [comment]
That actually doesn't work.  Class variables are only shared by all instantiations of that class if they are initialized as mutable types (lists or dicts).

I am fixing it by just creating a global variable ```_afwCamera = LsstSimMapper().camera``` and having ```__init__``` set ```self._camera = _afwCamera```.

About to push.

**Comment by danielsf** [comment]
This fixed the ```scons tests``` problem (at least on my laptop)

**Comment by rmjarvis** [comment]
I think I'd rather it be a class variable than a global.  Probably for no good reason other than aesthetics.  But you're right that I didn't quite have the syntax correct.  You need to refer to it as `LsstCamera._camera`, not `self._camera`.  See my latest commit 355f083.

**Comment by rmjarvis** [comment]
BTW, I'm still getting freezing when I run nosetests in parallel, even with the singleton camera construction.  I haven't managed to pare it down to a manageable minimum test case for someone to look into it.  But I can say that if I comment out the lines:
```
#import lsst.meas.astrom as measAstrom
#from lsst.obs.lsstSim import LsstSimMapper
```
in lsst_wcs.py, then `scons tests` runs through to completion (albeit with some expected failures).  Without these lines commented, it reliably freezes when using 4 processes.

**Comment by danielsf** [comment]
@rmjarvis What system are you running?  I was able to get ```scons tests``` to work on my Mac (Yosemite).

```
Ran 384 tests in 197.295s

OK
Nosetests finished successfully.

Starting cpp tests...
Running 12 test cases...

*** No errors detected
test_main finished successfully.

scons: done building targets.
```

It did slow way down towards the end, and I got a lot of messages like

```
Exception AttributeError: AttributeError('_coldefs',) in <bound method FITS_rec.__del__ of FITS_rec([(array([59,  0], dtype=int32),)], 
      dtype=(numpy.record, [('COMPRESSED_DATA', '>i4', (2,))]))> ignored
Exception AttributeError: AttributeError('_coldefs',) in <bound method FITS_rec.__del__ of FITS_rec([(array([14,  0], dtype=int32),), (array([14, 28], dtype=int32),),
       (array([14, 56], dtype=int32),), (array([14, 84], dtype=int32),),
       (array([ 14, 112], dtype=int32),)], 
      dtype=(numpy.record, [('COMPRESSED_DATA', '>i4', (2,))]))> ignored
Exception AttributeError: AttributeError('_coldefs',) in <bound method FITS_rec.__del__ of FITS_rec([(array([14,  0], dtype=int32),), (array([14, 28], dtype=int32),),
       (array([14, 56], dtype=int32),), (array([14, 84], dtype=int32),),
       (array([ 14, 112], dtype=int32),)], 
      dtype=(numpy.record, [('COMPRESSED_DATA', '>i4', (2,))]))> ignored
Exception AttributeError: AttributeError('_coldefs',) in <bound method FITS_rec.__del__ of FITS_rec([(array([14,  0], dtype=int32),), (array([14, 28], dtype=int32),),
       (array([14, 56], dtype=int32),), (array([14, 84], dtype=int32),),
       (array([ 14, 112], dtype=int32),)], 
      dtype=(numpy.record, [('COMPRESSED_DATA', '>i4', (2,))]))> ignored
```

which I don't know how to interpret.

**Comment by rmjarvis** [comment]
> AttributeError('_coldefs',) in <bound method FITS_rec.__del__

The errors are because you apparently have astropy 1.1.2, which [has a bug](https://github.com/astropy/astropy/issues/4690).

> What system are you running?

OSX Yosemite (10.10.5)
XCode version 7.2
LSST stack is w.2016.12+1
Python is miniconda installed by lsst: 3.19.0.lsst4
TMV is installed by lsst: v0.72
FFTW is installed by lsst: 3.3.3
Boost is installed by lsst: 1.55.0.1.lsst2
g++ is Mac's g++ 4.2.1, which is really clang: Apple LLVM version 7.0.2 (clang-700.1.81)

**Comment by rmjarvis** [comment]
> Ran 384 tests in 197.295s
>
> OK
> Nosetests finished successfully.

@danielsf Are you sure you're running in parallel.  I noticed that this is rather a long time for `scons tests`, similar to how long my machine takes to run `scons tests -j1`.  So I wonder if it's not detecting multiple cpus on your system and just running in single process mode.  

The way you can tell is before running the tests, it should output lines that look something like this:
```
Using nosetests from:  /Users/Mike/lsst/DarwinX86/miniconda2/3.19.0.lsst4/bin/nosetests
nosetests is version 1.3.7... running tests in parallel with 4 jobs
If you have trouble with the parallel test run, you may disable it by typing:
    scons tests -j1
```

If it isn't telling you it's doing a parallel run, then you can force it to run with (e.g.) 4 processes by doing `scons tests -j4`.

**Comment by danielsf** [comment]
I definitely got the "If you have trouble with the parallel test run" message

I also had a different terminal open in which I was running ```top```.  While the tests were running, I noticed 4 active instances of python.

**Comment by danielsf** [comment]
I am running a slightly newer version of the lsst stack than you are (though I would be shocked if that matters).

Can you try running with the conda installed stack?

https://confluence.lsstcorp.org/display/SIM/Catalogs+and+MAF#CatalogsandMAF-BinaryInstallation

**Comment by rmjarvis** [comment]
> Can you try running with the conda installed stack?

OK.  That seemed to work.  (Using w.2016.15)  So maybe some bug in the w.2016.12+1 version of the stack that was fixed recently?  Or something different about when it is installed from source vs the binary installation?

Also, it takes a really long time to complete `scons tests` for me too (168 sec), going very slow at the end particularly.  The master branch only takes 87 sec using the same conda python and nosetests versions.  So adding in the lsst tests almost doubles the running time.

Which is weird, since just doing `time python test_lsst.py` only takes about 6 seconds.  But if I get rid of test_lsst.py temporarily, then `scons tests` goes back to only taking ~90 sec.  So I don't understand what's going on here with the timings.  But at least it's working now.

**Comment by rmjarvis** [comment]
@jmeyers314 I guess you probably have the LSST stack installed on your machine, right?  Could you try this out on your system too to make sure things are working for you?

**Comment by jmeyers314** [comment]
>@jmeyers314 I guess you probably have the LSST stack installed on your machine, right? Could you try this out on your system too to make sure things are working for you?

Hmmm.... `scons tests` hangs after about the 30th '.' for me, though `scons tests -j1` works (in 156s).  Running `time python test_lsst.py` runs fine in about 4 seconds.  Deleting test_lsst.py and running `scons tests` takes 43s.  Everything passes.

Probably not relevant, but I should note that the instructions for loading the appropriate lsst packages were slightly different than advertised for me.  I had to use `setup obs_lsstSim` instead of `setup obs_lsstSim -t sims`.

**Comment by danielsf** [comment]
hmmm the difference in the setup commands is expected, though I'm not sure how to deal with it.  If you install the LSST sims stack from source, the command will be

```
setup obs_lsstSim -t sims
```

if you use the conda install, the command will be

```
setup obs_lsstSim
```
or
```
setup obs_lsstSim -t conda
```
This is an EUPS question.  I'm not sure how to succinctly answer it without just explaining how EUPS works.

**Comment by danielsf** [comment]
and if you have a DM distribution of ```lsst_apps``` installed (rather than the simulations stack), it will also be

```
setup obs_lsstSim
```

**Comment by rmjarvis** [comment]
> I'm not sure how to succinctly answer it without just explaining how EUPS works.

Perhaps change the instructions slightly to 
```
If you have installed the stack, run

    source $LSST_HOME/loadLSST.bash
    setup obs_lsstSim

Note: you may need to add `-t sims` or `-t conda` to the last line depending on your setup.
```
That probably covers everything, right?

And I think given that some installations seem to have multi-processing issues, we should add a warning message near the top of test_lsst.py:
```
if have_lsst_stack:
    warnings.warn("Some installations of the LSST stack have trouble with multiprocessing runs of nosetests.\n"
                  "If `scons tests` freezes up, you may need to run `scons tests -j1` instead.")
```

danielsf:
That all sounds good.  One question:

Being perpetually leery of asking people to "just trust me; this works", is it worth including a link to this EUPS tutorial

https://developer.lsst.io/build-ci/eups_tutorial.html

for those who are curious what the ```setup -t sims``` versus ```setup -t conda``` versus ```setup``` is about?

rmjarvis:
> is it worth including a link to this EUPS tutorial

Sure.  More information never hurts.  :)

danielsf:
I've added the changes to the warning  message.

I ended up having to tweak the "how to install the stack" warning a bit more than we discussed.  I remembered that even ```source loadLSST.bash``` assumes that you installed from source.  If you installed from conda, it's ```source eups-setups.sh```.  Oy.

rmjarvis:
All looks fine to me.  Thanks Scott.  Anyone else planning to review this?  Or are we ok to merge now.

rmandelb:
Apologies - I wanted to review this but simply don't have time.  I think you should proceed without me.

rmjarvis:
OK.  Let's merge tomorrow night unless we hear any protestations for more time.  :)

jmeyers314:
I just tried a `scons tests` on the new master branch (which now includes this PR) and was tracking down all the "skipped" results, which showed up b/c I hadn't `setup obs_lsstSim` yet.  In the process, I've learned a few more things about the hanging unit tests I thought were worth sharing:

`scons tests` still hangs, but `nosetests --processes=8 test_lsst.py` works fine.  In fact, `scons tests` works fine if you delete all the non `test_lsst.py` unit tests.  So apparently nose multiprocessing + lsst stack isn't the problem per se, but some other interaction with the rest of the GalSim unit tests.

-Josh

Issue #691, opened 2015-11-16
Make it easier to produce great3-like simulations


rmjarvis:
One of the takeaways from the recent "Great4" planning meeting at Penn last week was that people would like it to be easier to produce simulations along the lines of what was done for Great3, and then to improve on that in various ways for sims that test blending, selection, etc.  

There are four principal improvements to the config structure that I think would go a long way towards this goal:
- [x] Incorporate some of the recent features to GalSim, most notably the `COSMOSCatalog` class, to simplify the config files that were used to generate the Great3 sims.  In particular, I think all the parameter calculations that we wrote to catalogs and then read in as `type : Catalog` entries can be done directly in the yaml file.  This would make it significantly easier for people to see what is actually being done for these and modify the scripts appropriately for their use case.  I propose cleaning these up, adding comments, and then putting them in the examples directory (or better, in a sub-directory examples/great3) as cgc.yaml, csc.yaml, etc. so people can more easily find useful examples for making these types of sims.
- [x] Make truth catalogs.  This is already an open issue #301.  Part of the reason for our decision to make the catalogs and then read the values from there was that the catalogs would serve as truth information.  If we can make the truth catalogs directly, then we wouldn't need this two-step process.  I have an idea about how to do that, but I'll comment further on that issue page.
- [x] Add a command-line option to split the work into multiple jobs.  We put some hacks into the config generation scripts to make several config files per branch to make it easier to run on multiple machines.  If you're starting with vsc.yaml, say, it's not super trivial to make the necessary changes to split it into multiple runs.  There are a number of values that would need to be adjusted correctly.  However, internal to the config processing, it wouldn't be that hard actually.  It can just skip the appropriate output files and increment the numbers appropriately.  So I propose adding an option along the lines of `galsim vsc.yaml -n 10 -j 1` to split the work into 10 jobs, where this would only do the work of job number 1.  This would then be easy to run on multiple machines.
- [x] Finally, I think it would be useful to be able to define templates for various parts of the config file, so it would be easier to do some standard kinds of galaxies or psfs or image types, etc.  I'm not completely settled on what the best API for this would be, but here is a possibility.  The file great3_constant_space_psf.yaml could have the `psf` part of the config for the constant-PSF space branches.  Then rather than duplicate that in csc.yaml, csv.yaml, rsc.yaml, rsv.yaml, msc.yaml, and msv.yaml, all 6 of these files could just do:
```
psf :
    template : great3_constant_space_psf
```
[...]

Issue #301, opened 2012-10-18
generate output truth catalog


rmjarvis:
Make an easy way to generate output catalogs from the (often random) values uses for each object.  Useful both for a truth table, and also for the catalogs that users will use with centroids, etc.

rmjarvis:
Here is my proposal for the structure of a truth catalog output type from the config processing:
```
output : 
    truth :
        file_name : truth.fits
        columns :
            num : obj_num
            g1 : gal.shear.g1
            g2 : gal.shear.g2
            x : image_posx
            y : image_posy
            psf_e1 : 
                type : Eval
                str : 'psf_e * math.cos(psf_beta.rad())`
                fpsf_e : { type : Current, key : psf.ellip.e }
                apsf_beta : { type : Current, key : psf.ellip.beta }
            psf_e2 : 
                type : Eval
                str : 'psf_e * math.sin(psf_beta.rad())`
                fpsf_e : { type : Current, key : psf.ellip.e }
                apsf_beta : { type : Current, key : psf.ellip.beta }
```
The `columns` item is a dict whose keys will be the names of the columns in the output file, and the values are the values in each row for that column.  Typically, the value is taken to be the "Current" value for that item in the config dict.  i.e. it works just like the `type : Current` value type.  So these would look like `gal.shear.g1` and `gal.shear.g2` above.

However, I think it also makes sense to allow some special values that config knows about for each object.  These are the same items that are allowed to be in `Eval` items.  In the above example, these include `obj_num`, `image_posx`, and `image_posy`.

And finally, it's probably worth allowing calculated values here too.  In the above example, we convert from an EBeta specification of the PSF ellipticity to e1,e2.  I'm not sure how useful this really is, so I'd be ok with not including this option.  But I think it would be easy to allow.

Comments welcome.  I'll probably be doing the coding for this issue in the #691 branch, since it is related to things I'll be doing for that.  But discussion about this feature can be kept here.

rmjarvis:
There is one other detail that I thought of.  In Python, a dict doesn't keep any particular order, so the order of the output columns won't necessarily match the order given in the yaml file.  This is probably ok in most cases, especially if the output file is a FITS binary table.  These are usually accessed by column name, so it doesn't matter what order they are in.  

However, some users may desire a particular output order, especially for ASCII output files where it is more common to access columns by number.  In this case, I think we can let the user provide the desired order of the columns in a list item, called 'order' perhaps:
```
output :
    truth :
[...]

Issue #654, opened 2015-03-19
Bug in des_meds


ckrawiec:
Hi All,

I'm attempting to use the des_meds module in galsim.des to write a MEDS file using galsim images. To do this, I tried to create a MultiExposureObject as follows:

```
gal = galsim.Gaussian(flux=1.e5,sigma=2.)
psf = galsim.Gaussian(flux=1., sigma=0.2)
final = galsim.Convolve([gal,psf])
image = galsim.ImageF(48,48,scale=0.2)
final.drawImage(image)
obj = des_meds.MultiExposureObject([image])
```

which gives me this error:

```
in __init__(self, images, weights, badpix, segs, wcs, id)
    135         else:
    136             # Get the wcs from the images.  Probably just the pixel scale.
--> 137             self.wcs = [ im.wcs.jacobian().setOrigin(im.trueCenter()) for im in self.images ]
    138 
    139          # check if weights,segs,jacks are lists

AttributeError: 'JacobianWCS' object has no attribute 'setOrigin'
```

which seems to be a problem outside of my grasp. Thanks for your help.

-Christina (grad student at UPenn)

rmjarvis:
Thanks for reporting this, Christina.

The proximate problem here is that `setOrigin` should be `withOrigin`.  

But this raises the larger issue that we don't actually have any unit tests of the des module.  Well, we do, but it is in the galsim/des directory rather than in our regular test directory, which means that it doesn't get run regularly.  And running it now, it seems to have other problems as well.

So I propose that we move that test to the main test directory and beef it up so it actually constitutes a fairly complete unit test of the code in the des module to keep the code here from getting out of date with any future changes in GalSim syntax or functionality.

rmandelb:
+1 for that, Mike.  Even as a non-user of that module, I'm happy to have the unit tests run as part of the standard set of unit tests so that it reminds us to keep it up to date.

rmjarvis:
@beckermr: I have two examples of making stamps with blends on this branch.
See [blend.yaml](https://github.com/GalSim-developers/GalSim/tree/%23654/examples/des/blend.yaml) and [blendset.yaml](https://github.com/GalSim-developers/GalSim/tree/%23654/examples/des/blendset.yaml).

[...]

Issue #534, opened 2014-04-21
Support python 3?


rmjarvis:
I'm not sure if there is much demand for this yet, but I was thinking about it recently, so I'm adding the issue so people can chime in here if they want us to pursue it.

If we do want to support python 3, I think we should go the route of trying to dual support both 2 and 3.  i.e. Have our code work no matter which version of python the user has.  A good description of how to do this (or at least a good starting point) is [here](http://python3porting.com/noconv.html).

To see how bad it might be, I ran 2to3 on the code, and the main thing it seems that we need to change is our print statements.  Some other things are the integer division, and apparently the import locations have rearranged a bit.  Also, basestr, which we check for at times is now str. But that link I gave above suggests ways to deal with all of these issues.  

There is one thing that seems to be a bit trickier, which is the syntax for metaclasses has changed.  But I'm sure StackOverflow will tell me how to deal with that.  :)

Anyway, I'll probably play around with this and see if I can get everything to work with both python varieties.  But it may not be terribly high priority for a while unless someone comments here that they want us to make it a high priority.

rmjarvis:
FYI, I did this for TreeCorr (with some prompting and help from Geogiana Ogrean) [here](https://github.com/rmjarvis/TreeCorr/pull/35) and it wasn't too painful.  So I'll probably do this here as well for the next release.

rmjarvis:
FYI, I think this is ready to go.  Once '#549' is merged, this is ready for a PR.

rmjarvis:
Done.  PR #761

PR #695, opened 2015-12-24
#691 Updates to config functionality to make it easier to run great3-like sims

rmjarvis:
This is a pretty big pull request.  Not only have I added a bunch of new features to the config apparatus to make it more functional, but I also took the opportunity to completely redesign the back end config layer.  In fact, since the PR was originally way too big for GitHub's diff screen to show all of it, I put a lot of the boring functionality-neutral changes on master already so the diff here mostly shows the interesting new features that are worth code reviewing.  But for full disclosure, here is a [listing of the changes that are already on master](https://github.com/GalSim-developers/GalSim/compare/7bd70046a731588181c3e25da32b401cac4d5a72...e0cd6920b75bd867ce6175cf96d698b3f6459486).

Updates to galsim executable:
- Dropped default verbosity from 2 to 1, since for real simulations, 2 is usually too much output.
- Added ability to easily split the total work into several jobs with `galsim -n njobs -j jobnum`.  This will just do a single job numbered jobnum out of a total of njobs jobs.  This makes it much easier to split the work of a single config file over several machines.
- Added `galsim -p` to perform profiling on the run.  Sometimes that's helpful to see what might be taking most of the running time.  Although, I've found that the output isn't always as enlightening as I would have hoped.

New config features:
- Truth catalogs can be output as specified in an `output.truth` field.  cf. demo9.yaml, demo10.yaml, cgc.yaml
- External modules can be specified in the config file in a top-level 'modules' field, rather than just with `galsim -m module`.  cf. examples/des/draw_psf.yaml
- Fields can be specified using `template` to read either a whole file or just a given field from a file and use that for the field in question.  This can significantly reduce repeated code for similar simulations.  cf. rgc.yaml, cgc_psf.yaml
- If a value field for something is a string and it starts with "$", then the rest of the string is treated as an Eval string.  Also, the character "@" in an Eval string is the equivalent of a Current type specification.  This ends up being very readable in most cases.  Often better than explicitly using the Eval and Current types for everything.  cf. cgc.yaml 
- In addition to values that are generated elsewhere in the config, now gsobjects can also be referenced from an Eval string, either using the @ notation or a regular Current type.  cf. demo10.yaml, output.truth.columns.cosmos_index.xgal
- A RandomDistribution can now specify a LookupTable using x and f lists, rather than requiring either a file with these columns or a lambda function.  cf. cgc.yaml, psf.items.0.fwhm

Notable features of the backend redesign:
- The config file is now read as an OrderedDict.  This shouldn't matter for anything except the truth catalog, which can now have the output columns in the same order as specified in the config file.
- I made pretty much everything customizable.  There is a basic concept of input objects, output files, images, stamps, gsobjects, values, etc, but the exact implementation of each is something the user can change if they want.  They can do this by writing their own "type" for anything they want to work differently and then register it with the appropriate Register function.  e.g. RegisterImageType will let you define a new `image.type` and specify how it works in each of several point in the code (setup, build, adding noise, number of objects).  (Note, this had been partially true previously, but the redesign makes it much easier and extends the modularity to more things.)  I think this enables all of the custom python code extensions envisioned by @TallJimbo for Great4 sims.
- There had been a lot of repeated code, which I've now made into helper functions that are called by multiple functions.  So most of the code is now packaged in smaller, more focused functions, which are easier to understand, I hope.  Also, for people designing their own type, they can probably use some of these functions as well.  (TODO: A task for me here is to write some good documentation on the wiki explaining how one might do this.  E.g. @esheldon already asked if it would be possible to draw from a log-normal distribution, so I explained how he could write his own value generator to do this.  Better if I could just point to good documentation about this.)
- All the logger commands are now guarded to check whether they will produce output.  This looks redundant, but when the loggers are communicating through a proxy, the arguments had to be sent through the multiprocessing pipe to get to the real logger object, who would usually then not do anything with it, which is inefficient.  So now I guard them all with an appropriate if statement to see if they will produce any output, which sped things up significantly in some cases.
- I changed how config handles random number seeds.  It had been the case that the objects in a file would use (seed + 0 .. seed + nobj-1), and anything at the file or image level would use (seed + nobj).  However, this wasn't really a good solution, since you might need the rng to determine what nobj is.  In fact, we do this in demo9.  So the rng seeding was a bit incoherent in this case.  It seemed better to change to using (seed + 0) for all file/image-level needs, and use (seed + 1 ... seed + nobj) for the objects.
- I don't pass the weight, badpix, and psf images around anymore (which are usually None) as return values from BuildImages and similar commands.  Now these are handled using a new "extra_output" concept.  There are now hooks into various places in the code where these "extra" output types can do some additional processing.  The images they make if they are active are stored in a special spot in the config dict, so final processing and output can be done at the top level.  This means that one can register additional extra output types that do other things than these.  See extra*.py for more details.
- I also added the concept of a stamp type.  Currently, there is only one stamp type, which I called Basic.  But there are now hooks into the stamp creation to do things differently.  e.g. @beckermr wants to be able to draw blended objects on stamps, which hadn't been possible, but now we can make a custom stamp type that would do this.  TODO: I've also promised him an example that will showcase this feature so he can see how to extend it to other similar simulations.

Updates to demos:
- Added examples/great3 directory with cgc.yaml, cgc_psf.yaml, rgc.yaml, rgc_psf.yaml.  I was originally planning on doing all 20, including the deep fields, but this PR is already too big, so I think I'll make another issue to flesh out the rest of these.
- Added truth catalog output in demo9, demo10.
- Introduce Current type in demo10.
- Moved explanation of `eval_variables` from demo10 to demo11.
- Set image.nproc : -1 in demo11, which didn't used to work correctly for subtle reasons, that I fixed on this branch.  (Related to the rng syncing for PowerSpectrum.buildGrid.  See below.)
- Updated demo8.py to reflect the simpler usage that is now possible with the refactored config code.
- Updated all the demo*.py files to match the new random number seed convention.
- Removed most of the comment lines from json files.  cf. discussion here: #301 
- A few minor changes to the order of some calculations in the .py files to match up with some refactoring changes I made in the config layer.

Updates to regular GalSim code:
- Added OutputCatalog class, used to implement truth catalogs.
- galsim.fits.writeMulti allows some "images" to be already-built hdus that are included as is.  Used for the truth catalog hdu in demo10.
- image.resize now allows an optional wcs argument
- Made a method equivalent for image[bounds] = rhs, called image.setSubImage(bounds,rhs).  Useful if accessing an image through a proxy object that doesn't allow the regular syntax.
- Added a function PowerSpectrum.nRandCallsForBuildGrid, which returns how many calls to the rng were made by ps.buildGrid.  Used to keep the rng synced when the ps is a proxy object, so the rng used by buildGrid is not the same as the local rng.
- Added rng.discard and rng.raw.  The former allows one to quickly discard a large number of random values from the rng.  Used in conjunction with ps.nRandCallsForBuildGrid() to keep the rng in sync when multiprocessing. For reference, discarding 1 million values takes 3 ms on my laptop.  So it's pretty quick.  rng.raw() isn't really used here, but it was a feature I wanted recently in another context, so I added it too. It returns the raw long int value generated by the underlying random number generator, not converted to a float.
- Added min/max hlr and min/max flux for COSMOSCatalog.  Currently this gives a rough approximation to the S/N cuts made by Great3 in cgc.yaml.  But it really should be improved further to allow for cuts that are closer to what Great3 did.  cf. Issue #693.
- Added a `sersic_prec` option to COSMOSCatalog.makeGalaxy to round off the Sersic index n values to some precision.  Since GalSim caches a lot of information about each Sersic n it uses, using fewer n values runs a lot faster.  Default is 0.05, which I imagine is usually plenty accurate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,4 +52,8 @@ New config features
   2 times the current value of image.pixel_scale.  (#691)
 - Allowed gsobjects to be referenced from Current types. (#691)
 - Added x,f specification for a RandomDistribution. (#691)
-
+- Added a new 'stamp' top level field and moved some of the items that had
+  belonged in 'image' over to 'stamp'.  Notably, 'draw_method', 'offset', and
+  'gsparams', among other less commonly used parameters.  However, for
+  backwards compatibility, they are all still allowed in the image field
+  as well. (#691)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,17 @@
 Changes from v1.3 to v1.4
 =========================
 
+API Changes
+-----------
+
+- Changed the default shift and/or offset for the output.psf field in a config
+  file to not do any shift or offset.  It had been the default to match what
+  was applied to the galaxy (cf. demo5).  However, we thought that was probably
+  not the most intuitive default.  Now, matching the galaxy is still possible,
+  but requires explicit specification of output.psf.shift = "galaxy" or
+  output.psf.offset = "galaxy". (#691)
+
+
 Bug Fixes
 ---------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,13 +12,15 @@ Bug Fixes
   ChromaticObjects and image-setup keyword arguments (#683)
 - Added ability to manipulate the width of the moment-measuring weight function
   for the KSB shear estimation method of the galsim.hsm package. (#686)
-- Fixed bug the (undocumented) function COSMOSCatalog._makeSingleGalaxy,
+- Fixed bug in the (undocumented) function COSMOSCatalog._makeSingleGalaxy,
   where the resulting object did not set the index attribute properly. (#694)
 
 
 New Features
-------------
+<<<<<<< HEAD
 - Added OutputCatalog class (#301, #691)
+- Added methods calculateHLR, calculateMomentRadius, and calculateFWHM to both
+  GSObject and Image. (#308)
 - Changed `galsim.fits.writeMulti` to allow any of the "image"s to be
   already-built hdus, which are included as is.. (#691)
 - Added optional `wcs` argument to `Image.resize()`. (#691)
@@ -57,3 +59,4 @@ New config features
   'gsparams', among other less commonly used parameters.  However, for
   backwards compatibility, they are all still allowed in the image field
   as well. (#691)
+
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ New Features
 - Added methods calculateHLR, calculateMomentRadius, and calculateFWHM to both
   GSObject and Image. (#308)
 - Changed `galsim.fits.writeMulti` to allow any of the "image"s to be
-  already-built hdus, which are included as is.. (#691)
+  already-built hdus, which are included as is. (#691)
 - Added optional `wcs` argument to `Image.resize()`. (#691)
 - Added `BaseDeviate.discard(n)` and `BaseDeviate.raw()`. (#691)
 - Added `sersic_prec` option to COSMOSCatalog.makeGalaxy(). (#691)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,10 @@ Bug Fixes
 
 
 New Features
-<<<<<<< HEAD
-- Added OutputCatalog class (#301, #691)
+------------
+
+- Added OutputCatalog class, which can be used to keep track of and then output
+  truth information.  cf. demos 9 and 10. (#301, #691)
 - Added methods calculateHLR, calculateMomentRadius, and calculateFWHM to both
   GSObject and Image. (#308)
 - Changed `galsim.fits.writeMulti` to allow any of the "image"s to be
--- a/bin/galsim.py
+++ b/bin/galsim.py
@@ -103,7 +103,7 @@ def parse_args():
             'Used in conjunction with -j (--job)')
         parser.add_argument(
             '-j', '--job', type=int, action='store', default=1,
-            help='set the job number for this particular run. Must be in (1 .. njobs). ' +
+            help='set the job number for this particular run. Must be in [1,njobs]. ' +
             'Used in conjunction with -n (--njobs)')
         parser.add_argument(
             '--version', action='store_const', default=False, const=True,
@@ -149,7 +149,7 @@ def parse_args():
             'Used in conjunction with -j (--job)')
         parser.add_option(
             '-j', '--job', type=int, action='store', default=1,
-            help='set the job number for this particular run. Must be in (1 .. njobs). ' +
+            help='set the job number for this particular run. Must be in [1,njobs]. ' +
             'Used in conjunction with -n (--njobs)')
         parser.add_option(
             '-p', '--profile', action='store_const', default=False, const=True,
--- a/bin/galsim.py
+++ b/bin/galsim.py
@@ -278,7 +278,7 @@ def main():
         sortby = 'tottime'
         ps = pstats.Stats(pr, stream=s).sort_stats(sortby).reverse_order()
         ps.print_stats()
-        logger.error(s.getvalue)
+        logger.error(s.getvalue())
  
 
 if __name__ == "__main__":
--- a/examples/demo10.py
+++ b/examples/demo10.py
@@ -250,15 +250,22 @@ def main(argv):
         # Draw the image
         final.drawImage(sub_gal_image)
 
-        # For the PSF image, we also shift the PSF by the same amount.
-        psf = psf.shift(dx,dy)
+        # For the PSF image, we don't match the galaxy shift.  Rather, we use the offset
+        # parameter to drawImage to apply a random offset of up to 0.5 pixels in each direction.
+        # Note the difference in units between shift and offset.  The shift is applied to the
+        # surface brightness profile, so it is in sky coordinates (as all dimension are for
+        # GSObjects), which are arcsec here.  The offset though is applied to the image itself,
+        # so it is in pixels.  Hence, we don't multiply by pixel_scale.
+        psf_dx = ud() - 0.5
+        psf_dy = ud() - 0.5
+        psf_offset = galsim.PositionD(psf_dx, psf_dy)
 
         # Draw the PSF image:
         # We use real space integration over the pixels to avoid some of the 
         # artifacts that can show up with Fourier convolution.
         # The level of the artifacts is quite low, but when drawing with
         # so little noise, they are apparent with ds9's zscale viewing.
-        psf.drawImage(sub_psf_image, method='real_space')
+        psf.drawImage(sub_psf_image, method='real_space', offset=psf_offset)
 
         # Build the noise model: Poisson noise with a given sky level.
         sky_level_pixel = sky_level * pixel_scale**2
--- a/examples/demo10.py
+++ b/examples/demo10.py
@@ -282,7 +282,7 @@ def main(argv):
 
     logger.info('Done making images of postage stamps')
 
-    # In this case, we'll attach the truth catalog as an aditional HDU in the same file as
+    # In this case, we'll attach the truth catalog as an additional HDU in the same file as
     # the image data.
     truth_hdu = truth_catalog.write_fits_hdu()
 
--- a/examples/demo10.py
+++ b/examples/demo10.py
@@ -284,7 +284,7 @@ def main(argv):
                 psf_shape.e1, psf_shape.e2, psf_fwhm,
                 id_list[index], cosmos_index[index], (theta_deg % 360.),
                 alt_g1, alt_g2, dx, dy ]
-        truth_catalog.add_row(row)
+        truth_catalog.addRow(row)
 
         logger.info('Galaxy (%d,%d): position relative to center = %s', ix,iy,str(pos))
 
@@ -292,7 +292,7 @@ def main(argv):
 
     # In this case, we'll attach the truth catalog as an additional HDU in the same file as
     # the image data.
-    truth_hdu = truth_catalog.write_fits_hdu()
+    truth_hdu = truth_catalog.writeFitsHdu()
 
     # Now write the images to disk.
     images = [ gal_image , psf_image, truth_hdu ]
--- a/examples/demo10.py
+++ b/examples/demo10.py
@@ -28,7 +28,8 @@ is 48 x 48 pixels.  Instead of putting the PSF images on a separate image, we pa
 as the second HDU in the file.  For the galaxies, we use a random selection from 5 specific
 RealGalaxy objects, selected to be 5 particularly irregular ones. (These are taken from 
 the same catalog of 100 objects that demo6 used.)  The galaxies are oriented in a ring 
-test (Nakajima & Bernstein 2007) of 20 each.
+test (Nakajima & Bernstein 2007) of 20 each.  And we again output a truth catalog with the
+correct applied shear for each object (among other information).
 
 New features introduced in this demo:
 
--- a/examples/demo10.py
+++ b/examples/demo10.py
@@ -268,7 +268,7 @@ def main(argv):
         # See demo5.py for more info about how this works.
         sub_psf_image.addNoiseSNR(noise, psf_signal_to_noise)
 
-        # And also to the galaxy image using its signal-to-noise..
+        # And also to the galaxy image using its signal-to-noise.
         sub_gal_image.addNoiseSNR(noise, gal_signal_to_noise)
 
         # Add the truth values to the truth catalog
--- a/examples/demo10.yaml
+++ b/examples/demo10.yaml
@@ -72,9 +72,11 @@ psf :
             #     b = bool
             #     s = string
             #     a = Angle
-            #     p = Position
+            #     p = PositionD
             #     g = Shear
+            #     x = GSObject
             # Thus, fr here is a float variable named r.
+            # (Examples of i, p, g, and x codes can be seen below in the output.truth field.)
             fr : { type : Eval , str : '(world_pos.x**2 + world_pos.y**2)**0.5' } 
             # Now we can use this variable (as r, not fr) in our string:
             str : '0.4 * (r/100)**1.5'   # large value at the edge, so visible by eye.
@@ -266,14 +268,17 @@ output :
             cosmos_id : gal.first.id
 
             # Even though we accessed the catalog by id, we can still get the index number
-            # in the catalog via the gal.index attribute.
+            # in the catalog via the RealGalaxy index attribute.
             cosmos_index :
                 type : Eval
-                str : 'gal.index'
-                # There is a special Eval code we haven't mentioned yet.  If you precede a
-                # variable name with x, then that means it's an object, rather than a value.
-                # So in this case, get the index from the attribute of the object, gal.first.
-                xgal : { type : Current, key : 'gal.first' }
+                str : 'orig_gal.index'
+                # The orig_gal variable uses the x code, which is a little different from the
+                # other codes we've seen.  It refers to a GSObject, which needs to already be
+                # built by some other part of the code; the only type that works for these then
+                # is Current.  Here we take the current value stored in gal.first, which is the
+                # RealGalaxy object that was built in that field before the transformations
+                # were applied in the main gal field.
+                xorig_gal : { type : Current, key : 'gal.first' }
 
             # Fields that are automatically generated are also available.
             # List, Ring, and Catalog all have an index parameter, but it is often omitted
--- a/examples/demo10.yaml
+++ b/examples/demo10.yaml
@@ -28,7 +28,8 @@
 # as the second HDU in the file.  For the galaxies, we use a random selection from 5 specific
 # RealGalaxy objects, selected to be 5 particularly irregular ones.  (These are taken from 
 # the same catalog of 100 objects that demo6 used.)  The galaxies are oriented in a ring 
-# test (Nakajima & Bernstein 2007) of 20 each.
+# test (Nakajima & Bernstein 2007) of 20 each.  And we again output a truth catalog with the
+# correct applied shear for each object (among other information).
 #
 # New features introduced in this demo:
 #
--- a/examples/demo10.yaml
+++ b/examples/demo10.yaml
@@ -258,8 +258,8 @@ output :
             cosmos_index :
                 type : Eval
                 str : 'gal.index'
-                # There is a special Eval code we haven't mentioned yet.  If you name precede
-                # a variable name with x, then that means it's an object, rather than a value.
+                # There is a special Eval code we haven't mentioned yet.  If you precede a
+                # variable name with x, then that means it's an object, rather than a value.
                 # So in this case, get the index from the attribute of the object, gal.first.
                 xgal : { type : Current, key : 'gal.first' }
 
--- a/examples/demo10.yaml
+++ b/examples/demo10.yaml
@@ -41,7 +41,7 @@
 # - pos_value : RTheta (r, theta)
 # - image type : Tiled (..., order)
 # - input : power_spectrum (e_power_function, b_power_function)
-# - output.psf : hdu, signal_to_noise, draw_method
+# - output.psf : hdu, signal_to_noise, draw_method, offset
 # - output.truth : hdu
 #
 # - Evaluated values in output.truth.columns
@@ -218,6 +218,18 @@ output :
         # If there is a signal_to_noise value, then it uses the same kind of noise as in the 
         # main image.
         signal_to_noise : 1000
+
+        # Offset the PSF images randomly within the central pixel to better sample the
+        # sub-pixel surface brightness profile.  Note: as with the output.psf.shift parameter
+        # (cf. demo5.yaml), the output field may also take the value 'galaxy', which would mean
+        # to apply the same offset as was applied to the galaxy.  In this case, there is no
+        # such offset, so it wouldn't apply here, but it's an option if you want to match the
+        # offsets of the PSF and galaxy.
+        offset:
+            type : XY
+            x : { type: Random, min: -0.5, max: 0.5 }
+            y : { type: Random, min: -0.5, max: 0.5 }
+
 
     # Make a truth catalog as a binary table in hdu 2
     truth :
--- a/examples/demo11.yaml
+++ b/examples/demo11.yaml
@@ -248,6 +248,14 @@ image :
         ra : 19.3 hours
         dec : -33.1 degrees
 
+    random_seed : 24783923
+
+    nproc : -1
+
+
+# Define some image properties that are specific to the stamp generation phase.
+stamp :
+
     # Normally, the draw routine accounts for a convolution by the pixel.  In this case we 
     # we don't want it to do that, since the psf here is what is sometimes called an 
     # "effective PSF".  It is already the convolution of the real PSF with a square pixel.
@@ -255,9 +263,6 @@ image :
     # draw_method = no_pixel.
     draw_method : no_pixel
 
-    random_seed : 24783923
-
-    nproc : -1
 
 # Define the input files
 input :
--- a/examples/demo11.yaml
+++ b/examples/demo11.yaml
@@ -42,7 +42,7 @@
 # We've also seen (in demo4.yaml) that YAML lets you tag values with an & in one place and
 # use them later with *.  It can be convenient to combine this notation with eval_variables
 # to let you define all the useful variables near the top of the config file and use them
-# either in Eval strings or as YAML tags.  Here, the only one use use both ways is pixel_scale,
+# either in Eval strings or as YAML tags.  Here, the only one we use both ways is pixel_scale,
 # but we show it for all of them to give you the idea.
 eval_variables :
 
--- a/examples/demo8.py
+++ b/examples/demo8.py
@@ -141,14 +141,14 @@ def main(argv):
     # save the catalog (or whatever) in the config dict in the way that further processing
     # functions expect.  However, we don't have any input field, so we don't need it here.
     #
-    # To build the files, the Process function then calls the following to build each file:
+    # The Process function then calls the following function to build each file:
     #
     #     galsim.config.BuildFile(config, file_num)
     #
     # Again, we'll forego that option here, so we can see how to use the config machinery
     # to produce images that we can use from within python.
     #
-    # This functions then call the following function to process the image field and actually
+    # This function then calls the following function to process the image field and actually
     # build each image:
     #
     #     image = galsim.config.BuildImage(config, image_num)
--- a/examples/great3/cgc.yaml
+++ b/examples/great3/cgc.yaml
@@ -65,7 +65,7 @@ psf:
         # The aberrations are taken to have an rms of 0.41 waves
         # Great3 put a lot more weight into the defocus than the others.
         # Unfortunately, this makes a couple PSFs into donuts, which was a problem.
-        # According to Aaron Roodma, for DES the appropriate rms values are closer to
+        # According to Aaron Roodman, for DES the appropriate rms values are closer to
         # [ 0.13, 0.13, 0.14, 0.06, 0.06, 0.05, 0.06, 0.03 ]
         # which has an rms of 0.26 waves, so probably 0.41 was too large a target.
         defocus: { type: RandomGaussian, sigma: 0.36 }
--- a/examples/great3/cgc.yaml
+++ b/examples/great3/cgc.yaml
@@ -1,3 +1,20 @@
+# Copyright (c) 2012-2015 by the GalSim developers team on GitHub
+# https://github.com/GalSim-developers
+#
+# This file is part of GalSim: The modular galaxy image simulation toolkit.
+# https://github.com/GalSim-developers/GalSim
+#
+# GalSim is free software: redistribution and use in source and binary forms,
+# with or without modification, are permitted provided that the following
+# conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this
+#    list of conditions, and the disclaimer given in the accompanying LICENSE
+#    file.
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions, and the disclaimer given in the documentation
+#    and/or other materials provided with the distribution.
+
 # The Great3 control-ground-constant branch configuration file
 # Note: This configuration will not produce images _identical_ to the ones used for Great3, 
 # since the order of random number generation is different.  However, it is intended to
--- a/examples/great3/rgc.yaml
+++ b/examples/great3/rgc.yaml
@@ -1,3 +1,20 @@
+# Copyright (c) 2012-2015 by the GalSim developers team on GitHub
+# https://github.com/GalSim-developers
+#
+# This file is part of GalSim: The modular galaxy image simulation toolkit.
+# https://github.com/GalSim-developers/GalSim
+#
+# GalSim is free software: redistribution and use in source and binary forms,
+# with or without modification, are permitted provided that the following
+# conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this
+#    list of conditions, and the disclaimer given in the accompanying LICENSE
+#    file.
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions, and the disclaimer given in the documentation
+#    and/or other materials provided with the distribution.
+
 # The Great3 real-ground-constant branch configuration file
 # Note: This configuration will not produce images _identical_ to the ones used for Great3, 
 # since the order of random number generation is different.  However, it is intended to
--- a/galsim/catalog.py
+++ b/galsim/catalog.py
@@ -468,7 +468,7 @@ class OutputCatalog(object):
         new_cols = []
         for col, name, t in zip(cols, self.names, self.types):
             name = str(name)  # numpy will barf if the name is a unicode string
-            dt = numpy.dtype(t) # just used to catagorize the type into int, float, str
+            dt = numpy.dtype(t) # just used to categorize the type into int, float, str
             if dt.kind in numpy.typecodes['AllInteger']:
                 dtypes.append( (name, int) )
                 new_cols.append(col)
--- a/galsim/catalog.py
+++ b/galsim/catalog.py
@@ -381,20 +381,14 @@ class OutputCatalog(object):
     """
     # Watch out for this "Gotcha".  Using _rows=[] as the default argument doesn't work!
     # https://pythonconquerstheuniverse.wordpress.com/2012/02/15/mutable-default-arguments/
-    def __init__(self, names, types=None, _rows=None, _sort_keys=None):
+    def __init__(self, names, types=None, _rows=(), _sort_keys=()):
         self.names = names
         if types is None:
             self.types = [ float for i in names ]
         else:
             self.types = types
-        if _rows is None:
-            self.rows = []
-        else:
-            self.rows = _rows
-        if _sort_keys is None:
-            self.sort_keys = []
-        else:
-            self.sort_keys = _sort_keys
+        self.rows = list(_rows)
+        self.sort_keys = list(_sort_keys)
 
     @property
     def nobjects(self): return len(self.rows)
--- a/galsim/catalog.py
+++ b/galsim/catalog.py
@@ -446,7 +446,6 @@ class OutputCatalog(object):
         file_type = file_type.upper()
         if file_type not in ['FITS', 'ASCII']:
             raise ValueError("file_type must be either FITS or ASCII if specified.")
-        self.file_type = file_type
 
         if file_type == 'FITS':
             self.writeFits(file_name)
--- a/galsim/catalog.py
+++ b/galsim/catalog.py
@@ -376,7 +376,7 @@ class OutputCatalog(object):
         ncols       The number of columns in the catalog.
         names       The names of the columns.
         types       The types of the columns.
-        cols        The columns of data that have been accumulated so far.
+        rows        The rows of data that have been accumulated so far.
 
     """
     # Watch out for this "Gotcha".  Using _rows=[] as the default argument doesn't work!
--- a/galsim/catalog.py
+++ b/galsim/catalog.py
@@ -83,9 +83,9 @@ class Catalog(object):
         self.hdu = hdu
 
         if file_type == 'FITS':
-            self.read_fits(hdu, _nobjects_only)
+            self.readFits(hdu, _nobjects_only)
         elif file_type == 'ASCII':
-            self.read_ascii(comments, _nobjects_only)
+            self.readAscii(comments, _nobjects_only)
         else:
             raise ValueError("Invalid file_type %s"%file_type)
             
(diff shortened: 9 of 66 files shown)


rmandelb:
Whoa.  OK, I will try to take a look at this over break. I assume this is the branch you were recommending I work off of when I do #693, but please let me know if I've misunderstood.

rmjarvis:
>  I assume this is the branch you were recommending I work off of when I do #693, but please let me know if I've misunderstood.

Yes.  Ideally, on that branch we can try to make cgc.yaml and rgc.yaml really produce something equivalent to the Great3 branches.  I _think_ the only thing missing is the selection of galaxies from the COSMOS catalog, which so far is only an approximation to what was done for Great3.

rmandelb:
Should I be getting a "WARNING: Overwriting existing file 'output/catalog.fits'. [astropy.io.fits.file]" when running the unit tests?

rmjarvis:
> Should I be getting a "WARNING: Overwriting existing file 'output/catalog.fits'. [astropy.io.fits.file]" when running the unit tests?

I don't get that.  But maybe depends on the astropy version.  I have 1.0.1.

rmjarvis:
cf. https://github.com/astropy/astropy/issues/1963

I think the warning should go away if you upgrade to a newer astropy.

rmandelb:
Done.  I now have 1.0.7, and the warning is gone.  Weirdly enough, I can no longer import astropy in IPython (if it's in IPython then it looks for IPython.version, which my version of IPython doesn't have; I guess go update that too but I'm lazy).  Works fine in python though.

rmjarvis:
I've made some further edits on this branch since my initial PR.  Mostly code simplifications and bug fixes related to things I discovered working on issue "#654" where I'm putting in some more complicated config examples using the DES modules.  

The change that is most worth calling out explicitly is in commit 5cd96da.  I moved a few of the fields that had been in `image` to `stamp`.  For backwards compatibility, it will still look in image too, but `draw_method`, `retry_failures` and `gsparams` are now moved to the `stamp` field in demos 7, 8, and 11.

rmandelb:
Question about this:

> Added ability to easily split the total work into several jobs with galsim -n njobs -j jobnum. This will 
> just do a single job numbered jobnum out of a total of njobs jobs. This makes it much easier to split 
> the work of a single config file over several machines.

Let's say I want to make simulations on a cluster with multiple nodes.  Each node has 16 cores that can be used to split up the postage stamp rendering within a single large image.  So let's say we want to use 10 nodes and make 200 images, dividing these up into 20 images per node (and then do the postage stamp-level sub processing automatically using all cores on the node).  Is this now feasible using this functionality?  I would make n versions of the same yaml file and just change the value of jobnum?

rmjarvis:
> I would make n versions of the same yaml file and just change the value of jobnum?

You only need the one yaml file.  You would change jobnum on the command line.  So the 10 jobs that you would run on the 10 different machines are:
```
galsim -n 10 -j 1 config.yaml
galsim -n 10 -j 2 config.yaml
galsim -n 10 -j 3 config.yaml
...
galsim -n 10 -j 10 config.yaml
```
All using the same yaml file.

rmjarvis:
BTW, I haven't really benchmarked whether it is better to run one job per node, letting it use all the cores with the GalSim multiprocessing, or if it is better to turn off multiprocessing and just run 200 jobs each to build one file.  So far, I've done the latter, letting the cluster queue basically do my multiprocessing for me, but I should really compare the two to see which ends up being faster.

rmandelb:
> You only need the one yaml file. 

OK, I'm glad I misunderstood, that's better.

> BTW, I haven't really benchmarked whether it is better to run one job per node, letting it use all the 
> cores with the GalSim multiprocessing, or if it is better to turn off multiprocessing and just run 200 
> jobs each to build one file.

I might test this out in a few weeks, once we are running some sims for HSC.  If I do I will let you know :)

CHANGELOG.md
```diff
@@ -12,10 +12,51 @@ Bug Fixes
...
 - Added ability to manipulate the width of the moment-measuring weight function
   for the KSB shear estimation method of the galsim.hsm package. (#686)
+- Fixed bug in the (undocumented) function COSMOSCatalog._makeSingleGalaxy,
+  where the resulting object did not set the index attribute properly. (#694)
 
 
 New Features
-------------
-
+<<<<<<< HEAD
```
rmandelb: merge failure?

CHANGELOG.md
```diff
@@ -12,10 +12,51 @@ Bug Fixes
...
+- Added OutputCatalog class (#301, #691)
```
rmandelb: say (briefly) what this is, for those who might find it useful but weren't necessarily looking at those issues?

examples/demo10.py
```diff
@@ -166,14 +171,21 @@ def main(argv):
     # applied.
     galsim.random.permute(rng, ix_list, iy_list)
 
-    # Now have the PowerSpectrum object build a grid of shear values for us to use.
-    grid_g1, grid_g2 = ps.buildGrid(grid_spacing=stamp_size*pixel_scale, ngrid=n_tiles,
-                                    rng=rng)
+    # Initialize the OutputCatalog for the truth values
```
rmandelb: Should the docstring for the demo mention that this demo shows the use of the OutputCatalog functionality?

examples/demo10.yaml
```diff
@@ -216,3 +218,77 @@ output :
...
+            psf_fwhm : psf.fwhm
+
+            cosmos_id : gal.first.id
+
+            # Even though we accessed the catalog by id, we can still get the index number
+            # in the catalog via the gal.index attribute.
+            cosmos_index :
+                type : Eval
+                str : 'gal.index'
+                # There is a special Eval code we haven't mentioned yet.  If you name precede
```
rmandelb: Should this be "If you precede"?

examples/demo11.yaml
```diff
@@ -25,17 +25,25 @@
...
+# Rather than defining the variable multiple times, once for each time you use it,
+# GalSim allows you to put such variables in a special top-level field called eval_variables.
+# Anything defined in eval_variables will be available for all Eval commands, so you don't 
+# have to repeat them for each one.  In this case, pixel_scale and theta are used below in the
+# Eval strings of dudx, dudy, dvdx, and dvdy for the field image.wcs.  
+#
+# We've also seen (in demo4.yaml) that YAML lets you tag values with an & in one place and
+# use them later with *.  It can be convenient to combine this notation with eval_variables
+# to let you define all the useful variables near the top of the config file and use them
+# either in Eval strings or as YAML tags.  Here, the only one use use both ways is pixel_scale,
```
rmandelb: "use use" -> "we use"?

examples/great3/cgc.yaml
```diff
@@ -0,0 +1,251 @@
+# The Great3 control-ground-constant branch configuration file
```
rmandelb: should have license info in all the files in this dir?

galsim/catalog.py
```diff
@@ -343,3 +349,251 @@ def __hash__(self): return hash(repr(self))
...
+    about a simulation.
+
+    Each row corresponds to a different object, and each column stores some item of
+    information about that object (e.g. flux or half_light_radius).
+
+    Initialization
+    --------------
+
+    @param names    A list of names for the output columns.
+    @param types    A list of types for the output columns. [default: None, which assumes all
```
rmandelb: Perhaps the comment below about lack of type-checking should be mentioned here?

galsim/config/extra_psf.py
```diff
@@ -17,62 +17,100 @@
...
             raise AttributeError("Invalid draw_method: %s"%method)
     else:
         method = 'auto'
 
-    # Special: if the galaxy was shifted, then also shift the psf 
-    if 'shift' in config['gal']:
-        gal_shift = galsim.config.GetCurrentValue(config['gal'],'shift')
-        if False:
-            logger.debug('obj %d: psf shift (1): %s',config['obj_num'],str(gal_shift))
+    # Special: if the galaxy was shifted, then also shift the psf
```
rmandelb: Wait... really?  Why?  I think I am missing something.

galsim/config/input_fitsheader.py
```diff
@@ -15,7 +15,16 @@
 #    this list of conditions, and the disclaimer given in the documentation
 #    and/or other materials provided with the distribution.
 #
+
+import os
 import galsim
+import logging
+
+# This file adds input type fits_header and value types NFWHaloShear and NFWHaloMagnification.
```
rmandelb: Copy/paste error?  Should this be FitsHeader?

galsim/config/input_powerspectrum.py
```diff
@@ -15,15 +15,44 @@
 #    this list of conditions, and the disclaimer given in the documentation
 #    and/or other materials provided with the distribution.
 #
+
+import os
 import galsim
+import logging
+
+# This file adds input type nfw_halo and value types NFWHaloShear and NFWHaloMagnification.
```
rmandelb: Copy/paste error?

galsim/config/process.py
```diff
@@ -20,8 +20,20 @@
 import galsim
 import logging
 
-from .input import *
-from .output import *
+
+# Python 2.6 doesn't include OrderdDict natively.  There is a package ordereddict that you
+# can pip install.  But if the user hasn't done that, we'll just read into a regular dict.
+# The only feature that requires the OrderedDict is the truth catalog output.  With a regular
+# dict the columns will appear in arbitrary order.
```
rmandelb: Should galsim emit some kind of warning in the case that this is going to happen (i.e., if we use a regular dict instead of an ordered dict)?

galsim/config/value.py
```diff
@@ -618,51 +708,74 @@ def _GenerateFromSum(config, base, value_type):
...
+    2. The base parameter is the original config dict being processed.
+    3. The value_type parameter is the intended type of the generated value.  It should
+       be one of the values that you specify as valid in valid_types.
+    4. The return value of gen_func should be a tuple consisting of the value and a boolean,
+       safe, which indicates whether the generated value is safe to use again rather than
+       regenerate for subsequent objects.  This will be used upstream to determine if 
+       objects constructed using this value are safe to keep or if they have to be rebuilt.
+
+    The allowed types to include in valid_types are: float, int, bool, str, galsim.Angle,
+    galsim.Shear, galsim.PositionD.  In addition, including None in this list means that
```
rmandelb: no PositionI?

galsim/scene.py
```diff
@@ -170,37 +180,73 @@ def __init__(self, file_name=None, image_dir=None, dir=None, preload=False, nois
                 param_file_name = param_file_name[:k] + '_fits' + param_file_name[k:]
                 self.param_cat = pyfits.getdata(param_file_name)
 
-        # If requested, select galaxies based on existence of a usable fit.
+        # NB. The pyfits FITS_Rec class has a bug where it makes a copy of the full
+        # record array in each record (e.g. in getParametricRecord) and then doesn't 
+        # garbage collect it until the top-level FITS_Record goes out of scope.  
+        # This leads to a memory leak of order 10MB or so each time we make a parametric
+        # galaxy.
```
rmandelb: Wow!

rmandelb:
OK, this is quite an impressive piece of work.  Sadly I feel that I'm not the most qualified version to review the updates to the config backend.  Not sure what to do about that.

It might be worth mentioning the GREAT3-like config scripts in the README where we talk about demos?

TallJimbo:
> Sadly I feel that I'm not the most qualified version to review the updates to the config backend. Not sure what to do about that.

I've actually been meaning to do my own code review on this, too, but I haven't been able to get to it yet.  Is mid-next-week too late?

rmjarvis:
> I've actually been meaning to do my own code review on this, too, but I haven't been able to get to it yet. Is mid-next-week too late?

Certainly not!  I would definitely appreciate a review of this from you.  Thanks, Jim.

And thanks Rachel for your comments.  I'll address them tonight probably.

CHANGELOG.md
rmjarvis: Done. ad235ec65842dc562ded37c6b46d1588da9834f8

examples/demo10.py
rmjarvis: Done. f66ab9146f0ed6c1c913cef15e97af2040115cf1

examples/great3/cgc.yaml
rmjarvis: Done. 1a5ccb6b1e5d1508c4248b731a5162242ed7c4e5

galsim/catalog.py
rmjarvis: Done. b91edbf6035635da4c71514d9969632d40c5f18c

galsim/config/input_fitsheader.py
rmjarvis: Done.  Along with other typos.  447fb0020f1ea6f51b906f39124d87836c2a4c32

galsim/config/process.py
rmjarvis: Not here I think, since it would usually be irrelevant (if the user isn't outputting a truth catalog), but I added one where the column names are processed in extra_truth.py.  7803620a7b1581a143f2ddf01f81f54308a19f88

galsim/config/value.py
rmjarvis: > no PositionI?

No.  We never needed one.  All the position items we build are floats, not integers.  So PositionD, not PositionI.  If a need arises, it wouldn't be hard to add this, but there hasn't been any need yet.

galsim/config/extra_psf.py
rmjarvis: This was existing functionality that I preserved.  It's only relevant for demo5 I think, where we draw the psfs in the output image at the same (shifted) position as the galaxies.  However, I'd be happy to revisit this, since it's not necessarily what the user would probably expect.

Maybe have a special option for output.psf.shift = 'galaxy' or 'match' meaning to match the galaxy's shift, but otherwise allow any normal shift instead.  So demo5.yaml's output field would look like:
```
output :
    ...

    psf :
        file_name : g08_psf.fits
        shift : galaxy
```
Currently, `offset` works the same way.  If there is an `offset` parameter in config['image'] or config['stamp'], it applies here as well.  However, this doesn't actually show up in any of the demos currently.  We could make that work the same way with an `offset : galaxy` option, or allow a different offset value if the user specifies something else.

And if neither `offset` nor `shift` are in the `output.psf` field, then just leave the PSF centered.

That seems like a sensible behavior.  Do you agree?  It's a non-backward compatible API change, but I'm ok with that.

rmandelb: > However, I'd be happy to revisit this, since it's not necessarily what the user would probably expect.

Yes, that's exactly what I was worried about.  I like your proposed solution.

rmjarvis: Done.  655428d

I kept demo5 with the same functionality, so it uses the new psf.shift = galaxy option.  

It turns out this change also affected demo10, which has a psf output and a galaxy shift.  So in that case, I took the opportunity to switch to an explicit offset parameter rather than copying the galaxy's shift, which both seems pedagogically useful and covers the missing functionality test.

bin/galsim.py
```diff
@@ -95,6 +95,17 @@ def parse_args():
...
         parser.add_argument(
+            '-p', '--profile', action='store_const', default=False, const=True,
+            help='output profiling information at the end of the run')
+        parser.add_argument(
+            '-n', '--njobs', type=int, action='store', default=1, 
+            help='set the total number of jobs that this run is a part of. ' + 
+            'Used in conjunction with -j (--job)')
+        parser.add_argument(
+            '-j', '--job', type=int, action='store', default=1,
+            help='set the job number for this particular run. Must be in (1 .. njobs). ' +
```
TallJimbo: This is really trivial, but I'd recommend `[1, njobs]` for the range notation.  The use of regular parenthesis to mean a closed interval (even though it's obvious what you mean with a little thought) threw me off for just a second.

rmjarvis: Done.  6a10783

examples/demo10.py
```diff
@@ -201,10 +217,11 @@ def main(argv):
         # orientations stepped uniformly in angle, making a ring in e1-e2 space.
         # We're drawing each profile at 20 different orientations and then skipping to the
         # next galaxy in the list.  So theta steps by 1/20 * 360 degrees:
-        theta = k/20. * 360. * galsim.degrees
+        theta_deg = (k%20) * 360. / 20
+        theta = theta_deg * galsim.degrees
 
         # The index needs to increment every 20 objects so we use k/20 using integer math.
-        index = k / 20
+        index = k // 20
```
TallJimbo: Since you've switched here (and elsewhere in the file, I assume), maybe do `from __future__ import division` at the top of the file?

examples/demo10.yaml
```diff
@@ -100,12 +103,18 @@ gal :
...
-            # We again use the convenient YAML variable name notation to do so.
-            index : { type : Sequence , repeat : *num_in_ring }
+            index :
+                type : Sequence
+                # You can access the current value of any other item in the config dict using
+                # type = Current.  The syntax for the key is to treat the config levels as
+                # attributes.  So the number of galaxies in the ring defined above is gal.num.
+                # You can use any value this way.  If it hasn't actually been evaluated yet,
+                # it will evaluate it as needed for the current object.
+                repeat : { type : Current, key : gal.num }
```
TallJimbo: Does the old raw-YAML approach work too?  If so, is there an advantage to using one or the other, besides readability?

examples/demo10.yaml
```diff
@@ -216,3 +219,89 @@ output :
...
+
+            # Even though we accessed the catalog by id, we can still get the index number
+            # in the catalog via the gal.index attribute.
+            cosmos_index :
+                type : Eval
+                str : 'gal.index'
+                # There is a special Eval code we haven't mentioned yet.  If you precede a
+                # variable name with x, then that means it's an object, rather than a value.
+                # So in this case, get the index from the attribute of the object, gal.first.
+                xgal : { type : Current, key : 'gal.first' }
```
TallJimbo: I found this part of the example a bit confusing, for two reasons:
 - The string `gal` is overloaded to mean both the original config file element and the alias to its "first" attribute.
 - "an object, rather than a value" didn't fully make sense to me at first, because it may "object" sound like more than just a different allowed type for variables.  Maybe it'd be better to include "x = object" in the original list of allowed variable types in the comment for `psf.ellip.e`, and just refer to this as a variable with `object` type here.

TallJimbo: Ah, I see now that an `object` type variable is in fact very different from the other types.  Another sentence of explanation might be helpful.

rmjarvis: Good point on the overloaded name here.  Maybe calling the variable `first_gal` would be clearer?

And yes, the `x` code is a bit different from the rest, since it refers to a GSObject, which is what I meant by "object" here.  I'll add a bit more to try to make that clearer.

examples/demo10.yaml
rmjarvis: The YAML version works too.  I just wanted to introduce the "Current" type here for pedagogical reasons, not because the other way doesn't work any more.  

In particular, it seemed useful to have a couple Current types early in this demo before the ones in the truth catalog section.  This and pixel_scale later are both constants, but the ones in the truth output are not.  That's the main advantage of Current over the YAML `*` notation.  The former gets the current value of things that may change from one galaxy to another.  The `*` notation only works for constants within the YAML file.

I also moved the introduction of `eval_variables` to demo11, since its use is a bit richer in that demo than what we had here.  It had been a fairly lame example in this demo, imo.  But combining it with the YAML tags is a nice design pattern that I've used in practice to put all the constants you might want to modify all together near the top of the config file.

examples/demo8.py
```diff
@@ -144,28 +141,17 @@ def main(argv):
...
     # functions expect.  However, we don't have any input field, so we don't need it here.
     #
-    # To build the files, the Process function then calls one of the following:
+    # To build the files, the Process function then calls the following to build each file:
     #
-    #     galsim.config.BuildFits(file_name, config)               -- build a regular fits file
-    #     galsim.config.BuildMultiFits(file_name, nimages, config) -- build a multi-extension fits 
-    #                                                                 file
-    #     galsim.config.BuildDataCube(file_name, nimages, config)  -- build a fits data cube
+    #     galsim.config.BuildFile(config, file_num)
```
TallJimbo: Now that there's just single element here, it's probably best to rephrase this sentence.

examples/demo8.py
```diff
@@ -144,28 +141,17 @@ def main(argv):
...
     #
     # Again, we'll forego that option here, so we can see how to use the config machinery
     # to produce images that we can use from within python.
     #
-    # Each of these functions call the following function to process the image field
-    # and actually build the images:
+    # This functions then call the following function to process the image field and actually
+    # build each image:
```
TallJimbo: singular/plural mismatch.

examples/demo10.py
rmjarvis: This was a bit of pre-planning for making GalSim compatible with [Python 3](https://github.com/GalSim-developers/GalSim/issues/534).  The `//` operator has the same meaning in both Python 2 and 3.  But `/` does not.

Ideally, I'd like to never use `from __future__ import division` if possible, since I don't actually like having integer division return a float as it does in Python 3.  That runs against all my instincts as a C/C++ programmer.  `5 / 3` is 1, not 1.66666667.   But writing `//` at least makes me more cognizant of the issue without any ambiguity.

examples/great3/cgc.yaml
```diff
@@ -0,0 +1,268 @@
...
+        obscuration: { type: Random, min: 0.1, max: 0.5 }
+
+        # Despite sort of modeling off of DECam (with 4 struts), they didn't use 
+        # any struts for the ground psfs.
+        nstruts: 0
+
+        # The aberrations are taken to have an rms of 0.41 waves
+        # Great3 put a lot more weight into the defocus than the others.
+        # Unfortunately, this makes a couple PSFs into donuts, which was a problem.
+        # According to Aaron Roodma, for DES the appropriate rms values are closer to
```
TallJimbo: "Roodma[n]"

examples/great3/rgc.yaml
```diff
@@ -0,0 +1,40 @@
...
+# everything from cgc.yaml and change a couple items, we use the template feature of GalSim
+# config.
+
+# This copies over everything from cgc.yaml
+template : cgc.yaml
+
+# Now we can change the things we need to change:
+
+# 1. Use real galaxies rather than parametric
+gal.first.gal_type : real
```
TallJimbo: Just curious: this kind of override notation is something you added on top of YAML, not something a YAML parser understands intrinsically, right?

rmjarvis: Yes.  The YAML parser gets you to `config['gal.first.gal_type'] = 'real'`.  Then I have to parse it to convert it to `config['gal']['first']['gal_type'] = 'real'`.

galsim/catalog.py
```diff
@@ -343,3 +349,255 @@ def __hash__(self): return hash(repr(self))
...
+    def ncols(self): return len(self.names)
+
+    # Again, when we use this through a proxy, we need getters for the attributes.
+    def getNames(self): return self.names
+    def getTypes(self): return self.types
+    def setTypes(self, types): self.types = types
+    def getNObjects(self): return self.nobjects
+    def getNCols(self): return self.ncols
+
+    def add_row(self, row, sort_key=None):
```
TallJimbo: I see a mix of camelcase and underscore+lower method names here.  Is this intentional, to be consistent with similar method names elsewhere?

galsim/catalog.py
```diff
@@ -343,3 +349,255 @@ def __hash__(self): return hash(repr(self))
...
+
+        nobjects    The number of objects so far in the catalog.
+        ncols       The number of columns in the catalog.
+        names       The names of the columns.
+        types       The types of the columns.
+        cols        The columns of data that have been accumulated so far.
+
+    """
+    # Watch out for this "Gotcha".  Using _rows=[] as the default argument doesn't work!
+    # https://pythonconquerstheuniverse.wordpress.com/2012/02/15/mutable-default-arguments/
```
TallJimbo: FWIW, my preferred workaround for this is use `_rows=()` as the argument and default value, then `self.rows = list(_rows)` to set the attribute.  But whether that's better or not depends on how much you trust the user to pass a list rather than some other sequence, and given the underscore in the name, I'm guessing that's not a concern here.

galsim/catalog.py
rmjarvis: We're not actually very consistent about this in GalSim unfortunately.  But mixing within the same class is pretty bad, I admit.  Overall, we probably have more camelCase than underscores, so I'll switch to that.

galsim/catalog.py
```diff
@@ -343,3 +349,255 @@ def __hash__(self): return hash(repr(self))
...
+    Attributes
+    ----------
+
+    After construction, the following attributes are available:
+
+        nobjects    The number of objects so far in the catalog.
+        ncols       The number of columns in the catalog.
+        names       The names of the columns.
+        types       The types of the columns.
+        cols        The columns of data that have been accumulated so far.
```
TallJimbo: Should this be "rows"?  I don't see an attribute assignment or property that provides "cols".

galsim/catalog.py
```diff
@@ -343,3 +349,255 @@ def __hash__(self): return hash(repr(self))
...
+            import os
+            name, ext = os.path.splitext(file_name)
+            if ext.lower().startswith('.fit'):
+                file_type = 'FITS'
+            else:
+                file_type = 'ASCII'
+        file_type = file_type.upper()
+        if file_type not in ['FITS', 'ASCII']:
+            raise ValueError("file_type must be either FITS or ASCII if specified.")
+        self.file_type = file_type
```
TallJimbo: Why set an instance attribute here?  I don't see it used downstream, and the way you've just persisted it doesn't seem like it belongs as an attribute of the in-memory object.

galsim/config/extra.py
```diff
@@ -0,0 +1,377 @@
...
+        for h in range(first,len(hdus)+first):
+            if h not in hdus.keys():
+                raise ValueError("Cannot skip hdus.  Not output found for hdu %d"%h)
+        # Turn hdus into a list (in order)
+        hdulist = [ hdus[k] for k in range(first,len(hdus)+first) ]
+        return hdulist
+    else:
+        return []
+
+valid_extra_outputs = {}
```
TallJimbo: Maybe move this to the top of the file?  I was a bit confused by some logic above that referenced it before I realized it was a module-level variable.

galsim/config/extra.py
```diff
@@ -0,0 +1,377 @@
...
+def RegisterExtraOutput(key, init_func=None, kwargs_func=None, setup_func=None, stamp_func=None,
+                        image_func=None, final_func=None, write_func=None, hdu_func=None):
+    """Register an extra output field for use by the config apparatus.
+
+    @param key              The name of the output field in config['output']
+    @param init_func        A function or class name to use to build the output object. 
+                            [default: None, in which case the output "object" will be a list
+                            of length nimages that can be used to construct what is needed.]
+    @param kwargs_func      A function to get the initialization kwargs. [default: None, which
+                            means initialize with no arguments.]
```
TallJimbo: No call signature for `kwargs_func`?

galsim/config/extra.py
```diff
@@ -0,0 +1,377 @@
...
+    valid_extra_outputs[key] = {
+        'init' : init_func,
+        'kwargs' : kwargs_func,
+        'setup' : setup_func,
+        'stamp' : stamp_func,
+        'image' : image_func,
+        'final' : final_func,
+        'write' : write_func,
+        'hdu' : hdu_func
+    }
```
TallJimbo: While this mechanism with registered free functions is more general, this seems like place where a traditional inheritance pattern would be cleaner (i.e. make all of these callbacks no-op methods on a base class, then derived classes override the ones they want to implement, and just register the whole class).

galsim/config/gsobject.py
```diff
@@ -79,25 +61,49 @@ def BuildGSObject(config, key, base=None, gsparams={}, logger=None):
...
 
-    # Ring is only allowed for top level gal (since it requires special handling in 
-    # multiprocessing, and that's the only place we look for it currently).
-    if type_name == 'Ring' and key != 'gal':
-        raise AttributeError("Ring type only allowed for top level gal")
+    # Check if we can use the current cached object
+    if ('current_val' in param and 
+            (param['current_safe'] or param['current_index']//repeat == index//repeat)):
+        # If logging, explain why we are using the current object.
+        if logger and logger.isEnabledFor(logging.DEBUG):
```
TallJimbo: This seems like the kind of check that it'd be better to put inside some sort of proxy object, so you don't have to keep repeating it in code that uses the logger.  I gather that the Python logging framework's built-in support for filtering log messages isn't enough to avoid the performance slowdowns you mentioned in the PR overview, but is it also impossible (or not useful) to have a proxy object to do just this check?

galsim/config/gsobject_ring.py
```diff
@@ -44,7 +45,7 @@ def _BuildRing(config, base, ignore, gsparams, logger):
         full_rotation = math.pi * galsim.radians
 
     dtheta = full_rotation / num
-    if logger and logger.isEnabledFor(logging.DEBUG):
+    if False:
```
TallJimbo: Not sure how much you care about leaving `if False:` statements in; just wanted to make sure it wasn't accidental.

galsim/config/input.py
```diff
@@ -294,6 +298,50 @@ def GetInputObj(input_type, config, base, param_name):
...
+    @param input_type       The name of the type in config['input']
+    @param init_func        A function or class name to use to build the input object.
+    @param types            A list of value or object types that use this input type.
+                            These items will have their "current" values invalidated when the
+                            input object changes.
+    @param kwargs_func      A function to get the initialization kwargs if the regular
+                            _req, _opt, etc. kind of initialization will not work. The call
+                            signature is:
+                                kwargs, safe = GetKwargs(config, base)
+                            [default: None, which means use the regular initialization]
```
TallJimbo: A pointer to documentation on the "regular" kind of initialization would be helpful here (maybe a signature for `init_func`?)

galsim/config/input.py
```diff
@@ -294,6 +298,50 @@ def GetInputObj(input_type, config, base, param_name):
...
+                            [default: None]
+    """
+    valid_input_types[input_type] = {
+        'init' : init_func,
+        'types' : types,
+        'kwargs' : kwargs_func,
+        'nobj' : has_nobj,
+        'file' : file_scope,
+        'setup' : setup_func
+    }
```
TallJimbo: Again, I wonder if a class hierarchy would make it easier for extenders to understand and implement the interface they need to support.

galsim/config/input_powerspectrum.py
```diff
@@ -55,6 +84,23 @@ def PowerSpectrumInit(ps, config, base):
     ps.buildGrid(grid_spacing=grid_spacing, ngrid=ngrid, center=world_center,
                  rng=base['rng'], interpolant=interpolant)
 
+    # Make sure this process gives consistent results regardless of the number of processes
+    # being used.
+    if not isinstance(ps, galsim.PowerSpectrum):
+        # Then ps is really a proxy, which means the rng was pickled, so we need to
+        # discard the same number of random calls from the one in the config dict.
```
TallJimbo: It seems like there out to be a more direct way to determine whether an object is a proxy.  If it's something you have control over (I haven't followed the logic carefully enough to know if it's something from the built-in `multiprocessing` module), maybe an `is_proxy` class attribute?

rmjarvis: The multiprocessing module makes these proxy objects automatically for us.  I think it uses introspection to create something with all the same methods, but which pickles the args and kwargs to these methods and sends them through the multprocessing.Pipe back to the original object, which then calls the method.  Then the return value is again pickled and sent back through the pipe to the process that called it.

It's possible there is a more elegant way to check if we have a proxy object.  I'm certainly not a python multiprocessing expert, but I couldn't come up with anything more straightforward than this.  I suppose you could try to access an attribute and catch the exception that a proxy would raise, since it doesn't copy over all the attributes, just the public methods.  But that seems even more convoluted than this isinstance check.

galsim/config/input.py
rmjarvis: I'm considering this idea, but I'm not completely sure yet.  It seems like if we go with the hierarchy approach for input and extra, we should do it for all of the different registrations.  That might be a win across the board, but I'm still thinking about the consequences of it.

examples/demo10.yaml
rmjarvis: Fixed.  08d7db7c322c006918b5a0b0088af7aa93c1d3ea

examples/demo8.py
rmjarvis: Done. 4b3fc25ab2e224c270edae4726329c26d07526e2

examples/great3/cgc.yaml
rmjarvis: Done.  b1730c47d7c2e4693516dc10139e6082314bb2e0

galsim/catalog.py
rmjarvis: Done.  b018de82b96419731b5e1c0299ff312d8471a85c

galsim/catalog.py
rmjarvis: I agree.  That is more elegant.  Done.  63a5361b80ee02fdd8c29e47348b6950bcacb304

galsim/catalog.py
rmjarvis: Yes.  Fixed.  fb5c58992fc1a3c7613b49f54ee5cf375dc141e4

galsim/catalog.py
rmjarvis: That was a residue of copying aspects of the Catalog class.  Not necessary or appropriate here.  a2e3bb1a116d32ed0bb3110eff92fdb77a76e811

galsim/config/gsobject.py
rmjarvis: I agree.  I made a LoggerWrapper that wraps the multiprocessing proxy object and does these checks before calling the proxy object's function.  So most places in the code can skip the `isEnabledFor` part of this line.  f62f5f47e1102a2180bfe902b4755c9b86371301

galsim/config/extra.py
rmjarvis: Done. 38582f60c782e0784fc275c906cdb89998da39a0

galsim/config/extra.py
rmjarvis: Done.  5426f30b4074146c6726e82d0b18256edf47253a

galsim/config/gsobject_ring.py
rmjarvis: Thanks.  This (and a few others like it) happened when I was trying to figure out why the logging commands were so slow at times.  I put them all back in now as part of this commit: f62f5f47e1102a2180bfe902b4755c9b86371301

galsim/config/extra.py
rmjarvis: OK.  I switched the extra output stuff to use the inheritance model, and I agree that it is a nicer approach, so I'm convinced.  

I had to be a bit careful about some of the multiprocessing communication, but I think I got it all sorted out.  The key is to give the builder the scratch and data objects to use for writing anything that you want to persist in later stages.

I'll work on adopting this model for the other registrations.  For now if you want to check it out, here is the commit for the extra output types: f058762dab0415073b2e57dd6881b672f10b4567

galsim/config/stamp.py
```diff
@@ -404,61 +387,166 @@ def BuildStamp(config, xsize=0, ysize=0,
...
+    elif 'world_pos' in config['image']:
+        world_pos = galsim.config.ParseValue(
+            config['image'], 'world_pos', config, galsim.PositionD)[0]
+    else:
+        world_pos = None
+
+    return xsize, ysize, image_pos, world_pos
+
+
+def ProfileBasic(config, psf, gsparams, logger):
```
TallJimbo: Maybe add a verb to this method name?  Without one, it sounds like "profile" *is* the verb, and that implies something very different.

galsim/config/image.py
```diff
@@ -242,27 +194,67 @@ def BuildImage(config, logger=None, image_num=0, obj_num=0,
...
+    # It calls out to the appropriate build function for this image type.
+    build_func = valid_image_types[image_type]['build']
+    image = build_func(config, image_num, obj_num, logger)
+
+    # Store the current image in the base-level config for reference
+    config['current_image'] = image
+
+    # Mark that we are no longer doing a single galaxy by deleting image_pos from config top
+    # level, so it cannot be used for things like wcs.pixelArea(image_pos).
+    if 'image_pos' in config: del config['image_pos']
```
TallJimbo: I'm sure isn't important from a performance standpoint, but you can write this "delete if exists" operation more simply and with one less lookup as `config.pop("image_pos", None)`.

galsim/config/input_powerspectrum.py
TallJimbo: Fair enough.  If you're not the one making the objects, I agree that what you have is fine.

galsim/config/image.py
rmjarvis: Done. fed2624

rmjarvis:
Thanks Jim.  I've finished the refactoring to use the Builder classes approach you suggested.  There are a lot of whitespace edits, since a lot of functions got pushed in one indentation level, so it killed the github diff again, since there are too many lines changed now.  It might be easier to just look at the last few commits individually.

galsim/config/extra.py
```diff
@@ -0,0 +1,412 @@
...
+    The builder parameter should be a subclass of galsim.config.ExtraOutputBuilder.
+    See that class for the functions that should be defined and their signatures.
+    Not all functions need to be overridden.  If nothing needs to be done at a particular place
+    in the processing, you can leave the base class function, which doesn't do anything.
+
+    @param key              The name of the output field in config['output']
+    @param builder          A builder object to use for building the extra output object.
+                            It should be an instance of a subclass of ExtraOutputBuilder.
+    """
+    valid_extra_outputs[key] = builder
```
TallJimbo: Unnecessary but neat: make this a decorator.
```
def RegisterExtraOutput(key):
    def decorate(builder):
        valid_extra_outputs[key] = builder

@RegisterExtraOutput("badpix")
class BadPixBuilder(ExtraOutputBuilder):
    ...
```

TallJimbo:
I've now finished going through the latest versions of things, with my only comment the one above on decorators.  Looks great!

galsim/config/extra.py
rmjarvis: I think I'll pass on this recommendation.  No offense.  :)  I'm not super familiar with decorators yet, and I'm worried that this will be intimidating to other users who want to use this registration functionality, who may be even less familiar with decorator syntax.

Also, I think as written, this will assign the _class_ to the `valid_extra_outputs` dict rather than an _instance_ of that class.  For the `ExtraOutputBuilder`s, we could do it that way: have the class be the thing that is registered and make an instance later when we need it.  

But some of the other things I register use the same class with different initialization parameters for different types.  So I decided it made more sense to have everything register instances rather than classes to be consistent.  I think in the decorator syntax, this would look like:
```
def RegisterExtraOutput(key):
    def decorate(builder):
        valid_extra_outputs[key] = builder

class BadPixBuilder(ExtraOutputBuilder):
    ...

@RegisterExtraOutput("badpix")
BadPixBuilder()
```
which doesn't seem like a big improvement over the existing syntax using the function call.

TallJimbo: Fair enough.  I had indeed missed the point about instances vs. classes here, and even if familiarity with decorators wasn't a concern I think that would shoot down this proposal.

rmjarvis:
OK.  Thanks.  I think this is all done then, unless anyone else would like to take a look.  I'll plan to merge this tomorrow night unless I hear otherwise.

Issue #308, opened 2012-10-23
Have a calculateHalfLightRadius available for all GSObjects


rmjarvis:
Currently most of our simple GSObjects have a `getHalfLightRadius`, but not our more complicated ones.  The rationale for this is that it is not even well-defined what the half-light radius means for something that isn't radially symmetric.  And even if it were well-defined, it would require some work to calculate it for many of our objects.

However, this restricts our ability to set the size of an object according to a resolution parameter.  There is currently no easy way to set the size of a RealGalaxy (for example) according to a desired resolution.

I think a reasonable solution is to add a new function `calculateHalfLightRadius` that will check for a `getHalfLightRadius` function and return that if it's available.  But if not, then make an image and calculate the value using some reasonable definition of half-light radius.

Then we could draw objects with a specific resolution by constructing them with an arbitrary size value, calculating the half-light radius, and then dilating the profile appropriately.

Furthermore, half-light radius was a fairly arbitrary choice for how we defined the resolution.  Another reasonable option is Irr = Ixx + Iyy (or its square root).  This would be even easier to implement, since I think we can have a `calculateIrr` function that would just use the existing HSM code when there isn't an analytic option.  I don't know if there are other common definitions of resolution that we might want to enable, but these two would be a good start I think.

PR #697, opened 2015-12-29
#308 Code to calculate HLR, Sigma, FWHM

rmjarvis:
(Take 2, this time to merge into master.)

Adding three new methods of GSObject to calculate the half-light radius, sigma=sqrt(T/2), and the fwhm for profiles that don't have these already as attributes.

We're thinking of using these on branch #693. They aren't necessarily hugely optimized for either speed or accuracy. Tests indicate that at least for Gaussian and Exponential, the defaults are accurate to 1 decimal place, which is probably sufficient for the kinds of uses we are planning for them. There are parameters to allow for greater accuracy if desired.

### DIFF
--- a/galsim/base.py
+++ b/galsim/base.py
@@ -403,6 +403,56 @@ class GSObject(object):
         return hlr
 
 
+    def calculateSigma(self, size=None, scale=None):
+        """Returns the sqrt of the radial second moment of the object, sigma = sqrt(T/2), where
+        T = Irr = Ixx+Iyy.
+
+        If the profile has a sigma attribute (only true for a Gaussian), it will just return that,
+        but in the general case, we draw the profile and estimate T directly.
+
+        The function optionally takes size and scale values to use for the image drawing.
+        The default scale is the nyquist scale, which generally produces results accurate
+        to about 1 decimal place.  Using a smaller scale will be more accurate at the expense
+        of speed.  The default size is None, which means drawImage will choose a size designed
+        to contain around 99.5% of the flux.  Using a larger size will again be more accurate
+        at the expense of speed.
+
+        Note: The results from this calculation should be taken as approximate at best.
+              They should usually be acceptable for things like testing that a galaxy has a
+              reasonable resolution, but they should not be trusted for very fine grain
+              discriminations.  For a more accurate estimate, see galsim.hlm.FindAdaptiveMom.
+
+        @param size         If given, the stamp size to use for the drawn image. [default: None,
+                            which will let drawImage choose the size automatically]
+        @param scale        If given, the pixel scale to use for the drawn image. [default:
+                            self.nyquistScale()]
+
+        @returns an estimate of sigma = sqrt(T/2)
+        """
+        if hasattr(self, 'sigma'): 
+            return self.sigma
+
+        if scale is None:
+            scale = self.nyquistScale()
+
+        # Draw the image.  Note: need a method that integrates over pixels to get flux right.
+        im = self.drawImage(nx=size, ny=size, scale=scale)
+
+        # Use radii at centers of pixels as approximation to the radial integral
+        x,y = np.meshgrid(range(im.array.shape[0]), range(im.array.shape[1]))
+        x = x - (im.trueCenter().x - im.bounds.xmin)
+        y = y - (im.trueCenter().y - im.bounds.ymin)
+        rsq = x*x + y*y
+
+        # Accumulate Irr
+        Irr = np.sum(rsq * im.array) / self.flux
+
+        # This has all been done in pixels.  So normalize according to the pixel scale.
+        sigma = np.sqrt(Irr/2.) * im.scale
+
+        return sigma
+
+
     @property
     def flux(self): return self.getFlux()
     @property
--- a/galsim/base.py
+++ b/galsim/base.py
@@ -338,6 +338,71 @@ class GSObject(object):
         """
         return self.SBProfile.getGSParams()
 
+    def calculateHLR(self, size=None, scale=None):
+        """Returns the half-light radius of the object.
+
+        If the profile has a half_light_radius attribute, it will just return that, but in the
+        general case, we draw the profile and estimate the half-light radius directly.
+
+        The function optionally takes size and scale values to use for the image drawing.
+        The default scale is half the nyquist scale, which generally produces results accurate
+        to about 1 decimal place.  Using a smaller scale will be more accurate at the expense
+        of speed.  The default size is None, which means drawImage will choose a size designed
+        to contain around 99.5% of the flux.  This is overkill for this calculation, so 
+        choosing a smaller size than this may speed up this calculation somewhat.
+
+        Note: The results from this calculation should be taken as approximate at best.
+              They should usually be acceptable for things like testing that a galaxy has a
+              reasonable resolution, but they should not be trusted for very fine grain
+              discriminations.
+
+        @param size         If given, the stamp size to use for the drawn image. [default: None,
+                            which will let drawImage choose the size automatically]
+        @param scale        If given, the pixel scale to use for the drawn image. [default:
+                            0.5 * self.nyquistScale()]
+
+        @returns an estimate of the half-light radius
+        """
+        if hasattr(self, 'half_light_radius'): 
+            return self.half_light_radius
+
+        if scale is None:
+            scale = self.nyquistScale() * 0.5
+
+        # Draw the image.  Note: need a method that integrates over pixels to get flux right.
+        # The offset is to make all the rsq values different to help the precision a bit.
+        offset = galsim.PositionD(0.2, 0.33)
+        im = self.drawImage(nx=size, ny=size, scale=scale, offset=offset)
+
+        # Use radii at centers of pixels as approximation to the radial integral
+        x,y = np.meshgrid(range(im.array.shape[0]), range(im.array.shape[1]))
+        x = x - (im.trueCenter().x - im.bounds.xmin + offset.x)
+        y = y - (im.trueCenter().y - im.bounds.ymin + offset.y)
+        rsq = x*x + y*y
+
+        # Sort by radius
+        indx = np.argsort(rsq.flatten())
+        rsqf = rsq.flatten()[indx]
+        data = im.array.flatten()[indx]
+        cumflux = np.cumsum(data)
+
+        # Find the first value with cumflux > 0.5 * flux
+        k = np.argmax(cumflux > 0.5 * self.flux)
+        flux_k = cumflux[k] / self.flux  # normalize to unit total flux
+
+        # Interpolate (linearly) between this and the previous value.
+        if k == 0:
+            hlrsq = rsqf[0] * (0.5 / flux_k)
+        else:
+            flux_km1 = cumflux[k-1] / self.flux
+            hlrsq = (rsqf[k-1] * (flux_k-0.5) + rsqf[k] * (0.5-flux_km1)) / (flux_k-flux_km1)
+
+        # This has all been done in pixels.  So normalize according to the pixel scale.
+        hlr = np.sqrt(hlrsq) * im.scale
+
+        return hlr
+
+
     @property
     def flux(self): return self.getFlux()
     @property
--- a/galsim/base.py
+++ b/galsim/base.py
@@ -453,6 +453,79 @@ class GSObject(object):
         return sigma
 
 
+    def calculateFWHM(self, size=None, scale=None):
+        """Returns the full-width half-maximum (FWHM) of the object.
+
+        If the profile has a fwhm attribute, it will just return that, but in the general case,
+        we draw the profile and estimate the FWHM directly.
+
+        The function optionally takes size and scale values to use for the image drawing.
+        The default scale is half the nyquist scale, which generally produces results accurate
+        to about 1 decimal place.  Using a smaller scale will be more accurate at the expense
+        of speed.  The default size is None, which means drawImage will choose a size designed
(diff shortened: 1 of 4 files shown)


rmandelb:
As a matter of notation / terminology, I think sigma as defined here is slightly confusing.  I'm used to calling ```sigma=[det(T)]^(1/4)```, rather than having ```sigma``` be defined using the trace of ```T```.  Aside from the issue of terminology, I think it would be good to have the determinant radius be an option for this routine; if you have a galaxy that is a sheared version of a round one, the determinant radius is less sensitive to the shearing than the trace radius.  So there are reasons to want to use it for defining the resolution.  Shouldn't be hard to add, I would think, once you've already added the trace radius.

gbernstein:
Also, to chime in here, it would worry me to have default calculations be good to only 1 decimal place for some types and more for others.  It’s something that could catch a user by surprise.  Can you make the default to be accurate and slow, then allow people to speed it up if they are willing to sacrifice accuracy?

rmjarvis:
> Can you make the default to be accurate and slow, then allow people to speed it up if they are willing to sacrifice accuracy?

I think it's hard to know a priori how accurate these are going to be for an arbitrary profile.  I guess you could see how much the answer changes as you cut scale in half or something.  But even then it's not like the half-light radius and fwhm are even all that well defined for complicated (i.e. non-radially symmetric) profiles.  So I'm not sure how to structure this to reliably give a much more accurate value for arbitrary profiles.

Also, is there actually any demand for the default to be more accurate than a few percent accuracy?

gbernstein:
> On Dec 29, 2015, at 3:46 PM, Mike Jarvis <notifications@github.com> wrote:
> 
> Can you make the default to be accurate and slow, then allow people to speed it up if they are willing to sacrifice accuracy?
> 
> I think it's hard to know a priori how accurate these are going to be for an arbitrary profile. I guess you could see how much the answer changes as you cut scale in half or something. But even then it's not like the half-light radius and fwhm are even all that well defined for complicated (i.e. non-radially symmetric) profiles. So I'm not sure how to structure this to reliably give a much more accurate value for arbitrary profiles.
> 
> Also, is there actually any demand for the default to be more accurate than a few percent accuracy?
> 
I don’t know - it’s hard to predict what other people will use it for, which I guess is the basis of my worry.
The new routines must use some operational definition for these quantities though - is it consistent across all profiles?

rmjarvis:
> The new routines must use some operational definition for these quantities though - is it consistent across all profiles?

The routines all draw the profile with some size and scale and then try to estimate the appropriate value from the drawn image.  So as the scale -> 0 and size -> inf, it gets perfectly accurate. (The size only being relevant for Sigma.)  But I don't know any way to determine in advance what scale and size would give a given desired level of accuracy.  Happy to take suggestions if you know of a better algorithm for any of these.

rmjarvis:
> I'm used to calling sigma=[det(T)]^(1/4), rather than having sigma be defined using the trace of T.

I guess by T here, you don't mean the trace T = Ixx + Iyy (Kaiser's definition), but rather the full second moment matrix (sometimes denoted Q).

I originally named the method `calculateIrr`, which I thought was clearer, but if we plan to use these for resolution calculations, it seemed like maybe they should all return linear measures, rather than an area.  So I switched it.  But I admit that Sigma is a bit weird for anything other than a Gaussian, so I'm open to other naming options for this.

The determinant is also doable of course.  I thought you said in Great3 you used T, which is why that choice.  But we can do both.  Either as another method or with a kwarg to indicate which kind of second moment radius is desired.

rmandelb:
> I guess by T here, you don't mean the trace T = Ixx + Iyy (Kaiser's definition), but rather the full 
> second moment matrix (sometimes denoted Q).

Oops.  Yes.  (And here I was talking about notational confusion!)  I meant the full matrix.

Naming: how about ```calculateMomentRadius``` and having it return both the determinant and trace radius?   I would imagine the expensive part of the calculation is drawing the image, not doing the calculation, so it seems worth doing both.  Or, if I'm wrong, have a kwarg for the type of radius with 'trace' and 'determinant' options.

Yes, in GREAT3 we used the trace, but it does give ellipticity-dependent selection effects.  Thinking beyond the attempt to mimic what GREAT3 did, I would be more likely to use a determinant radius than a trace radius in future.

rmjarvis:
I think the default at least should only return a single value for ease of swapping in different radius measures.  Also, I imagine that normally people will only want one or the other, not both.  But we can have `rtype='both'`mean to return both measures in case people want them both.

arunkannawadi:
I was actually using a routine similar to the `calculateHLR` myself. What I found useful in such a routine was to let the user specify the centroid of the profile optionally. For light profiles that do not have circular symmetry, one might want to choose the image center or occasionally centroid as measured from some weighted moments. 

Another minor modification which is easy to do is to extend the routine to calculate radius like r_90 and so on by having another optional argument `f` with its default value being 0.5.

rmjarvis:
Thanks for the suggestions.  I've enabled all of these feature requests.

galsim/base.py
```diff
@@ -338,6 +338,258 @@ def getGSParams(self):
         """
         return self.SBProfile.getGSParams()
 
+    def calculateHLR(self, size=None, scale=None, centroid=None, flux_frac=0.5):
```
rmandelb: Mention these new features in CHANGELOG? (nothing to do with the specifics of this line of code, sorry)

galsim/base.py
```diff
@@ -338,6 +338,258 @@ def getGSParams(self):
...
+        """Returns the half-light radius of the object.
+
+        If the profile has a half_light_radius attribute, it will just return that, but in the
```
rmandelb: What about cases where the profile has already been drawn?  Could we enable a version of this that takes an already-existing image?  (a method of the image class)

galsim/base.py
```diff
@@ -338,6 +338,258 @@ def getGSParams(self):
...
+
+        @param size         If given, the stamp size to use for the drawn image. [default: None,
+                            which will let drawImage choose the size automatically]
+        @param scale        If given, the pixel scale to use for the drawn image. [default:
+                            0.5 * self.nyquistScale()]
+        @param centroid     The position to use for the centroid. [default: self.centroid()]
+        @param flux_frac    The fraction of light to be enclosed by the returned radius.
+                            [default: 0.5]
+
+        @returns an estimate of the half-light radius
```
rmandelb: Say whether units are those of the ```scale``` parameter or pixels

galsim/base.py
```diff
@@ -338,6 +338,258 @@ def getGSParams(self):
...
+            hlrsq = (rsqf[k-1] * (fk-f) + rsqf[k] * (f-fkm1)) / (fk-fkm1)
+
+        # This has all been done in pixels.  So normalize according to the pixel scale.
+        hlr = np.sqrt(hlrsq) * im.scale
+
+        return hlr
+
+
+    def calculateMomentRadius(self, size=None, scale=None, centroid=None, rtype='det'):
+        """Returns an estimate of the radius based on second moments.
```
rmandelb: "based on unweighted second moments"

galsim/base.py
```diff
@@ -338,6 +338,258 @@ def getGSParams(self):
...
+
+        The second moments are defines as:
```
rmandelb: defined

galsim/base.py
```diff
@@ -338,6 +338,258 @@ def getGSParams(self):
...
+        If I(x,y) is a Gaussian, then T = Tr(Q) = Qxx + Qyy = 2 sigma^2.  Thus, one reasonable
+        choice for a "radius" for an arbitrary profile is sqrt(T/2).  
+
+        In addition, det(Q) = sigma^4.  So another choice for an arbitrary profile is det(Q)^1/4.
+
+        This routine can return either of these measures according to the value of the `rtype`
+        parameter.  `rtype='trace'` will cause it to return sqrt(T/2).  `rtype='det'` will cause
+        it to return det(Q)^1/4.  And `rtype='both'` will return a tuple with both values.
+
+        Note that for the special case of a Gaussian profile, no calculation is necessary, and
```
rmandelb: Technically if you have a Gaussian, the moments-based radius is what you've set it to when you ignore the pixel response.  But if something is not a Gaussian, the image is drawn with a pixel response before determining the radius.  So it seems like we're inconsistent in whether to include the pixel response or not depending on whether the object is a Gaussian vs. anything else.  Have I interpreted this right?  I'm not sure that is very desirable behavior.

galsim/base.py
```diff
@@ -338,6 +338,258 @@ def getGSParams(self):
...
+        The default scale is the nyquist scale, which generally produces results accurate
+        to about 1 decimal place.  Using a smaller scale will be more accurate at the expense
+        of speed.  The default size is None, which means drawImage will choose a size designed
+        to contain around 99.5% of the flux.  Using a larger size will again be more accurate
+        at the expense of speed.
+
+        Note: The results from this calculation should be taken as approximate at best.
+              They should usually be acceptable for things like testing that a galaxy has a
+              reasonable resolution, but they should not be trusted for very fine grain
+              discriminations.  For a more accurate estimate, see galsim.hlm.FindAdaptiveMom.
```
rmandelb: hlm -> hsm

galsim/base.py
rmandelb: Same comment applies to the moment radius routine (though in that case, we have to make sure there's no noise to prevent divergence of the unweighted moments).

galsim/base.py
rmandelb: Same comment applies for other routine too, so the user doesn't have to read the code.

rmjarvis: It's not in units of either the scale parameter or pixels.  It's in whatever physical units the object used for its distance measures when it was constructed.  Typically arcsec.

galsim/base.py
rmjarvis: I don't think there is any inconsistency here.  The pixel response is equivalent to integrating the flux over the pixels.  So the approximation when you don't have a Gaussian is just the normal approximation of converting an integral over dxdy into a sum over pixels.  The unit tests check that a Gaussian gets pretty close to the same answer done either way.

galsim/base.py
rmandelb: Yeah, sorry, I meant same units as scale parameter.

galsim/base.py
rmandelb: If you have a Gaussian with sigma that is >> the pixel scale, they should be consistent.  But if sigma is not that much more than the pixel scale, then inclusion/exclusion of the pixel response does matter.

I do think that, conceptually, something different is happening mathematically.  If you prefer not to change it, then we should at least acknowledge it in the dox.

rmjarvis: Well, mathematically the approximation is that we take int_pixel (I(x,y) i j) dx dy ~= (int_pixel I(x,y) dx dy) * i_pixel_center * j_pixel_center.  It's not that we are using a different profile.  It seems to me that this is already implicit in the description of the method.  Especially that it gets more accurate as the pixel scale decreases.  But I'll add this caveat explicitly.

rmjarvis:
Finished addressing the comments so far.  I made image versions of each (which the GSObject versions call after drawing the image) and tested them for non-square images.  Seems to all be working correctly.

gbernstein:
> On Dec 29, 2015, at 3:56 PM, Mike Jarvis <notifications@github.com> wrote:
> 
> The routines all draw the profile with some size and scale and then try to estimate the appropriate value from the drawn image. So as the scale -> 0 and size -> inf, it gets perfectly accurate. (The size only being relevant for Sigma.) But I don't know any way to determine in advance what scale and size would give a given desired level of accuracy. Happy to take suggestions if you know of a better algorithm for any of these.
> 
> 

OK, it sounds like there is a universal definition for HLR and FWHM at least, which is to use circular apertures and radially averaged quantities even for asymmetric objects, with the only ambiguities being about pixelization and the ultimate extent of the “total” flux integration.  I think that would be pretty clear to a user.  These are always hard to get right on a coarsely pixelized image, so if that is the source of the “1 decimal place” accuracy that you quoted, that’s fine with me.  You could try your chosen pixel scale and then try again at half the scale until there is convergence to some optionally specified accuracy.  Or maybe you already do this… I did not look at the code.

rmjarvis:
Yes, the pixelization is the source of the imprecision in these algorithms.

I don't currently do any iteration.  I did some accuracy tests on Gaussians and Exponentials where I know the exact values.  For FWHM, I found that using the Nyquist scale, which is the default for GalSim's drawImage routine was accurate to.much less than 1%.  And for the moments, it was accurate to a few percent.  So in these two cases, I just let the default scale be Nyquist.

For HLR, on the tests I did, it was only good to a little worse than 10 percent with the Nyquist scale, but using 1/2 that was accurate to a few percent.  So I made that the default.  Going to around 0.03 * Nyquist was good to 4 dp.

I'm inclined to just document that the default pixel scale is only expected to be accurate to a few percent, possibly worse depending on the profile, and tell the user how to improve on this if they desire (by specifying a smaller scale value).  The routines are already not super fast, so I'm not sure I want to have the default be a complicated iterative procedure to achieve a given accuracy.

rmandelb:
> I'm inclined to just document that the default pixel scale is only expected to be accurate to a few 
> percent, possibly worse depending on the profile, and tell the user how to improve on this if they 
> desire (by specifying a smaller scale value).

I think that's okay.

gbernstein:
Sure.

rmjarvis:
Done.  8d8707a

rmandelb:
This looks good to me now, and tests pass on both of my machines.

rmjarvis:
Any other comments?  OK to merge this?

PR #702, opened 2016-01-21
pip should install pyyaml instead of yaml

surhud:


rmjarvis:
Thanks.  Quite right.

rmandelb:
Question: this merged it into releases/v1.3, but don't we also need it to go onto master?

rmjarvis:
Yes.  I cherry picked it onto master.

Issue #694, opened 2015-12-22
COSMOS galaxies built with _makeSingleGalaxy always have index attribute set to zero


beckermr:
COSMOS galaxies made with _makeSingleGalaxy always have their index set to zero. 

I think in order to preserve the static methods, the index will need to be passed out of this function 

https://github.com/GalSim-developers/GalSim/blob/releases/1.3/galsim/scene.py#L502

to the real galaxy maker and added there.

@mjarvis: Is this fixed #691? If not, is there a bug in the truth data feature you are adding? I was using this feature in my sims and getting all zero indexes.

rmjarvis:
OK.  The gal.index attribute isn't documented functionality, so I'm not sure it counts as a but to not have it set correctly in some circumstances.  But it's an easy enough fix to do so.  Commit 6a7fc34 on branch #691.

BTW, there wasn't any bug in the truth values for cgc, since this wasn't where it was getting the index numbers.

beckermr:
Good point and thanks for the fix!

beckermr:
I won't be bothered if you decide not to support this. Def not a bug, more a feature request as you say.

rmjarvis:
Sorry.  This should work: 8bfecb0

I actually did some testing this time before committing. :-S

Issue #602, opened 2014-10-15
Let galsim.fits.read be more forgiving of bad FITS files


rmjarvis:
When reading some DES images, I got an error trying to use `im = galsim.fits.read(file_name)`.

It turns out that some of the header items have bad values: NAN.0, which means that astropy (or pyfits) chokes on it:
```
>>> f[1].data
WARNING: AstropyDeprecationWarning: The key function is deprecated and may be removed in a future version.
        Use the `.keyword` attribute instead. [astropy.utils.misc]
ERROR: VerifyError: Unparsable card (G-MEANX), fix it first with .verify('fix'). [astropy.io.fits.card]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/astro/SL64/packages/astropy/0.3.1/lib/python2.7/site-packages/astropy-0.3.1-py2.7-linux-x86_64.egg/astropy/utils/misc.py", line 279, in __get__
    val = self._fget(obj)
 ... [snip] ...
  File "/opt/astro/SL64/packages/astropy/0.3.1/lib/python2.7/site-packages/astropy-0.3.1-py2.7-linux-x86_64.egg/astropy/io/fits/card.py", line 543, in value
    self._value = self._parse_value()
  File "/opt/astro/SL64/packages/astropy/0.3.1/lib/python2.7/site-packages/astropy-0.3.1-py2.7-linux-x86_64.egg/astropy/io/fits/card.py", line 975, in _parse_value
    ".verify('fix')." % self.key)
astropy.io.fits.verify.VerifyError: Unparsable card (G-MEANX), fix it first with .verify('fix').
```

Fortunately, it turns out that the recommended action of doing `f[1].verify('fix')` did work:
```
>>> f[1].verify('fix')
WARNING: VerifyWarning: Output verification result: [astropy.io.fits.verify]
WARNING: VerifyWarning: Card 80: [astropy.io.fits.verify]
WARNING: VerifyWarning:     Card 'G-MEANX' is not FITS standard (invalid value string: NAN.0 / [arcsec] Guider x-axis mean offset).  Fixed 'G-MEANX' card to meet the FITS standard. [astropy.io.fits.verify]
WARNING: VerifyWarning: Card 81: [astropy.io.fits.verify]
[...]

PR #706, opened 2016-01-30
#654 Updates to DES module

rmjarvis:
This PR is mostly about updates to the DES module, but there are a few things of interest to non-DES folks as well.

* The original issue #654 was based on a bug @ckrawiec found in the MEDS code, where the code had gotten out of sync with other changes we had made in GalSim, which she fixed.  I added a unit test for the bug she found and moved some other tests that had been in the des module into tests/test_des.py, so they are now part of the standard test suite to make sure we notice if things get out of sync again.  I also added more unit tests of the various classes in the des module to get better coverage of that module including tests using @esheldon's meds code to make sure the meds files we write stay in sync with what that code is expecting to read in.

* I added a feature to our fits reader code to catch pyfits.VerifyErrors when reading an hdu.  We had an image from SV data that had some bad header entries (with NAN.0 values), which pyfits didn't like.  However, `hdu.verify('fix')` works to fix them as the VerifyError error message kindly informed me about.  So now GalSim will try to catch these errors and if it catches one, it runs verify('fix') before moving on.  (cf. Issue #602)  Reading this file is now a unit test in test_des.py.

* After I upgraded my astropy version a month or so ago, I started getting warnings that look like
```
/Library/Python/2.7/site-packages/galsim/fits.py:793: UserWarning: No C++ Image template instantiation for pixel type <type 'numpy.uint16'>
/Library/Python/2.7/site-packages/galsim/fits.py:794: UserWarning:    Using numpy.float64 instead.
```
I get these warnings even if I go way back to old versions of GalSim.  I'm not sure why these warnings didn't show up before, but we definitely do have unit tests that try to read in images with unsigned int data.  Some of the new des meds tests do.  But also some hsm tests.  So I now explicitly recast such images into the corresponding signed integer types and use that.  I think that's ok, but we might want to consider if we should add unsigned int types to the list of `Image<T>` templates we instantiate.

* Making the above change broke one of the hsm tests that reads in a FITS file with a uint16 image.  It had been converting the unit16 data to float64, which is what it does when it doesn't know how to handle an input type.  But the hsm code is only instantiated for int32, float32, and float64.  Not int16, which is what this is converted to now.  So I added a `_convertImage` function in hsm.py to convert int16 images into int32 before passing them down into the C++ layer.  @rmandelb you might want to check if you think this is ok, or if it would be better to add the int16 instantiation.  The line in particular that failed is [here](https://github.com/GalSim-developers/GalSim/blob/master/tests/test_hsm.py#L221).  The data arrays in the images here are all uint16.

* I added a number of example configuration files in examples/des that show some more sophisticated kinds of customizations that are now possible using the new refactored config framework from #691.  In particular, blendset.yaml is pretty complex.  It creates sets of nearby (blended) objects and puts each one in the center of a postage stamp as though cutout from a larger observation image (like we do in DES with the MEDS files), so the same noise realization for each stamp in the set, just recentered.  In addition, it produces a second file that has all the same galaxies and noise, but this time without the neighboring flux.  So the only impact of the neighbors is the additional Poisson noise from their flux.  This simulation was requested by @beckermr to make sure the new config code had enough hooks in place to be able to do something this complicated.
@TallJimbo I think you'll be interested in checking out this and the other files in this directory with an eye towards the kinds of simulations we'll need for Great4.  
@beckermr You might also want to take a look, since the syntax has changed a bit from what you saw a few weeks ago for this.  In particular, Jim's suggestion of using builder objects rather than free functions here allows for variables to be saved in `self` attributes, rather than in the config dict, which is a bit cleaner.

* Parametric galaxies created by COSMOSCatalog now set an index attribute with their index into the full COSMOS catalog just like the attribute that gets set for real galaxies.  This is related to @beckermr's request in #694, but extended to the parametric galaxies.  It seemed appropriate to just have all galaxies made by COSMOSCatalog set this attribute.  In the examples/des/meds.yaml simulation, the index is written to the truth file, and it's convenient to have it still work if the galaxies are switched from real to parametric.

* I made a few more minor changes to the config processing code.  Most of the changes I needed for these simulations, I already put into #691, but I had a few more after code review had already started there.  The most significant one wasn't really required here, but I realized that I had some logic wrong.  It had been the case that input types needed to register all the value and gsobject types that would use that input type.  But I realized that the logic here is backwards.  A user might make a new builder that uses an existing input type in a different way, so that's where the registration of this connection should be, so I switched it to have the builder types register the connection to any input types they use rather than the other way around.


rmjarvis:
Is anyone interested in reviewing this?  I'm ok if no one wants to look at the DES module parts.  But I'll just call out a couple bits that people might want to review:

- @rmandelb You should take a look at the edits I made to the HSM module here.  Apparently, we didn't enable those functions for int16, which is probably fine.  I recast int16 images to int32 in this branch.  I think that's reasonable, but you might want to consider if we want to instead instantiate these functions for ImageViewS, rather than do this recast.

- Both @rmandelb and @jmeyers314 You might be interested in checking out the examples I wrote here in the directory examples/des.  Not so much for the DES-specific functionality, but they constitute more complex examples of ways one can write custom modules.  I reference them as "Advanced Simulations" [here](https://github.com/GalSim-developers/GalSim/wiki/Tutorials#des-simulations).

rmandelb:
I'll take a look in the next week.

examples/des/blend.yaml
```diff
@@ -0,0 +1,118 @@
...
+#    file.
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions, and the disclaimer given in the documentation
+#    and/or other materials provided with the distribution.
+
+
+# A simulation of blended pairs of galaxies at various separations and flux ratios.
+# There are two parallel runs, one normal and the other where the neighbor only contributes
+# to the noise, not the flux.  So this simulates a perfect deblender that removes the flux
+# correctly, but there is still the extra noise from the neighbor.
```
rmandelb: I think it would be worth mentioning dependencies here, particularly since some aren't listed explicitly (e.g., I tried to run this but it complained that I don't have ngmix).

examples/des/blendset.yaml
```diff
@@ -0,0 +1,205 @@
...
+    # image files.  On a typically large simulation, this time gets amortized over a lot
+    # of objects, so the difference isn't as huge.  So to keep the running time reasonable,
+    # we use parameteric here.
+    gal_type: parametric
+    #gal_type: real
+
+    # Randomly spin the original profile
+    rotate: { type: Random }
+
+    # The COSMOS galaxy sample is shallower than we want, so make them a bit smaller to simulate
```
rmandelb: Update this now that the fainter sample is the default?

examples/des/blendset.yaml
```diff
@@ -0,0 +1,205 @@
...
+        whiten: True  # Only relevant for gal_type = real, but it doesn't hurt to leave this in.
+
+
+input:
+    cosmos_catalog :
+        # Use the same COSMOS catalog choices we made in meds.yaml
+        template : "meds.yaml:input.cosmos_catalog"
+
+        # Also impose a minimum size, which we didn't there, to avoid very small galaxies.
+        min_hlr : 1.0  # arcsec
```
rmandelb: Really? Won't that chuck out nearly everything?  (I'm looking at fig 12 in the GREAT3 handbook)  Also, that isn't particularly small.

examples/des/meds.yaml
```diff
@@ -0,0 +1,291 @@
...
+#    and/or other materials provided with the distribution.
+
+
+# A simulation based on the Great3 rgc branch, but tailored to DES specifics including
+# outputing to MEDS files.
+#
+# For the psf, we use a combination of Kolmogorov for the atmospheric component, and an
+# OpticalPSF for the optics, using parameters appropriate for DECam and the Blanco telescope.
+#
+# The galaxies are real galaxies from COSMOS, modified to be slightly smaller to simulate a
```
rmandelb: Also need to update this now that the default catalog is the fainter one.

examples/des/meds.yaml
```diff
@@ -0,0 +1,291 @@
...
+    # will be done over files instead.  But if you modify this to only do one file or
+    # set output.nproc=1, then this stamp-based parallelization will turn on.
+    nproc: -1
+
+    # Great3 found that some stamps needed a larger fft size than default.  I don't think it's
+    # necessary here, but it doesn't hurt to put this in.
+    gsparams: {maximum_fft_size: 10240}
+
+input:
+    cosmos_catalog :
```
rmandelb: Update these cuts too?

rmandelb:
The HSM changes seem fine.  I like the new examples - just minor comments!

I did not go over the config changes or DES changes in great detail.

examples/des/blend.yaml
rmjarvis: I instead made the ngmix dependency optional.  If ngmix is there, it uses it, otherwise, it has its own implementation.  961883a60b58cec9a311c35481e634be622d785c

I also fixed a few things that got messed up in the merge from #693.  So I think all the scripts in the des directory should be runnable now.

examples/des/blendset.yaml
rmjarvis: Yes.  Thanks for the reminder.  I updated these to use the 25.2 sample.

examples/des/blendset.yaml
rmjarvis: This was mostly a cosmetic choice so the galaxies were all easy to see in the images.  I lowered it to 0.5, which is still visually fine, and I added a comment pointing out that this choice is not intended to be realistic.

examples/des/meds.yaml
rmjarvis: Yes.  Updated to the 25.2 catalog in this file as well.

rmjarvis:
@rmandelb Thanks.  I addressed all your comments here.  One thing I noticed after switching to the 25.2 sample is that a couple of objects failed the min_flux_frac check because they had negative flux.  Most of the negative flux was there, which means it's less than 0.99 * flux.  So they were rejected with lines like this:
```
Object 2523: Measured flux = -1.022624 < 0.99 * -1.032531.
```

I checked, and the parametric fit for this object has positive flux.  Also, the snr was positive (but low -- 3.74 for this one), so this must be something goofy on the edges of the postage stamp cutouts to make the total flux negative.  Bad sky subtraction maybe?  I would have thought the bad_ps exclusion would have included this test, but apparently not.  Maybe we should add that?

examples/des/blend.yaml
rmandelb: Nice!

examples/des/blendset.yaml
rmandelb: I see.  Thanks.

rmandelb:
@rmjarvis - thanks.  I suspect the ```bad_ps``` exclusion did not have to include this before, because it was originally based on the I<23.5 sample.  For that one, you would need a pretty big sky subtraction or other error to get negative flux.  For the I<25.2 sample, a smaller error could give you a negative flux.

However, before blindly imposing a flux>0 cut, I do want to check what fraction would be cut by this, and what is actually going on in those postage stamps.  Will try to investigate in the next day or two.

rmandelb:
OK.  In the 23.5 and 25.2 samples, 11 and 21 objects (respectively) would be eliminated by a "total flux in original postage stamp > 0" cut.  In nearly all cases, it's due to a sky subtraction or deblending failure.  In a few cases it seems to just be due to noise fluctuations in an already low-S/N detection, but I'm not bothered by chucking out a few good objects.

However - how do we impose this?  the total flux in the postage stamp would ideally be pre-tabulated in the files distributed with the real galaxy catalogs.  Since we didn't do that, the only way I can think of to do this would be to do what I did just now, which is rejection after making the galaxy:
```
gal = cat.makeGalaxy(index=ind_n, gal_type='real')
tot_flux = gal.gal_image.array.sum()
```

So I think our options are:
- remake the files to store the flux in the original postage stamp, and then make ```bad_ps``` include this cut.  Seems like overkill for something that affects such a small fraction of the galaxies, but I could do it.
- Add it as some kind of post-selection?

Or do you have some other ideas?

rmjarvis:
The post-selection already exists in the config layer if you know about it.  And post-selection doesn't really make sense in the COSMOSCatalog class itself based on the way we do things there.

So I guess I'd prefer adding the total postage stamp flux as a column to the catalog, which `bad_ps` could use to make this cut.  But I do hear your point about it being a really minor problem, so if you think it's not worth the effort of remaking catalogs *again*, then I'm also ok with just adding a caveat somewhere in the docs to mention that some postage stamps have negative flux, so if you care, you'll want to impose some check for that.

rmandelb:
No, I think you're right that this is what we should do.  I was just being lazy.  I'm traveling today+tomorrow, but once I'm back I'll:

- add that info to both catalogs and ask Joe to update the tarballs
- make the code check for that info and either (a) issue a warning that we can't do this cut if it's not there or (b) use the info when imposing ```bad_ps``` or ```marginal``` cuts.
- update documentation accordingly.

If you would prefer to keep that separate from this PR, given the rather different topic, I could open an issue and do this on a separate branch.

rmjarvis:
Sounds good.  You can do that here I guess.  We've never been all that strict about one-subject-at-a-time for PRs.  :)

rmandelb:
OK.  Since this is a postage stamp issue rather than a parametric fits issue, I am adding this to the real galaxy catalog itself (not to the parametric fits file).  I have a script doing that now, and once it's done I will tar these up and ask Joe to help putting them on the server.

Then in scene.py I will make the usual updates, i.e., checking whether the catalog has the new ```ps_flux``` attribute and issuing a warning if not, then using the attribute if it is there to eliminate the negative flux postage stamps (depending on the ```exclusion_level```).

rmjarvis:
Sounds good.  Thanks Rachel.

rmandelb:
Done.  https://github.com/GalSim-developers/GalSim/pull/706/commits/843b49f8b10a5ab395ecc1d6c0ce0de97e4d133b .

@joezuntz - I am updating the COSMOS training sample tarballs again (sorry...).  They are currently trickling onto my file-sharing site, and I will send you a direct link to download them once they are ready.  You can replace the current tarballs with these versions.

joezuntz:
@rmandelb no problem - downloading these to the server now.

rmandelb:
Thanks!

On Mar 17, 2016, at 1:30 PM, joezuntz <notifications@github.com> wrote:

> @rmandelb no problem - downloading these to the server now.
> 
> —
> You are receiving this because you were mentioned.
> 


-------------------------------

Rachel Mandelbaum
The Falco-DeBenedetti Career Development Professor in Physics
rmandelb@andrew.cmu.edu
http://www.andrew.cmu.edu/~rmandelb/

rmjarvis:
@rmandelb Don't forget to also update the small version of the real galaxy catalog in the examples/data directory.  I'm getting warnings now running the unit tests.

rmandelb:
Done.

rmjarvis:
Great.  Was that the last thing on this PR then?

Can I propose a plan to merge on Wednesday?  So another quick round of checks of the recent changes or if people wanted to check it out and haven't yet, do so by then...

rmandelb:
That was the last point related to anything we've discussed.  But like I said, I did not go over the DES or config changes that carefully.  Do you want to wait for me to do that?  Or ask someone else to?  Or go ahead without that.

joezuntz:
@rmandelb  can you confirm these MD5s:
    b9da57149f21238e4544125dad27933b  COSMOS_23.5_training_sample.tar.gz
    e05cfe60c037c645d61ac70545cc2a99  COSMOS_25.2_training_sample.tar.gz

rmandelb:
Yes, that's right.  Thank you!

joezuntz:
Okay, files are both up and the text for them updated.

Issue #682, opened 2015-09-18
clearly document that x->col and y->row


esheldon:
In C/C++ and numpy the order of indexing is
```
image[row, col]
```
which by convention is understood as "y, x".  I think this ordering is also followed by galsim Image objects since you can get a direct view as a numpy array.

A number of galsim functions (eg. shift()) take a tuple of arguments listed in the docs as (x,y) but it isn't clear if that is meant to follow the same ordering as images (row,col) or be equivalent to (col, row).  By word of mouth I have heard it is indeed (col,row), so the (row->y)  (col->x) relation holds, it is just the arguments are given in reverse.

Since this ordering is reverse of C/C++/python/galsim.Image conventions I think it is worth a big notice in the docs up front.  Sorry if this is clearly stated and I just missed it.

esheldon:
As a correlary to this, it might be useful to provide keyword arguments that make the meaning explicit, e.g
```
shift(drow = drow, dcol=dcol)
```

rmjarvis:
The (x,y) notation in GalSim matches that of ds9, SExtractor, SAOImage, etc.  This is the standard way of referring to positions on an astronomical image.

We do document in the `Image` class docstring that this is backwards of how numpy accesses elements:
```
    The `array` attribute is a NumPy array of the Image's pixels.  The individual elements in the
    array attribute are accessed as `image.array[y,x]`, matching the standard NumPy convention,
    while the Image class's own accessor uses `(x,y)`.
```
We could probably add a bit more explanation of this difference, but I definitely do not want to use the terms `drow` and `dcol` for this.  That IMO would only serve to add to the confusion.  

The API should insulate the user from caring about how x and y are stored internally.  That's why we currently have the above explanation in our description of the `array` attribute, since that's where the user is accessing the implementation, so they need to know.  At other points in the code, (x,y) should be abstractions that the code will implement however is appropriate.

esheldon:
Right, I understand this.

What I'm asking is for is a big up-front explanation at the very front of the galsim docs saying that everywhere a user sees x listed in the code or documentation, they can safely assume this means "column in the image" and that y means "row in the image" so they can do the translation in their head between the different ordering as appropriate.

Issue #657, opened 2015-04-04
Angular size units


dkirkby:
GalSim parameters like FWHM and half-light-radius that specify transverse sizes are generally treated as dimensionless, with the implicit assumption that the user is free to work in arcsec, microrad, microns, etc, although most (all?) examples use arcsecs.  However, the Airy, Kolmogorov and OpticalPSF profiles are not scale free so there is some potential confusion.  Specifically, the Airy.getFWHM and Kolmogorov.getFWHM return sizes in radians, which is probably not what the user is expecting if they are consistently using arcsecs for their galaxy profile sizes.  This confusion is probably not helped by the following docstring for the Image ctor scale parameter:

        scale       You can optionally specify a pixel scale to use.  This would normally have
                    units arcsec/pixel, but it doesn't have to be arcsec.  If you want to
                    use different units for the physical scale of your galsim objects, then
                    the same unit would be used here.

The simplest solution is probably to just clearly document that the Airy and Kolmogorov profiles work in radians, but that will still trip  someone up who makes reasonable assumptions and doesn't read the docs carefully.

A more radical solution would be to require explicit units on all size and scale parameters so that inconsistent units generate runtime errors, e.g. using astropy.units (which I didn't like at first, but is growing on me).

rmjarvis:
For Airy and OpticalPSF, I believe #618, which has been merged into master, and will be part of the next release 1.3, may address the concern you are raising.  

These classes only return sizes in radians if the value of `lam_over_diam` is given in radians as well.  In other words, they also follow the general pattern that the size units are arbitrary and can be specified by the user.  The new functionality allows the user to separately specify `lam`, `diam`, and `scale_units` which will convert the angular `lam / diam` into whatever angle units you want.  `arcsec` is the default.

cf. https://github.com/GalSim-developers/GalSim/blob/master/galsim/base.py#L1455

We forgot to add the same functionality to Kolmogorov, where it would also make sense.  So if we add the same thing there, do you think that would satisfy the concerns raised on this issue?

dkirkby:
The new functionality works numerically, although lumping the angular unit
conversion into the dimensionless quantity (lambda/D) seems non-intuitive.

dkirkby:
> The point is that lambda / D is not dimensionless. It is an angle.

[...]

PR #707, opened 2016-01-30
#682 Improved documentation about coordinates and units.

rmjarvis:
This PR addresses issues #682 and #657, both related to documentation.

For #682, I added documentation about what our convention for x,y is in terms of the rows and columns in the numpy array. I also added some concrete examples that will hopefully make clear what some of the image methods that shift around the coordinates do.  @esheldon Please see if the documentation update here addresses your concerns.

For #657, I tried to improve the documentation about the units used in Airy, OpticalPSF and Kolmogorov.  In particular with respect to the units of the `lam_over_diam` and `lam_over_r0` parameters.  I also added a calculation of lam / r0, giving those values directly for Kolmogorov, analogous to the lam / diam calculation that Rachel added to Airy and OpticalPSF.  @dkirkby Please see if this adequately addresses the concerns you raised about the units of these parameters.

In addition, I added some more documentation about the fact that GSObjects are always in sky coordinates.  In particular that `GSObject.shift` is necessarily in terms of sky coordinates, since the GSObject doesn't know about the pixel scale yet, but that `drawImage(offset)` is in terms of pixels.  This is a question that has come up a number of times, so it seems like our documentation on this had been insufficient.

### DIFF
--- a/galsim/base.py
+++ b/galsim/base.py
@@ -1911,8 +1911,8 @@ class Kolmogorov(GSObject):
 
     The Kolmogorov profile is normally defined in terms of the ratio lambda / r0, where lambda is
     the wavelength of the light (say in the middle of the bandpass you are using) and r0 is the
-    Fried parameter.  Typical values for the Fried parameter are on the order of 10cm for
-    most observatories and up to 20cm for excellent sites. The values are usually quoted at
+    Fried parameter.  Typical values for the Fried parameter are on the order of 0.1 m for
+    most observatories and up to 0.2 m for excellent sites. The values are usually quoted at
     lambda = 500nm and r0 depends on wavelength as [r0 ~ lambda^(6/5)].
 
     The natural units for this ratio is radians, which is not normally a convenient unit to use for
@@ -1921,13 +1921,13 @@ class Kolmogorov(GSObject):
     should convert this to arcsec as well:
 
         >>> lam = 700  # nm
-        >>> r0 = 15 * (lam/500)**1.2  # cm
-        >>> lam_over_r0 = (lam * 1.e-7) / r0  # radians
+        >>> r0 = 0.15 * (lam/500)**1.2  # meters
+        >>> lam_over_r0 = (lam * 1.e-9) / r0  # radians
         >>> lam_over_r0 *= 206265  # Convert to arcsec
         >>> psf = galsim.Kolmogorov(lam_over_r0)
 
     To make this process a bit simpler, we recommend instead providing the wavelength and Fried
-    parameter separately using the parameters `lam` (in nm) and `r0` (in cm).  GalSim will then
+    parameter separately using the parameters `lam` (in nm) and `r0` (in m).  GalSim will then
     convert this to any of the normal kinds of angular units using the `scale_unit` parameter:
 
         >>> psf = galsim.Kolmogorov(lam=lam, r0=r0, scale_unit=galsim.arcsec)
@@ -1952,9 +1952,9 @@ class Kolmogorov(GSObject):
     @param lam              Lambda (wavelength) in units of nanometers.  Must be supplied with
                             `r0`, and in this case, image scales (`scale`) should be specified in
                             units of `scale_unit`.
-    @param r0               The Fried parameter in units of centimeters.  Must be supplied with
-                            `lam`, and in this case, image scales (`scale`) should be specified in
-                            units of `scale_unit`.
+    @param r0               The Fried parameter in units of meters.  Must be supplied with `lam`,
+                            and in this case, image scales (`scale`) should be specified in units
+                            of `scale_unit`.
     @param flux             The flux (in photons) of the profile. [default: 1]
     @param scale_unit       Units to use for the sky coordinates when calculating lam/r0 if these
                             are supplied separately.  Note that the results of calling methods like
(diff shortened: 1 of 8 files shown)


dkirkby:
This looks good to me, thanks.

galsim/base.py
```diff
@@ -58,13 +58,21 @@ class GSObject(object):
...
+    GSObjects are always defined in sky coordinates.  So all sizes and other linear dimensions
+    should be in terms of some kind of units on the sky, arcsec for instance.  Only later (when
+    they are drawn) is the connection to pixel coordinates established via a pixel scale or WCS.
+    (See the documentation for galsim.BaseWCS for more details about how to specify various kinds
+    of world coordinate systems more complicated than a simple pixel scale.)
+
+    For instance, if you eventually draw onto an image that has a pixel scale of 0.2 arcsec/pixel,
+    then the normal thing to do would be to define your surface brightness profiles in terms of
+    arcsec and then draw with `pixel_scale=0.2`.  However, nowhere in GalSim do we enforce that sky
+    coordinates need to be in arcsec.  If you wanted, you could instead define the sizes of all
```
rmandelb: Don't we have a few places where we enforce units, for example in the use of the COSMOS catalog? (noise pad size, target pixel size, half-light radius cuts).  Also NFW halo, and ChromaticAtmosphere.

galsim/base.py
```diff
@@ -1732,15 +1759,14 @@ class Airy(GSObject):
         >>> lam_over_diam *= 206265  # Convert to arcsec
         >>> airy = galsim.Airy(lam_over_diam)
 
-    Or, use separate keywords for the telescope diameter and wavelength in meters and nanometers,
-    respectively:
-
-        >>> airy = galsim.Airy(lam=lam, diam=diam)
+    To make this process a bit simpler, we recommend instead providing the wavelength and diameter
+    separately using the `lam` and `diam` parameters.  GalSim will then convert this to any of
```
rmandelb: mention these are required to be in units of nm and m, respectively?

rmandelb:
Other than my nitpicky comments above, this looks good to me.

galsim/base.py
rmjarvis: - COSMOSCatalog: Yes, we should at least add some documentation that the objects built by COSMOSCatalog will have their linear dimensions in arcsec.  This seems to be poorly documented at the moment.
- NFW halo: I think the documentation on this is pretty good already that the positions should be given in arcsec.  We have a hook for units in a bunch of places to extend it to allow other units, although we say that only arcsec are implemented so far.  So I think this is ok, but there is room for converting them to use a similar kind of conversion calculation as we do in these classes.
- ChromaticAtmosphere: It would be easy to convert this to take a scale_unit parameter like these classes.  The arcsec is explicitly implemented [here](https://github.com/GalSim-developers/GalSim/blob/master/galsim/chromatic.py#L1113), so it would be very easy to divide by scale_unit instead.  I'll make that change.

rmjarvis: COSMOSCatalog: Done d12e907
ChromaticAtmosphere: Done 0ab1f68

galsim/base.py
rmjarvis: Done: dfa4cd5

galsim/base.py
rmjarvis: And I changed the language here to be less dogmatic about nowhere enforcing arcsec.  21796d1

rmandelb: Thanks - the change in language here was what I had in mind, but the other changes are good!  However, for the ChromaticAtmosphere change, since you added a bit of functionality, perhaps it would be good to use that functionality in a unit test?

galsim/base.py
```diff
@@ -1887,30 +1911,56 @@ class Kolmogorov(GSObject):
...
+                            [One of `lam_over_r0`, `fwhm`, `half_light_radius`, or both `lam` and
+                            `r0` is required.]
     @param half_light_radius  The half-light radius of the profile.  Typically given in arcsec.
-                            [One of `lam_over_r0`, `fwhm`, or `half_light_radius` is required.]
+                            [One of `lam_over_r0`, `fwhm`, `half_light_radius`, or both `lam` and
+                            `r0` is required.]
+    @param lam              Lambda (wavelength) in units of nanometers.  Must be supplied with
+                            `r0`, and in this case, image scales (`scale`) should be specified in
+                            units of `scale_unit`.
+    @param r0               The Fried parameter in units of meters.  Must be supplied with `lam`,
```
jmeyers314: When constructing the realistic atmospheric PSF code, I found it convenient to have a `r0_500` argument, the Fried parameter at 500nm.  This made it conceptually easier to investigate wavelength dependence of the PSF, since I could look through the same atmosphere at different wavelengths just by changing `lam` instead of needing to change both `lam` and `r0`.  I think having both options would be reasonable here (and in the real atmosphere code; need to do that).  I'll stop short of recommending a `lam_over_r0_500` parameter, though.  That seems too complicated.

jmeyers314:
Looks go to me, just the one minor comment above.

galsim/base.py
rmjarvis: Good idea.  Done.  b33f8d6

rmandelb:
Any reason not to merge this one?

rmjarvis:
I guess I was waiting for @esheldon to take a look.  He's been pretty sick for the past month though.  Erin, did you want to see if this addresses your concerns about col,row vs x,y?  I don't mind waiting if you don't have time yet, but want to get to it in a week or so.

esheldon:
What I've read is helpful.  I don't know how to check if the (x,y)->(col,row) convention is documented comprehensively throughout the code, but these examples are good.

rmjarvis:
OK.  Thanks.  If down the line you notice a doc string that still seems confusing to you, let us know.

PR #708, opened 2016-02-05
#666 deInterleaveImage

arunkannawadi:
This PR is to merge the `deInterleaveImage' routine on to the master branch. The routine does the opposite of what `interleaveImages' routine does, i.e., it takes an image and outputs a list of lower resolution images. 

A bug in the `interleaveImages' routine is also fixed.


galsim/utilities.py
```diff
@@ -418,6 +418,119 @@ def _gammafn(x):
...
+            for img in im_list:
+                img.wcs = galsim.JacobianWCS(scale*n1, 0., 0., scale*n2)
+    elif isinstance(wcs,galsim.JacobianWCS): # from say, a previously interleaved Image.
+        dudx, dudy, dvdx, dvdy = wcs.dudx, wcs.dudy, wcs.dvdx, wcs.dvdy
+        if (dudx*n1==dvdy*n2) and (dudy==0.0) and (dvdx==0.0):
+            for img in im_list:
+                img.wcs = galsim.PixelScale(dudx*n1)
+        else:
+            for img in im_list:
+                img.wcs = galsim.JacobianWCS(dudx*n1,dudy*n2,dvdx*n1,dvdy*n2)
```
rmjarvis: All these different wcs checks are a bit overly complicated.  You can distill them all down to the following (removing the whole scale bit):
```
if wcs is not None and wcs.isLocal():
    jac = wcs.jacobian()
    for img in im_list:
        img.wcs = galsim.JacobianWCS(jac.dudx*n1,jac.dudy*n2,jac.dvdx*n1,jac.dvdy*n2)
```
There's not really any reason to try to make sure you get a PixelScale WCS when the pixels are square.  JacobianWCS will do this case perfectly well.  Probably non-local, but still uniform WCS's could also be handled properly.  You'd just have to calculate the right origin to use.  I didn't try to figure out what that would be.

galsim/utilities.py
```diff
@@ -418,6 +418,119 @@ def _gammafn(x):
...
+    wcs = image.wcs
+    if wcs is not None:
+        if wcs.isPixelScale():
+            scale = wcs.scale
+    else:
+        scale = None
+
+    im_list, offsets = [], []
+    for i in xrange(n1):
+      for j in xrange(n2):
```
rmjarvis: Looks like you have some 2-space intents here.  These should be 4 spaces.

rmjarvis:
This seems fine as far as it goes.  I still don't really understand why anyone would think interleaving images would be a good idea for real data.  Better to use the original images and get the exact spacing, rather than approximating them to be perfectly uniformly dithered.  But it seems like the routines here would accurately simulate the dither sequence as advertised.

arunkannawadi:
The interleaving technique is not expected to be performed on real images.
It's simply an easy and legitimate technique by which we can incorporate
detector effects in simulations of individual exposures and then combine
multiple simulated exposures to get Nyquist sampled images. The
interleaving technique might not capture some of the complications that
might arise from combining random dither sequence, but it does capture the
effects of detector non-idealities in the images.

On Sat, Feb 6, 2016 at 1:18 AM, Mike Jarvis <notifications@github.com>
wrote:

> This seems fine as far as it goes. I still don't really understand why
> anyone would think interleaving images would be a good idea for real data.
> Better to use the original images and get the exact spacing, rather than
> approximating them to be perfectly uniformly dithered. But it seems like
> the routines here would accurately simulate the dither sequence as
> advertised.
>
> —
> <https://github.com/GalSim-developers/GalSim/pull/708#issuecomment-180702449>
> .
>



-- 
Regards

Arun Kannawadi Jayaraman
Carnegie Mellon University,
Pittsburgh, PA 15213

galsim/utilities.py
arunkannawadi: Agreed that the WCS assignment was a bit complicated. I replaced the existing code with pixel scale and JacobianWCS with the snippet above. Thanks!

galsim/utilities.py
arunkannawadi: Oops, fixed.

galsim/utilities.py
```diff
@@ -418,6 +418,103 @@ def _gammafn(x):
...
+    Here is an example script using this routine:
+
+    Interleaving four Gaussian images
+    ---------------------------------
+
+        >>> n = 2
+        >>> gal = galsim.Gaussian(sigma=2.8)
+        >>> gal_pix = galsim.Convolve([gal,galsim.Pixel(scale=1.0)])
+        >>> img = gal_pix.drawImage(gal_pix,scale=1.0*n,method='no_pixel')
+        >>> im_list, offsets = galsim.utilities.deInterleaveImage(gal_pix,N=n)
```
rmandelb: I'm confused, shouldn't the ```gal_pix``` in this line be ```img```?

rmandelb: Also I'm a little confused about the philosophy of how this was set up.  If you are setting this up as a  way of drawing a single image that is sampled more finely than the native pixel scale, and then recovering multiple images at the native pixel scale by de-interleaving, then I would think you would have a pixel response that corresponds to a **larger** physical scale than the drawn image.  (i.e., your native pixel scale is x but you draw it into a pixel scale that is x/2, then de-interleave.)   Right now you have a 1 arcsec pixel response, then draw into an image with 2 arcsec pixels, so you're sampling *less* finely; what is the use case for that?

galsim/utilities.py
```diff
@@ -418,6 +418,103 @@ def _gammafn(x):
...
+            dx,dy = -(i+0.5)/n1+0.5,-(j+0.5)/n2+0.5
+            offset = galsim.PositionD(dx,dy)
+            img_arr = image.array[j::n2,i::n1].copy()
+            img = galsim.Image(img_arr)
+            if conserve_flux is True:
+                img *= n1*n2
+            im_list.append(img)
+            offsets.append(offset)
+
+    wcs = image.wcs
```
rmandelb: I see these WCS checks, but I don't see bounds checks.  I don't think that if the original image has some non-trivial bounds, that the new one will respect that.

galsim/utilities.py
```diff
@@ -418,6 +418,103 @@ def _gammafn(x):
...
+    if x_size%n1 or y_size%n2:
+        raise ValueError("The value of 'N' is incompatible with the dimensions of the image to "+
+                         +"be 'deinterleaved'")
+
+    im_list, offsets = [], []
+    for i in xrange(n1):
+        for j in xrange(n2):
+            # The tricky part - going from array indices to Image coordinates (x,y)
+            # DX[i'] = -(i+0.5)/n+0.5 = -i/n + 0.5*(n-1)/n
+            #    i  = -n DX[i'] + 0.5*(n-1)
```
rmandelb: Do the signs work out right here?  It seems to me that for example if n1=n2=2, then for i=(0,1), you get dx=(0.25,-0.25), but the image subscripts in those cases will be 0::2 and 1::2 (which seems like it's taking an offset of -0.25 and 0.25 in that order, not 0.25 followed by -0.25 as implied by the formula).

galsim/utilities.py
arunkannawadi: Oops, yes. It is the image that is to be deInterleaved, not the profile.

On Tue, Feb 16, 2016 at 11:32 AM, Rachel Mandelbaum <
notifications@github.com> wrote:

> In galsim/utilities.py
> <https://github.com/GalSim-developers/GalSim/pull/708#discussion_r53036264>
> :
>
> > +    high resolution image of the same profile, even if the total number of pixels is the same. A
> > +    uniformly offset dither sequence can be extracted from a well-resolved image that is drawn by
> > +    convolving the surface brightness profile explicitly with the native pixel response and setting
> > +    a lower sampling scale (or higher sampling rate) using the `pixel_scale' argument in drawImage()
> > +    routine and setting the `method' parameter to `no_pixel'.
> > +
> > +    Here is an example script using this routine:
> > +
> > +    Interleaving four Gaussian images
> > +    ---------------------------------
> > +
> > +        >>> n = 2
> > +        >>> gal = galsim.Gaussian(sigma=2.8)
> > +        >>> gal_pix = galsim.Convolve([gal,galsim.Pixel(scale=1.0)])
> > +        >>> img = gal_pix.drawImage(gal_pix,scale=1.0*n,method='no_pixel')
> > +        >>> im_list, offsets = galsim.utilities.deInterleaveImage(gal_pix,N=n)
>
> I'm confused, shouldn't the gal_pix in this line be img?
>
> —
> <https://github.com/GalSim-developers/GalSim/pull/708/files#r53036264>.
>



-- 
Regards

Arun Kannawadi Jayaraman
Carnegie Mellon University,
Pittsburgh, PA 15213

galsim/utilities.py
arunkannawadi: As it stands, the routine does not care about the bounds. I will make the
routine respect the bounds and include those in the unit tests as well.

On Tue, Feb 16, 2016 at 11:45 AM, Rachel Mandelbaum <
notifications@github.com> wrote:

> In galsim/utilities.py
> <https://github.com/GalSim-developers/GalSim/pull/708#discussion_r53038449>
> :
>
> > +    im_list, offsets = [], []
> > +    for i in xrange(n1):
> > +        for j in xrange(n2):
> > +            # The tricky part - going from array indices to Image coordinates (x,y)
> > +            # DX[i'] = -(i+0.5)/n+0.5 = -i/n + 0.5*(n-1)/n
> > +            #    i  = -n DX[i'] + 0.5*(n-1)
> > +            dx,dy = -(i+0.5)/n1+0.5,-(j+0.5)/n2+0.5
> > +            offset = galsim.PositionD(dx,dy)
> > +            img_arr = image.array[j::n2,i::n1].copy()
> > +            img = galsim.Image(img_arr)
> > +            if conserve_flux is True:
> > +                img *= n1*n2
> > +            im_list.append(img)
> > +            offsets.append(offset)
> > +
> > +    wcs = image.wcs
>
> I see these WCS checks, but I don't see bounds checks. I don't think that
> if the original image has some non-trivial bounds, that the new one will
> respect that.
>
> —
> <https://github.com/GalSim-developers/GalSim/pull/708/files#r53038449>.
>



-- 
Regards

Arun Kannawadi Jayaraman
Carnegie Mellon University,
Pittsburgh, PA 15213

galsim/utilities.py
arunkannawadi: I had written the docstring long back and did not check it carefully now.
You are correct, the image should be drawn with `scale=1.0/n` and not 1.0*n.

On Tue, Feb 16, 2016 at 11:36 AM, Rachel Mandelbaum <
notifications@github.com> wrote:

> In galsim/utilities.py
> <https://github.com/GalSim-developers/GalSim/pull/708#discussion_r53036930>
> :
>
> > +    high resolution image of the same profile, even if the total number of pixels is the same. A
> > +    uniformly offset dither sequence can be extracted from a well-resolved image that is drawn by
> > +    convolving the surface brightness profile explicitly with the native pixel response and setting
> > +    a lower sampling scale (or higher sampling rate) using the `pixel_scale' argument in drawImage()
> > +    routine and setting the `method' parameter to `no_pixel'.
> > +
> > +    Here is an example script using this routine:
> > +
> > +    Interleaving four Gaussian images
> > +    ---------------------------------
> > +
> > +        >>> n = 2
> > +        >>> gal = galsim.Gaussian(sigma=2.8)
> > +        >>> gal_pix = galsim.Convolve([gal,galsim.Pixel(scale=1.0)])
> > +        >>> img = gal_pix.drawImage(gal_pix,scale=1.0*n,method='no_pixel')
> > +        >>> im_list, offsets = galsim.utilities.deInterleaveImage(gal_pix,N=n)
>
> Also I'm a little confused about the philosophy of how this was set up. If
> you are setting this up as a way of drawing a single image that is sampled
> more finely than the native pixel scale, and then recovering multiple
> images at the native pixel scale by de-interleaving, then I would think you
> would have a pixel response that corresponds to a *larger* physical scale
> than the drawn image. (i.e., your native pixel scale is x but you draw it
> into a pixel scale that is x/2, then de-interleave.) Right now you have a 1
> arcsec pixel response, then draw into an image with 2 arcsec pixels, so
> you're sampling *less* finely; what is the use case for that?
>
> —
> <https://github.com/GalSim-developers/GalSim/pull/708/files#r53036930>.
>



-- 
Regards

Arun Kannawadi Jayaraman
Carnegie Mellon University,
Pittsburgh, PA 15213

galsim/utilities.py
arunkannawadi: If I understand your concern correctly, the difference in the sign of the ```offset``` happens for the following reason. While calling ```drawImage```, when you provide an offset value of 0.25, it offsets the centroid of the profile by 0.25 pixels, which corresponds to moving the image boundary by -0.25, which is what you are seeing here. One can pass these offsets while calling ```drawImage``` and pass the resulting images to ```interleaveImage``` and get the original image back. I know this can get confusing, which is why I have some directional cases in my unit tests.

rmandelb: OK.

galsim/utilities.py
arunkannawadi: I have handled this issue not by matching the bounds but by taking care of the origin of the images. Since the image to be deinterleaved is bigger than the set of images the routine produces, the bounds will not match obviously. However, if you interleave them again after adding the detector effects, the interleaved image will have the same bounds as the original image.

galsim/utilities.py
arunkannawadi: How important is it to look if the WCS is local? I ask this because I had a test case where I give only a pixel scale and then change the origin of the image to have some non-trivial bounds. The WCS of the image becomes an instance of the `OffsetWCS` class which is non-local.

rmjarvis: Well, if it's not local, then I think the calculation you do isn't correct to get the final image to have the right wcs.  However, you might not care, in which case you could just check for `wcs.isUniform()` rather than `wcs.isLocal()`.  

Maybe issue a warning in this case that the output image has a local WCS, so it lost the offset information from the originals.

rmjarvis:
Is there anything else to do on this branch?  Or is this ready to merge?

arunkannawadi:
This branch is ready to be merged. Can someone check if the tests pass on Mac OS?

galsim/utilities.py
```diff
@@ -571,20 +679,31 @@ def interleaveImages(im_list, N, offsets, add_flux=True, suppress_warnings=False
...
-    if scale is not None:
-        if n1==n2:
-            img.wcs = galsim.PixelScale(1.*scale/n1)
-        else:
-            img.wcs = galsim.JacobianWCS(1.*scale/n1, 0., 0., 1.*scale/n2)
-    elif isinstance(wcs,galsim.JacobianWCS): # from say, a previously interleaved Image.
-        dudx, dudy, dvdx, dvdy = wcs.dudx, wcs.dudy, wcs.dvdx, wcs.dvdy
-        if (1.0*dudx/n1==1.0*dvdy/n2) and (dudy==0.0) and (dvdx==0.0):
-            img.wcs = galsim.PixelScale(1.*dudx/n1)
+    if wcs is not None: # and wcs.isLocal():
```
rmjarvis: You at least need to check `if wcs.isUniform()`, since if it is not, the call to `wcs.jacobian()` without any position argument will raise an exception.  Probably better to fall through to your warning about being unable to assign a WCS automatically instead.

Likewise in `deinterleaveImage` above.

galsim/utilities.py
```diff
@@ -525,8 +632,9 @@ def interleaveImages(im_list, N, offsets, add_flux=True, suppress_warnings=False
     y_size, x_size = im_list[0].array.shape
     wcs = im_list[0].wcs
 
-    if wcs.isPixelScale():
-        scale = wcs.scale
+    if wcs is not None:
+        if wcs.isPixelScale():
+            scale = wcs.scale
```
rmjarvis: I guess `scale` isn't used anymore.  Could remove these lines.

rmjarvis:
Tests pass on my Mac.  Just a couple of minor comments.

arunkannawadi:
Thanks Mike.
If there are no more comments/concerns, I'll merge this branch around noon
tomorrow.

On Thu, Mar 17, 2016 at 6:43 PM, Mike Jarvis <notifications@github.com>
wrote:

> Tests pass on my Mac. Just a couple of minor comments.
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/GalSim-developers/GalSim/pull/708#issuecomment-198112291>
>



-- 
Regards

Arun Kannawadi Jayaraman
Carnegie Mellon University,
Pittsburgh, PA 15213

Issue #705, opened 2016-01-26
checksum failure for galsim cosmos download?


beckermr:
I keep getting this error

```
beckermr@ki-ls07 $ galsim_download_cosmos -v 3 -f 
Downloading from url:
  http://great3.jb.man.ac.uk/leaderboard/data/public/COSMOS_23.5_training_sample.tar.gz
Target location is:
  /u/ki/beckermr/.local/share/galsim/COSMOS_23.5_training_sample.tar.gz

Meta information about url:
Date: Tue, 26 Jan 2016 21:13:42 GMT
Server: Apache/2.2.15 (Scientific Linux)
Last-Modified: Thu, 14 Jan 2016 10:47:44 GMT
ETag: "d5d5-10754aeff-5294904470e00"
Content-Length: 4417957631
Connection: close
Content-Type: text/plain; charset=UTF-8


Size of COSMOS_23.5_training_sample.tar.gz: 4213 MBytes

Target file has already been downloaded and unpacked.  Forced re-download.

Download complete.6 / 4213 MBytes  [25.54%]
Unpacking the tarball...
-rwxr-xr-x rmandelb/staff          0 2016-01-12 19:26:50 COSMOS_23.5_training_sample/
-rw-r--r-- rmandelb/staff      57600 2016-01-12 19:24:47 COSMOS_23.5_training_sample/acs_I_unrot_sci_20_cf.fits
-rw-r--r-- rmandelb/staff       6036 2016-01-12 19:26:01 COSMOS_23.5_training_sample/README
-rw-r--r-- rmandelb/staff    9371520 2016-01-08 08:51:32 COSMOS_23.5_training_sample/real_galaxy_catalog_23.5.fits
-rw-r--r-- rmandelb/staff   14808960 2016-01-08 08:51:32 COSMOS_23.5_training_sample/real_galaxy_catalog_23.5_fits.fits
-rw-r--r-- rmandelb/staff    2473920 2015-12-31 09:52:02 COSMOS_23.5_training_sample/real_galaxy_galsim_selection.fits
-rw-r--r-- rmandelb/staff   74802240 2016-01-08 08:52:02 COSMOS_23.5_training_sample/real_galaxy_images_23.5_n1.fits
-rw-r--r-- rmandelb/staff   74275200 2016-01-08 08:52:02 COSMOS_23.5_training_sample/real_galaxy_images_23.5_n10.fits
[...]

PR #709, opened 2016-02-06
Add epsilon as a parameter of HSMParams

rmjarvis:
This one is really quick.  I didn't even bother making an issue for it.  The branch to check is #705.

I recently wanted to see if a residual I'm seeing in hsm measurements of PSF models might be due to the convergence criterion of the hsm measurements themselves (probably not -- probably an error in the PSF fitting routine).  I found in the HSM code an epsilon parameter to find_mom_2.  But that parameter was hard-coded to be 1.e-6 in both places that call find_mom_2.

In this PR, I made epsilon a parameter in hsmparams, so we can control it from the python layer.


rmjarvis:
> The branch to check is #705.

Oops.  I thought 705 was the next number.  Forgot to check open PRs and closed issues.  Anyway, 705 is still the branch even though I should have used 709.

include/galsim/hsm/PSFCorr.h
```diff
@@ -122,6 +122,10 @@ namespace hsm {
      *                           KSB method of PSF correction.  Warning: deviating from default
      *                           value of 2 results in code running more slowly, and results have
      *                           not been significantly tested.
+     * @param epsilon            Accuracy (in x0, y0, and sigma as a fraction of sigma) when
```
jmeyers314: maybe use "x0, y0, and sigma, each as a fraction of sigma"  (if that's accurate).  It wasn't completely clear to me whether the x0 and the y0 were included in the fraction of sigma or not.

Other than that, looks good to me.

rmjarvis: I think you're right that it should be that, but I'm not sure if it actually is.  I tried to follow the code in hsm, but got a little lost, so I think I'll punt and let Rachel weigh in on whether this text should be changed.  (I just copied it from a comment in the C++ code.)  

I.e. is the x0,y0 convergence relative to sigma, or it units of pixels?

rmandelb:
Thanks for this, Mike.  My intention had been that the HSMParams would suck up any magic numbers, and apparently I missed this one.

I have a slight concern about the name ```epsilon``` for this new parameter.  If someone is looking at the source code and comparing it to the original Hirata & Seljak definition of the algorithm in their 2003 paper, that description has a quantity called epsilon (which is a residual image).  So for example in some of the comments on PSFCorr.cpp it mentions that quantity epsilon, which is different from this one and which has a different name in the code.  But still I think it's confusing and would prefer if this new quantity had a different name like ```accuracy```.

rmjarvis:
Sure.  Any name is fine with me.  I just used epsilon, since that's what it's called in the C++ code. In many codes, `threshold` is the correlate for a parameter like this.  Or `tolerance` is also common.  Or we can go verbose, something like `convergence_threshold` maybe.

rmandelb:
Yeah, the code was already confusing (you can find a comment referring to the epsilon that is in the H&S paper, for example, and it's clearly something different).  This will make it cleaner.  I think ```convergence_threshold``` would be best.

rmjarvis:
Done. b29f9e1

include/galsim/hsm/PSFCorr.h
rmjarvis: I looked into this a bit more, and I convinced myself that it is indeed correct that dx and dy are in fractions of sigma, so I updated the doc string for this to be: "Accuracy (in x0, y0, and sigma, each as a fraction of sigma) when calculating adaptive moments. ..." as you suggested, Josh.

rmjarvis:
I think this is good to merge unless anyone else wanted to take a look.  (No hurry if someone else wanted to review it.)

include/galsim/hsm/PSFCorr.h
```diff
@@ -122,6 +122,10 @@ namespace hsm {
      *                           KSB method of PSF correction.  Warning: deviating from default
      *                           value of 2 results in code running more slowly, and results have
      *                           not been significantly tested.
+     * @param convergence_threshold Accuracy (in x0, y0, and sigma as a fraction of sigma) when
+     *                           calculating adaptive moments. The value of sigma used for the
+     *                           convergence criterion is the minimum of the "guessed" value and
+     *                           the "current" value.
```
rmandelb: I don't think the second sentence is always true.  It's only true if ```std::abs(dsigma)>convergence_factor```.

It also seems like perhaps more information than is necessary, in any case?

rmandelb:
Other than my one small complaint, I think this is fine and I'm happy to have it merged.

include/galsim/hsm/PSFCorr.h
rmjarvis: OK.  I'll just remove that second sentence then.  Thanks.

Issue #712, opened 2016-02-19
CCD noise model


dkirkby:
I was confused by the ``electrons per ADU`` units of the CCDNoise gain parameter since I normally think of an amplifier gain as output / input, so I wanted to check how the gain is actually used in the code.  I couldn't find the relevant equation anywhere in the docs, but I eventually found this in `Noise.h`:
```
        double getVariance() const 
        {
            if (_gain > 0) return (_sky_level + _read_noise*_read_noise) / _gain;
            else return _read_noise * _read_noise;
        }
```
If the piece in parentheses is a variance in `electrons**2`, shouldn't it be divided by `gain**2` here?

rmjarvis:
Not quite.  Although, you're right that this line is wrong.  The correct formula is
```
sky_level / gain + (read_noise / gain)**2
```
And I agree with you about the definition of gain being weird.  ADU/e- would be much more natural, but e-/ADU is the standard definition for astronomy detectors.  Or CCDs at least.  cf. http://www.mirametrics.com/tech_note_ccdgain.htm

Unfortunately, it looks like our unit test of this function just checked that the output was this formula, rather than adding noise to an image and checking that the variance was right (or close).  I just did that, and the variance is indeed close to the formula I gave, but not close to `ccd_noise.getVariance()`.  I'll fix this and update the unit test.

rmjarvis:
BTW, the link I gave above has this quote: 

> Note: This use of the term "gain" is in the opposite sense to the way a circuit designer would use the term since, in electronic design, gain is considered to be an increase in the number of output units compared with the number of input units.

As usual, astronomers do everything backwards.  :)

dkirkby:
Thanks for the reference Mike.

Your formula assumes that `sky_level` is in ADU/pixel, but the docstring says it is in electrons/pixel:
```
@param sky_level    The sky level in electrons per pixel that was originally in the input
                    image, but which is taken to have already been subtracted off.
                    [default: 0.]
```
In this case, I believe my formula is still correct:
```
variance = (sky_level + read_noise ** 2) / (gain ** 2)
```

rmjarvis:
Ah, yes.  That seems to be another error.  The code assumes sky_level is in ADU/pixel (which is actually how it is usually given, so I think we want to keep that). The doc string is wrong here.  But you're right, if we did what the doc says, your formula is correct.  I'll fix the doc string about this too.

dkirkby:
[...]

PR #713, opened 2016-02-19
#712 Fix error in CCDNoise getVariance function

rmjarvis:
@dkirkby pointed out an error in the CCDNoise getVariance function.  The correct formula is:
```
var = sky_level/gain + (read_noise/gain)**2
```
This PR fixes this error.  It also adds some unit tests that check that when CCDNoise adds noise to an image, this is actually the variance of the resulting noise field.

For completeness, I also added the same check for Gaussian and Poisson noise, which both passed.


rmandelb:
Looks good to me.

rmandelb:
... except I just noticed that I have a test failure (CCDNoise variance).  Is it not deterministic?

rmjarvis:
I think it is deterministic.  Which test is failing for you?  Maybe you didn't recompile?  The change is in the C++ layer, so perhaps scons caching is messing something up and it didn't compile the new code.

rmandelb:
Looking back through the terminal history, I confirm that i ran scons and scons install before running the test suite.  And it appears to have recognized it had to recompile pysrc/Noise.cpp.

I just re-scons'ed and ran the tests again - same exact failure with same exact numbers, so yep, deterministic.  :)

Then I used scons -c to clean up, re-compiled everything, and still I get the exact same failure:
```
FAIL: Test CCD Noise generator
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/rmandelb/git/GalSim/tests/test_random.py", line 1449, in test_ccdnoise
    err_msg='Realized variance for CCDNoise did not match getVariance()')
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/site-packages/numpy/testing/utils.py", line 468, in assert_almost_equal
    raise AssertionError(msg)
AssertionError: 
Arrays are not almost equal to 1 decimals
Realized variance for CCDNoise did not match getVariance()
 ACTUAL: 19.333929061889648
 DESIRED: 19.444444444444446
variance =  19.3339290619
getVar =  19.4444444444

--------------------- >> end captured stdout << ----------------------
```

Am I missing something?

rmandelb:
I should add: this is on my mac.  On the linux cluster here, it all worked fine when I did scons / scons install / scons tests, no need to do any cleanup.

rmjarvis:
OK, it doesn't seem to be a cleanup issue then.  If you still had the old code, it would be very different from that.  On my machine the measured value 19.42, so I'm not sure why you get a different value here.

Update: I found a machine that gets the same answer you do.  (A linux box, in my case.)  So I can at least investigate and try to see why they get different answers.

rmandelb:
In case it�s relevant, I have numpy 1.6.1.

rmjarvis:
Yep.  I think that's it.  The one that fails for me is numpy 1.7.0.  The others are newer.  I think numpy changed their definition of var at some point.  The noise arrays in the two cases seem to be identical.  But the variance that numpy returns is different.  I'll see if I can figure out what the right function is to compute the variance we want here.  (Might be related to mean subtraction I'm guessing.  Still looking into it though...)

rmjarvis:
I figured it out.  Older versions of numpy don't use double precision in calculating statistics of single precision arrays.  Apparently with 4 million values, the errors added up enough to give the wrong answer for the variance.  Switching to dtype=float fixed it.

rmandelb:
Yep, that works, thanks!

rmjarvis:
Since @dkirkby pointed out even more errors in the docs related to CCDNoise as well as inconsistencies with the units in demo3, I went through everything associated with CCDNoise, and I think the docs are all correct now.  I also discovered some similar errors in the config-layer CCDNoise handling, but in code paths that we don't use anywhere.  To remedy that, I added unit tests to check those code paths.

I think things are all consistent now.  So David, please take a look and see if you think I covered everything.  And thanks very much for catching these errors!

dkirkby:
I didn't try running the new code but, reading it through, I didn't spot any lingering problems.  Thanks!

(For future consideration: these types of problems could be mitigated by attaching units to constants like read noise and gain.  See [here](https://github.com/desihub/specsim/blob/228e503b44960412db81fea26955ddca2d5641b6/specsim/data/config/desi.yaml#L80) for example.)

Issue #698, opened 2016-01-06
Make Ring a stamp type, not a gal type


rmjarvis:
Now that I've introduced `stamp` as a new top-level config field in #691, I've been thinking that our 'Ring' galaxy type really ought to be a stamp type.  I think it makes more logical sense, but since it is an API change, I figured I should raise the idea as an issue before implementing it to see if anyone had comments about it.

The new syntax I'm proposing for demo5 say is:
````
stamp:
    type: Ring
    num: 2

    # Put the transformations to do after the Ring rotation here:
    shear :
        type : G1G2
        g1 : 0.013
        g2 : -0.008

    shift :
        type : RandomCircle
        radius : 1 # arcsec

gal :
    type : Exponential
    resolution : 0.98

    # Only include the transformation that happen for the first galaxy in the ring, prior to the ring rotation.
    ellip :
        ...
```

It seems like this is a bit more logical than having Ring be the "galaxy type".  That never really matched the concept we had for all the other galaxy types very well.  But since I hadn't thought of the stamp type yet, there wasn't really a better place for it.  But now I think there is.

Thoughts?  

If we do this, I'd like to actually deprecate the Ring galaxy type, so it would be backwards compatible for now, but it would give a warning.  Unlike the other changes I've made recently to the config stuff that were all expansions in functionality and are fully backwards compatible.

rmandelb:
I think this is more logical than the current setup, and agree about deprecating the Ring galaxy type.

rmjarvis:
This is done on branch #693, since some of the implementation of this was connected to the reject feature we added there, so it was simpler to combine it with that.

Issue #660, opened 2015-04-17
Multiple PSFs input as images


pgee2000:
This is really a question.  I see from one of the demos that GalSim can read its PSF as an InterpolatedImage from a fits file image.  Is it also possible to use a multiple hdu fits file with many PSF images, and get it to use a different PSF for each galaxy through the yaml interface?

rmjarvis:
Hmm.  It seems we don't have the option of using an hdu other than the first one for an InterpolatedImage (either in the python layer or with a yaml file).  That's an oversight that we could certainly fix.  

However, another option that might work for you would be to use a different file for each object.  Something like the following should work:
```
psf:
    type: InterpolatedImage
    image:
        type: FormattedStr
        format: "psffile_%03d.fits"
        items:
            - # Only one item here, which is the number to use for %03d
                type: Sequence
                first: 1
                # You could also use last or nitems if you wanted the sequence to repeat.
                # Otherwise, this will use the next number for each successive galaxy.
```

pgee2000:
It is not terribly important.  As you say, I can always expand them into a
PSF directory of fits files.  It would be tiddier to use a multiple hdu
fits file, that's all.

Thanks,
Perry

pgee2000:
One other question for you.  When I run the experiment "real_galaxy", it
does not product a file called star_test_images.fits.  This causes the tar
file creation at the end of the run to fail.  It doesn't look fatal, since
everything else seems to get generated.  Is this a known bug, or am I
misparameterizing?

I am going through great3sims, so it might be a bug in their code.

Not sure how to report bugs.  Through github, in the same way as I sent
this last issue?

rmjarvis:
Sounds like a bug in their scripts.  Probably should post the details of your issue on the Great3 issues page:

https://github.com/barnabytprowe/great3-public/issues

rmandelb:
@pgee2000 , please do post it on the issues page that Mike linked.  If you've modified the scripts in any way, please include a link to them as well.

rmandelb:
FYI, this is done now on branch #693.  I plan to finish that functionality shortly and make a combined PR once the PR for #691 has been merged.

Issue #510, opened 2014-03-01
Add Chromatic stuff to config


rmjarvis:
So far, none of the chromatic functionality has been put into the config framework.  This issue is to do that.

I'd also like to refactor demo12 a bit.  Currently, it is three separate examples with separate (sets of) output files.  I'd like to turn this into a single integrated example that uses all the same functionality in the course of a single use case.  I had mentioned that I wanted to do this on #467, but it probably makes more sense to do this while I'm working on getting it to match up with an equivalent config version.

rmjarvis:
Closing this in preference to #969 where @sweverett is working on this.

rmjarvis:
Done.  #1087

PR #714, opened 2016-02-22
#693 and #660 and #698

rmandelb:
This branch includes the following updates:

- The ability to initialize an InterpolatedImage from a specific HDU of a fits file (#660).
- Some updates to the COSMOSCatalog and the config interface to allow pre-selection of galaxies being used for simulations, either using parametric fits or realistic galaxies (#693).
- Some updates to the catalogs themselves to enable that pre-selection to happen correctly and efficiently (which means that taking advantage of some updates requires re-downloading the catalogs; GalSim will identify whether that is needed, and the ```galsim_download_cosmos``` utility can be used for the downloading and unpacking).  Plus there's a new COSMOS catalog available, which isn't technically a GalSim update except that it is now the default one that GalSim will use unless the original one is specified, and there are some new kwargs that can be used to choose between the currently available catalogs.

Since both Mike and I worked on parts of this, it would be helpful if someone else like @jmeyers314 could take a look.  And I think @beckermr was interested in this, so perhaps he'd like to try it out?


rmjarvis:
I also did #698 on this branch: deprecating the config 'Ring' object type and putting this functionality in a stamp type (again called 'Ring') instead.

jmeyers314:
>it would be helpful if someone else like @jmeyers314 could take a look

Sure thing.

jmeyers314:
I just started looking this over by cleaning out my GalSim directory with:

```
scons -c
rm -rf .scon*
```

and then compiling with plain `scons`.

For quite some time now (certainly nothing specific to this PR), this has been generating warnings on my laptop; all of which look like:

```
1 warning generated.
In file included from pysrc/CorrelatedNoise.cpp:28:
In file included from include/galsim/CorrelatedNoise.h:29:
In file included from /Users/josh/miniconda2/opt/lsst/tmv/include/TMV_Sym.h:27:
In file included from /Users/josh/miniconda2/opt/lsst/tmv/include/TMV.h:30:
In file included from /Users/josh/miniconda2/opt/lsst/tmv/include/tmv/TMV_Matrix.h:495:
/Users/josh/miniconda2/opt/lsst/tmv/include/tmv/TMV_Permutation.h:465:57: warning: dependent nested
      name specifier 'VectorView<T, A>::' for friend class declaration is not supported; turning off
      access control for 'Permutation' [-Wunsupported-friend]
        friend inline VectorView<T,A>& VectorView<T,A>::sort(
                                       ~~~~~~~~~~~~~~~~~^
```

rmjarvis:
> For quite some time now (certainly nothing specific to this PR), this has been generating warnings on my laptop; all of which look like:

Yes.  Sorry.  I know about those.  clang added this warning recently, and I haven't gotten around to fixing TMV to avoid it yet.  Until I do, you'll need to compile with WARN=false unfortunately.

galsim/scene.py
```diff
@@ -33,7 +33,8 @@
 
 class COSMOSCatalog(object):
     """
-    A class representing a sample of galaxies from the COSMOS sample with F814W<23.5.
+    A class representing a random subsample of galaxies from the COSMOS sample with F814W<25.2
```
jmeyers314: Should probably indicate at which galsim version number the default catalog changes, i.e. the meaning of "previous" here.

galsim/scene.py
```diff
@@ -86,46 +92,58 @@ class COSMOSCatalog(object):
...
+                            F814W<25.2 COSMOS catalog in $PREFIX/share/galsim.  It will raise an
+                            exception if the catalog is not there telling you to run
+                            galsim_download_cosmos.]
+    @param sample           A keyword argument that can be used to specify the sample to use, i.e.,
+                            "23.5" or "25.2".  At most one of `file_name` and `sample` should be
+                            specified.
+                            [default: None, which results in the same default as `file_name=None`.]
+    @param image_dir        Keyword that is only used for real galaxies, not parametric ones, to
+                            specify the directory of the image files.
+                            If a string containing no `/`, it is the relative path from the location
```
jmeyers314: Is there a reason not to follow the normal *nix relative-vs-absolute path rule here?  I.e., relative paths can't *start* with '/', but could contain '/' at later positions?  I see there's even an os.path.isabs() utility for this.

galsim/scene.py
```diff
@@ -86,46 +92,58 @@ class COSMOSCatalog(object):
...
+                            of the catalog file to the directory containing the galaxy/PDF images.
+                            If a path (a string containing `/`), it is the full path to the
+                            directory containing the galaxy/PDF images. [default: None]
+    @param dir              The directory of catalog file. [default: None]
+    @param preload          Keyword that is only used for real galaxies, not parametric ones, to
+                            choose whether to preload the header information.  If `preload=True`,
+                            the bulk of the I/O time is in the constructor.  If `preload=False`,
+                            there is approximately the same total I/O time (assuming you eventually
+                            use most of the image files referenced in the catalog), but it is spread
+                            over the various calls to getGal() and getPSF().  [default: False]
```
jmeyers314: Should `getGal()` and `getPSF()` just be `makeGalaxy()` here?

galsim/scene.py
```diff
@@ -86,46 +92,58 @@ class COSMOSCatalog(object):
...
+    @param noise_dir        Keyword that is only used for real galaxies, not parametric ones.
```
jmeyers314: Does `noise_dir` need any relative-vs-absolute path guidance like `image_dir`?

galsim/scene.py
```diff
@@ -86,46 +92,58 @@ class COSMOSCatalog(object):
...
+                            The directory of the noise files if different from the directory of the 
+                            image files.  [default: image_dir]
+    @param use_real         Enable the use of realistic galaxies?  [default: True]
```
jmeyers314: Is there a speed or RAM benefit to having `use_real=False` when making parametric galaxies?  If so, then should add this to the documentation.  If not, then why have this kwarg?

galsim/scene.py
```diff
@@ -86,46 +92,58 @@ class COSMOSCatalog(object):
...
+                            If this parameter is False, then `makeGalaxy(gal_type='real')` will
+                            not be allowed.
+    @param exclusion_level  Level of cuts to make on the galaxies based on the quality of postage
+                            stamp definition and/or parametric fit quality.  Options:
+                            "none": No cuts.
```
jmeyers314: I assume there were some cuts in place when making the catalogs.  So should this be "No additional cuts", maybe?

galsim/scene.py
```diff
@@ -86,46 +92,58 @@ class COSMOSCatalog(object):
...
+                            "bad_ps": Apply cuts to eliminate galaxies that have failures in postage
```
jmeyers314: I see `bad_ps` and immediately think that something is wrong with a power spectrum somewhere.  Suggest `bad_stamp` as an alternative.

galsim/scene.py
```diff
@@ -194,47 +271,104 @@ def __init__(self, file_name=None, image_dir=None, dir=None, preload=False, nois
         self.orig_index = np.arange(len(self.param_cat))
         mask = np.ones(len(self.orig_index), dtype=bool)
 
-        # If requested, select galaxies based on existence of a usable fit.
-        if exclude_fail:
+        # If requested, select galaxies in a way that excludes suspect postage stamps (e.g., with
+        # deblending issues), suspect parametric model fits, or both of the above plus marginal
+        # ones.
+        if exclusion_level == 'bad_ps' or exclusion_level == 'marginal':
```
jmeyers314: Looks like this is the same if statement as line 239.  Could these be streamlined?

galsim/scene.py
```diff
@@ -194,47 +271,104 @@ def __init__(self, file_name=None, image_dir=None, dir=None, preload=False, nois
...
+                sn_limit = 12.0
+            if self.selection_cat is not None:
+                div_val = self.selection_cat['peak_image_pixel_count']
+                div_val[div_val == 0.] = 1.e-5
+                mask &= ( (self.selection_cat['sn_ellip_gauss'] >= sn_limit) &
+                          ((self.selection_cat['min_mask_dist_pixels'] > 11.0) |
+                           (self.selection_cat['average_mask_adjacent_pixel_count'] / \
+                               div_val < cut_ratio)) )
+
+        if exclusion_level == 'bad_fits' or exclusion_level == 'marginal':
```
jmeyers314: Personally, I like the earlier style: `if exclusion_level in ['bad_fits', 'marginal']`.

galsim/scene.py
```diff
@@ -194,47 +271,104 @@ def __init__(self, file_name=None, image_dir=None, dir=None, preload=False, nois
...
+                      (sersicfit_status < 5)) &
+                      ((bulgefit_status > 0) &
+                      (bulgefit_status < 5)) )
+        
+            # Some fit parameters can indicate a likely sky subtraction error
             hlr = self.param_cat['sersicfit'][:,1]
             n = self.param_cat['sersicfit'][:,2]
-            mask &= ( (n < 5) | (hlr < 1./cosmos_pix_scale) ) 
-            # May add more cuts here if we discover other kinds of problematic objects.
+            mask &= ( (n < 5) | (hlr < 1./cosmos_pix_scale) )
```
jmeyers314: I think this is saying that the HLR should be smaller than 1 arcsec.  This would be clearer with something like
```
hlr_pix = self.param_cat['sersicfit'][:, 1]
max_hlr_arcsec = 1.0
mask &= ((n < 5) | (hlr_pix < max_hlr_arcsec/cosmos_pix_scale))
```

jmeyers314: Also, I see down below that there's also a factor of `np.sqrt(q)` in the formula for hlr in arcsec.  Maybe a comment would be good here explaining why that's important down below but not here.

galsim/scene.py
```diff
@@ -194,47 +271,104 @@ def __init__(self, file_name=None, image_dir=None, dir=None, preload=False, nois
...
+                hlr_pix = sparams[:,1]
+                n = sparams[:,2]
+                q = sparams[:,3]
+                hlr = cosmos_pix_scale*hlr_pix*np.sqrt(q)
+                if min_flux > 0. or max_flux > 0.:
+                    flux_hlr = sparams[:,0]
+                    # The prefactor for n=4 is 3.607.  For n=1, it is 1.901.
+                    # It's not linear in these values, but for the sake of efficiency and the 
+                    # ability to work on the whole array at once, just linearly interpolate.
+                    # Hopefully, this can be improved as part of issue #693.  Maybe by storing the
```
jmeyers314: I assume this comment is outdated?

galsim/scene.py
```diff
@@ -401,7 +549,7 @@ def _makeParametric(self, indices, chromatic, sersic_prec, gsparams):
 
         gal_list = []
         for index in indices:
-            record = self.param_cat[self.orig_index[index]]
+            record = self.getParametricRecord(index)
```
jmeyers314: Looks like I've been commenting on a fair amount of old code that just happened to show up in the imperfectly processed github diff.  Hope that's not too annoying.  I actually have a comment for some old code on line 751 in `getParametricRecord`, which isn't imperfectly diffed, so github won't let me comment on that line, so I'll place it here instead.  I was happy to see a generator expression:

```
record_dict = dict( ((k,record[k]) for k in record.dtype.names) )
```

I just wanted to point out that a generator expression passed as a single argument to a function doesn't need its own parens.  I.e., the following works and, at least to me, looks a bit cleaner:

```
record_dict = dict((k,record[k]) for k in record.dtype.names)
```

galsim/scene.py
rmandelb: Indeed.  Done!

galsim/scene.py
rmandelb: Yes they should.  That was clearly copied from the RealGalaxyCatalog description.  Fixed.

galsim/scene.py
rmandelb: Fixed.

galsim/scene.py
rmandelb: Yes, fixed.

galsim/scene.py
rmandelb: Both the sqrt(q) issue and the nature of this cut itself are clarified in https://github.com/GalSim-developers/GalSim/commit/e7761cdd750d921e786282aae1349ee6942811f5

galsim/scene.py
rmandelb: Good catch.  Done: https://github.com/GalSim-developers/GalSim/commit/7e3b9aa4889cd4d976daf7af47150bd4662abf3f

galsim/scene.py
rmandelb: Fair enough.  Done.

galsim/scene.py
rmandelb: Hmmm, interesting question.   @rmjarvis , what are your thoughts?

galsim/scene.py
rmandelb: Well - those cuts were really minimal because we were trying to get as close as we could to a flux-limited sample.  But I suppose technically we should mention them and point to their description.  Done.

galsim/scene.py
rmjarvis: I've always found the way these dir kwargs are parsed a bit confusing, to be honest.  I'd actually rather have just a single dir for everything and let the user use symlinks if they want to put things in different directories than the standard layout.  Seems to me that would be less confusing than having 3 different dir kwargs here.  Then there wouldn't be any need to distinguish between absolute and relative paths.

rmandelb: I'm fine with that in principle.  But it might be tricky to do this in a way that doesn't break people's old code when they go to v1.4.  If you have ideas for how to cope with that, I would be fine with updating both COSMOSCatalog and RealGalaxyCatalog to this new model.

galsim/scene.py
rmandelb: Done.

galsim/scene.py
rmandelb: Well - in the very original version of the code I think there may have been more benefit because it was structured differently.  Now, there is a quite small benefit in terms of RAM and speed (don't have to read in the real catalog) and a possible logistical benefit, like if you never want to use realistic galaxies and deleted the real catalog from your system, you could still use COSMOSCatalog with ```use_real=False``` and not have it complain that it couldn't find the catalog.  I just added a note on this.

galsim/scene.py
rmjarvis: It would mean deprecating the old functionality, rather than removing it.  For now, probably just leave the code the same, but give deprecation warnings if either `image_dir` or `noise_dir` is not None.  I don't think there is any change required for the case where `dir` is given, but not `image_dir` or `noise_dir`.

galsim/scene.py
```diff
@@ -350,13 +504,24 @@ def makeGalaxy(self, index=None, gal_type=None, chromatic=False, noise_pad_size=
...
+                    'You seem to have an old version of the COSMOS parameter file.\n'+
+                    'Please run `galsim_download_cosmos` to re-download the COSMOS catalog\n' +
+                    'and take advantage of pre-computation of many quantities..')
+
             gal_list = self._makeParametric(indices, chromatic, sersic_prec, gsparams)
 
-        # If deep, rescale the size and flux
+        # If trying to use the 23.5 sample and "fake" a deep sample, rescale the size and flux as
+        # suggested in the GREAT3 handbook.
         if deep:
```
jmeyers314: Should the `deep` kwarg be disabled if using the 25.2 catalog?  (Same question for `_makeSingleGalaxy` down below).

rmandelb: Yes.  Done.

galsim/scene.py
rmandelb: OK.  I will take a stab at it this weekend; will post when it�s ready to go over.

examples/great3/cgc.yaml
```diff
@@ -16,9 +16,11 @@
 #    and/or other materials provided with the distribution.
 
 # The Great3 control-ground-constant branch configuration file
-# Note: This configuration will not produce images _identical_ to the ones used for Great3, 
+# Note: This configuration will not produce images _identical_ to the ones used for Great3,
 # since the order of random number generation is different.  However, it is intended to
 # produce an _equivalent_ simulation as the one produced for Great3.
+modules:
```
jmeyers314: I know that this example isn't really part of the tutorial series, but I think it would be nice for some more documentation on the `modules` top-level config type here.  (I think the only other example of this is `examples/des/draw_psf.yaml`.)  Or alternatively, maybe a new tutorial that more simply shows off the `modules` config type (or augment an existing tutorial?).

examples/great3/cgc_faint.yaml
```diff
@@ -0,0 +1,52 @@
...
+#    file.
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions, and the disclaimer given in the documentation
+#    and/or other materials provided with the distribution.
+
+# An alterate version of the Great3 control-ground-constant branch configuration file, 
+# using an I<25.2 sample (instead of faking a deeper sample using I<23.5 images).
+
+# Start with the cgc.yaml file, which we'll modify below.
+template: cgc.yaml
```
jmeyers314: This top-level config type is also really cool.  We should advertise it in the tutorial series!

examples/great3/cgc.yaml
rmjarvis: The DES examples are a more complete set of examples that use the modules field in various ways.  cf. the PR for #654.  I've called both the DES and the Great3 examples "Advanced Simulations" [here](https://github.com/GalSim-developers/GalSim/wiki/Tutorials) and mention the top-level modules field as a significant feature of both.

I'm not averse to including the modules in the main set of numbered demos.  I think the natural place will likely be demo13.  But that will have to wait until the chromatic stuff is folding into the config framework, which I haven't done yet.

examples/great3/cgc.yaml
```diff
@@ -122,16 +132,32 @@ gal:
...
-    # lower S/N.  To compensate, they added less noise in those cases.  However, the
-    # exact formula they used is still a bit of a mystery to me.
-    # They did this as a LookupTable, but the formula below is pretty close.
-    fmax_var: '$0.004 * math.exp(2.9 * (1.1-@psf.items.0.fwhm) * (0.95-@psf.items.0.fwhm))'
+    # different seeing sizes.  When the seeing is larger, a given flux galaxy will have
+    # lower S/N.  To compensate, they added less noise in those cases.  The formula that was used
+    # was based on running experiments with different seeing values, and checking how the S/N varied
+    # at fixed noise variance.  The results of those experiments were used to create a LookupTable
+    # in the GREAT3 scripts, but the formula below is pretty close.
+    fmax_var: '$0.004 * math.exp(2.9 * (1.1-psf_fwhm) * (0.95-psf_fwhm))'
```
jmeyers314: Are the '$' and '@' Eval shorthands documented somewhere?

rmjarvis: It's not live yet.  But yes.  Here: https://github.com/GalSim-developers/GalSim/wiki/Config-Values#shorthand-notation

jmeyers314:
Hi Rachel, Mike,

I've looked through most of the code now.  Looks good!  One caveat, I'm not as familiar with the config code, so you should only consider that "lightly" reviewed.

One last question/comment though:

I tried the following experiment to check on the tasks/jobs machinery.  From directory `$GALSIM_DIR/examples/great3/`, I ran (in three different terminals):

```
galsim cgc.yaml output.nfiles=10 output.dir=all10
galsim cgc.yaml output.nfiles=10 output.dir=5by5 -n 2 -j 1
galsim cgc.yaml output.nfiles=10 output.dir=5by5 -n 2 -j 2
```

My expectation was that the results in all10/ would bit-for-bit match those in 5by5/.  What I found was that all10/image-00[0-7]-0.fits matched 5by5/image-00[0-7]-0.fits, and similarly for the epoch_catalog files, but that numbers 8 and 9 in the series differed (rather significantly when I ds9'd the images).  It's as if a random number somewhere got skipped.  Is this surprising behavior?

examples/great3/cgc.yaml
jmeyers314: Great.  Thanks.

rmjarvis:
> My expectation was that the results in all10/ would bit-for-bit match those in 5by5/. What I found was that all10/image-00[0-7]-0.fits matched 5by5/image-00[0-7]-0.fits, and similarly for the epoch_catalog files, but that numbers 8 and 9 in the series differed (rather significantly when I ds9'd the images). It's as if a random number somewhere got skipped. Is this surprising behavior?

Yes, that's probably indicating a bug.  Thanks for checking that.  I'll investigate and see if I can figure out what is going on there.  They should have been identical.

rmjarvis:
> numbers 8 and 9 in the series differed (rather significantly when I ds9'd the images). It's as if a random number somewhere got skipped. Is this surprising behavior?

I figured it out.  The problem was with the cosmos_cat.  The number of objects being rejected was coming out wrong.  I tracked it down to where `psf_fwhm` gets the current value of `psf.items.0.fwhm`.  Querying this when it hadn't already been calculated meant that it had to calculate it then, which should have been fine, but it jumped over the place where we tell it to use `index_key : image_num` (at the top of `psf`).

TL;DR, I fixed it here: 44ca3067cb70ebaa99375b09cafb01dfc2624dc8

rmandelb:
I wonder if there is a short/quick unit test we can add that will be able to catch the kind of issue that Josh found?

rmandelb:
I deprecated ```image_dir``` and ```noise_dir```, and updated the documentation to reflect this, in https://github.com/GalSim-developers/GalSim/commit/56e04d0c6b3b84a0cf493198d36dfd04d5d448c8 . This caused an issue for ```__repr__``` and therefore for pickling, which I fixed in a subsequent commit (I hope...).

@rmjarvis , @jmeyers314 , can you please check this out and see what you think?

rmjarvis:
You don't need to split off the directory in the `__repr__` function.  Providing just a catalog name, complete with the directory, will do that correctly already.  cf. 7c30fdf

rmandelb:
Good point.  Thanks.

rmjarvis:
> I wonder if there is a short/quick unit test we can add that will be able to catch the kind of issue that Josh found?

Good idea.  Done.  	bb87871

rmjarvis:
I think maybe all the comments have been dealt with, except for the suggestion to add template and modules examples to the demos.  Like I said, I think those will be most appropriate in demo12 and/or 13, so I'd like to push that into issue #510.

If that's ok with you (@jmeyers314 and @rmandelb), then I think we're ready to merge this?

jmeyers314:
>I think we're ready to merge this?

Yep.  Looks good to me.  All tests pass.

rmandelb:
+1

Issue #721, opened 2016-03-12
Make GalSim compilable under Anaconda on OS X El Capitan


jmeyers314:
The OS X El Capitan System Integrity Protection (SIP) feature breaks the ability to install GalSim under Anaconda python.  We hacked together a solution during DESC hack day, and will PR it here.

drphilmarshall:
Jolly good!

Issue #724, opened 2016-03-12
adjust GalSim eups table to make setup command `setup galsim` instead of `setup GalSim`


jmeyers314:
This issue is to change the name of `ups/GalSim.table` to `ups/galsim.table`, so that the eups name of the package is all lowercase rather than camelCase, which is more in line with other LSST eups package names.  This also helps avoid an eups bug on OS X related to that os's filesystem's lack of case sensitivity.

If no one objects in the next few days, I'll do this on master, and then also create and tag some revision releases 1.2.1 and 1.3.1.

danielsf:
Speaking for LSST (because I am totally empowered to do that...) I have no objections.

cwwalter:
So, I'm having problems making this work so I can actually run GalSim..  The LD_LIBRARY_PATH is being set but not the DYLD_LIBRARY_PATH.  

This is what is in the repo now:

```
envPrepend(PYTHONPATH, ${PRODUCT_DIR})
envPrepend(LD_LIBRARY_PATH, ${PRODUCT_DIR}/lib)
envPrepend(PATH, ${PRODUCT_DIR}/bin)
```

And this is what is in Scott's stack version from (https://github.com/lsst/galsim/blob/master/ups/GalSim.table):

```
envPrepend(PYTHONPATH, ${PRODUCT_DIR})
envPrepend(LD_LIBRARY_PATH, ${PRODUCT_DIR}/lib)
envPrepend(DYLD_LIBRARY_PATH, ${PRODUCT_DIR}/lib)
envPrepend(PYTHONPATH, ${PRODUCT_DIR}/lib/python)
envPrepend(PATH, ${PRODUCT_DIR}/bin)
``

The effect of not having the DYLD is that at run time it can't load the module.  I think we might have fixed this by hand at SLAC.  I'm not sure why it is working for Josh..

**Comment by cwwalter** [comment]
One note is that after I fixed this by hand after some testing I found a DYLD_LIBRARY_PATH line in gs_scons.conf.

So Josh may want to see if he has something actually in there instead.

**Comment by jmeyers314** [comment]
>So Josh may want to see if he has something actually in there instead.

My gs_scons.conf is still empty, as is my $DYDL_LIBRARY_PATH before running "setup GalSim".

**Comment by cwwalter** [comment]
what does it look like after you do the setup of your own version?

**Comment by jmeyers314** [comment]
>what does it look like after you do the setup of your own version?

```
[...]

Issue #717, opened 2016-02-25
SConstruct on darwin insists on Xcode and ignores command line tools


timj:
I am having trouble building Galsim on my Mac because I do not have Xcode installed but I am using the Apple command line tools. The problem is that an attempt is made to determine the Xcode version by invoking `xcodebuild` and this fails. Can the Xcode version determination be deferred until it might be needed? If TMV is not linking against BLAS then the whole test is irrelevant. The code is currently:
```python
p = subprocess.Popen(['xcodebuild','-version'], stdout=subprocess.PIPE)
xcode_version = p.stdout.readlines()[0].split()[1]
print 'XCode version is',xcode_version
if (platform.mac_ver()[0] >= '10.7' and #xcode_version < '5.1' and
    '-latlas' not in tmv_link and ('-lblas' in tmv_link or '-lcblas' in tmv_link)):
    print 'WARNING: The Apple BLAS library has been found not to be thread safe on'
    print '         Mac OS versions 10.7+, even across multiple processes (i.e. not'
    print '         just multiple threads in the same process.)  The symptom is that'
    print '         `scons tests` may hang when running nosetests using multiple'
    print '         processes.'
    if xcode_version < '5.1':
        print '         This seems to have been partially fixed with XCode 5.1, so we'
        print '         recommend upgrading to the latest XCode version.  However, even'
        print '         with 5.1, some systems still seem to have problems.'
        env['BAD_BLAS'] = True
    else:
        print '         This seems to have been partially fixed with XCode 5.1, so there'
        print '         is a good chance you will not have any problems.  But there are'
        print '         still occasional systems that fail when using multithreading with'
        print '         programs or modules that link to the BLAS library (such as GalSim).'
        print '         If you do have problems, the solution is to recompile TMV with'
        print '         the SCons option "WITH_BLAS=false".'
```
but changing it to something like:
```python
if (platform.mac_ver()[0] >= '10.7' and
    '-latlas' not in tmv_link and ('-lblas' in tmv_link or '-lcblas' in tmv_link)):
    p = subprocess.Popen(['xcodebuild','-version'], stdout=subprocess.PIPE)
    xcode_version = p.stdout.readlines()[0].split()[1]
    print 'XCode version is',xcode_version
   if xcode_version < '5.1':
```
would at least defer the breakage to a situation where BLAS is involved. I think the only real fix for this is to treat the clang version as a proxy for Xcode version.

rmjarvis:
[...]

PR #725, opened 2016-03-12
#721 Fix SCons for El Capitan when DYLD_LIBRARY_PATH is needed

rmjarvis:
At the DESC hack day today, we tried to install GalSim with Anaconda Python on @cwwalter's laptop.  But we ran into a new problem introduced by El Capitan, the latest MacOS.  They decided to institute a new policy called "System Integrity Protection" (SIP).  Probably, this is really important for stopping terrorists from taking over the world.  But for running SCons properly, it was a big headache.

Basically, SIP involved removing dangerous environment variables like LD_LIBRARY_PATH and DYLD_LIBRARY_PATH from system calls.  So within SCons, or when SCons calls system commands to see if things can run properly, it fails if you system is relying on these variables to find the right libraries at runtime.

The solution is twofold:
1. Add scons options to let the user explicitly give LD_LIBRARY_PATH or DYLD_LIBRARY_PATH so SCons can set them back into its own environment after SIP has stripped them out of os.environ.  
2. Wheneven SCons calls a shell command to run something, add `DYLD_LIBRARY_PATH=env['DYLD_LIBRARY_PATH']` before the command to bypass the SIP stripping that happens there.

I think all this is working properly.  We got it to work on Chris's laptop before he left for the airport.  But it would be worth people quickly checking that I haven't broken anything for other systems in the process.

Also, @cwwalter, please make sure that my cleaning up of the SConstruct file from what you sent me didn't make anything stop working for you.  You should be able to do:
```
cd /path/to/GalSim
git pull
git checkout '#721'
scons
```
and just make sure this still works for you.  Thanks!

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -341,6 +341,17 @@ than the default `g++`, type
 
     scons CXX=icpc
 
+On El Capitan, Apple instituted a new security measure wherein system calls
+lose some of the system environment variables, including DYLD_LIBRARY_PATH
+among others.  If your system is set up to use that environment variable to
+resolve library locations at runtime, then this will cause problems when SCons
+is trying to figure out if things are installed correctly.  To override this
+behavior, you can explicitly send this environment variable to SCons by writing
+
+    scons DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH
+
+and it will be able to re-set this value within the SCons processing.
+
 One nice feature of SCons is that once you have specified a parameter, it will
 save that value for future builds in the file `gs_scons.conf`, so once you have
 the build process working, for later builds you only need to type `scons`. It
--- a/SConstruct
+++ b/SConstruct
@@ -239,6 +239,22 @@ def ErrorExit(*args, **kwargs):
         out.write("Error trying to get output of conftest executables.\n")
         out.write(sys.exc_info()[0])
 
+    # Give a helpful message if running El Capitan.
+    if sys.platform.find('darwin') != -1:
+        import platform
+        major, minor, rev = platform.mac_ver()[0].split('.')
+        print 'Mac version: ',major,minor
+        if int(major) > 10 or int(minor) >= 11:
+            print
+            print 'Starting with El Capitan (OSX 10.11), Apple instituted a new policy called'
+            print '"System Integrity Protection" (SIP) where they strip "dangerous" environment'
+            print 'variables from system calls (including SCons).  So if your system is using'
+            print 'DYLD_LIBRARY_PATH for run-time library resolution, then SCons cannot see it'
+            print 'so that may be why this is failing.  cf. Issues #721 and #725.'
+            print 'You should try executing:'
+            print
+            print '    scons DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH'
+
     print
     print 'Please fix the above error(s) and rerun scons.'
     print 'Note: you may want to look through the file INSTALL.md for advice.'
@@ -1620,7 +1636,8 @@ int main()
             # Don't require the user to have xcode installed.
             xcode_version = None
             print 'Unable to determine XCode version'
-        if (platform.mac_ver()[0] >= '10.7' and '-latlas' not in tmv_link and
+        major, minor, rev = platform.mac_ver()[0].split('.')
+        if ((int(major) > 10 or int(minor) >= 7) and '-latlas' not in tmv_link and
                 ('-lblas' in tmv_link or '-lcblas' in tmv_link)):
             print 'WARNING: The Apple BLAS library has been found not to be thread safe on'
             print '         Mac OS versions 10.7+, even across multiple processes (i.e. not'
@@ -1825,7 +1842,6 @@ Help(opts.GenerateHelpText(env))
 env['final_messages'] = []
 # Everything we are going to build so we can have the final message depend on these.
 env['all_builds'] = []
-
 
 if not GetOption('help'):
 
--- a/SConstruct
+++ b/SConstruct
@@ -217,8 +217,9 @@ def ErrorExit(*args, **kwargs):
                 cmd = conftest
             else:
                 cmd = env['PYTHON'] + " < " + conftest
-            p = subprocess.Popen([cmd], stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
-                                 shell=True)
+            cmd = PrependLibraryPaths(cmd,env)
+            p = subprocess.Popen(['bash','-c',cmd], stdout=subprocess.PIPE,
+                                 stderr=subprocess.STDOUT, shell=False)
             conftest_out = p.stdout.readlines()
             out.write('Output of the command %s is:\n'%cmd)
             out.write(''.join(conftest_out) + '\n')
--- a/SConstruct
+++ b/SConstruct
@@ -1892,6 +1892,7 @@ if not GetOption('help'):
     env['_RunInstall'] = RunInstall
     env['_RunUninstall'] = RunUninstall
     env['_AddRPATH'] = AddRPATH
+    env['_PrependLibraryPaths'] = PrependLibraryPaths
 
     # Both bin and examples use this:
     env['BUILDERS']['ExecScript'] = Builder(action = BuildExecutableScript)
--- a/SConstruct
+++ b/SConstruct
@@ -1,4 +1,4 @@
-# vim: set filetype=python et ts=4 sw=4:
+
 
 # Copyright (c) 2012-2015 by the GalSim developers team on GitHub
 # https://github.com/GalSim-developers
@@ -43,7 +43,7 @@ subdirs=['src', 'pysrc', 'bin', 'galsim', 'share']
 config_file = 'gs_scons.conf'
 
 # Default directory for installation.
-# This is the only UNIX specific things I am aware
+
 # of in the script.  On the other hand, these are not required for the
 # script to work since prefix can be set on the command line and the
 # extra paths are not needed, but I wish I knew how to get the default
@@ -104,7 +104,8 @@ opts.Add(PathVariable('DYLD_LIBRARY_PATH',
          'Set the DYLD_LIBRARY_PATH inside of SCons.  '+
          'Particularly useful on El Capitan (and later), since Apple strips out '+
          'DYLD_LIBRARY_PATH from the environment that SCons sees, so if you need it, '+
-         'this option enables SCons to set it back in for you.',
+         'this option enables SCons to set it back in for you by doing '+
+         '`scons DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH`.',
          '', PathVariable.PathAccept))
 opts.Add('NOSETESTS','Name of nosetests executable','')
 opts.Add(BoolVariable('CACHE_LIB','Cache the results of the library checks',True))
--- a/SConstruct
+++ b/SConstruct
@@ -715,20 +715,13 @@ def ReadFileList(fname):
 def PrependLibraryPaths(pname, env):
     """Turn a system command, pname, into "DYLD_LIBRARY_PATH=blah "+pname
 
-    env is the relevant SCons environment.  But also check os.environ if nothing is there.
+    env is the relevant SCons environment.
     """
-    osenv = os.environ
     if 'DYLD_LIBRARY_PATH' in env and env['DYLD_LIBRARY_PATH'] != '':
         pre = 'DYLD_LIBRARY_PATH=%r'%env['DYLD_LIBRARY_PATH']
         pname = "%s %s"%(pre,pname)
-    elif 'DYLD_LIBRARY_PATH' in osenv and osenv['DYLD_LIBRARY_PATH'] != '':
-        pre = 'DYLD_LIBRARY_PATH=%r'%osenv['DYLD_LIBRARY_PATH']
-        pname = "%s %s"%(pre,pname)
     if 'LD_LIBRARY_PATH' in env and env['LD_LIBRARY_PATH'] != '':
         pre = 'LD_LIBRARY_PATH=%r'%env['LD_LIBRARY_PATH']
-        pname = "%s %s"%(pre,pname)
-    elif 'LD_LIBRARY_PATH' in osenv and osenv['LD_LIBRARY_PATH'] != '':
-        pre = 'LD_LIBRARY_PATH=%r'%osenv['LD_LIBRARY_PATH']
         pname = "%s %s"%(pre,pname)
     return pname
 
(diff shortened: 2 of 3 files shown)


jmeyers314:
Hmm...  This broke my installation.  I cleaned up with
```
rm -rf .scon*
scons -c
```
and then tried to compile.  Scons hangs on "Checking for correct TMV linkage..."
Here are my config.log and gs.error files (with `.txt` extension appended to make github happy.)

[config.log.txt](https://github.com/GalSim-developers/GalSim/files/170464/config.log.txt)
[gs.error.txt](https://github.com/GalSim-developers/GalSim/files/170465/gs.error.txt)

rmjarvis:
Sorry Josh.  I think I figured out what I did wrong.  Please try it now.

cwwalter:
Sorry... got home at 4:00am last night... I was trying to tidy everything up before I boarded last night for my flight and ran across the same thing as Josh.

I thin maybe TMV checks on my laptop had already been phase when were just bypassing the check.    Is that consistent with what you saw Mike?  Was there something missing?

cwwalter:
With this new version for me it is immediately stopping on the FFTW check.  

'''
wasabi:GalSim % scons
scons: Reading SConscript files ...
SCons is version 2.4.1 using python version 2.7.11
Python is from /Users/Walter/LSST/Anaconda/include/python2.7
Using the default scons options
Using python =  /Users/Walter/LSST/Anaconda/bin/python
Using default PYPREFIX =  /Users/Walter/LSST/Anaconda/lib/python2.7/site-packages
Using compiler: /usr/bin/g++
compiler version: 10.11
Determined that a good number of jobs = 8
Checking for C++ header file fftw3.h... yes
Checking for correct FFTW linkage... 
Error: fftw file failed to link correctly
Check that the correct location is specified for FFTW_DIR

Please fix the above error(s) and rerun scons.
Note: you may want to look through the file INSTALL.md for advice.
Also, if you are having trouble, please check the INSTALL FAQ at 
   https://github.com/GalSim-developers/GalSim/wiki/Installation%20FAQ
'''

'''
wasabi:GalSim % echo $DYLD_LIBRARY_PATH 
/Users/walter/LSST/Anaconda/opt/lsst/GalSim/lib:/Users/walter/LSST/Anaconda/opt/lsst/tmv/lib:/Users/walter/LSST/Anaconda/opt/lsst/fftw/lib:/Users/walter/LSST/Anaconda/opt/lsst/boost/lib
'''

rmjarvis:
Chris, it seems to be losing the DYLD_LIBRARY_PATH from the environment now.  I thought yesterday, it wasn't doing that (we thought because scons wasn't using `/usr/bin/env python`), but now it's gone from both SCons's environment and `os.environ`.  So I think you will need to do either
```
scons DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH
```
or if you prefer
```
scons DYLD_LIBRARY_PATH=$LSST_LIBRARY_PATH
```
It's possible that the LSST lines that we put in there that we thought weren't doing anything actually were working to repopulate the DYLD_LIBRARY_PATH variable, and we just misdiagnosed what was going on under the hood.

jmeyers314:
>Sorry Josh. I think I figured out what I did wrong. Please try it now.

Thanks, that fixed it.  All tests pass.

rmandelb:
I'm feeling even less eager than before about upgrading to El Capitan (or its predecessor).

Anyway - I just confirmed that this doesn't break anything on my Mac (10.9.5) or Linux cluster.

cwwalter:
OK, I can confirm that the first one works.  Thanks.

Do you think we could check for the existence of  LSST_LIBRARY_PATH and set if it is exists?

rmjarvis:
> Do you think we could check for the existence of LSST_LIBRARY_PATH and set if it is exists?

I'm a little averse to having something quite so LSST-stack specific in the normal GalSim operations.  I'd rather have the LSST stack's setup of GalSim use this new DYLD_LIBRARY_PATH option when it installs GalSim if that's possible.  

I think it should be possible for the LSST stack's distro of GalSim to use
```
scons DYLD_LIBRARY_PATH=$LSST_LIBRARY_PATH
```
as its compile command.  (Possibly with other options as well.)  Does this sound reasonable?  Or do you think there is some reason this either won't work or will be annoying in some respect?

cwwalter:
Well, this is really an issue for people compiling their own development version that the cloned from the repo and switching to it instead of the distributed one.  So, its certainly possible to just tell people to compile things this way but it annoying in the sense that I have already forgotten to do it several times tonight alone :)

So, I think for the stack install itself that gets distributed it is no problem to just do it this way. Rather, it is an issue for people trying to compile GalSim in the stack environment.  In that environment the LSST_LIBRARY_PATH will be set.

rmjarvis:
Once you do it once, scons should remember it, so you shouldn't have to keep doing it each time.  Are you recloning or something?  Or deleting the gs_scons.conf file?

cwwalter:
I see, OK.  Yes, I was deleting the file since I was trying to start cleanly in my tests.  It did record it as you say.

So, people will need to know to do it the first time.

rmjarvis:
Right.  I'll add something to our [Installation FAQ](https://github.com/GalSim-developers/GalSim/wiki/Installation%20FAQ) about this.  Also we should put it on [your page about using GalSim with the LSST stack](https://github.com/GalSim-developers/GalSim/wiki/Building-GalSim-with-the-LSST-stack), which I think may need some more general updating.  It looks like this is out of date now.

And if there is another place where we are accumulating wisdom within DESC or LSST generally about running GalSim, we should make sure to mention it there as well.

cwwalter:
So actually this is not just going to be a LSST issue right?  Anyone running El Capitan is going to have this happen.  The only thing special about LSST is that there is a guaranteed work around we could use in the code with the special variable set up to deal with this.

cwwalter:
By anyone I mean anyone with libraries not in the standard area.  That includes macports, fink, etc.

rmjarvis:
Well, mostly anyone running Anaconda on El Capitan.  But in general anyone who is relying on the DYLD_LIBRARY_PATH environment variable for runtime resolution of library locations will hit this.  

The normal Mac way of doing libraries doesn't require DYLD_LIBRARY_PATH for anything, so for my normal setup, it will be fine.  My Enthought python setup also doesn't require it.  Same with fink.  Not sure about macports, since I don't use that.  Anaconda is the only python setup I have on my Mac that needs DYLD_LIBRARY_PATH to be set properly for it to work at run time.

On Linux, it is more common to need the environment variables set properly, although there it is LD_LIBRARY_PATH, not DYLD_LIBRARY_PATH.  However, so far I guess the common Linux varieties aren't doing this SIP thing.

cwwalter:
So..  I think it is the stack part that is mostly an issue here.  Not (mostly?) anaconda.  

/Users/walter/LSST/GalSim/lib:/Users/walter/LSST/Anaconda/opt/lsst/tmv/lib:/Users/walter/LSST/Anaconda/opt/lsst/fftw/lib:/Users/walter/LSST/Anaconda/opt/lsst/boost/lib

shows as being in Anaconda only because that is how the conda version of the stack is distributed.  eups distrib installs would have the same problem.  I'm not sure how it would interact with an anaconda only part.

rmjarvis:
Even without the LSST stack installed, my Anaconda python installation needs to have DYLD_LIBRARY_PATH set correctly for galsim to work properly.  That's the only set up I have that requires it.

rmjarvis:
I added a [FAQ entry](https://github.com/GalSim-developers/GalSim/wiki/Installation-FAQ#are-you-using-el-capitan-or-later) about this.

cwwalter:
OK.  I will try to document this well somewhere else for when it fails for LSST people. I'm a bit worried people aren't going to easily find the solution buried in the FAQ like that.

BTW, @timj points out that the DYLD issue is also important for the unit tests.

rmjarvis:
> BTW, @timj points out that the DYLD issue is also important for the unit tests.

Does `scons tests` work for you?  If not, can you go to the tests directory and run `nosetests`?

If the former fails, but the latter succeeds, I think I know what to do.  But if the latter fails, then Apple really did break everything, and the better solution is to tell people to disable SIP.

timj:
`nosetests` being called directly will work fine (noting that I can't build it at all at the moment because the LSST version doesn't have the fix for #717 in yet) because `nosetests` has an explicit shebang for a non-/usr binary so the environment is not stripped. Running nose from within SCons requires that the `DYLD_LIBRARY_PATH` is included in the executable line (which is what LSST does when running tests) because the Scons environment will not forward through the `DYLD_LIBRARY_PATH` to subproceses.

rmjarvis:
I added the prefix to the nosetests call.  So I think `scons tests` should work now. I also added a message for El Capitan users telling them to do 
```
scons DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH
```
if the build fails.  Let me know if you notice anything else that seems to not work correctly for you.

cwwalter:
Nice on adding the message for El Capitan.

I'm trying to run the tests (which I haven't before) and I am running into some issues which I suspect may be unrelated to the DYLD problem.

after a successful build I do a 'scons tests' and get:

```
run_tests(["tests/tests.log"], ["bin/test_main"])
Using nosetests from:  /Users/Walter/LSST/Anaconda/bin/nosetests
nosetests is version 1.3.7... running tests in parallel with 8 jobs
If you have trouble with the parallel test run, you may disable it by typing:
    scons tests -j1

Starting python tests...
scons: *** [tests/tests.log] No such file or directory
```

cwwalter:
tests/tests.log does exist and has this in it:

```
wasabi:GalSim % cat tests/tests.log 
Nosetests version is: 1.3.7
Nosetests command is:
DYLD_LIBRARY_PATH='/Users/walter/LSST/GalSim/lib:/Users/walter/LSST/Anaconda/opt/lsst/tmv/lib:/Users/walter/LSST/Anaconda/opt/lsst/fftw/lib:/Users/walter/LSST/Anaconda/opt/lsst/boost/lib' /Users/Walter/LSST/Anaconda/bin/nosetests --processes=8 --process-timeout=60
```

rmjarvis:
Hmm..  It looks like it's dying as soon as nosetests starts.  Could you try this command directly and see if that works.  (From in the tests directory)
> DYLD_LIBRARY_PATH='/Users/walter/LSST/GalSim/lib:/Users/walter/LSST/Anaconda/opt/lsst/tmv/lib:/Users/walter/LSST/Anaconda/opt/lsst/fftw/lib:/Users/walter/LSST/Anaconda/opt/lsst/boost/lib' /Users/Walter/LSST/Anaconda/bin/nosetests --processes=8 --process-timeout=60

You could also just try `nosetests` without all the extras (still in the tests directory) to see if that works.  Or /Users/Walter/LSST/Anaconda/bin/nosetests if it's not in your path.

cwwalter:
Yes, from in the tests directory the command works OK and gives only one warning.   I actually had tried it before and got a ton of errors but wasn't in the test directory.

But: (because of the syntax) I needed to go into a bash shell first to execute the command.

I tried doing the 'scons test' from a bash shell too but got the same error.

rmjarvis:
Oh, I bet that's what it is.  I forgot to add the `bash -c` part here.  Give me a minute...

rmjarvis:
OK.  See if that does it.  (And if the warning is about no Image types for uint16, we have a fix for that on another branch that isn't merged in yet.)

cwwalter:
OK that worked.  Yes the warning was for uint16.  Next I get:

```
Starting cpp tests...
dyld: Library not loaded: /usr/local/lib/libgalsim.1.4.dylib
  Referenced from: /Users/walter/LSST/GalSim/bin/test_main
  Reason: image not found
test_main terminated by signal  5

scons: done building targets.
```

timj:
Presumably because you aren't running it as `scons tests DYLD=$DYLD` ?

cwwalter:
No, I don't think so (just checked that didn't fix it).  The unit tests are working.  Also gs_scons.conf has DYLD... in it. I guess what ever step is calling these cpp tests is once again not including the $DLYD= line.

timj:
Ah yes. The C++ test execution will need the explicit DYLD path included.

cwwalter:
BTW just to put this in the record:  Yesterday I asked on HipChat if we could think of transition eups to some scheme built on links instead of environment variables.  Apple doesn't want us to do this and it basically is going to keep causing us pain forever.  Robert said:

```
I think it's possible to get an install from e.g. conda and make play with eups.  The DYLD_LIBRARY_PATH thing from apple is a real pain,  and we should be able to make eups use links instead of DYLD_LIBRARY_PATH.  I went some of the way with a script that builds a soft-link farm based on the environment (hmm, $SCONSUTILS_DIR/bin/makeLinkFarm), and if that gets around the apple issues it'd be easy to add an option to eups to do this directly (it builds the changes to the environment that it knows it needs to make, and then generates the commands on sh/csh but it could easily generate the link commands).
The problem has always been handling cleanup of the per-shell environments.  trap 0 might work, but I'm not sure how watertight it is.
```

rmjarvis:
Yep.  There was another subprocess there.  I added the prefix to that one too.  See if that works now.

cwwalter:
Success :+1: 

```
Ran 343 tests in 44.491s

OK
Nosetests finished successfully.

Starting cpp tests...
Running 12 test cases...

*** No errors detected
test_main finished successfully.

scons: done building targets.
wasabi:GalSim % 
```

rmjarvis:
Awesome.  Thanks for your help on this!

timj:
Great. Now I just need @danielsf to copy a snapshot to the LSST EUPS repo...

cwwalter:
You will also want to hold off until #724 is taken care of.  That has two things in it (as per the title we need things to be lowercase) and also we need some lines added to the table file.  Currently that second part is added by @danielsf 's patch file.

cwwalter:
Umm OK.  I wanted to do an extra careful check.  So I deleted my directory, re-cloned and started from scratch.  I am getting this:

```
wasabi:GalSim % scons DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH
scons: Reading SConscript files ...
SCons is version 2.4.1 using python version 2.7.11
Python is from /Users/Walter/LSST/Anaconda/include/python2.7
Unknown variables: ['DYLD_LIBRARY_PATH']
If you are sure these are right (e.g. you want to set some SCons parameters
that are not in the list of GalSim parameters given by scons -h)
then you can override this check with USE_UNKNOWN_VARS=true
```

I wonder if we had defined things earlier when we were doing a bunch of things manually but that is not set anymore.  Or am I just forgetting something totally simple?

cwwalter:
hmm... hold on it looks this branch is not remote tracking for some reason.  I must have done some stupid git thing... sigh.

rmjarvis:
I think you forgot to go to branch '#721'?

cwwalter:
No I did do that.  But I think I might have done it wrong for some reason I can't imagine... retrying again.

cwwalter:
Ah... I think I typed 'git branch #721' instead of 'git checkout 721' thereby giving a local copy of master with the same name.  Sorry for the noise.

rmjarvis:
I think this is all confirmed to be working now, right?  Are we ready to merge this?

cwwalter:
Yes, this should be ready.  

To really use this with eups etc #724 still needs to be addressed including the text in the comments about adding the two line into table file in addition to lower casing the file name.

PR #726, opened 2016-03-15
Modernize Python usage

timj:
+ Print function
+ Absolute import
+ Modern exceptions

This should simplify a port to Python 3 but step 1 is to get to Python 2.7.


rmjarvis:
Thanks Tim.  However, our plan was to wait until a number of outstanding branches get merged into master before doing this to make the merge process simpler (especially with regards to the print function, which is pretty pervasive).  Then we'll do all these kinds of changes before releasing the next version, 1.4.  cf. discussion on issue #534.

rmjarvis:
Or, never mind.  I guess the conversation @rmandelb and I had about this wasn't on the issue page.  Anyway, we were planning to do this once the current swath of PRs are resolved, and before we do the next version release.

timj:
Thanks, so long as the answer wasn't "we have to support python 2.4" then I'm fine. LSST are moving to Python 3 in the summer (at least data management are). It was a few minutes work to generate this PR so I can do it again any time.

rmjarvis:
OK.  Good.  The plan is to support both 2.7 and 3.4+ using the normal tricks of using six where necessary and try/except blocks.  All the stuff you did here will certainly be included in that.

timj:
Ok. Regarding `six`, I don't like it much. I've tended towards `future` and that's what we are planning to use at LSST.

rmjarvis:
OK.  Thanks for the advice.  I don't have a lot of experience with either yet, so if I get stuck on something I'll ping you.  For TreeCorr, it was all pretty straightforward.  I think the only place I used six was for `iteritems`, which looks like it's also in future.utils, so either way I guess.

Issue #732, opened 2016-03-29
bug in correlated noise in config


rmandelb:
When I try to use COSMOSNoise, and set the `file_name` attribute, I was getting an exception that I didn't understand.  To try to track this down, I modified demo1.yaml so that the image.noise now looks like this:

```
    noise : 
        type : COSMOS
        variance: 900
        file_name: my_correlation.fits
```

I think this is legitimate use of image.noise, but here is the exception that gets thrown:

```
Traceback (most recent call last):
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/site-packages/galsim/config/process.py", line 840, in MultiProcess
    result = job_func(**kwargs)
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/site-packages/galsim/config/output.py", line 222, in BuildFile
    data = builder.buildImages(output, config, file_num, image_num, obj_num, ignore, logger)
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/site-packages/galsim/config/output.py", line 427, in buildImages
    image = galsim.config.BuildImage(base, image_num, obj_num, logger=logger)
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/site-packages/galsim/config/image.py", line 229, in BuildImage
    image = builder.buildImage(cfg_image, config, image_num, obj_num, logger)
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/site-packages/galsim/config/image.py", line 423, in buildImage
    base, obj_num=obj_num, xsize=xsize, ysize=ysize, do_noise=True, logger=logger)
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/site-packages/galsim/config/stamp.py", line 395, in BuildStamp
    im, current_var = builder.addNoise(stamp,config,im,skip,current_var,logger)
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/site-packages/galsim/config/stamp.py", line 836, in addNoise
    current_var = galsim.config.AddNoise(base,image,current_var,logger)
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/site-packages/galsim/config/noise.py", line 92, in AddNoise
    var = builder.addNoise(noise, config, im, rng, current_var, draw_method, logger)
[...]

PR #733, opened 2016-03-29
#732- quick correlated noise fix

rmandelb:
This is a one-line fix for the bug noted in #732.  (I also mentioned it in CHANGELOG because in v1.3 it looks like the same error would have occurred.)  

Should I add any unit tests, or is this all good?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ API Changes
 Bug Fixes
 ---------
 
+- Fixed bug in config that did not allow users to pass in a filename for
+  COSMOS (correlated) noise.  (#732)
 - Improved ability of galsim.fits.read to handle invalid but fixable FITS
   headers. (#602)
 - Fixed bug in des module related to building meds file with wcs taken from
--- a/galsim/config/image.py
+++ b/galsim/config/image.py
@@ -121,6 +121,9 @@ def SetupConfigImageNum(config, image_num, obj_num):
     if not isinstance(image, dict):
         raise AttributeError("config.image is not a dict.")
 
+    if 'file_num' not in config:
+        config['file_num'] = 0
+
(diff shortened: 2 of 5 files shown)


rmjarvis:
This is fine.  I noticed that the `file_name` is listed as a required parameter here, but it's not actually required anymore (now that we have it in the share directory), so I put it in the `opt` list rather than `req`.

If you feel up to it, it would probably be worth having a quick unit test for this, since it's not in any demo script anymore (which serves as unit tests for most of the config stuff).

rmandelb:
Yeah, I agree; was just being lazy.  I tried to make one in https://github.com/GalSim-developers/GalSim/pull/733/commits/696e7075b6448205a9f586acb9bdb191d1c6b394 but I'm getting a somewhat obscure error that I don't understand (in the config innards).  Can you take a look?  If you can get it to run past the point where it currently breaks, then I can sort out any bugs that come in later on that make it not work...

rmjarvis:
Fixed.

rmandelb:
Awesome, thanks.  And the new test passes now, but there's an error when running it on the old version of the code, so that's good.

Do you have any objections to my merging this without another review, given that it's relatively minor?
Should we cherry-pick this onto releases/v1.3 given that it's a bug fix for a problem that was present already in that release?

rmjarvis:
I'm happy to merge now.

And if you want, I can cherry pick this onto the 1.3 branch.  I'm making a 1.3.2 for Scott anyway, since he wants the El Capitan fixes on an official release to use for LSST.  Easy enough to include this too if you want.

rmandelb:
Okay.  I'm merging now.  Please do go ahead and cherry pick onto 1.3, once you're making a 1.3.2 anyway.  Thanks!

Issue #723, opened 2016-03-12
better `__repr__` for Image and deviates


jmeyers314:
The current GalSim policy for having `eval(repr(obj))` being able to recreate any object is great for interrogating most types, but cumbersome for images and random deviates, where the repr string will quickly explode a terminal screen.  This issue is to propose that we truncate the repr for Image and deviate types so they don't totally fill the screen, (although still printing the first few array elements, similar to `ndarray.__repr__,` would still be useful for debugging).  The python standard library has a repr module that may be helpful for this.

rmjarvis:
I agree.  I'm totally fine with backing off on these two cases where the fully eval-able repr is more annoying than helpful.

jmeyers314:
The simplest way to redo `Image.__repr__` is to just let the ndarray repr itself:

```
    def __repr__(self):
        return 'galsim.Image(bounds=%r, \n%r, wcs=%r)'%(
                self.bounds, self.array, self.wcs)
```

It looks like this is going to break `Image.__hash__`, which just returns the hash of the repr string.  However, since they're mutable, I think it might have been a mistake to make Images hashable in the first place.  Quoting from the [python data model](https://docs.python.org/2/reference/datamodel.html#object.__hash__):

>If a class defines mutable objects and implements a `__cmp__()` or `__eq__()` method, it should not implement `__hash__()`, since hashable collection implementations require that a object’s hash value is immutable (if the object’s hash value changes, it will be in the wrong hash bucket).

rmjarvis:
Sounds good.  The other mutable objects are deviates and your new phase generators.  Should double check that they don't have a hash function either.

jmeyers314:
Does `SBInterpolatedImage::SBInterpolatedImageImpl::repr()` ever get called?  It looks like `repr(interp_image._sbii)` gets dispatched to `_galsim.SBInterpolatedImage.__repr__` in `interpolated image.py`, but maybe there's another path to the method?

rmjarvis:
> Does SBInterpolatedImage::SBInterpolatedImageImpl::repr() ever get called?

Yes.  SBProfile's repr is a virtual method, so `SBProfile::repr()` will track down to this one if the python later doesn't actually know that a given sbprofile is really an sbinterpolatedimage.  However, if it does know (which InterpolatedImage does), then it can do it directly in python and I think look a little nice if I recall.

jmeyers314:
[...]

PR #734, opened 2016-03-30
#723 better repr

jmeyers314:
This PR is for a better `__repr__` for Images, InterpolatedImages, and BaseDeviates.  In this case, better just means shorter.  For Images and InterpolatedImages, this means using `ndarray.__repr__` and subsequently whatever options for array length the user sets with numpy.set_printoptions().  For BaseDeviates, I shortened the random state to include just the first 3 and last 3 integers (similar to the behavior of `ndarray.__repr__`).  In the c++ layer, I rewrote the SBInterpolatedImage and SBInterpolatedKImage reprs to mimic `ndarray.__repr__` with the default printoptions.

These changes required rewriting the above classes' `__eq__` and `__hash__` methods, since the old ones depended on `__repr__`.  While updating these, I went ahead and updated all GSObjects and ChromaticObjects `__eq__` and `__hash__` methods.  The new versions don't depend on `__repr__`, but instead compare object attributes directly, so should be somewhat quicker.  

I also added a number of new tests, mostly checking that `__eq__` doesn't incorrectly return True for objects that should compare False.  I also checked that there are no hash collisions for these test objects (there weren't).

At first, I tried to make `__eq__` respect the commutativity and associativity of Sums and Convolutions, but ran into problems with photon shooting.  I believe (but didn't check) that the issue is that Sum(a, b) first draws deflections from profile a followed by deflections from profile b, so that Sum(a, b) will produce a different photon-shooting image than Sum(b, a) even when the random number generators are synchronized.  In the end I reverted back to Sum(a, b) != Sum(b, a), though I'd be curious if anyone had any ideas to make commutativity and associativity work.

Finally, I also made a number of PEP8 changes, which in hindsight, I probably should have made on a different branch since they muddy the diffs a bit here on some of the changed files.  I'll create a new issue for this, but I think it would be nice if our python files satisfied something like [flake8](https://pypi.python.org/pypi/flake8), presumably with some specific "ignore-this-error" settings.


rmjarvis:
Hmm.  There's a lot about the pep8 style that I don't like, so I'm afraid I have issues with a lot of the style changes you made here.  I find their whitespace requirements in particular way too heavy handed, often making things less readable, not more.  I'll point them out in code, but I kind of wish you had at least made the style changes separate commits rather than intertwining them with the code changes you did here.

galsim/bandpass.py
```diff
@@ -93,13 +94,13 @@ def __init__(self, throughput, blue_limit=None, red_limit=None, wave_type='nm',
         # `wave_list` is normally constructed (see `Bandpass.__mul__` below)
 
         self._orig_tp = throughput  # Save this for pickling.
-        self._tp = _tp              # This will normally become orig_tp turned into an actual 
-                                    # function (see _initialize_tp()), although in some cases,
-                                    # it can be supplied directly as a constructor argument.
-
```
rmjarvis: First example, this comment is eminently more readable than what you changed it too.  Code on the left, comments on the right is a nice paradigm.  So let's ignore the flake8 proscription against hanging indented comments.

galsim/bandpass.py
```diff
@@ -229,15 +229,15 @@ def __mul__(self, other):
...
         elif isinstance(self._tp, galsim.LookupTable):
-            # If other is not a function, then there is no loss of accuracy by applying the 
+            # If other is not a function, then there is no loss of accuracy by applying the
             # factor directly to the LookupTable, if that's what we are using.
             # Make sure to keep the same properties about the table, wave_type.
             if self.wave_factor == 10.0:
                 wave_type = 'Angstroms'
             x = self._tp.getArgs()
-            f = [ val * other for val in self._tp.getVals() ]
+            f = [val * other for val in self._tp.getVals()]
```
rmjarvis: Here's another one.  I don't understand why pep8 doesn't like spaces at the start and end of lists.  It makes them much easier to read with the initial space after the `[` or `(` or `{`.

galsim/base.py
```diff
@@ -1462,10 +1462,13 @@ def drawKImage(self, re=None, im=None, nx=None, ny=None, bounds=None, scale=None
...
-    # Quick and dirty.  Just check reprs are equal.
-    def __eq__(self, other): return repr(self) == repr(other)
+    def __eq__(self, other):
+        if repr(self) != repr(other):
+            return False
+        return self.SBProfile == other.SBProfile
+
     def __ne__(self, other): return not self.__eq__(other)
-    def __hash__(self): return hash(repr(self))
+    def __hash__(self): return hash((repr(self), self.SBProfile.serialize()))
```
rmjarvis: Do you need the repr here now that you have serialize?  I would have thought making the hash using just serialize would be sufficient.

galsim/chromatic.py
```diff
@@ -118,9 +118,10 @@ def __init__(self, obj):
         elif isinstance(obj, ChromaticObject):
             self.separable = obj.separable
             self.wave_list = obj.wave_list
-            if self.separable: self.SED = obj.SED
+            if self.separable:
+                self.SED = obj.SED
```
rmjarvis: Not a fan of this rule either, but I guess I can live with it.  Sometimes a one-liner is just nicer.

galsim/chromatic.py
```diff
@@ -166,15 +167,21 @@ def _fiducial_profile(self, bandpass):
...
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+    def __hash__(self):
+        return hash(("galsim.ChromaticObject", self.obj))
+
     def __repr__(self):
-        return 'galsim.ChromaticObject(%r)'%self.obj
+        return 'galsim.ChromaticObject(%r)' % self.obj
```
rmjarvis: This seems gratuitous to me.  But I guess I can manage.

galsim/chromatic.py
```diff
@@ -643,18 +654,17 @@ def rotate(self, theta):
 
         @returns the rotated object.
         """
-        import math
         if hasattr(theta, '__call__'):
             def buildRMatrix(w):
                 sth, cth = theta(w).sincos()
                 R = np.array([[cth, -sth],
-                              [sth,  cth]], dtype=float)
+                              [sth, cth]], dtype=float)
```
rmjarvis: This one is ridiculous.  We used to have a nicely aligned matrix, but you changed it to be unaligned to satisfy some one-space-only rule.  We have to ignore that one and revert this change.

galsim/chromatic.py
```diff
@@ -679,10 +689,14 @@ def transform(self, dudx, dudy, dvdx, dvdy):
...
-            if not hasattr(dvdx, '__call__'): _dvdx = lambda w: dvdx
-            if not hasattr(dvdy, '__call__'): _dvdy = lambda w: dvdy
+            if not hasattr(dudx, '__call__'):
+                _dudx = lambda w: dudx
+            if not hasattr(dudy, '__call__'):
+                _dudy = lambda w: dudy
+            if not hasattr(dvdx, '__call__'):
+                _dvdx = lambda w: dvdx
+            if not hasattr(dvdy, '__call__'):
+                _dvdy = lambda w: dvdy
```
rmjarvis: Here's an example where the one-liners are more readable.  Should revert this change.

galsim/chromatic.py
```diff
@@ -721,16 +735,17 @@ def shift(self, *args, **kwargs):
                     args[0](700.).x
                     # If the function returns a Position, recast it as a function returning
                     # a numpy array.
+
                     def offset_func(w):
                         d = args[0](w)
-                        return np.asarray( (d.x, d.y) )
+                        return np.asarray((d.x, d.y))
```
rmjarvis: Here's another case where the extra spaces add clarity.  With the spaces, it is much clearer that you are making an array from a tuple.  It's harder to notice what you are doing with the parentheses right up against each other.

galsim/chromatic.py
```diff
@@ -804,50 +819,60 @@ def __init__(self, obj, waves, oversample_fac=1.0):
...
         # to be saved for later, unlike the images.
-        objs = [ obj.evaluateAtWavelength(wave) for wave in self.waves ]
+        objs = [obj.evaluateAtWavelength(wave) for wave in self.waves]
 
         # Check the fluxes for the objects.  If they are unity (within some tolerance) then that
         # makes things simple.  If they are not, however, then we have to reset them to unity, and
         # modify the SED attribute, which now refers to the total flux at a given wavelength after
         # integrating over the whole light profile.
-        fluxes = np.array([ obj.getFlux() for obj in objs ])
+        fluxes = np.array([o.getFlux() for o in objs])
```
rmjarvis: Is this a rule?  That you can't use `obj` as a variable?  (a) Let's ignore that rule.  (b) `o` is a much worse variable name.

galsim/chromatic.py
```diff
@@ -1250,23 +1298,23 @@ class ChromaticTransformation(ChromaticObject):
     @param gsparams         An optional GSParams argument.  See the docstring for GSParams for
                             details. [default: None]
     """
-    def __init__(self, obj, jac=np.identity(2), offset=(0,0), flux_ratio=1., gsparams=None):
-        if isinstance(offset,galsim.PositionD) or isinstance(offset, galsim.PositionI):
+    def __init__(self, obj, jac=np.identity(2), offset=(0, 0), flux_ratio=1., gsparams=None):
```
rmjarvis: Unusual, but in this case, leaving out the space is probably clearer.

galsim/chromatic.py
```diff
@@ -1587,11 +1680,10 @@ def drawImage(self, bandpass, image=None, integrator='trapezoidal', **kwargs):
         add_to_image = kwargs.pop('add_to_image', False)
         # Use given add_to_image for the first one, then add_to_image=False for the rest.
         image = self.objlist[0].drawImage(
-                bandpass, image=image, add_to_image=add_to_image, **kwargs)
+            bandpass, image=image, add_to_image=add_to_image, **kwargs)
```
rmjarvis: In most cases, I prefer 8 spaces for a line continuation to better differentiate it from a new code block.  I think pep8 has the schizophrenic rule that line continuations should be 4 unless the next line is an intended code block, in which case it should be 8.  But having the general rule of 8 for all continuations seems cleaner.  In any case, I wouldn't want to proscribe 8-space indents such as this.

galsim/compound.py
```diff
@@ -384,24 +392,36 @@ def __init__(self, *args, **kwargs):
...
         # Then finally initialize the SBProfile using the objects' SBProfiles.
-        SBList = [ obj.SBProfile for obj in args ]
+        SBList = [obj.SBProfile for obj in args]
         sbp = galsim._galsim.SBConvolve(SBList, real_space, gsparams)
         galsim.GSObject.__init__(self, sbp)
         if noise is not None:
             self.noise = noise
 
     @property
-    def obj_list(self): return self._obj_list
```
rmjarvis: Another type of case where the one liner is clearer.

galsim/image.py
```diff
@@ -382,8 +382,8 @@ def __init__(self, *args, **kwargs):
             self.wcs = wcs
 
     def __repr__(self):
-        return 'galsim.Image(bounds=%r, array=array(%r, dtype=%s), wcs=%r)'%(
-                self.bounds, self.array.tolist(), self.array.dtype, self.wcs)
+        return 'galsim.Image(bounds=%r, \n%r, wcs=%r)'%(
```
rmjarvis: You should still keep the `array=` prelude to the numpy array.

galsim/interpolatedimage.py
```diff
@@ -683,13 +725,31 @@ def __init__(self, real_kimage, imag_kimage, k_interpolant=None, stepk=None,
...
     def __repr__(self):
-        return ('galsim.InterpolatedKImage(%r, %r, %r, stepk=%r, gsparams=%r)')%(
+        return ('galsim.InterpolatedKImage(\n%r,\n%r,\n%r, stepk=%r, gsparams=%r)') % (
             self._real_kimage, self._imag_kimage, self.k_interpolant,
             self._stepk, self._gsparams)
 
-    def __str__(self): return 'galsim.InterpolatedKImage(real_kimage=%s)'%(
-            self._real_kimage)
+    def __str__(self):
+        return 'galsim.InterpolatedKImage(real_kimage=%s)' % self._real_kimage
```
rmjarvis: Seems like an oversight (on my part no doubt) not to list imag_kimage here too.  Probably ought to add it.

galsim/random.py
```diff
@@ -133,10 +133,19 @@
 next output value.
 """
 
-# Quick and dirty.  Just check reprs are equal.
-_galsim.BaseDeviate.__eq__ = lambda self, other: repr(self) == repr(other)
+def _gs_BD__eq__(self, other):
```
rmjarvis: Can we please use a less obfuscatory name here?  Say `BaseDeviate_eq` maybe?

galsim/random.py
```diff
@@ -133,10 +133,19 @@
...
+    if repr(self) != repr(other):
+        return False
+    return self.serialize() == other.serialize()
+_galsim.BaseDeviate.__eq__ = _gs_BD__eq__
 _galsim.BaseDeviate.__ne__ = lambda self, other: not self.__eq__(other)
-_galsim.BaseDeviate.__hash__ = lambda self: hash(repr(self))
+_galsim.BaseDeviate.__hash__ = None
+
+def _gs_BD_seed_repr(self):
```
rmjarvis: likewise.

galsim/sed.py
```diff
@@ -146,12 +147,12 @@ def _initialize_spec(self):
         # Do some SED unit conversions to make internal representation proportional to photons/nm.
         # Note that w should have units of nm below.
         c = 2.99792458e17  # speed of light in nm/s
-        h = 6.62606957e-27 # Planck's constant in erg seconds
+        h = 6.62606957e-27  # Planck's constant in erg seconds
```
rmjarvis: These comments should be aligned.  If you (or flake8) feel the need to add a space here, add it before the `c` comment too.

galsim/table.py
```diff
@@ -20,6 +20,8 @@
 addition of the docstring and few extra features.
 """
 from . import _galsim
+import galsim
```
rmjarvis: It's generally not a good idea to import the base module from a submodule, since galsim in turn imports things from table.py, which can lead to circular dependencies.  But in this case, I don't see where you even use this.  So probably can just remove this line.

galsim/transform.py
```diff
@@ -20,7 +20,7 @@
 """
 
 import galsim
-import numpy
+import numpy as np
```
rmjarvis: I may be outvoted on this one, but I kind of hate this tendency of people to redefine modules with `import` as `some-two-character-abbreviation`.  `np` for `numpy` is pretty common, so it's not particularly unclear.  But in general I think it's worth the extra typing (only three characters in this case!) to be explicit about the real module name.

src/Random.cpp
```diff
@@ -85,13 +88,43 @@ namespace galsim {
         clearCache();
     }
 
+    // Next two functions shamelessly stolen from
+    // http://stackoverflow.com/questions/236129/split-a-string-in-c
+    std::vector<std::string> &split(const std::string &s, char delim,
```
rmjarvis: I normally wouldn't mention this, but since we're discussing style here, I generally prefer the space to be after the &.  The return type here is `std::vector<std::string>&`.  So it makes things clearer to connect the `&` there rather than to adjoin it to the function name.  Likewise with the argument types.  No need to change it though.  Just putting that out there...  :)

src/SBAdd.cpp
```diff
@@ -43,6 +43,19 @@ namespace galsim {
...
+        oss << "galsim._galsim.SBAdd([";
+        ConstIter sptr = _plist.begin();
+        oss << sptr->serialize();
+        for (++sptr; sptr!=_plist.end(); ++sptr) oss << ", " << sptr->serialize();
+        oss << "], galsim.GSParams("<<*gsparams<<"))";
+        return oss.str();
+    }
+
+    // We can't rely on SBProfileImpl::repr for SBAdd b/c it will end up calling the
+    // summands serialize method instead of repr method.  So we override here.
```
rmjarvis: I don't understand this.  Isn't serialize the same as repr for the SB types?  So why do we need this distinction?

src/SBInterpolatedImage.cpp
```diff
@@ -426,17 +433,101 @@ namespace galsim {
...
+        oss << "galsim.Interpolant('"<<xinterp->makeStr()<<"', "<<xinterp->getTolerance()<<"), ";
+        oss << "galsim.Interpolant('"<<kinterp->makeStr()<<"', "<<kinterp->getTolerance()<<"), ";
+
+        oss << "1., "<<stepK()<<", "<<maxK()<<", galsim.GSParams("<<*gsparams<<"))";
+        return oss.str();
+    }
+
+    // Alternative to repr that's more annoying for a human to read, but can be used to *exactly*
+    // reconstruct the SBInterpolatedImage object when pickling.
+    std::string SBInterpolatedImage::SBInterpolatedImageImpl::serialize() const
```
rmjarvis: Maybe this is why you overrode the Add, Convolve, etc. reprs.  I'm not sure that we need the reprs for SBInterpolatedImage to be particularly small.  It's extremely rare for that to ever get triggered by someone.  Certainly not by accident.  We do our best to keep the C++ classes well hidden from the typical user.  So I think I'd prefer to just have `serialize` for all of these and have `SBProfile` use that for the repr.

tests/galsim_test_helpers.py
```diff
@@ -262,14 +267,114 @@ def do_pickle(obj1, func = lambda x : x):
...
+            else:
+                # sys.stderr.write("Unknown type: {}\n".format(args[i]))
+                continue
+            try:
+                obj6 = eval('galsim.' + classname + repr(tuple(newargs)))
+            except:
+                try:
+                    obj6 = eval('galsim._galsim.' + classname + repr(tuple(newargs)))
+                except:
+                    pass
```
rmjarvis: This bit is very clever.  However, I'm a little concerned about this line.  What kinds of objects didn't work for constructing `obj6`?

jmeyers314:
>Hmm. There's a lot about the pep8 style that I don't like, so I'm afraid I have issues with a lot of the style changes you made here. 

Right.  I have my own ignore list for flake8 too.  It just looks a bit different than yours.  For the record, here's my list:

- C901 function is too complex
- E125 continuation line with same indent as next logical line
- E129 visually indented line with same indent as next logical line
- E226 missing whitespace around arithmetic operator
- E261 at least two spaces before inline comment
- E402 module level import not at top of file
- E731 do not assign a lambda expression, use a def

and here's the list you're proposing to add:

- E116 unexpected indentation (comment)
- E126 continuation line over-indented for hanging indent
- E201 whitespace after '('
- E202 whitespace before ')'
- E228 missing whitespace around modulo operator
- E231 missing whitespace after ‘,’
- E241 multiple spaces after ‘,’
- E701 multiple statements on one line (colon)

I’ll note that 126 and 241 are “standard”, and I think a few of your objections are probably special cases and not general rules.

The remaining guidelines I think I address most often in GalSim are:

- E251 unexpected spaces around keyword / parameter equals
- E301 expected 1 blank line, found 0
- E302 expected 2 blank lines, found 0
- E303 too many blank lines

Mostly, I’d just like to have the style more consistent throughout GalSim, but I’m not overly attached to what that style should be.

galsim/chromatic.py
jmeyers314: The problem here was that the `obj` in the listcomp was overwriting the `obj` in the method argument list.  I'll change the method argument to `original` and then `o` back to `obj`.

galsim/bandpass.py
jmeyers314: I'm not so attached to this one so I'll revert it (and any other similar cases).  I think the argument against is maintainability.  If you change the code or the comment, then it's (slightly) more work to realign than if the block comment occupied separate lines.

galsim/bandpass.py
jmeyers314: I have to admit, this one still looks weird to me with the extra spaces, but I'll revert.

galsim/chromatic.py
jmeyers314: No problem.  Will revert.

galsim/chromatic.py
jmeyers314: I like this one so I think I'll keep it unless others chime in.

galsim/chromatic.py
jmeyers314: Sure.  Reverted.

galsim/chromatic.py
jmeyers314: Done.

galsim/chromatic.py
jmeyers314: Done.

galsim/chromatic.py
```diff
@@ -1542,12 +1627,20 @@ def __init__(self, *args, **kwargs):
...
-        return 'galsim.ChromaticSum(%r, gsparams=%r)'%(self.objlist, self.gsparams)
+        return 'galsim.ChromaticSum(%r, gsparams=%r)' % (self.objlist, self.gsparams)
 
     def __str__(self):
-        str_list = [ str(obj) for obj in self.objlist ]
-        return 'galsim.ChromaticSum([%s])'%', '.join(str_list)
+        str_list = [str(obj) for obj in self.objlist]
+        return 'galsim.ChromaticSum([%s])' % ', '.join(str_list)
 
     def evaluateAtWavelength(self, wave):
```
jmeyers314: Here's a place (just below) where we're inconsistent with the spaces around the listcomp.

galsim/chromatic.py
jmeyers314: Sure.  Reverting.

src/Random.cpp
jmeyers314: Easiest one to change in the bunch!  Done.

jmeyers314:
@rmjarvis, I think I've finished reverting the style changes you highlighted (except for the % operator).

I'd still like to make the remaining changes a bit more standard throughout GalSim, but we can save that for another issue.

galsim/base.py
jmeyers314: >Do you need the repr here now that you have serialize? I would have thought making the hash using just serialize would be sufficient.

I guess this depends on if you want `galsim.Gaussian(fwhm=1)` and `galsim.GSObject(galsim.Gaussian(fwhm=1))` to be equal or not.  Since they have different methods, I'd vote they should be unequal.  I'll change `repr(self)` to `"galsim.GSObject"` though to be more consistent with the rest of the hashes, and update `__eq__` to be more consistent too.

galsim/image.py
jmeyers314: Yep.  Done.

galsim/interpolatedimage.py
jmeyers314: Since imag_kimage has to have the same metadata has real_kimage, I don't think we lost much without it, but I guess it's good to remind users that both are required for InterpolatedKImage.  Fixed.

galsim/random.py
jmeyers314: Done.

galsim/random.py
jmeyers314: and done.

galsim/table.py
jmeyers314: Done.

galsim/transform.py
jmeyers314: I think this one is standard enough to leave.  Mostly though, I'd just like this to be consistent so when I write new code I don't have to lookup if it's going to be `numpy` or `np`.

galsim/random.py
jmeyers314: Actually, I just went ahead and removed the repr dependencies for deviates too.

tests/galsim_test_helpers.py
jmeyers314: Looks like it was just SBInterpolatedImage and SBInterpolatedKImage, which failed here since the `repr(tuple(newargs))` bit ends up using `ndarray.__repr__()` and the output gets truncated.  I've now wrapped this in a printoptions context to turn it into an actual test.

src/SBInterpolatedImage.cpp
jmeyers314: Okay.  This is hard enough to figure out how to trigger on purpose, so it's fair to let it spit out the full repr.  Fixed.

jmeyers314:
Okay @rmjarvis , I think I've hit all your comments now.  Sorry again about the mess, I'll try to atomize my commits better in the future.

CHANGELOG.md
```diff
@@ -33,6 +33,8 @@ Bug Fixes
   where the resulting object did not set the index attribute properly. (#694)
 - Fixed an error in the `CCDNoise.getVariance()` function, as well as some
   errors in the documentation about the units of CCDNoise parameters. (#713)
+- Fixed a bug where InterpolatedKImage incorrectly checked that its input real
```
rmjarvis: The grammar of this sounds like it used to check that the image was Hermitian, but that that check was incorrect, so now it doesn't.  That's now what you meant though.  (It was an and/or mixup.)  Maybe
"Fixed a bug in how InterpolatedKiImage checked for properly Hermitian input images."

galsim/chromatic.py
```diff
@@ -1677,12 +1758,15 @@ def __init__(self, *args, **kwargs):
...
-            if not obj.separable and not isinstance(obj, galsim.ChromaticSum): n_nonsep += 1
-            if isinstance(obj, InterpolatedChromaticObject): n_interp += 1
-        if n_nonsep>1 and n_interp>0:
+        # n_nonsep = 0
+        # n_interp = 0
+        # for obj in self.objlist:
+        #     if not obj.separable and not isinstance(obj, galsim.ChromaticSum): n_nonsep += 1
+        #     if isinstance(obj, InterpolatedChromaticObject): n_interp += 1
+
+        n_nonsep = sum(o.separable and not isinstance(o, galsim.ChromaticSum) for o in self.objlist)
```
rmjarvis: This still has `o` as a variable.  IMO, `o` and `l` should never be used as variables.

galsim/compound.py
```diff
@@ -734,23 +768,32 @@ def __init__(self, obj, real_space=None, gsparams=None):
...
-                self.orig_obj, self.real_space, self._gsparams)
+        return 'galsim.AutoCorrelation(%r, real_space=%r, gsparams=%r)' % (
+            self.orig_obj, self.real_space, self._gsparams)
 
     def __str__(self):
-        s = 'galsim.AutoCorrelate(%s'%self.orig_obj
-        if self.real_space:
-            s += ', real_space=True'
+        s = 'galsim.AutoCorrelate(%s' % self.orig_obj
+        if self.real_space: s += ', real_space=True'
```
rmjarvis: This on the other hand, is not a place where the one liner is clearer.  This one should be two lines.  Mostly, you should just trust the original author's choice. That's most often going to be a good, readable style.  In this case, this was originally two lines, and IMO it should stay that way.

galsim/interpolatedimage.py
```diff
@@ -638,24 +679,24 @@ def __init__(self, real_kimage, imag_kimage, k_interpolant=None, stepk=None,
...
 
         # Check for Hermitian symmetry properties of real_kimage and imag_kimage
         shape = real_kimage.array.shape
         # If image is even-sized, ignore first row/column since in this case not every pixel has
         # a symmetric partner to which to compare.
-        bd = galsim.BoundsI(real_kimage.xmin + (1 if shape[1]%2==0 else 0),
+        bd = galsim.BoundsI(real_kimage.xmin + (1 if shape[1] % 2 == 0 else 0),
                             real_kimage.xmax,
-                            real_kimage.ymin + (1 if shape[0]%2==0 else 0),
+                            real_kimage.ymin + (1 if shape[0] % 2 == 0 else 0),
```
rmjarvis: Here's another case where the original was better.  Leaving out the spaces in `shape[0]%2==0` helps the reader easily see what's going on with the `if`/`else`.  Adding in spaces actually makes harder to parse the logical parts of this line.  Again, I wish you could just trust the style sensibilities of the original coder rather than some black box linter that completely ignores the first rule in pep8: "A Foolish Consistency is the Hobgoblin of Little Minds".  

Flake8 doesn't know what to do with that rule, so it just skips it.  But it's actually the most important rule in the whole thing.  And it's why I can't stand style linters.  Code linters (like pyflake) are fine and helpful.  But flake8 is not as far as I'm concerned.  I haven't seen a single style change you made here to satisfy flake8 that makes anything clearer, and I've seen lots that make things just a little bit worse.  I haven't commented on most of them, because they just aren't a big enough deal for me to bother, but there are at least a dozen or two more that I think are better the way they were for reasons similar to this one.

galsim/shapelet.py
```diff
@@ -125,31 +128,29 @@ def __init__(self, sigma, order, bvec=None, gsparams=None):
...
     @property
     def bvec(self): return self.getBVec()
 
-    def getPQ(self,p,q):
-        return self.SBProfile.getBVec().getPQ(p,q)
-    def getNM(self,N,m):
-        return self.SBProfile.getBVec().getPQ((N+m)/2,(N-m)/2)
+    def getPQ(self, p, q): return self.SBProfile.getBVec().getPQ(p, q)
+
+    def getNM(self, N, m): return self.SBProfile.getBVec().getPQ((N+m)/2, (N-m)/2)
```
rmjarvis: Again, these don't belong as one-liners.  (unlike the properties)  Leave these the way they were please.

galsim/table.py
```diff
@@ -213,34 +216,57 @@ def isLogF(self):
...
 _galsim._LookupTable.__getinitargs__ = lambda self: \
-        (self.getArgs(), self.getVals(), self.getInterp())
-_galsim._LookupTable.__repr__ = lambda self: \
-        'galsim._galsim._LookupTable(array(%r), array(%r), %r)'%(
-            self.getArgs(), self.getVals(), self.getInterp())
-_galsim._LookupTable.__eq__ = lambda self, other: repr(self) == repr(other)
+    (self.getArgs(), self.getVals(), self.getInterp())
+
+
+def _gs_LT_repr(self):
```
rmjarvis: Please give this a less-obfuscated name.  And the eq and hash.

galsim/transform.py
rmjarvis: I think this is the only good argument I've heard for any of these style changes.  It's a fair point to have this be consistent in the code so that we know which one to type.  I prefer `numpy` but both you and Rachel both seem to prefer `np`, so I'm fine with make that a standard throughout GalSim.

src/Random.cpp
```diff
@@ -85,13 +88,43 @@ namespace galsim {
...
+        return elems;
+    }
+
+    std::vector<std::string> split(const std::string& s, char delim) {
+        std::vector<std::string> elems;
+        split(s, delim, elems);
+        return elems;
+    }
+
+    std::string seedstring(std::vector<std::string> seed) {
```
rmjarvis: Just noticed that this is passed by value.  The argument should be a const reference to avoid the gratuitous copy.

CHANGELOG.md
jmeyers314: Sure. Changed.

src/Random.cpp
jmeyers314: Sure.  Done.

galsim/interpolatedimage.py
jmeyers314: Alright.  You've convinced me to drop the flake8.  I think I've reverted most of the style changes now.

jmeyers314:
Alright.  Third time's the charm.  I think this is ready to look at again.

I also went ahead and made the "import numpy" -> "import numpy as np" switch in a few more places where the actually numpy code was far from the import statement.  I left alone all the cases where the import was close to the code though.

rmjarvis:
Looks good to me now.  Thanks Josh.  And sorry being a little acerbic on the whole style thing.  I tried flake8 myself some time back and just got more and more annoyed with it for recommending changes that I thought made things look worse, so I gave it up and now have pyflake as the running linter in my vim sessions.

You should of course still feel entitled to make style-related changes in the code if you think things will be clearer or more legible, especially on code you're actively working on.  And also definitely make style suggestions on PRs.  I think that's a good place to give each other feedback on coding style.  I just don't want to conform to a black box is all.  :)

galsim/base.py
```diff
@@ -1462,10 +1462,12 @@ def drawKImage(self, re=None, im=None, nx=None, ny=None, bounds=None, scale=None
 
         return re,im
 
-    # Quick and dirty.  Just check reprs are equal.
-    def __eq__(self, other): return repr(self) == repr(other)
+    def __eq__(self, other):
```
rmandelb: Why do we check only these things and not, say, the noise attribute (if it has one)?  If two objects have the same SBProfile and type GSObject, but different ```noise``` attribute, should they be considered equal?

rmjarvis: I guess it's a judgement call, but I think those two objects should still be considered equal.  Anything you do with them would produce the same result.  The only thing that would be different is explicitly accessing the `noise` attribute.  

In other words, there is nothing in GalSim that uses the `noise` attribute of an object.  It's only there for convenience of the user if they want to do something with it.  So if the user cares, they can explicitly check the equality of two objects' `noise` attributes.  And the only way they could possibly end up different is for the user to have explicitly added or deleted it from one of them somewhere along the way.

galsim/transform.py
rmandelb: I slightly prefer ```np```, but I strongly prefer consistency throughout GalSim.

galsim/base.py
jmeyers314: >And the only way they could possibly end up different is for the user to have explicitly added or deleted it from one of them somewhere along the way.
 
That was my thought here.  Maybe there's a case for comparing the `noise` in the future though if we ever introduce competing noise correlation function estimators for `RealGalaxy`s.

rmandelb: OK, fair enough.

jmeyers314:
Were you finished with this PR @rmandelb ?  Is this ready to merge?

rmandelb:
Oh yes.  Sorry for not commenting when I was done going over it!  I think it's fine.

jmeyers314:
>Oh yes. Sorry for not commenting when I was done going over it! I think it's fine.

No worries, just wanted to be sure.  Merging.

Issue #720, opened 2016-03-09
Investigate correlated noise from metacal and strategies to whiten it 


rmjarvis:
The naive way to write metacal and whiten the noise doesn't work.

Erin found a strategy that does, but when I adapt it to running on a non-square wcs, it still doesn't work if the wcs has any shear component.

This issue is for investigating how to do this correctly.  Possibly involving edits to GalSim code.  Or possibly not.

rmjarvis:
@esheldon @rmandelb Here is a view that shows the code I've written to test several different strategies to whiten the noise induced by metacal.

https://github.com/GalSim-developers/GalSim/compare/master...%23720

So far, the only real successful strategy is Strategy #3, which is the one Erin mentioned in his talk.  However, so far, I haven't gotten that strategy working when the WCS has a shear component.  I did get it to work if the WCS has a rotation though.

Also, it's written as a unit test, but I've turned off the asserts for the moment, so strategies 1 and 2 just report as failures, rather than aborting.

I'm keep working on trying to figure out how to get strategy 3 to work with arbitrary wcs.  It seems like that's the first order of business, since it's what we need to run metacal on real data.  But Rachel, if you have any ideas for how to make strategies 1 or 2 working, please feel free to chime in.

rmandelb:
Is that a bug-fix in wcs.py?  (the ```g2``` to ```g1``` replacement)  If so, should we cherry-pick that elsewhere, so as to get it in place before this issue is resolved?

rmjarvis:
It's an error in the doc.  Not a bug.

rmandelb:
Oy.  Right.

rmandelb:
I just looked through and have a question based on the comments:

1) Strategy 1 - I'm surprised you're finding symmetrizing works even with a wrong description of the correlation fcn.  But I'm also surprised that this entire thing is necessary, because I would think that imposing symmetry on the noise (not whitening) is all that's needed for MetaCal to work.  Do you understand this?  Yesterday I was relieved because I thought you meant neither symmetrization nor whitening was working, but now I am confused again.  For multiple reasons.  :)

2) Also strategy 1 - can we use gsparams to force a higher maxk and see if that fixes the issue?  It would be nice if the fix was as easy as that... but I'm guessing you may have already tried this?

rmjarvis:
[...]

PR #736, opened 2016-04-08
#720 A few bug fixes relevant for metacal runs

rmjarvis:
This PR won't close issue #720, but given that we'll be releasing version 1.4 soon, I wanted to make sure the bug fixes I've made while investigating the metacal noise properties get included in that release.  Here are the bug fixes I've made on this branch:

1. There was an error in drawKImage when giving a non-default scale, but letting it automatically set the size.  This is fixed.
2. There was an assert statement in the C++ layer that could be triggered (leading to an ugly abort) when making an InterpolatedImage from an empty input image.  This now gives a graceful Python exception.
3. When deconvolving things that have zeros if fourier space and then reconvolving by something that wasn't perfectly matched due to numerical approximations, the noise could be massively amplified.  This isn't really a bug in anything (that's just the math), but I changed the SBDeconvolve function to max out the inverse at 1/kvalue_accuracy.  This seems appropriate, since that's as accurate as we know any of the zeros most of the time.  And it significantly improved the noise properties of the resulting images.

This PR also includes a new test_metacal.py, which is where I'm still working on trying to understand the noise whitening procedures and why strategies 1 and 2 don't work.  This is still work in progress, but for those who are interested, here is a summary of the state of things there:

There are several strategies to remove the noise correlations induced by the metacal procedure.  Only strategy 4 works.  I believe I understand why strategy 1 doesn't work, but I haven't figured out a good fix for it yet.  I think strategy 2 should work, but it doesn't.  I don't yet know if this is an error in the test or a bug in GalSim.  Strategy 3 was the one that @esheldon found worked on his tests, but it turns out it only works on square pixels.  We understand why, so there's no bug in the fact that that one fails.  Strategy 4 is our current successful strategy that Erin is using in production metacal code (which lowers the pressure to figuring out how to get 1 or 2 working).  Strategy 5 was an idea I had that I thought might work, but it also only works for square pixels.  I think I understand why that is now, and I'm not sure it can be made to work with arbitrary wcs.


esheldon:
I think raising a `RuntimeError` for the zero image can be problematic.  I can't generically catch that exception, because such an exception might come from anywhere in the run. This exception (zero image) I would like to catch and keep running, but `RuntimeError` is generic and so ambiguous it is unwise to catch it.

Would you consider making a specific exception for this?  Something indicating it comes from galsim?  e.g. `Galsim*Error` where `*` is something meaningful that you choose?

rmjarvis:
> Would you consider making a specific exception for this? Something indicating it comes from galsim? e.g. Galsim*Error where * is something meaningful that you choose?

We don't have any GalSim-specific error classes like this.  I could do ValueError if you prefer.  But I'm not particularly inclined to create a new Error class just for this.

rmandelb:
Tests pass on both my systems.  

GitHub is claiming there are conflicts that prevent an automatic merge.

galsim/interpolatedimage.py
```diff
@@ -435,6 +435,10 @@ def __init__(self, image, x_interpolant=None, k_interpolant=None, normalization=
                 pad_image.image, self.x_interpolant, self.k_interpolant, pad_factor,
                 _force_stepk, _force_maxk, gsparams)
 
+        if (calculate_stepk or calculate_maxk) and sbii.getFlux() == 0.:
```
rmandelb: Why only raise this exception if ```calculate_stepk or calculate_maxk```?  I agree that if those aren't set, it won't raise immediate problems to have an empty surface brightness profile.  But doesn't this just kick the problems further down the line?

rmjarvis: I didn't want to completely disallow using InterpolatedImage with a zero image in case people are using it in a non-standard way where maybe it would be ok.  But the other thing that will mess up if getFlux() == 0 is if the user supplies a given flux value.  So I also raise the exception in that case now.  (Just added: 2c5330e)

rmjarvis:
Did anyone else want to take a look at this?  OK to merge?

esheldon:
I've been using it in production without any issues.

rmandelb:
I think you're OK to merge.

PR #737, opened 2016-04-10
Pass DYLD_FALLBACK_LIBRARY_PATH to scons

mjuric:
This makes it easier to build GalSim against buggy Python builds (such as the one shipping with Anaconda; see https://github.com/ContinuumIO/anaconda-issues/issues/498).

Note: in most cases one can get away with adding the path to libpython*.dyld onto `DYLD_LIBRARY_PATH`, but having it on `DYLD_FALLBACK_LIBRARY_PATH` ensures nothing else gets picked up from it that would otherwise get picked up from system paths. E.g., Anaconda also ships with `libz`, but one probably wants to build against the system version, not the Anaconda one.

Related issues: https://jira.lsstcorp.org/browse/SIM-1895, https://github.com/lsst/galsim/pull/9

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -107,6 +107,13 @@ opts.Add(PathVariable('DYLD_LIBRARY_PATH',
          'this option enables SCons to set it back in for you by doing '+
          '`scons DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH`.',
          '', PathVariable.PathAccept))
+opts.Add(PathVariable('DYLD_FALLBACK_LIBRARY_PATH',
+         'Set the DYLD_FALLBACK_LIBRARY_PATH inside of SCons.  '+
+         'Particularly useful on El Capitan (and later), since Apple strips out '+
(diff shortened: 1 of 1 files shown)


rmjarvis:
All looks fine to me.  Thanks, Mario.  I wasn't aware of this FALLBACK set of environment variables.  I certainly don't have a problem with treating this one the same way as the others we do this for.

Do you need me to cherry-pick this over to the 1.3 branch (and make a new revision there)?  Or is this ok just being included in version 1.4 (probably <~ 1 month away)?

mjuric:
Hi @rmjarvis -- merging to master only is fine, thanks!

rmandelb:
Sounds good to me.

cwwalter:
Hi All,

Do we need have both  this and the DYLD_LIBRARY_PATH option which we added at SLAC?

rmjarvis:
I think most of the time, DYLD_LIBRARY_PATH is sufficient.  But Mario wanted the ability to put things in the DYLD_FALLBACK_LIBRARY_PATH as well.  I guess it can make it easier to build in some cases.

PR #740, opened 2016-04-13
Minor bug fixes in getting parametric galaxies from COSMOSCatalog

arunkannawadi:
This is very short PR to fix a few bugs that occurs when one tries to get the parametric fits for galaxies from the recently updated COSMOS Catalog.

P.S.: Sorry, I am sleepy and have pushed the fix in existing branches :-/. It shouldn't bother with anything but let me know if it should be reverted.

### DIFF
--- a/tests/test_scene.py
+++ b/tests/test_scene.py
@@ -56,25 +56,25 @@ def test_cosmos_basic():
     # Try making galaxies
     gal_real = cat2.makeGalaxy(index=0,gal_type='real',chromatic=False)
     if not isinstance(gal_real, galsim.RealGalaxy):
-        raise TypeError("COSMOS Catalog makeGalaxy routine does not return an instance of"+\
+        raise TypeError("COSMOS Catalog makeGalaxy routine does not return an instance of "
                         "'galsim.RealGalaxy'")
 
     gal_param = cat.makeGalaxy(index=10,gal_type='parametric',chromatic=True)
     if not isinstance(gal_param, galsim.ChromaticObject):
-        raise TypeError("COSMOS Catalog makeGalaxy routine does not return an instance of"+\
+        raise TypeError("COSMOS Catalog makeGalaxy routine does not return an instance of "
                         "'galsim.ChromaticObject' for parametric galaxies")
 
     gal_real_list = cat.makeGalaxy(index=[3,6],gal_type='real',chromatic=False)
     for gal_real in gal_real_list:
         if not isinstance(gal_real, galsim.RealGalaxy):
-            raise TypeError("COSMOS Catalog makeGalaxy routine does not return a list of instances"\
-                            +" of 'galsim.RealGalaxy'")
+            raise TypeError("COSMOS Catalog makeGalaxy routine does not return a list of instances "
+                            "of 'galsim.RealGalaxy'")
 
     gal_param_list = cat.makeGalaxy(index=[4,7],gal_type='parametric',chromatic=False)
     for gal_param in gal_param_list:
         if not isinstance(gal_param, galsim.GSObject):
-            raise TypeError("COSMOS Catalog makeGalaxy routine does not return a list of instances"\
-                            +" of 'galsim.GSObect'")
+            raise TypeError("COSMOS Catalog makeGalaxy routine does not return a list of instances "
+                            "of 'galsim.GSObect'")
 
     # Check for parametric catalog
     cat_param = galsim.COSMOSCatalog(file_name='real_galaxy_catalog_example_fits.fits',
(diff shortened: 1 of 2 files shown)


arunkannawadi:
Disclaimer: I am fixing bugs as and when I see my script that was working fine earlier crashing. I haven't checked for other use cases.

rmandelb:
These look good to me.  Thanks for catching them.

Not sure what the best thing to do is about this appearing on other branches; if we merge it into master relatively soon given that it's pretty simple (I imagine someone else could also take a look pretty soon) then it won't cause any issue with other PRs.

arunkannawadi:
These are only about 4 lines. So if someone else as well, Mike or Josh,
could take a quick look at it, we can push it onto the master immediately.

jmeyers314:
Looks good to me.  Can you add a short unit test though that would have caught these before you fixed them?

rmjarvis:
Agreed.  A could additional unit tests that would touch these lines of code would be helpful.

arunkannawadi:
I added some unit tests. I was having some trouble getting the Chromatic galaxies as my Pandas were throwing some error. I have Pandas v0.13.1, should I update it to be able to use it?

rmandelb:
What kind of errors?  With old versions, I was getting AttributeErrors, so the part of table.py that imports pandas looks like
```
try:
    import pandas
    ...other stuff...
except (ImportError, AttributeError):
   ...use np.loadtxt() instead...
```
So I think you can just add your type of exception to the list of exceptions that gets caught (and is used to decide not to use pandas).

rmandelb:
BTW I pushed a fix to address the way warnings were being given (the new code to address the case where there's only a parametric catalog was causing problems).

rmjarvis:
> So I think you can just add your type of exception to the list of exceptions that gets caught (and is used to decide not to use pandas).

I also ran across a machine which gave pandas errors.  I fixed this on master: b01b124c77f3f93519d0a090681a41daf6418142

rmjarvis:
All looks fine to me.  I also noticed a slight logic error here, with respect to the nobjects_only option, which I fixed.  I'm happy to merge this now.

rmandelb:
Thanks for adding the unit tests.  It looks good to me.  @arunkannawadi , please go ahead and merge as soon as you want.

arunkannawadi:
The pandas error seems to be resolved. But when I try to run 
```gal_param = cat.makeGalaxy(index=10,gal_type='parametric',chromatic=True)```,
 I get this error: 
```
  File "/usr/lib/python2.7/dist-packages/galsim/table.py", line 190, in __call__
    self.table.interpMany(x.astype(float),f)
AttributeError: '_LookupTable' object has no attribute 'interpMany'
```

**Comment by rmandelb** [comment]
Are you saying you try to change the line in ```test_scene.py``` from 
```gal_param = cat.makeGalaxy(index=10,gal_type='parametric',chromatic=False)`` by changing the ```False``` to ```True```, and get an exception?  That runs just fine for me.  Have you definitely rerun ```scons``` and ```scons install``` on the latest version?

**Comment by arunkannawadi** [comment]
Yes, I figured out that I haven't run scons recently, so I guess it's fine.
I will change the ```chromatic=False``` to ```chromatic=True``` for this
one case alone and merge with the master.

On Sun, Apr 17, 2016 at 10:51 AM, Rachel Mandelbaum <
notifications@github.com> wrote:

> Are you saying you try to change the line in test_scene.py from
> gal_param =
> cat.makeGalaxy(index=10,gal_type='parametric',chromatic=False)`` by
> changing theFalsetoTrue, and get an exception? That runs just fine for
> me. Have you definitely rerunsconsandscons install``` on the latest
> version?
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/GalSim-developers/GalSim/pull/740#issuecomment-211038031>
>



-- 
Regards

Arun Kannawadi Jayaraman
Carnegie Mellon University,
Pittsburgh, PA 15213

**Comment by rmandelb** [comment]
Well, no, if you do that then the rest of the test will fail.

**Comment by rmandelb** [comment]
(Because it's not a GSObject, it's a ChromaticObject.)  I was saying that I don't get the same error as you, not that the test passes as-is.  The test needs to be changed to make sure it does the proper check for this one.

**Comment by arunkannawadi** [comment]
Right. I fixed the complete test for Chromatic Object. I also fixed some
spacing issues with the error messages. Tests pass on my systems and I am
happy to merge now.

On Sun, Apr 17, 2016 at 11:28 AM, Rachel Mandelbaum <
notifications@github.com> wrote:

> (Because it's not a GSObject, it's a ChromaticObject.) I was saying that I
> don't get the same error as you, not that the test passes as-is. The test
> needs to be changed to make sure it does the proper check for this one.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/GalSim-developers/GalSim/pull/740#issuecomment-211044088>
>



-- 
Regards

Arun Kannawadi Jayaraman
Carnegie Mellon University,
Pittsburgh, PA 15213

File: tests/test_scene.py [hunk 1]
```diff
@@ -53,6 +53,45 @@ def test_cosmos_basic():
     except ImportError:
         print 'The assert_raises tests require nose'
 
+    # Try making galaxies
+    gal_real = cat2.makeGalaxy(index=0,gal_type='real',chromatic=False)
+    if not isinstance(gal_real, galsim.RealGalaxy):
+        raise TypeError("COSMOS Catalog makeGalaxy routine does not return an instance of"+\
+                        "'galsim.RealGalaxy'")
```
rmjarvis: You don't need the `\` here.  Or the `+` for that matter.  Python automatically concatenates two consecutive strings.

However, you should put a space at the end of the first string.  Otherwise you'll get "... of'galsim.RealGalaxy'".

arunkannawadi: Ah, I didn't know that.

If there are no more comments, I'll merge it by tonight.

arunkannawadi:
Merging.

Issue #741, opened 2016-04-13
Problem with near-flat probabilities in DistDeviate


msimet:
There is a bug in the DistDeviate implementation for near-flat probabilities.  I have a fix for this already and will be making a PR in a few minutes; I include the details here for historical reference.

As a reminder, the DistDeviate works by building the cumulative distribution function from a given PDF, then inverting it to draw an x value given a probability between 0 and 1.  Because we invert the function, it has to be always increasing; however, the true PDF may have regions of p=0.  We can't just collapse them to a single point. If we have a two-peaked distribution, say a series of p-values 
`0.5, 0, 0, 0, 0, 0.5`
for x-values of
`1, 2, 3, 4, 5, 6`
then collapsing to, say, only the x=2 point would mean a constantly-increasing probability between 2 and 6.  But we also can't keep the end points as-is, because x=2 and x=5 would both have a value of 0, so the function would not be invertible.  In the current implementation, we add epsilon to the upper edge of a range of zeros and delete all the points in the middle. That makes the function always increasing, but the region of zeroes is essentially never probed by the inversion, and we maintain the two-peaked distribution.

Here's the problem I just found: p-values of something like
```0.5, 0, 0, 0, 0, epsilon, 0.5-epsilon.```
Now when we add epsilon to the upper end of the range of zeroes, we still produce something non-increasing--we've just shifted up by one the indices where the function is constant.  

The solution here, I think, is to delete the upper edge `n` and let that epsilon at `n+1` stand for the whole range.

rmandelb:
> Here's the problem I just found: p-values of something like

You mean when you give it input that already looks like this, that's when the problem kicks in?  So the idea is to check whether a sequence of zeros has a tiny number after it, and if so, the tiny number can stand for the whole range?  But if there isn't a tiny number after the sequence of zeros then you enforce one, as before?

Just trying to make sure I understand what you're trying to do before I read the code.

msimet:
@rmandelb : Yes, exactly.

PR #742, opened 2016-04-13
#741 Problem with near-flat probabilities in DistDeviate

msimet:
And here's the fix for the problem identified in issue #741, with details given there.  This also switches to using the system's particular value of epsilon rather than hard-coding it in.


rmandelb:
Can you update CHANGELOG, and include a unit test that the old code would fail but the new code passes?

rmjarvis:
I just took a look, and I think the code here is probably fine, but I'll reiterate the need for a unit test on this that would have failed with the old code, but works now.

msimet:
Okay, I think this has a working test that would have caught the bug I found before.  (In making it I found an indexing error in my fix, too.)  My apologies for the delay.

tests/test_random.py
```diff
@@ -120,6 +120,18 @@ def dfunction(x):
...
+dp1_eps = np.zeros(dx_eps.shape)
+dp2_eps = np.zeros(dx_eps.shape)
+eps = np.finfo(dp1_eps[0].dtype).eps
+dp1_eps[0] = 0.5
+dp2_eps[0] = 0.5
+dp1_eps[-1] = 0.5
+dp2_eps[-2] = eps
+dp2_eps[-1] = 0.5-eps
+dLookupTableEps1 = galsim.LookupTable(x=dx_eps, f=dp1_eps, interpolant='linear')
+dLookupTableEps2 = galsim.LookupTable(x=dx_eps, f=dp2_eps, interpolant='linear')
```
rmandelb: These aren't getting used, so is the problem that it would die when trying to initialize the LookupTable?

rmandelb: Oh never mind, I see they get used later.

rmandelb:
OK, thanks for adding the test.  In my opinion, this is fine to merge once CHANGELOG is updated, but hopefully Mike will check it out too.

tests/test_random.py
rmjarvis: Minor style point, but I would probably put these into the test that uses them, since they are only used in one test.  IMO, it's generally better to keep these things together so someone looking at that test doesn't have to do too much scrolling around to see what is going on.  Other than that, this looks fine to me.

msimet:
Okay, updated the test definition location & the CHANGELOG.  Anything else I need to do?

rmjarvis:
Looks good to me.

rmandelb:
Me too, feel free to merge whenever you want.

Issue #738, opened 2016-04-11
problem with image.order=random in config


rmandelb:
I think I found either (a) a problem with the implementation of image.order=random in config or (b) an undesirable feature of how the GREAT3-like example scripts ```cgc.yaml``` etc. were set up.

Here is the issue:
- ```cgc.yaml``` has ```image.order=random``` so that the 90-degree rotated pairs aren't right next to each other.
- The output images indeed don't have 90 degree pairs next to each other.
- The output truth tables DO list the 90 degree pairs right next to each other, and there's no information that one could use to figure out where they actually live in the image.

So I'm not sure if this is a bug in ```image.order=random``` implementation, where the locations aren't getting propagated into the truth catalog part of the code, or if it's just that the truth catalogs should have another output column in this example that tells you where to find the galaxies.  But something weird is definitely going on.

rmjarvis:
I see.  I guess we didn't include the real x,y position.  Just the offset.  Probably ought to make x,y = image_pos.x, image_pos.y, and give another name (dx,dy?) to the offset values.  Assuming I didn't screw something up, that should accurately tell you which object is which in the image.

rmjarvis:
Oh, and currently, the cosmos_index value in the truth catalog will tell you which pairs go together.  They'll have the same index.

rmandelb:
> I see. I guess we didn't include the real x,y position. Just the offset. Probably ought to make x,y = 
> image_pos.x, image_pos.y, and give another name (dx,dy?) to the offset values. Assuming I didn't 
> screw something up, that should accurately tell you which object is which in the image.

I tried doing
```
            x: image_pos.x
            y: image_pos.y
            dx: image.offset.x
            dy: image.offset.y
```
but that failed with an exception:
```
Traceback (most recent call last):
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/site-packages/galsim/config/process.py", line 840, in MultiProcess
    result = job_func(**kwargs)
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/site-packages/galsim/config/output.py", line 222, in BuildFile
    data = builder.buildImages(output, config, file_num, image_num, obj_num, ignore, logger)
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/site-packages/galsim/config/output.py", line 427, in buildImages
    image = galsim.config.BuildImage(base, image_num, obj_num, logger=logger)
[...]

PR #743, opened 2016-04-17
#738 - small bug in writing output truth tables

rmandelb:
As described in #738, there was a bug in tracking the original COSMOS catalog index for parametric galaxies that use the Sersic galaxy model.  If a transformation (like lensing) was later applied to those objects then the original catalog index was not retained.  On this branch:

- That bug was fixed.  I also made a unit test that would catch the issue on the python side (not the config side), but did not add the bug fix to the CHANGELOG since the truth table functionality is new in v1.4.
- Some of the demos were updated to include more information in the output truth tables (e.g., location of the galaxy within a larger image, and index in the COSMOS catalog).
- Mike made some of the error messages when using the wrong syntax for output truth tables more informative.

### DIFF
--- a/galsim/config/value.py
+++ b/galsim/config/value.py
@@ -219,17 +219,25 @@ def GetCurrentValue(key, config, value_type=None, base=None, return_safe=False):
 
         if chain: 
             # If there are more keys, just set d to the next in the chain.
-            d = d[k]
+            try:
+                d = d[k]
+            except TypeError:
+                raise ValueError("Invalid key in GetCurrentValue = %s"%key)
 
             # One subtlety here.  Normally the normal tree traversal will keep track of the index
             # key so that all lower levels inherit an index_key specification at a higher level.
             # This can circumvent that, so we need to do it here as well.  The easiest way to
             # handle it is to watch for an index_key specification along our chain, and if there
             # is one, set that in the final dict.
-            if 'index_key' in d:
+            if isinstance(d,dict) and 'index_key' in d:
                 use_index_key = d['index_key']
                 #print 'Set use_index_key = ',use_index_key
         else:
+            try:
+                dk = d[k]
+            except TypeError:
+                raise ValueError("Invalid key in GetCurrentValue = %s"%key)
+
             if not isinstance(d[k], dict):
                 if value_type is None:
                     # If we are not given the value_type, and it's not a dict, then the
@@ -261,7 +269,7 @@ def GetCurrentValue(key, config, value_type=None, base=None, return_safe=False):
             else:
                 return val
 
-    raise ValueError("Invalid key = %s"%key)
+    raise ValueError("Invalid key in GetCurrentValue = %s"%key)
 
 
 def SetDefaultIndex(config, num):
--- a/galsim/config/value.py
+++ b/galsim/config/value.py
@@ -221,7 +221,9 @@ def GetCurrentValue(key, config, value_type=None, base=None, return_safe=False):
             # If there are more keys, just set d to the next in the chain.
             try:
                 d = d[k]
-            except TypeError:
+            except (TypeError, KeyError):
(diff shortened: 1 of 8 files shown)


galsim/scene.py
```diff
@@ -842,10 +843,16 @@ def _makeSingleGalaxy(cosmos_catalog, index, gal_type, noise_pad_size=5, deep=Fa
...
+        # Store the orig_index as gal.index, since the above RealGalaxy initialization just sets it
+        # as 0.  Plus, it isn't set at all if we make a parametric galaxy.  And if we are doing the
+        # deep scaling, then it gets messed up by that.  If we have done some transformations, and
+        # are also doing later transformation, it will take the `original` attribute that is already
+        # there.  So having `index` doesn't help, and we also need `original.index`.
+        if not hasattr(gal, 'original'):
+            gal.index = cosmos_catalog.getOrigIndex(index)
+        else:
+            gal.index = cosmos_catalog.getOrigIndex(index)
+            gal.original.index = cosmos_catalog.getOrigIndex(index)
```
rmjarvis: There's a little bit of redundant code here.  Perhaps simplify this to:
```
gal.index = cosmos_catalog.getOrigIndex(index)
if hasattr(gal, 'original'):
    gal.original.index = cosmos_catalog.getOrigIndex(index)
```

rmjarvis:
Just one minor comment from me.  Looks good.

galsim/scene.py
rmandelb: Yes, that's closer to how I did it elsewhere.  Done.

rmandelb:
Thanks!

rmandelb:
Given that this is a small-ish bug fix plus demo updates, are you OK with my going ahead and merging?  Or shall I ping someone else (presumably Josh) to see if he is willing to check it out?

galsim/config/value.py
```diff
@@ -219,17 +219,25 @@ def GetCurrentValue(key, config, value_type=None, base=None, return_safe=False):
 
         if chain: 
             # If there are more keys, just set d to the next in the chain.
-            d = d[k]
+            try:
+                d = d[k]
+            except TypeError:
+                raise ValueError("Invalid key in GetCurrentValue = %s"%key)
```
jmeyers314: Minor, but should one or both of these be a KeyError?

rmjarvis: I'll add KeyError, since that would also be relevant.  But the type error is to catch the case where `d` is already the final object, say a float or a Position2D.  Then doing `d[k]` raises a TypeError.  And the error is not terribly informative about what when wrong.

But if the next key in the chain is not valid, that would raise a KeyError, which would probably also be worth catching here and telling the user what the real problem is.

Done here: 608d776

jmeyers314: Okay.  That makes sense.

jmeyers314:
I'm looking at the output of demo9 now; comparing the truth catalog "truth0001.dat" to the image "cluster0001.fits".  I can see by eye that image_pos.x and image_pos.y line up with the image coords.  The pos.x and pos.y don't seem to line up with the wcs coords though.  Is that expected?

rmjarvis:
> The pos.x and pos.y don't seem to line up with the wcs coords though. Is that expected?

Yes, in this case, the WCS is not a type that ds9 knows about.  It's our own custom UVFunction WCS type that lets you do any arbitrary function you want.  

We still put something semi-plausible into the FITS header, which is what ds9 is using, but I think it's just the local linear approximation at the center of the image.  This WCS function in this demo is rather extreme (to exaggerate the effects), so that's not a very good approximation, and the ds9 world positions aren't very close.

jmeyers314:
Okay, got it.  In that case, all looks good to me.

rmjarvis:
Great.  Thanks so much Josh.  Merging now.

rmandelb:
Thanks both!

Issue #745, opened 2016-05-05
Error while making some chromatic galaxies


arunkannawadi:
While trying to make chromatic galaxies from the COSMOS catalog, a few cases seem to try accessing wavelengths well beyond the range of SEDs. For example, the following code:

`import galsim`
`cc = galsim.COSMOSCatalog(use_real=False)`
`gal = cc.makeGalaxy(30105,chromatic=True)`

throws an error message:
` ValueError: Requested wavelength (400.0) is bluer than blue_limit (499.543110609) `

@jmeyers314 , do you know anything about this?

jmeyers314:
Interesting.  I think there are a couple of things going on here.  The SED is basically reporting that it isn't defined where you need it to be defined in order to compute a flux through a Bandpass.

Part of this may be that the Bandpass in question extends down to 400 nm, even though the bulk of the throughput doesn't actually start until ~700 nm.  So using the .truncate method [here](https://github.com/GalSim-developers/GalSim/blob/master/galsim/scene.py#L577) might be appropriate.  Actually, I'd probably just pre-thin and truncate this file (documenting what you do in its header) in the share directory so this doesn't need to happen when running scene.py.

More critically though, the SEDs aren't being read in right.  They're missing a `wave_type='ang'` kwarg [here](https://github.com/GalSim-developers/GalSim/blob/master/galsim/scene.py#L586) and below.  I probably should have caught this in an earlier review, sorry!  What you've effectively discovered is that GalSim refuses to compute the F814 flux for a z=26 elliptical spectrum, which I don't think is that surprising.

Finally, we should probably update the bandpass and sed files in the share dir so that they use the newer version of .thin().  I remembered to do this for examples/data/, but not for share.

rmandelb:
Thanks for looking into this Josh.

> Actually, I'd probably just pre-thin and truncate this file

Regarding the bandpass, I think we'd prefer to keep it as-is in share.  Right now the bandpasses get read in with proper zero points etc. by the ```galsim.wfirst.getBandpasses``` routine, which also thins them according to a user-specified value of ```thin_err```.  We'd like to make it possible for people to modify ```thin_err```, which would require the full, non-thinned bandpasses to be stored.

But it sounds like you recommend that we use the ```truncate``` method in ```getBandpasses``` too?  Right now we leave that up to the user, but perhaps we should make a kwarg for that (and have it be True by default).

[...]

PR #746, opened 2016-05-12
#745

rmandelb:
This pull request addresses the bug report in #745 and does several vaguely related changes:

- For ```galsim.Bandpass``` and ```galsim.SED```, ```wave_type``` and ```flux_type``` keywords are now required, instead of optional.  This is to avoid people simply using the defaults when they are wrong for their own bandpasses / SEDs.  Now they have to explicitly check and use the right values.

- Thinned and truncated SEDs are available in share/ .  They are used in scene.py when making parametric chromatic galaxy models.

- ```galsim.wfirst.getBandpasses()``` has been updated to propagate through the new keywords for the ```truncate``` and ```thin``` routines.


share/CWW_E_ext_more.sed
```diff
@@ -0,0 +1,278 @@
+#  E SED of Coleman, Wu, and Weedman (1980)
```
jmeyers314: This file looks good.  I think you still need to update the old file (CWW_E_ext.sed) in this directory too though, since the thinning algorithm recently changed.  In particular, I noticed that the "less" thinned version starts at 137 Ang, and the more thinned version starts at 91 Ang, which is a bit counter-intuitive.

rmandelb: Good point.  I will do that, thanks!  I assume I can still use the examples/data/getSEDs.py with the settings appropriately updated?

The other thing that I realized is pretty silly with the current setup is that the more thinned versions here in share/ correspond to the ones in examples/data/, yet the filenames are the same for the old (less thinned) versions in share/ and in examples/data.  Would you object to my changing the filenames of the ones in examples/data so be consistent with the more thinned ones in share/?  Or, is there any reason that demo12.py shouldn't just use the thinned versions in share/, and we can delete the copies of the SEDs from examples/data/?

rmjarvis: I'd prefer removing any files in examples/data that are duplicated in share.

rmandelb: OK - I pushed a few commits that include the following:

1) Self-consistent thinning for the more/less thinned ones.
2) Keep them only in share/, not examples/data/.
3) Once they aren't in examples/data/, I didn't think getSEDs.py should be either.  So I moved it to devel/modules/.  Its location is noted in the comments of the SEDs themselves.

galsim/deprecated/bandpass.py
```diff
@@ -39,7 +39,12 @@ def Bandpass_rdiv(self, other):
     else:
         tp = lambda w: other / self.func(w)
 
-    return galsim.Bandpass(tp, blue_limit, red_limit, _wave_list=wave_list)
+    if self.wave_factor == 10.0:
+        wave_type = 'Angstroms'
+    else:
+        wave_type = 'nm'
+    return galsim.Bandpass(tp, 'nm', blue_limit, red_limit,
```
rmjarvis: I think probably this `'nm'` should be `wave_type`?

rmjarvis: (Or if not, then the above definition of `wave_type` seems to be unused.)

galsim/wfirst/wfirst_bandpass.py
```diff
@@ -65,7 +66,10 @@ def getBandpasses(AB_zeropoint=True, exptime=None, thin_err=1.e-4):
                             True]
     @param exptime          Exposure time to use for setting the zeropoint; if None, use the default
                             WFIRST exposure time, taken from galsim.wfirst.exptime.  [default: None]
-    @param thin_err         Relative error allowed when thinning the bandpasses.  [default: 1e-4]
```
rmjarvis: The parameter `default_thin_trunc` should be listed explicitly here.  And I'd suggest adding after that
```
@params **kwargs           Other kwargs are passed to either `bandpass.thin()` or `bandpass.truncate()` as appropriate.
```
Also, I think some of the below would be slightly simpler if the `default_thin_trunc` kwarg was not part of the `kwargs` dict.  So I'd put it explicitly in the function signature before `**kwargs`.

galsim/wfirst/wfirst_bandpass.py
```diff
@@ -87,6 +91,32 @@ def getBandpasses(AB_zeropoint=True, exptime=None, thin_err=1.e-4):
...
+        if kwargs['default_thin_trunc']:
+            pass
+        else:
+            del kwargs['default_thin_trunc']
+    else:
+        if any(x in kwargs.keys() for x in truncate_kwargs):
+            for key in kwargs.keys():
+                if key in truncate_kwargs:
+                    tmp_truncate_dict[key] = kwargs[key]
+                    del kwargs[key]
```
rmjarvis: Minor simplification: These two lines can be combined as
```
tmp_truncate_dict[key] = kwargs.pop(key)
```
In fact, I think the whole block (including the if statement) can be just
```
tmp_truncate_dict = dict( ((key, kwargs.pop(key)) for key in kwargs if key in truncate_kwargs) )
```
But I'll let you decide if that dict comprehension is too confusing compared to the explicit loop, since efficiency isn't really a concern here.

galsim/deprecated/bandpass.py
rmandelb: It should actually be 'nm' which was the former default (and that's why the old code didn't have a unit attached to it).  The unit conversions had already been taken care of properly.  The real problem is that I didn't notice that at first, so I added the if/else block which is totally unnecessary.  I'll remove it.

rmandelb:
@jmeyers314 , when I updated the SEDs to use the latest version of the thinning routine, I found some test failures in the chromatic functionality.  I think they are not something to worry about too much (i.e., something got bumped just across the line from passing to failing) but can you please check it out and see what you think?

galsim/wfirst/wfirst_bandpass.py
rmandelb: I prefer your first option (I think it's easier to read).  I made that change here and one other place where a similar code block appears.

galsim/wfirst/wfirst_bandpass.py
rmandelb: Done.

jmeyers314:
Hi @rmandelb .  I took a look at the chromatic test failures.  In both cases, the problem was that the changes to the SEDs didn't get propagated into the saved reference images.  For test_ChromaticOpticalPSF, I took your comment about how the reference image was (slowly) created and instead made this part of the test itself, but only to be executed (and subsequently saved) if the reference image cannot be found.  For test_ChromaticAiry, I actually think the saved reference image comparison wasn't really necessary, since the code to generate the reference image was identical to the code to generate the test image.  So I just ripped this test out and the associated reference image.  You and @rmjarvis should definitely take a look.

Other than that, I've finished looking through the rest of your commits and have no additional suggestions.

tests/test_chromatic.py
```diff
@@ -1609,15 +1569,30 @@ def test_ChromaticOpticalPSF():
     # plot(waves, bandpass(waves)*disk_SED(waves), ls=' ', marker='o', c='r', markersize=10)
     # show()
 
+    star = galsim.Gaussian(fwhm=1.e-8) * disk_SED
     psf = galsim.ChromaticOpticalPSF(lam=lam, diam=diam, aberrations=aberrations,
                                      obscuration=obscuration, nstruts=nstruts)
     do_pickle(psf)
 
+    if not os.path.isfile(os.path.join(refdir, 'r_exact.fits')):
```
rmjarvis: Maybe put a warning here that this is happening.  nosetests will print warnings amidst the dots, so if this happened in a nosetests run for some reason (which shouldn't really happen), it would be noticed.  Something like
```
import warnings
warnings.warn("Could not find file r_exact.fits, so generating it from scratch.  This should only "
              "happen if you intentionally deleted the file in order to regenerate it!")
```

rmjarvis:
Just the one comment.  Otherwise, I think this is good to go.

tests/test_chromatic.py
jmeyers314: Sure.  Done.

rmandelb:
Great.  It seems like we're done here, so I will plan to merge this at the end of the day unless further objections are raised.

rmandelb:
Merging #745

Issue #748, opened 2016-05-17
Fourier-space Sqrt GSObject/SBProfile (proposed contribution)


TallJimbo:
I'd like to implement a the Kaiser algorithm for coaddition (see [Zackay & Ofek 2015](http://adsabs.harvard.edu/abs/2015arXiv151206879Z)) within GalSim.  I think all this requires is a Fourier-space `sqrt` operator on GSObjects, since all of the other operations (autocorrection and sums) already exist.

 - I'm thinking the `AutoConvolve` or `AutoCorrelate` `SBProfiles` look best as as examples to start from, but I'd love any recommendations for a simple Fourier-space only operator.
 - `Sqrt` probably isn't an appropriate name, since it's not a sqrt in real-space.  But it's also not an operation that has a name in real space.  Any other recommendations?

rmjarvis:
Maybe FourierSqrt would be clear and not too verbose?

I'd probably recommend Deconvolve as your template, since it can't be photon shot, and I guess that restriction applies to this too.

rmandelb:
+1 for FourierSqrt

TallJimbo:
Thanks for the excellent suggestions; copying `Deconvolve` made this really easy.  PR will be up shortly.

rmjarvis:
#749 merged.

PR #749, opened 2016-05-17
Add Fourier-space square root operation (#748)

TallJimbo:
This looks like a lot of new code, but it's actually an extremely simple change: I just copied the deconvolution code, changed some names, and replaced "1/x" with "sqrt(x)".  The only really new code is in the unit test.

That means there's a lot of cargo-culting, though, so I'd appreciate some eyeballs on this PR to tell me if I've copied anything silly.

One more note: I've not implemented the centroid and flux accessors (these just throw exceptions).  If these are considered important, I could probably implement them by drawing a Fourier-space image, but I thought that might be considered too heavyweight and approximate to match user expectations (as well as a lot of work).


rmjarvis:
I don't believe there will be any issue with `centroid` not being implemented, but I think `flux` is probably more important.  Fortunately, I think the flux of this is just the sqrt of the flux of the adaptee, so that's easy to implement (without having to draw anything).

rmjarvis:
Oh, and actually I think the centroid is easy too.  I believe it is just `0.5 * adaptee.centroid()`.

TallJimbo:
Hmm.  I had somehow convinced myself that the flux wasn't simple, but I can't reconstruct my reasoning, and that's a pretty good sign it was faulty.

Anyhow, thanks; I'll add those implementations and add some unit tests for them.

galsim/chromatic.py
```diff
@@ -2027,6 +2027,47 @@ def evaluateAtWavelength(self, wave):
...
+
+    Initialization
+    --------------
+
+    @param obj              The object to compute the Fourier-space square root of.
+    @param gsparams         An optional GSParams argument.  See the docstring for GSParams for
+                            details. [default: None]
+    """
+    def __init__(self, obj, **kwargs):
+        self.obj = obj.copy()
```
jmeyers314: I think you probably just copied this from `ChromaticDeconvolution`, but I'm wondering if the `.copy()` is actually necessary since all `ChromaticObject`s should be immutable?  If it's not needed, no need to change here, but I wanted to flag this for a future "cleanup the copying" issue.

galsim/compound.py
```diff
@@ -772,3 +773,91 @@ def __setstate__(self, d):
...
+                            details. [default: None]
+
+    @returns a FourierSqrtProfile or ChromaticFourierSqrtProfile instance as appropriate.
+    """
+    if isinstance(obj, galsim.ChromaticObject):
+        return galsim.ChromaticFourierSqrt(obj, gsparams=gsparams)
+    elif isinstance(obj, galsim.GSObject):
+        return FourierSqrtProfile(obj, gsparams=gsparams)
+    else:
+        raise TypeError("Argument to Deconvolve must be either a GSObject or a ChromaticObject.")
```
jmeyers314: Missed a "Deconvolve" -> "FourierSqrt".

galsim/compound.py
```diff
@@ -772,3 +773,91 @@ def __setstate__(self, d):
...
+
+
+class FourierSqrtProfile(galsim.GSObject):
+    """A class for deconvolving a GSObject.
```
jmeyers314: Here too.

include/galsim/SBFourierSqrt.h
```diff
@@ -0,0 +1,64 @@
...
+
+
+#include "SBProfile.h"
+
+namespace galsim {
+
+    /**
+     * @brief SBProfile adapter which computes the square root of its subject in k space.
+     *
+     * @param[in] adaptee   SBProfile for which to effect a deconvolution by k space inversion.
```
jmeyers314: and here.

src/SBFourierSqrt.cpp
```diff
@@ -0,0 +1,137 @@
...
+        dbg<<"SBFourierSqrt constructor: _maxksq = "<<_maxksq<<std::endl;
+    }
+
+    // xValue() not implemented for SBFourierSqrt.
+    double SBFourierSqrt::SBFourierSqrtImpl::xValue(const Position<double>& p) const
+    { throw SBError("SBFourierSqrt::xValue() not implemented (and not possible)"); }
+
+    std::complex<double> SBFourierSqrt::SBFourierSqrtImpl::kValue(const Position<double>& k) const
+    {
+        return (k.x*k.x + k.y*k.y <= _maxksq) ? 1./_adaptee.kValue(k) : 0.;
```
jmeyers314: I think this needs "1/blah" -> "sqrt(blah)" too.

jmeyers314:
Hi @TallJimbo , I think this looks basically right.  There are just a few minor comments above.

Since this is a somewhat unusual operation to perform, though, (in my head at least), I think it would be nice to briefly say somewhere -- probably the FourierSqrt factory function -- what the intended use is, and include the reference you cited in the initial issue page.

rmandelb:
The PR is asking to merge this into releases/1.3, whereas it should be master.  Can this be edited in the PR or does a new one have to be made?

rmjarvis:
> The PR is asking to merge this into releases/1.3, whereas it should be master. Can this be edited in the PR or does a new one have to be made?

[Apparently not](https://github.com/isaacs/github/issues/18).  But lots of people (>1500 comments, mostly saying +1) want it.  Here is a gem from that issue thread:

image

rmandelb:
So you're saying it's time to cause the death of a kitten?

TallJimbo:
Weird that it picked releases/1.3 instead of master; that certainly wasn't intentional.

In any case, maybe it's best if I just rebase and merge to master manually (after fixing up the branch, of course), rather than creating a new PR?

galsim/chromatic.py
TallJimbo: I did just copy from `ChromaticDeconvolution`, and if you're convinced it's unnecessary I'll at least change it here; no need to add to the future cleanup work if we don't have to.

TallJimbo:
Thanks for the close review, @jmeyers314; I think I've addressed most of your comments, and I'm *quite* embarrassed I let some of those slip through this far.

As for explaining why it's here, I'm thinking I might add a new example demonstrating how I'm actually using it right now.

rmandelb:
It looks like the branch itself was built off of releases/1.3 and not off of master, just based on the fact that most active branches are ~800 commits ahead of releases/1.3 and this one is just 2 ahead.  We do have releases/1.3 off the default branch, because most users will want the last release, but it is a bit of a pain for developers.

TallJimbo:
Oh, wow.  I didn't realize the branch was that far behind master, and I didn't realize one *could* set the default branch for a repo to be something other than master.  I should have paid more attention.

I'll work on finishing up the changes (just adding that demo now), then rebase.  I still suspect there won't be conflicts, but we'll see.

rmandelb:
I agree there are unlikely to be many conflicts.

Once you rebase, you should add this to the CHANGELOG.

galsim/compound.py
```diff
@@ -772,3 +773,91 @@ def __setstate__(self, d):
...
+    being operated on.
+
+    Initialization
+    --------------
+
+    The normal way to use this class is to use the FourierSqrt() factory function:
+
+        >>> b = galsim.FourierSqrt(a)
+
+    @param obj              The object to deconvolve.
```
rmjarvis: Another deconvolve oversight.

SConstruct
```diff
@@ -937,7 +937,7 @@ def TryScript(config,text,executable):
 
     # Run the given executable with the source file we just built
     output = config.sconf.confdir.File(f + '.out')
-    node = config.env.Command(output, source, executable + " < $SOURCE >& $TARGET")
+    node = config.env.Command(output, source, executable + " $SOURCE > $TARGET")
```
rmjarvis: This line will be a conflict when you rebase.  The '>&' isn't actually posix compliant, so we fixed that.  However, I'm curious why you removed the '<'.  It's probably fine this way too, but I wouldn't have thought it would be a problem.

TallJimbo: Oops.  I hadn't meant to commit this at all - this PR really is an embarrassment.  That said, I'm glad it's fixed downstream, because I did indeed have to make that change to get it working on my machine.  Removing the `<` was non necessary for that; I was just something I tried while trying to figure out what went wrong.

I'll remove this change entirely before I rebase.

include/galsim/SBFourierSqrtImpl.h
```diff
@@ -0,0 +1,80 @@
...
+
+        // xValue() not implemented for SBFourierSqrt.
+        double xValue(const Position<double>& p) const;
+
+        SBProfile getObj() const { return _adaptee; }
+
+        std::complex<double> kValue(const Position<double>& k) const;
+
+        double maxK() const { return _adaptee.maxK(); }
+        double stepK() const { return _adaptee.stepK(); }
```
rmjarvis: I think technically, this should be `_adaptee.stepK() / sqrt(2.)`.  Perhaps you could add a unit test that checks the (at least approximate) equivalence of these values for `galsim.AutoConvolve(galsim.FourierSqrt(obj))` and `obj`.  

They should be equivalent conceptual objects, so a few tests that they draw similar results would probably also be appropriate.

tests/test_compound.py
```diff
@@ -696,6 +696,62 @@ def test_autocorrelate():
...
+    t1 = time.time()
+
+    dx = 0.4
+    myImg1 = galsim.ImageF(80,80, scale=dx)
+    myImg1.setCenter(0,0)
+    myImg2 = galsim.ImageF(80,80, scale=dx)
+    myImg2.setCenter(0,0)
+
+    # Test trivial case, where we could (but don't) collapse the chain of SBProfiles
+    # by recognizing the inverse.
```
rmjarvis: Ah, you do have that second unit test I suggested.  However, here ("inverse") is another relic of the deconvolve copying.

rmjarvis:
Just a few minor comments.  Otherwise looks fine to me.

tests/test_compound.py
TallJimbo: Actually, here I meant "recognize that FourierSqrt is the inverse of AutoConvolve"; I'll expand the comment to clarify that.

include/galsim/SBFourierSqrtImpl.h
TallJimbo: `AutoConvolve` has `_adaptee.stepK() / sqrt(2.)`, so I think that means `FourierSqrt` should have `_adaptee.stepK() * sqrt(2.)`.  And I'll add that check to the tests.

TallJimbo:
This has been rebased on master, so the diff is no longer meaningful (though the per-commit diffs are).  I still need to update the changelog, and I'm going to do a diff of the things I copied against version/1.3 to see if I didn't copy something I should have.  After that this should be good to go, with no kittens harmed.

TallJimbo:
I believe this is now ready for merge.  I'll leave it up in case anyone wants to take a look at it in its final form.  As I said above, the full diff is now difficult to interpret but looking at the last three commits should be fine.

galsim/chromatic.py
```diff
@@ -2158,6 +2158,47 @@ def evaluateAtWavelength(self, wave):
...
+    @param obj              The object to compute the Fourier-space square root of.
+    @param gsparams         An optional GSParams argument.  See the docstring for GSParams for
+                            details. [default: None]
+    """
+    def __init__(self, obj, **kwargs):
+        self.obj = obj
+        self.kwargs = kwargs
+        self.separable = obj.separable
+        if self.separable:
+            self.SED = lambda w: 1./obj.SED(w)
```
rmjarvis: This seems wrong.  I guess this should be
```
self.SED = lambda w: obj.SED(w)**0.5
```

rmandelb:
I think README.md should be updated to refer to demo14 as the last demo (and as being one of the ones without a .yaml file).

rmandelb:
And maybe put the documentation for demo14 on the tutorial page: https://github.com/GalSim-developers/GalSim/wiki/Tutorials, unless you want me to do it?

rmjarvis:
My only comment at this point is that I'm not sure the example script you made is appropriate for the tutorial series.  This is kind of a niche application, not really something that most users would need to grok in order to "learn how to use GalSim".  

Perhaps rename it "make_coadd.py" or something along those lines?  And then maybe reference the example script in the doc string for FourierSqrt.

rmjarvis:
I went ahead and made the cosmetic change I suggested and merged this into master.  Thanks for the contribution, Jim!

TallJimbo:
Thanks, and sorry I moved too slowly to do it myself.

rmjarvis:
No worries!  We're looking to cut a new release soon, so it seemed helpful to get this merged now.  Thanks again.

Issue #711, opened 2016-02-16
Bandpass.thin() destroys zeropoint state


mdschneider:
I tried creating a `Bandpass` object from a lookup table as follows:

    bp = galsim.Bandpass(table)
    bp = bp.withZeroPoint(zeropoint='AB', effective_diameter=6.4, exptime=15)

rmandelb:
I see what you mean.  The thin() method returns a new Bandpass object and doesn't transfer over the information about the desired zeropoint.  

I think the fix could be as simple as changing the initialization of the new Bandpass to transfer over the zeropoint information from the original one.  @jmeyers314 , is that correct, or is it more complicated than that?

rmandelb:
(Also should include this in a unit test.)

jmeyers314:
>I think the fix could be as simple as changing the initialization of the new Bandpass to transfer over the zeropoint information from the original one. 

I think that sounds right, though I'd want to think a bit about whether or not a zeropoint should/could be recomputed using the thinned representation of the Bandpass.  Would it be worth keeping the "more accurate" zeropoint from the more finely sampled pre-thinned Bandpass even if this is different than what one would have computed on the post-thinned Bandpass, for instance?

rmandelb:
Hmmmm.  I think my answer will depend on the following: If you thin a typical bandpass using the defaults for how much to thin, how much do zero points typically change after recomputing with the thinned bandpass?

rmandelb:
Sorry for the delay, @mdschneider , this fell off my stack.  I just created a branch with a fix for this.  By default it gives the thinned bandpass the same zeropoint as the original.  The justification for this choice is that for the default thinning kwargs and typical optical bandpasses, recalculating the zeropoint with the thinned bandpass results in a change in zeropoint that is of order 10^-6.  However, I also added a keyword that people can use to tell it to forget about the bandpass.  This seemed unnecessary to me at first (because I figured that if someone really cares to recalculate, maybe because they thinned more aggressively, they could just recalculate the zeropoint themselves after thinning - it doesn't really matter that it's temporarily assigned a wrong one).  But then I realized it's an opportunity to explicitly flag the fact that the old zeropoint is not technically the correct one, which otherwise people might overlook.

@rmjarvis and @jmeyers314 - I made the new branch off of #549, so will wait on a PR until that one is merged.  But I've finished the work on here, so it's all ready to go.

jmeyers314:
Looks good to me @rmandelb !

[...]

PR #759, opened 2016-06-16
#711 - zeropoint preservation when thinning bandpasses

rmandelb:
This small PR addresses #711 (raised by @mdschneider ).  Based on @jmeyers314 's comment on the issue itself, I think he is OK with these changes, but it would be nice if @mdschneider or @rmjarvis could also check it out.


rmjarvis:
This all looks fine to me.

rmjarvis:
One thing just occurred to me.  `truncate` still destroys the zeropoint state.  Should we add a similar `preserve_zeropoint` option there?  

In that case, probably the default should be True if `blue_limit` and `red_limit` are None, but False if either of them are given explicitly.

rmandelb:
Good idea.  Will do this in the next day or so.

rmandelb:
Done.  @rmjarvis and @jmeyers314 , please let me know if you have any concerns about this final update.

galsim/bandpass.py
```diff
@@ -413,14 +413,37 @@ def truncate(self, blue_limit=None, red_limit=None, relative_throughput=None):
...
+                                    will have a zeropoint of None. If 'auto', the new truncated
+                                    Bandpass will have the same zeropoint as the original when
+                                    truncating using `relative_throughput`, but will have a
+                                    zeropoint of None when truncating using 'blue_limit' and/or
+                                   'red_limit'.  [default: 'auto']
 
         @returns the truncated Bandpass.
         """
+        if preserve_zp == 'auto':
+            if relative_throughput is not None: preserve_zp = True
```
rmjarvis: It think this should be
```
if blue_limit is None and red_limit is None: preserve_zp = True
```
since I don't think we currently limit the user from doing both the explicit limits (or one of them) as well as a relative throughput.  

OTOH, the doc string implies that only one or the other is allowed, so maybe just add that check to make sure the user doesn't provide both.  In which case the current version of this line is fine.

galsim/bandpass.py
```diff
@@ -413,14 +413,37 @@ def truncate(self, blue_limit=None, red_limit=None, relative_throughput=None):
...
-                                [default: None]
+                                    this relative throughput level.  (See above for details.)
+                                    [default: None]
+        @param preserve_zp          If True, the new truncated Bandpass will be assigned the same
+                                    zeropoint as the original.  If False, the new truncated Bandpass
+                                    will have a zeropoint of None. If 'auto', the new truncated
+                                    Bandpass will have the same zeropoint as the original when
+                                    truncating using `relative_throughput`, but will have a
+                                    zeropoint of None when truncating using 'blue_limit' and/or
+                                   'red_limit'.  [default: 'auto']
```
rmjarvis: One minor aesthetic point.  I'd rather the behavior you have using `'auto'` use `None` instead.  

And then in the doc string, you can just say `[default: False if either blue_limit or red_limit is given, otherwise True]`.

galsim/bandpass.py
rmandelb: Actually, regarding this:

> OTOH, the doc string implies that only one or the other is allowed, so maybe just add that check to make sure the 
> user doesn't provide both.

I too was a bit confused; I read the docstring, thought they were mutually exclusive, then saw it wasn't enforced in the code.  @jmeyers314 - is there a use case you had in mind where one might want to do both?  Or should I make it one or the other?

galsim/bandpass.py
rmandelb: I'm not sure I like that.  To me, ```None``` implies that you're not doing whatever it is.  But in the case of ```preserve_zp=None```, we would in fact be preserving the zeropoint some of the time.  It sounds counter-intuitive to me.

I guess your docstring suggestion is clear enough that if people are reading it, they will know exactly what's going on, but if someone is looking at what the default kwargs are and trying to work it out, I think ```None``` will suggest nothing is happening here, while ```'auto'``` will suggest that some automatic process is being done and they should read the docstring for info.

rmjarvis: To me, `None` means you aren't specifying whether or not to preserve the zeropoint, so it will do some default behavior.  But I won't press if you'd prefer to leave it as is.

galsim/bandpass.py
jmeyers314: I think it's probably least confusing if these are made mutually exclusive.  Also, we should guard against setting a `blue_limit` that's bluer than the original blue_limit and similarly for red_limit.

rmandelb: Cool.  Will make those changes.

galsim/bandpass.py
```diff
@@ -413,14 +413,37 @@ def truncate(self, blue_limit=None, red_limit=None, relative_throughput=None):
...
+        When truncating a bandpass that already has an assigned zeropoint, there are several
+        possibilities for what should happen to the new (returned) bandpass by default.  If red
+        and/or blue limits are given, then the new bandpass will have no assigned zeropoint because
+        it is difficult to predict what should happen if the bandpass is being arbitrarily
+        truncated.  If `relative_throughput` is given, often corresponding to low-level truncation
+        that results in little change in observed quantities, then the new bandpass is assigned the
+        same zeropoint as the original.  This default behavior is called 'auto'.  The user can also
+        give boolean True or False values.
+
+        @param blue_limit           Truncate blue side of bandpass here. [default: None]
```
jmeyers314: I'm sure this is my previous oversight, but I think it would be good to indicate the units for blue_limit and red_limit here, which I'm pretty sure are always nanometers (regardless of the wave_type of the bandpass).

galsim/bandpass.py
rmandelb: Done: https://github.com/GalSim-developers/GalSim/pull/759/commits/cc442bbedd674d48d908fef1daf6a427e0175b01

galsim/bandpass.py
rmandelb: Done.

rmandelb:
Thanks Josh!  I think I'm done here, if you two want to check out the final changes I made just now.

jmeyers314:
Looks ready to merge to me!

rmjarvis:
Looks fine to me.

rmandelb:
Great, thanks.  Merging #711.

PR #762, opened 2016-06-19
Fix bug in Image constructor where wcs wasn't copied if image is an Image

rmjarvis:
This is a tiny little bug fix that I ran across working on Piff.  If I used `im2 = galsim.Image(im, dtype=dtype)`, then the wcs attribute of `im` was not properly transfered over to `im2`.  I added some unit tests that check this and added 2 lines to `image.py` to fix the bug.


rmandelb:
Looks fine to me.

jmeyers314:
This one looks good too!

Issue #763, opened 2016-06-20
repr and str on transformed single summed galaxies 


ismael2395:
There seems to be a bug when trying to print the str or repr of an GSObject that consists of a sum of a single object. Example below. 
```
import galsim
gal1 = galsim.Sum([galsim.Transformation(galsim.Exponential(scale_radius=0.34586814810813454, flux=1049961.0753931333, gsparams=None), jac=[[0.5398181795385592, 0.17977371761576255], [0.17977371761576255, 1.9123449870247538]], offset=galsim.PositionD(x=0.0, y=0.0), flux_ratio=1.0, gsparams=None)], gsparams=None)
gal2 = gal1.dilate(1)
print str(gal1)
(galsim.Exponential(scale_radius=0.345868148108, flux=1049961.07539).shear(galsim.Shear(g1=-0.559720831877,g2=0.146624596656)))

print str(gal2)
(galsim.Exponential(scale_radius=0.345868148108, flux=1049961.07539).shear(galsim.Shear(g1=-0.559720831877,g2=0.146624596656))).shear(galsim.Shear(g1=-0.559720831877,g2=0.146624596656))

print repr(gal1)
galsim.Sum([galsim.Transformation(galsim.Exponential(scale_radius=0.34586814810813454, flux=1049961.0753931333, gsparams=None), jac=[[0.5398181795385592, 0.17977371761576255], [0.17977371761576255, 1.9123449870247538]], offset=galsim.PositionD(x=0.0, y=0.0), flux_ratio=1.0, gsparams=None)], gsparams=None)

print repr(gal2)
galsim.Transformation(galsim.Sum([galsim.Transformation(galsim.Exponential(scale_radius=0.34586814810813454, flux=1049961.0753931333, gsparams=None), jac=[[0.5398181795385592, 0.17977371761576255], [0.17977371761576255, 1.9123449870247538]], offset=galsim.PositionD(x=0.0, y=0.0), flux_ratio=1.0, gsparams=None)], gsparams=None), jac=[[0.5398181795385592, 0.17977371761576255], [0.17977371761576255, 1.9123449870247538]], offset=galsim.PositionD(x=0.0, y=0.0), flux_ratio=1.0, gsparams=None)

rmjarvis:
Thanks very much for the report Ismael. I found the bug and fixed it.  Turns out it was a bit of premature optimization -- skipping a step in the Sum and Convolution constructors that probably never leads to any speed up and led to this error.

[...]

PR #765, opened 2016-06-21
#763 Fix a bug in the Sum and Convolution constructors when the list has only one object.

rmjarvis:
@ismael2395 pointed out (cf. #763) a subtle bug where the str and repr of objects that are a Transform(Sum(Transform(obj))) were messed up.  

I tracked it back to an "optimization" we do in the Sum (and Convolution) constructors to skip much of the function if the list has only a single element.  But it turns out in this case to lead to the bug he found.

Seems like the easiest solution is to just not do that.  I'm sure this is never anything close to an actual optimization, so just let them both go through the motions of the full constructor even though some of the operations are trivial when there is only a single element.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,9 @@ Bug Fixes
   not quite, flat (#741)
 - Fixed a bug in chromatic parametric galaxy models based on COSMOS galaxies.
   (#745)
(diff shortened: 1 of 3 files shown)


rmandelb:
Looks good to me, thanks.

jmeyers314:
Looks good to me too.

PR #766, opened 2016-06-24
Quantizing the background image correctly in demo13

arunkannawadi:
This is a 1-line PR aimed to fix a minor bug in creating a background image that is subtracted from the galaxy images. The previous background image was quantized only once where as the galaxy images are quantized as multiple stages - integer photon counts, integer electron counts and quantization in ADU. Quantizing the background image only once could slightly underestimate the true background, which could have significant impact for galaxies with low flux.


rmjarvis:
Looks fine to me.

rmandelb:
Fine with me too.

Issue #764, opened 2016-06-21
boost 1.60 shared_ptr registration


jmeyers314:
I encountered the error 

```
TypeError: No to_python (by-value) converter found for C++ type: boost::shared_ptr<galsim::GSParams>
```

in dozens of unit tests when switching from boost 1.59.5 to boost 1.60.0.  This appears to be related to a change in the way the GSParams shared_ptr gets registered.  Inserting the following line at then end of `PyGSParams::wrap` solves the issue under boost 1.60:

```
            bp::register_ptr_to_python< boost::shared_ptr<GSParams> >();
```

but then I start getting the following warning (at least it's not an error!) back under boost 1.59.5:

```
RuntimeWarning: to-Python converter for boost::shared_ptr<galsim::GSParams> already registered; second conversion method ignored.
```

So probably there's a more elegant solution...

jmeyers314:
Just found relevant boost issue: https://github.com/boostorg/python/issues/56

rmjarvis:
I guess wrap that line in
```
#if BOOST_VERSION > 106000
    bp::register_ptr_to_python< boost::shared_ptr<GSParams> >();
#endif
```

PR #767, opened 2016-06-25
#764 boost shared ptr

jmeyers314:
Another very short PR to dodge an error that appears to only be present in boost version 1.60 concerning the registration of a boost python shared_ptr.  I branched off of #534 so I could test on python3 too.


pysrc/SBProfile.cpp
```diff
@@ -79,6 +79,9 @@ namespace galsim {
                 .def(bp::self == bp::other<GSParams>())
                 .enable_pickling()
                 ;
+#if BOOST_VERSION >= 106000 && BOOST_VERSION < 106100
```
rmjarvis: Do you really think we want the `< 106100` part of this?  Is this something they've already changed back in version 1.61?  I would have assumed that this line will be required for all versions >= 1.60.

jmeyers314: Yeah, I used boost 1.61.0 for my python3 test.  If I leave the `< 106100` out, I get the same warning as for 1.59.5: 

```
RuntimeWarning: to-Python converter for boost::shared_ptr<galsim::GSParams> already registered; second conversion method ignored.
```

So I think this is just something that sneaked past the boost devs for a single release.  We could almost just ignore it, except that boost-1.60 is the version that ships with the LSST conda stack at the moment, so I have a feeling others may run into this like I did.

rmjarvis: OK.  Thanks for checking that.  Maybe just add a comment in the code then to say that this is a workaround for what we believe is a bug in a single version of boost that was fixed in 1.61.  And reference this thread for more details.  Just in case it crops up again.

jmeyers314: Sure.  Added:

```
// Work around for "no to_python (by-value) converter found for C++ type: boost::shared_ptr<>"
// boost::python bug that seems to only exist for boost version 1.60.
// (GalSim Issue #764, https://github.com/GalSim-developers/GalSim/pull/767).
```

rmjarvis:
Great.  I'm happy to merge this.  Then we can start the final push to release version 1.4, asking people to test master on as many different systems as possible...

rmandelb:
I'm happy to merge this as well and start the v1.4 release prep.

jmeyers314:
Sounds good to me.  Let's get v1.4 out the door!

PR #769, opened 2016-07-01
Fix atmospheric screen size

jmeyers314:
I finally got around to comparing the simulated phase screen structure function out of AtmosphericScreen to an analytic prediction (as recommended by George Angeli and Bo Xin some weeks ago).  This turned up a small but important bug in the AtmosphericScreen constructor, where the screen size was tweaked to yield a good FFT size, but this wasn't properly reflected in the `.screen_size` attribute.  So I fixed that and added a couple of unit tests.

Also, here's the theory vs. simulated structure functions (D(r) = <|phi(x) - phi(x+r)|^2>) now:

unknown

Pretty decent agreement I think at scales far from the screen sampling interval (0.05 m) and screen size (100.0 m).

### DIFF
--- a/tests/test_phase_psf.py
+++ b/tests/test_phase_psf.py
@@ -67,6 +67,35 @@ def test_atm_screen_size():
 
 
 @timer
+def test_structure_function():
+    """Test that AtmosphericScreen generates approximately the right structure function for infinite
+    outer scale.
+    """
+    rng = galsim.BaseDeviate(4815162342)
+    r0_500 = 0.2
+    L0 = None
+    screen_scale = 0.05
+    screen_size = 100.0
+
+    # Theoretical pure Kolmogorov structure function (at 500 nm!):
+    D_kolm = lambda r: 6.8839 * (r/r0_500)**(5./3)
+
+    atm = galsim.AtmosphericScreen(screen_size=screen_size, screen_scale=screen_scale,
+                                   r0_500=r0_500, L0=L0, rng=rng)
+    phase = atm.tab2d.table.getVals()[:-1, :-1].copy()
+    phase *= 2 * np.pi / 500.0  # nm -> radians
+    im = galsim.Image(phase, scale=screen_scale)
+    D_sim = galsim.utilities.structure_function(im)
+
+    for r in [0.5, 2.0, 5.0]:  # Only check values far from the screen size and scale.
+        # We're only attempting to verify that we haven't missed a factor of 2 or pi or
+        # something like that here, so set the rtol below to be *very* forgiving.  Since the
+        # structure function varies quite quickly as r**(5./3), this is still a useful test.
+        np.testing.assert_allclose(D_kolm(r), D_sim(r), rtol=0.5,
+                                   err_msg="Simulated structure function not close to prediction.")
+
+
+@timer
 def test_phase_screen_list():
     """Test list-like behaviors of PhaseScreenList."""
     rng = galsim.BaseDeviate(1234)
@@ -387,6 +416,7 @@ def test_ne():
 if __name__ == "__main__":
     test_aperture()
     test_atm_screen_size()
+    test_structure_function()
     test_phase_screen_list()
     test_frozen_flow()
     test_phase_psf_reset()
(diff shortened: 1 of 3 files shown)


tests/test_phase_psf.py
```diff
@@ -52,8 +53,51 @@ def test_aperture():
...
 @timer
+def test_atm_screen_size():
+    """Test for consistent AtmosphericScreen size and scale."""
+    screen_size = 10.0
+    screen_scale = 0.1
+    atm = galsim.AtmosphericScreen(screen_size=screen_size, screen_scale=screen_scale)
+    # AtmosphericScreen will preserve screen_scale, but will adjust screen_size as necessary to get
+    # a good FFT size.
+    assert atm.screen_scale == screen_scale
+    assert atm.screen_size != screen_size
```
rmjarvis: Rather than just the `!=` test, maybe also confirm that `screen_size < atm.screen_size < 1.5 * screen_size`, which I think should be true based on how `goodFFTSize` works.

tests/test_phase_psf.py
```diff
@@ -52,8 +53,51 @@ def test_aperture():
...
+    phase = atm.tab2d.table.getVals()[:-1, :-1].copy()
+    phase *= 2 * np.pi / 500.0  # nm -> radians
+    im = galsim.Image(phase, scale=screen_scale)
+    D_sim = galsim.utilities.structure_function(im)
+
+    for r in [0.5, 2.0, 5.0]:  # Only check values far from the screen size and scale.
+        # We're only attempting to verify that we haven't missed a factor of 2 or pi or
+        # something like that here, so set the rtol below to be *very* forgiving.  Since the
+        # structure function varies quite quickly as r**(5./3), this is still a useful test.
+        np.testing.assert_allclose(D_kolm(r), D_sim(r), rtol=0.5,
```
rmjarvis: Just curious, but what is the error here for these values of r?  Maybe mention the values in a comment to motivate why rtol = 0.5 is necessary.

I'd at least add a print statement here giving D_kolm(r) and D_sim(r) so that people running from the command line (rather than via nosetests) can see the values directly.

rmjarvis:
Just minor comments.  Looks good.  We should probably get this into 1.4 as well I guess.

rmandelb:
Tests pass on both my systems.  I'm afraid I don't have time to go over the code before I leave for vacation, but I gather this is a small change so I recommend you go ahead without me on this one.

tests/test_phase_psf.py
jmeyers314: Yep.  I just replaced the old test since I think `!=` is implied in your suggested test.

tests/test_phase_psf.py
jmeyers314: Sure.  Here's the output of the print:

```
r   D_kolm   D_sim
0.5 31.7006305419 36.7609791124
2.0 319.522333717 322.01840719
5.0 1471.41292742 1274.34747692
```

So for these settings and random seed, it's actually good to about 15%.  Unfortunately, I don't have a good feeling for how consistent we expect these numbers to be, since I think that probably is sensitive to resolution, edge effects, and it's also stochastic.  I did add to the comment above the test:

```
        # We're only attempting to verify that we haven't missed a factor of 2 or pi or
        # something like that here, so set the rtol below to be *very* forgiving.  Since the
        # structure function varies quite quickly as r**(5./3), this is still a useful test.
        # For the parameters above (including the random seed), D_kolm(r) and D_sim(r) are actually
        # consistent at about the 15% level in the test below.  It's difficult to predict how
        # consistent they *should* be though, since the simulated structure function estimate is
        # sensitive to resolution and edge effects, as well as the particular realization of the
        # field.
```

jmeyers314:
>Just minor comments. Looks good. We should probably get this into 1.4 as well I guess.

Yes, I think that would be good.  Sorry for catching this so late!

>Tests pass on both my systems. I'm afraid I don't have time to go over the code before I leave for vacation, but I gather this is a small change so I recommend you go ahead without me on this one.

Yep.  I think I only changed one line of actual production code, all the rest are tests and utilities.  I'll go ahead and merge tonight if I don't hear anything more.

rmjarvis:
>  I'll go ahead and merge tonight if I don't hear anything more.

Looks good to me.  Please do.

PR #777, opened 2016-07-21
Add a makedirs helper function to deal with race conditions

esheldon:
when many parallel jobs are all trying to run `os.makedirs` on the same directory a race condition can occur; the directory is created between the `os.path.exists` check and when `makedirs` is called.  In that case an `OSError` is raised with error code 17 (File Exists).  This pull request adds a helper function to deal with this at the most basic level.

I'm just checking for the OSError and the File Exists error code.  I also make sure that, if the path does exist, it is in fact a directory.

One could get much fancier than this.


rmjarvis:
Looks good.  Thanks for this Erin.

Issue #772, opened 2016-07-13
Pupil plane sampling for WFIRST PSFs


arunkannawadi:
On trying to generate the PSF profiles for a given band, I get the following warning:

```
Input pupil plane image may not be sampled well enough!
Consider increasing sampling by a factor 98.131897, and/or check PhaseScreenPSF outputs for signs of folding in real space.
```
The sampling factor to be increased is ~100 ?! Here's a code snippet to get that warning:

```
import galsim
import galsim.wfirst as wfirst
PSFs = wfirst.getPSF(SCAs=7, wavelength=1840.)
```

jmeyers314:
Hi Arun,

This is part of the new API for OpticalPSF, which has stricter requirements for how to set the scale (in meters) for an image of the pupil plane.  Sorry that I didn't think about how this might affect the wfirst module.  From the `galsim.Aperture` documentation:

```
The pixel scale of the pupil plane can be specified in one of three ways.  In descending order 
of priority, these are:
      1.  The `pupil_plane_scale` keyword argument (units are meters).
      2.  The `pupil_plane_im.scale` attribute (units are meters).
      3.  If (1) and (2) are both None, then the scale will be inferred by assuming that the
          illuminated pixel farthest from the image center is at a physical distance of self.diam/2.
```

In this case, the image file of the WFIRST focal plane has a scale of 1.0 (I'm not sure if this is from the WCS or if it's a default), so the 4096x4096 image is interpreted as having physical extent of ~4km x ~4km.  While it would certainly be cool to have a ~km aperture optical/infrared telescope in space, I don't think this is in the plans for WFIRST!

Of the three ways listed above for setting the correct scale, I think in this case setting it as an image attribute (via the image file header WCS) makes the most sense.  (Anyone know if it's possible to indicate in the FITS header that the WCS units are in meters?)

That said, it would probably be a good idea to check in the code if the inferred diameter coming from the pupil image + scale at least roughly matches the diameter specified as an argument, and report a helpful warning if they don't.

Let me know if you want to tackle this one, or if you'd like me to.  If you want me to though, I'll need to know the precise pixel scale for the WFIRST pupil plane image.

rmjarvis:
Hm.  Seems like we should probably resolve this for version 1.4.

The WFirst mirror diameter is 2.4 m.  And the pupil image has non-zero values from x=1202 to x=2897. So the proper scale here is probably 2.4 m / 1696 = 0.00141509.

[...]

PR #778, opened 2016-07-22
#772

jmeyers314:
This PR fixes an issue @arunkannawadi noted regarding the generation of WFIRST PSFs under the new optics framework.  The new framework uses the `.scale` attribute of `galsim.Image`s used to create a pupil plane to set the scale of the pupil plane in meters.  The old framework simply ignored this keyword.  During the course of fixing this, @rmjarvis realized that the combination of `pupil_plane_im` and `lam_over_diam` was annoying (and imprecise) to get right, so we're disallowing it now unless someone raises an issue.  Probably most of the time when someone has a `pupil_plane_im` with definite scale, they also know the telescope diameter, and can therefore just use the `lam` and `diam` kwargs separately, which is still allowed.

### DIFF
--- a/examples/demo13.py
+++ b/examples/demo13.py
@@ -456,7 +456,8 @@ def main(argv):
 
         # Since many people are used to viewing background-subtracted images, we provide a
         # version with the background subtracted (also rounding that to an int).
-        tot_sky_image = (sky_image.quantize() + round(dark_current))/wfirst.gain
+        sky_image.quantize()
+        tot_sky_image = (sky_image + round(dark_current))/wfirst.gain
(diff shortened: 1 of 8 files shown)


rmjarvis:
@jmeyers314 I added back the tests you commented out, but made them work.  The config one in particular I didn't want to lose, since it's the only test there of pupil_plane_im.  But please check that that changes to test_optics are ok with you.  If so, I'm fine with merging this.

jmeyers314:
>But please check that that changes to test_optics are ok with you.

Yep.  Looks fine to me.

Issue #780, opened 2016-08-02
LSST module testing error


jmeyers314:
As part of my v1.4 checks, I tried running the lsst tests on the master branch and got the following test error:

```
ERROR: Test that you can pass a camera from one WCS to another
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/josh/src/GalSim/tests/galsim_test_helpers.py", line 412, in f2
    result = f(*args, **kwargs)
  File "/Users/josh/src/GalSim/tests/test_lsst.py", line 799, in test_passing_camera_by_hand
    self.assertEqual(str(ww[0].args[0]),
AttributeError: 'WarningMessage' object has no attribute 'args'
```

I'm guessing something in the LSST API changed and this needs to follow suit?  (I re-installed the sims from conda just before running the GalSim tests.)  @danielsf, can you take a look?  Also, should the `self.assertEqual`s in this method be inside the `with` block?

Finally, I'm also seeing a deprecation warning:

```
/Users/josh/src/GalSim/tests/test_lsst.py:377: DeprecationWarning: This function is deprecated. Please call randint(0, 3 + 1) instead
  for ii in rng.random_integers(0,3,n_samples)]
```

Looks like this would also be pretty easy to address now.

danielsf:
What version of python are you running?

jmeyers314:
Anaconda 2.7.11

danielsf:
This is not an LSST problem.  It is a shift in python warnings API between python 2 and python 3.  See commit 229045e5.  I will look into whether there is a python-version-agnostic way to fix it.

danielsf:
I can't even get as far as you did running test_lsst.py
```
D-173-250-159-162:GalSim danielsf$ python tests/test_lsst.py
Traceback (most recent call last):
  File "tests/test_lsst.py", line 24, in <module>
    import galsim
  File "/Users/danielsf/GalSim/galsim/__init__.py", line 100, in <module>
    from .table import LookupTable, LookupTable2D
  File "/Users/danielsf/GalSim/galsim/table.py", line 530, in <module>
    _galsim._LookupTable2D.__getinitargs__ = lambda self: \
AttributeError: 'module' object has no attribute '_LookupTable2D'
```

danielsf:
oops.  Sorry.  Forgot that this wasn't pure python

PR #781, opened 2016-08-02
#780

danielsf:
Someone running python 3 should check that I did not break that compatibility.

### DIFF
--- a/tests/test_lsst.py
+++ b/tests/test_lsst.py
@@ -796,22 +796,29 @@ class LsstWcsTestCase(unittest.TestCase):
             wcs1 = LsstWCS(galsim.CelestialCoord(0.0*galsim.degrees, 0.0*galsim.degrees),
                            self.rotation, chip_name='R:0,1 S:1,1', camera=self.wcs.camera)
 
-        self.assertEqual(str(ww[0].args[0]),
-                         "The camera you passed to LsstWCS does not have the same\n"
-                         "pointing and rotation angle as you asked for for this WCS.\n"
-                         "LsstWCS is creating a new camera with the pointing and\n"
-                         "rotation angle you specified in the constructor for LsstWCS.")
+        expected_message = "The camera you passed to LsstWCS does not have the same\n" \
+                           "pointing and rotation angle as you asked for for this WCS.\n" \
+                           "LsstWCS is creating a new camera with the pointing and\n" \
+                           "rotation angle you specified in the constructor for LsstWCS."
 
+        if hasattr(ww[0], 'args'):
+            self.assertEqual(str(ww[0].args[0]), expected_message)
+        else:
+            self.assertEqual(str(ww[0].message), expected_message)
 
         with warnings.catch_warnings(record=True) as ww:
             wcs1 = LsstWCS(self.pointing, 49.0*galsim.degrees,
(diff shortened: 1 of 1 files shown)


jmeyers314:
Thanks @danielsf , this all looks good to me; tests pass in python 2.7.  I don't have a python3 installation of the lsst stack to test, (is that on conda too?), but I don't think we need to hold this up for that.

danielsf:
Fair point.  There isn't a python 3 version of the stack yet.  We hope to be python 3 compliant by the end of the fall.

tests/test_lsst.py
```diff
@@ -374,7 +374,7 @@ def test_pupil_coords_from_pixel_coords(self):
         chip_name_possibilities = ('R:0,1 S:1,1', 'R:0,3 S:0,2', 'R:4,2 S:2,2', 'R:3,4 S:0,2')
 
         chip_name_list = [chip_name_possibilities[ii]
-                          for ii in rng.random_integers(0,3,n_samples)]
+                          for ii in rng.randint(0,3,n_samples)]
```
rmjarvis: I'm pretty sure this is wrong.  Or at least it is a change in what the test is doing to something now what you meant to do.

[numpy.random_integers](http://docs.scipy.org/doc/numpy/reference/generated/numpy.random.random_integers.html) selects from a closed range, [0,3] here, while [numpy.randint](http://docs.scipy.org/doc/numpy/reference/generated/numpy.random.randint.html) selects from a half-open range, [0,3) here.

Since the former is now deprecated (why, numpy?), I think you want to change the argument to `(0,4,n_samples)` for `randint`.

tests/test_lsst.py
```diff
@@ -796,22 +796,29 @@ def test_passing_camera_by_hand(self):
...
-                         "rotation angle you specified in the constructor for LsstWCS.")
+        expected_message = "The camera you passed to LsstWCS does not have the same\n" \
+                           "pointing and rotation angle as you asked for for this WCS.\n" \
+                           "LsstWCS is creating a new camera with the pointing and\n" \
+                           "rotation angle you specified in the constructor for LsstWCS."
 
+        if hasattr(ww[0], 'args'):
+            self.assertEqual(str(ww[0].args[0]), expected_message)
+        else:
+            self.assertEqual(str(ww[0].message), expected_message)
```
rmjarvis: I think this is my error.  The exceptions changed from allowing `message` in python 2 to requiring the use of `args` in python 3.  But the warning, as far as I can tell, still use `message` in python 3.  So I think this can just use the message check, without the args part.

rmjarvis:
@danielsf, please don't merge pull requests yourself without someone signing off that it is ok to merge.  We generally want two people to check over a PR before we merge, and as you can see, I have a few comments on this one.  The typical pattern is to either ask if it is ok to merge, or give a timeline for when you plan to merge so people can get any comments in before then.

danielsf:
Sorry.  I thought, when Josh signed off on the PR, that was an "okay to merge" statement.

tests/test_lsst.py
danielsf: Looks like you're right.  I don't think it really matters in this case, though.  All I am doing is drawing a random sample of chip names from `chip_name_possibilities`.  The code now ignores the last entry in `chip_name_possibilities`, which doesn't really affect the usefulness of the test.

PR #784, opened 2016-08-03
HSM: replace assertion with exception

PaulPrice:
This is an exceptional condition that can occur sometimes
(because it certainly does occur!), rather than something
that should bring the world to a screaming halt.

### DIFF
--- a/src/hsm/PSFCorr.cpp
+++ b/src/hsm/PSFCorr.cpp
@@ -728,7 +728,9 @@ namespace hsm {
         if (iy2 > ymax) iy2 = ymax;
         dbg<<"y1,y2 = "<<y1<<','<<y2<<std::endl;
         dbg<<"iy1,iy2 = "<<iy1<<','<<iy2<<std::endl;
-        assert(iy1 <= iy2);
+        if (iy1 > iy2) {
+             throw HSMError("Bounds don't make sense");
+        }
 
         //
         /* Use these pointers to speed up referencing arrays */


rmjarvis:
I'm fine with this change, but I did look quickly at how this exceptional circumstance can occur.  It seems that it would only happen when the initial y2, defined [here](https://github.com/GalSim-developers/GalSim/pull/784/files#diff-01d4f765f0fa15bf6b13a3ab7d184ae5L722) is < 0.5 and y0 is near 0.5 itself, so y1 and y2 end up being something like 34.2, 34.8.  Then the ceil and floor make them cross.

@rmandelb Is this a sign that the ceil and floor should be the opposite of what they currently are?  Then in the case I laid out, `iy1, iy2` would end up 34, 35 rather than 35, 34.  Probably in other more normal cases, this would also be preferred -- to expand the range rather than shrink it.  Or is there some reason that the current functionality is correct, and we want to just raise an exception as Paul proposes?

PaulPrice:
Cross reference to LSST ticket: [DM-7114](https://jira.lsstcorp.org/browse/DM-7114).

rmandelb:
>  Is this a sign that the ceil and floor should be the opposite of what they currently are? 

I don't think so.  Probably it's a sign that something slightly different from the current code should be done, but if you reverse them you will run into problems.  The goal of these statements as I recall (it's been a long time since I dug into this part of the code) is to identify a region lying within an ellipse of some radius set by ```max_moment_nsig2``` and orientation; if you swap them, you're outside the ellipse which by definition is not what you want.

rmandelb:
Oh, and I should have said, I'm fine changing this from assertion to exception.

rmjarvis:
OK.  In that case, let's just switch to an exception then.  Asserts should be reserved for things that cannot occur if the code is correct, rather than flagging bad input.  And since, as Paul points out, this can occur, it should be an exception rather than an assert.  Any objection to merging this?

PaulPrice:
Reviewer on the LSST side pointed out that the `assert` could be removed completely, but this patch only adds the exception.  I'm happy to submit a revised patch if desired.

rmjarvis:
If we omit the assert entirely, we'll eventually get to a division by zero, which will at least lead to nans if not worse problems.  Better to raise the exception here I think.

PaulPrice:
But if the exception is raised, then there's no need for the `assert`, right?

rmjarvis:
Oh, I see what you mean.  Right.  The assert can be removed once we have the exception.

PaulPrice:
Got rid of the `assert`.

PaulPrice:
@rmandelb previously said "I'm fine changing this from assertion to exception."  So I think we're good to merge, yes?

rmjarvis:
Thanks for this, Paul!

PR #788, opened 2016-08-14
INSTALL: Update OS X instructions with Homebrew

bfontaine:


rmjarvis:
@dstndstn Dustin, could you look over this change?  AFIK, you were previously (maybe still currently?) the curator of GalSim on Homebrew, so perhaps you can confirm that this is correct?

rmandelb:
@bfontaine - I just tried this out to see how it works on my Mac.  It appears to be installing GalSim v1.3 (plus dependencies like TMV).  Given that we are now releasing v1.4, is it possible to update it to install v1.4?

dstndstn:
Galsim is in the homebrew-science tap, right?  Don't you need to tap it explicitly?  Or maybe "brew install" is smart enough to find it automatically.

https://github.com/Homebrew/homebrew-science/blob/master/galsim.rb

It appears that you "borrowed" my gal-sim.rb formula,
https://github.com/camphogg/homebrew-science/blob/master/gal-sim.rb
without attribution or license.  That is generally frowned upon.

rmjarvis:
Also, scanning through the script, I note a few errors (at least for use with 1.4 -- some of these were fine with 1.3)

- WITH_OPENMP is not an option anymore.
- Additional dependencies: future and PyYaml.  Both installable via pip.
- scons tests should really be done after scons install, otherwise the tests could be using a previous installation of galsim
- The tests don't seem to include any check that the galsim executable is in the right place and working properly.  It's probably fine, but given your other tests, this would be of equal priority to the other things you test.  (Also the galsim_download_cosmos executable to be even more complete.)

bfontaine:
> Galsim is in the homebrew-science tap, right? Don't you need to tap it explicitly? Or maybe "brew install" is smart enough to find it automatically.

It’s an official tap so you don’t need to tap it explicitely. But you need to give its full path if it’s not already tapped: `brew install homebrew/science/galsim`. I updated this PR to reflect that.

> It appears that you "borrowed" my gal-sim.rb formula

No, you added it yourself in https://github.com/Homebrew/homebrew-science/pull/545.

dstndstn:
Ahhhh, please accept my apologies, I was wrong.  For the github.rb file, Github lists "5 contributors" (not incl me), and I also don't show up in the History.  I do show up in Blame though ;)

So, pretty much ignore everything I said, with apologies.

bfontaine:
Thank you @rmjarvis for the feedback; I’m working on an update that fixes these issues.

rmjarvis:
Any update on this?  This has been dormant for a while...

bfontaine:
No; I started to edit the formula but forgot about it since and I don’t currently have a lot of time to work on that 😞

rmjarvis:
I'm assuming from the lack of activity here that this effort has been abandoned, so I'm going ahead and closing it.  Feel free to reopen or resubmit a new PR if you get a chance update this based on the above comments.

dstndstn:
Sorry for dropping the ball.  I thought this change looked good -- we should definitely drop the camphogg/science tap in favor of the upstream homebrew/science repository.

I will submit a PR to update the galsim version in homebrew to 1.4.2.

Apologies for my delay and foolishness here.

Issue #782, opened 2016-08-02
3D disk galaxy models


rmandelb:
For some Euclid simulations, the group at Oxford (Lance Miller, Niraj Welikala) has been building 3D bulge+disk galaxies.  The bulge can be represented by GalSim already, but the 3D disk cannot be.  They model it as a disk of finite thickness when viewed edge-on, with a sech^2(z) surface brightness distribution as observed for real edge-on galaxies (the radial part of the profile is indeed exponential as you move along the disk at fixed z, so it's the sech^2(z) aspect that would be new to GalSim). Initially they did this via photon-shooting but they recently began representing it using DFTs. Lance has offered some C code that could be used to implement the 3D disk models in GalSim.  He does not plan to implement it in GalSim himself.

Unlike many of our profiles, the input parameter would be the disk inclination, with the ellipticity being a derived parameter.   I'm mentioning this not as a problem, but rather as an interesting feature.

Any thoughts, questions, concerns, etc.?

brgillis:
Unless someone already familiar with working on GalSim is keen to pick this up, I can try implementing this.

The use of inclination as an input parameter is actually more convenient for simulations like what I'm working on than ellipticity is. It might also be a good idea to implement a similar class for bulges, so inclination can be passed directly to GalSim rather than requiring a calculation of the resulting ellipticity.

rmandelb:
If you want to give this a try, then my suggestion is that you go through the C++ layer and look at the examples of base class implementations, come up with a proposal for how/where to add this in the C++ and python layers, and post it here.  I'm happy to iterate with you and make suggestions about code design before you start trying to actually incorporate it.

brgillis:
Alright, I've given this some thought about how to implement it now. What we have is a profile with an analytic form in Fourier space, but no analytic form in real space (aside from of course the Fourier transform integral). I can think of a couple ways to implement this. In both ways, implement in the C++ layer (like SBExponential.h/.cpp) and use the analytic form for kValue and fillKValue.

1. For xValue, perform the Fourier integral each time the method is called to calculate for that point. For fillXValue, perform an FFT on an array of k values.

2. Store a lookup table of x values (if fillXValue is called first, store a copy of the matrix made here, otherwise create one with an FFT of k values), and interpolate when xValue is called.

[...]

Issue #755, opened 2016-05-31
replace generic exceptions with galsim exceptions


esheldon:
There are many errors that should be recoverable but galsim throws a generic
error.  This is a problem because it may not be clear from where this error
originates.  One cannot distinguish the recoverable from non-recoverable
errors.

For example, a pipeline may run many calls to galsim methods; tens is not
unusual. Each of these may throw a `RuntimeError` or `ValueError`.   The
pipeline, and galsim itself, will intersperse other python function calls with
galsim calls, many of which may throw the same errors.

The difference is that most `RuntimeError` or `ValueError` thrown by the python
standard library are not recoverable, they signify a bug in the program, and
the program should crash and the user should fix the code.

But many times when galsim throws such an error it is not a bug, galsim simply
does not know what to do with the inputs or has some internal numerical
restrictions that are not met.  Often one can just re-call the method with
modified parameters, or otherwise deal with this an move on.  If I _knew_ the
exception was a galsim exception I could write some robust code to deal with
the situation.

An example came up today when the fft code threw a `RuntimeError` because the
sheared image would require too large an FFT.  This is not a bug, I can deal
with this quite cleanly at run time, but I can't tell if this `RuntimeError`
is coming from galsim or not.

I propose replacing some uses of `RuntimeError` or `ValueError` with something
like `GalsimValueError`, `GalsimRangeError` or whatever, inheriting from a base
`GalsimError` class.    These can then be caught and dealt with robustly.

rmjarvis:
I don't have a problem with this.  It would take some time to go through and figure out which exceptions ought to have new GalSim-specific names and which ones are bugs that should keep the current names.

As a first pass, I'd wager that all the `RuntimeError`s could get a GalSim-specific name.  But the rest would require some thought to determine if they are pointing to coding errors or just a problem with the input.

esheldon:
I'll spend some time on this

esheldon:
Here are two I ran into recently

This error I think comes from include/galsim/FFT.h
```
RuntimeError: FFT error: KTable size <=0
```

I'm not sure where this originates, but most likely from
src/SBSersic.cpp or src/SBMoffat.cpp
```
RuntimeError: Error in function boost::math::tgamma<d>(d): numeric overflow
```

esheldon:
The above occurred in a recent run of my metacalibration code.  Here are some that I have seen in the past

This occurred processing real DES data, from interpolatedimage.py
```
[...]

PR #791, opened 2016-08-23
#782

brgillis:
Implementation of an inclined exponential disk. Presently it's only analytic in k-space, and so photon-shooting for it is not yet enabled.


devel/external/.gitignore
```diff
@@ -1 +1,4 @@
 temp*
+/test_photon_vs_fft/
+/test_reconvolution/
+/test_sersic_highn/
```
rmjarvis: Why are you ignoring these?  We want to keep these in the repo.  Please revert this.

devel/external/test_photon_vs_fft/cosmos_sersics_sample_N300.asc
```diff
@@ -0,0 +1,301 @@
+# ident  n_sersic  half_light_radius [arcsec] |g|
```
rmjarvis: Why did you replace the symlink with a copy of the file (here and several more below in these devel/external directories)?  Please revert.  No need to clutter the repo with multiple copies of these files.  The symlinks work fine.

galsim/__init__.py
```diff
@@ -111,7 +111,7 @@
 
 # GSObject
 from .base import GSParams, GSObject, Gaussian, Moffat, Airy, Kolmogorov, Pixel, Box, TopHat
-from .base import Exponential, Sersic, DeVaucouleurs, Spergel
+from .base import Exponential, Sersic, DeVaucouleurs, Spergel, InclinedExponential
```
rmjarvis: This should probably get its own file, rather than add another class to base.  This isn't really a profile that we will consider a "base" class in GalSim, so better to have a new file for it.  (Spergel probably also ought to have gotten its own file as well for the same reason...)

galsim/base.py
```diff
@@ -2892,6 +2892,119 @@ def __str__(self):
...
+    
+    In this implementation, the profile is inclined along the y-axis. This means that it will likely
+    need to be rotated in most circumstances.
+    
+    At present, this profile is not enabled for photon-shooting.
+    
+    Initialization
+    --------------
+
+    @param i                The inclination angle, i
```
rmjarvis: Not `i`.  Parameters should be at least somewhat self-documenting via their names.  `inclination` would probably be appropriate.

Also, should mention here that it is an `Angle`.

galsim/base.py
```diff
@@ -2892,6 +2892,119 @@ def __str__(self):
...
+    _opt_params = { "flux" : float }
+    _takes_rng = False
+
+    def __init__(self, i, scale_radius, scale_height,
+                 flux=1., gsparams=None):
+        
+        if not isinstance(i, galsim.Angle):
+            raise TypeError("Input i should be an Angle")
+        
+        i_rad = i.rad()
```
rmjarvis: There's no need for this.  `Angle`s are python-wrapped, so you should be able to pass in the `Angle` object to C++ and use it as such there and keep the unit safety afforded by the `Angle` class.

galsim/base.py
```diff
@@ -2892,6 +2892,119 @@ def __str__(self):
...
+
+    def __eq__(self, other):
+        return ((isinstance(other, galsim.InclinedExponential) and
+                 (self.i == other.i) and
+                 (self.scale_radius == other.scale_radius) and
+                 (self.scale_height == other.scale_height) and
+                 (self.flux == other.flux) and
+                 (self._gsparams == other._gsparams))
+                or
+                (isinstance(other, galsim.Exponential) and
```
rmjarvis: I don't think we want to declare an `InclinedExponential` with `i==0` "equal" to an `Exponential`.  Just keep the first part of this function.

include/galsim/SBInclinedExponential.h
```diff
@@ -0,0 +1,98 @@
...
+        double getI() const;
+
+        /// @brief Returns the scale radius r0 of the disk profile
+        double getScaleRadius() const;
+
+        /// @brief Returns the scale height r0 of the disk profile
+        double getScaleHeight() const;
+
+        /// @brief Returns the truncation radius
+        double getTrunc() const;
```
rmjarvis: not applicable

include/galsim/SBInclinedExponentialImpl.h
```diff
@@ -0,0 +1,218 @@
...
+#include "SBProfileImpl.h"
+#include "SBInclinedExponential.h"
+#include "LRUCache.h"
+#include "Table.h"
+
+namespace galsim {
+
+    /// @brief A private class that caches the needed parameters for each inclined exponential angle 'i'
+    class InclinedExponentialInfo
+    {
```
rmjarvis: I don't think there is any use for this class.  As far as I can tell, you don't have any complex calculations to perform that will be true for all `InclinedExponential` objects.  Mostly they will have different values of `h_tani_over_r`, so there would need to be a separate Info object for each Impl object.  Better to just combine these into a single class, `SBInclinedExponentialImpl`.

include/galsim/SBInclinedExponentialImpl.h
```diff
@@ -0,0 +1,218 @@
...
+
+        double xValue(const Position<double>& p) const;
+        std::complex<double> kValue(const Position<double>& k) const;
+
+        double maxK() const;
+        double stepK() const;
+
+        void getXRange(double& xmin, double& xmax, std::vector<double>& splits) const
+        {
+            splits.push_back(0.);
```
rmjarvis: 0 isn't much of a special point here.  Should remove this line.  (And in getYRange).

include/galsim/SBInclinedExponentialImpl.h
```diff
@@ -0,0 +1,218 @@
...
+        // Helper functions used internally:
+        /* NYI
+        void buildIFT() const;
+        */
+    };
+
+    class SBInclinedExponential::SBInclinedExponentialImpl : public SBProfileImpl
+    {
+    public:
+        SBInclinedExponentialImpl(double i, double scale_radius, double scale_height, double flux,
```
rmjarvis: As mentioned above, `i` should (a) have a better name and (b) be an `Angle` type, not double.

src/SBInclinedExponential.cpp
```diff
@@ -0,0 +1,367 @@
...
+            _i = sbp::maximum_i;
+        }
+        else
+        {
+            // Normal setup here
+            _i = i;
+            _h_tani_over_r = scale_height*std::tan(i)/scale_radius;
+            _r0_cosi = _r0*std::cos(i);
+            _inv_r0_cosi = 1./(scale_radius*std::cos(i));
+        }
```
rmjarvis: A few points about this bit.

1. the logic here is a bit wrong.  You recompute `_h_tani_over_r`, `_r0_cosi`, and `_inv_r0_cosi` when `i` is within the allowed limits, so the calculation you did in the initializer list is still good.  But in the case where you changed `_i`, you don't recalculate, even though they are incorrect.

2. I'd recommend changing the limit to be on `cos(i)` rather than `i`.  The problem with `i` near pi/2 is that cos(i) == 0, so `inv_r0_cosi` is undefined.  Probably better to just set a minimum on `cosi` instead.

3. I don't think you need all the adjustments of the range then.  Just take the absolute value of cos(i) and tan(i).

4. Once `i` is switch to being an `Angle`, note that you can do the trig calls with `i.cos()` in lieu of `std::cos(i.rad())`.  It looks like we never wrote free functions `cos(i)`, etc. for `Angle` arguments.  That would be even nicer looking and of course easy to do, so if you feel like adding those, please go ahead.

src/SBInclinedExponential.cpp
```diff
@@ -0,0 +1,367 @@
...
+        /* Shooting NYI
+        _shootnorm = _flux * _info->getXNorm(); // For shooting, we don't need the 1/r0^2 factor.
+        _xnorm = _shootnorm * _inv_r0 * _inv_r0 ;
+        dbg<<"norms = "<<_xnorm<<", "<<_shootnorm<<std::endl;
+        */
+    }
+
+    double SBInclinedExponential::SBInclinedExponentialImpl::xValue(const Position<double>& p) const
+    {
+        throw std::logic_error("Real-space expression. of SBInclinedExponential NYI.");
```
rmjarvis: This isn't a logic error.  Probably a std::runtime_error would be appropriate.  At least until we do something about issue #755.

src/SBInclinedExponential.cpp
```diff
@@ -0,0 +1,367 @@
...
+    double InclinedExponentialInfo::kValue(double kx, double ky) const
+    {
+        // Calculate the base value for an exponential profile
+
+        double kysq = ky*ky;
+        double ksq = kx*kx + kysq;
+        double res_base;
+        if (ksq > _ksq_max)
+        {
+            res_base = 0.;
```
rmjarvis: No need to calculate `res_conv` in this case.  Just return early.

src/SBInclinedExponential.cpp
```diff
@@ -0,0 +1,367 @@
...
+        // Fraction excluded is thus (1+R) exp(-R)
+        // A fast solution to (1+R)exp(-R) = x:
+        // log(1+R) - R = log(x)
+        // R = log(1+R) - log(x)
+        double logx = std::log(gsparams->folding_threshold);
+        double R = -logx;
+        for (int i=0; i<3; i++) R = std::log(1.+R) - logx;
+        // Make sure it is at least 5 hlr
+        // half-light radius = 1.6783469900166605 * r0
+        const double hlr = 1.6783469900166605;
```
rmjarvis: This is the Exponential value, which is presumably not true here.  This merits at least a bit more discussion about the half-light-radius values of these inclined disks.  Are they really similar to a face-on exponential?  My guess is no, but that this a conservative choice (i.e. the inclined disks always have a smaller hlr than the face on value).

src/SBInclinedExponential.cpp
```diff
@@ -0,0 +1,367 @@
...
+        _maxk(0.), _stepk(0.)
+    {
+        dbg<<"Start InclinedExponentialInfo constructor for h_tani_over_r = "<<h_tani_over_r<<std::endl;
+
+        assert(h_tani_over_r >= 0); // Should only ever have non-negative
+
+        // For large k, we clip the result of kValue to 0.
+        // We do this when the correct answer is less than kvalue_accuracy.
+        // (1+k^2 r0^2)^-1.5 = kvalue_accuracy
+        _ksq_max = (std::pow(gsparams->kvalue_accuracy,-1./1.5)-1.);
```
rmjarvis: Is this calculation still correct?  I suspect the inclined disks need to go to higher k than face-on disks.  More power at small scales, which is higher k values.  Is the `_ksq_max` value that we used for Exponential still the relevant value here?  I would assume not, but maybe it's close?  At least do a bit more investigation of where the k values drop below `kvalue_accuracy` for various choices of inclination angle to see if you can come up with a heuristic or approximate choice for where it's ok to cut it off to zero.

src/SBInclinedExponential.cpp
```diff
@@ -0,0 +1,367 @@
...
+        double R = -logx;
+        for (int i=0; i<3; i++) R = std::log(1.+R) - logx;
+        // Make sure it is at least 5 hlr
+        // half-light radius = 1.6783469900166605 * r0
+        const double hlr = 1.6783469900166605;
+        R = std::max(R,gsparams->stepk_minimum_hlr*hlr);
+        _stepk = M_PI / R;
+        dbg<<"stepk = "<<_stepk<<std::endl;
+
+        _maxk = std::sqrt(_ksq_max);
```
rmjarvis: This isn't correct.  The `maxk` value should be based on `gsparams->maxk_threshold`, not `kvalue_accuracy`.  Probably, it's too large in fact (which will significantly slow down some uses of this class), since `maxk_threshold` is usually a couple orders of magnitude larger than `kvalue_accuracy`.  But also, the same points I made above about `ksq_max` apply here.  Just do the same calculation using `maxk_threshold` as the target value.

tests/test_inclined_exponential.py
```diff
@@ -0,0 +1,102 @@
...
+oversampling = 1.0
+
+@timer
+def test_inclined_exponential():
+    """Test that the inclined exponential profile matches the results from Lance Miller's code."""
+    
+    for inc_angle, scale_radius, scale_height, pos_angle in zip(image_inc_angles,image_scale_radii,
+                                                                image_scale_heights,image_pos_angles):
+        image_filename = "galaxy_"+inc_angle+"_"+scale_radius+"_"+scale_height+"_"+pos_angle+".fits"
+        image = galsim.fits.read(image_filename, image_dir)
```
rmjarvis: The regression tests here are good.  You should say where these comparison files come from though.  Are these from some external code?  You should probably have either a README in the inclinded_exponential_images directory saying how the files were generated, possibly with the code you used to do so if appropriate.

tests/test_inclined_exponential.py
```diff
@@ -0,0 +1,102 @@
...
+        image_core = image.array[image_ny//2-2:image_ny//2+3, image_nx//2-2:image_nx//2+3]
+        test_image_core = test_image.array[image_ny//2-2:image_ny//2+3, image_nx//2-2:image_nx//2+3]
+        
+        ratio_core = image_core / test_image_core
+        
+        # galsim.fits.write(test_image,"test_"+image_filename,image_dir)
+        
+        np.testing.assert_array_almost_equal(ratio_core, np.mean(ratio_core)*np.ones_like(ratio_core), decimal = 2,
+                                             err_msg = "Error in comparison of inclined exponential profile to samples.",
+                                             verbose=True)
```
rmjarvis: Need to add some more testing aside from just the regression tests though.

1. Need to check the accuracy of the repr, pickling etc.  cf. uses of the function `do_pickle` in other test files.
2. Some tests of edge-on inclination == Pi/2.
3. Should test close, but not quite edge-on.  Probably on both sides of the magic cut-off you impose on i.
4. Should test that `i==0` is functionally equivalent to an `Exponential`.
5. Check some basic properties like that kValue at k=maxK() is below `maxk_threshold` and that less than `folding_threshold` fraction of the light falls outside of r=Pi/stepK() for a few inclination angles.
6. Sanity checks like kValue(0,0) == flux, centroid() == (0,0), total flux draw with large enough image == flux.

include/galsim/SBInclinedExponentialImpl.h
brgillis: I left this in in case we want to enable photon-shooting at a later date, in which case caching would be useful (as it would require doing an IFT and then interpolating based off of the results to get x values). I could remove the use of this class, but that might be a waste of time if it'll just be reimplemented later

devel/external/test_photon_vs_fft/cosmos_sersics_sample_N300.asc
brgillis: I don't recall deliberately doing this. My guess is that my git process might have done this while changing branches at some point.

include/galsim/SBInclinedExponentialImpl.h
rmjarvis: I don't see any reason it would be useful for photon shooting.  You need a different IFT for each object don't you?  The Info classes are intended for calculations that are common to all or many objects.  e.g. for Sersic there are a bunch of calculations that need to be done once for each value of n.  Since many applications repeat the same n many times, it is useful to cache these values once.

As far as I can tell, the InclinedExponential calculations need to be redone for each value of h_tani_over_r, and there is no reason to expect these to be repeated very often in typical use cases.

rmandelb:
@brgillis - I'm currently in Japan for a meeting, and I see Mike has already given you a lot of helpful feedback.  I'm going to wait to look at this until after you've addressed his comments.  Please comment on this thread when you think you're at that stage.  Thanks!

include/galsim/SBInclinedExponentialImpl.h
brgillis: "As far as I can tell, the InclinedExponential calculations need to be redone for each value of h_tani_over_r" - This is correct.

As for how it might be useful, this is my imagined use case: Many simulated galaxies are needed, planned to be generated via photon shooting, with varying inclination angles (but the same h/r). If inclination angles are discretized to the nearest degree, then h_tani_over_r values will end up being reused. This means that the IFT would only need to be calculated 90 times at most (in this example), rather than once per galaxy. This could result in significant time savings.

Granted, this does depend on a few assumptions:
* The user would specifically want to do this with photon shooting
* Discretizing the inclination angle (or h_tani_over_r values) is acceptable
* There isn't a better way to do this

src/SBInclinedExponential.cpp
brgillis: For what we're using it for here (calculating _stepk), I think a slightly-conservative choice like this is fine. We're not outputting the hlr at any point, so I don't think we need to worry about it. If hlr is desired in the future (with some definition for it for a non-axisymmetric profile), perhaps we can worry about it then.

rmjarvis: I agree that this is probably conservative and fine.  Mostly, I think this needs an inline comment explaining that this hlr value is the value for Exponential, not the current profile, and that the real hlr is smaller, so this choice is conservative.  The current inline comment that you copied from the Exponential class is simply inaccurate in this context.  

Similarly, above you say "assuming that this is similar enough to an exponential disk", but really the salient point is that the flux enclosed in a radius R will be >= the flux from an exponential, so this is a conservative choice for that calculation as well.  

Both of these points should be explained in comments for the next person who comes along to look at this code.

brgillis:
@rmandelb Alright, I believe I've addressed all of @rmjarvis's comments. Please let me know of anything else you notice in the current version that needs to be addressed.

rmjarvis:
Some of the symlinks are still messed up.  If you look at the Files Changed tab, you should be able to confirm that everything there is stuff that you meant to change.

galsim/base.py
```diff
@@ -396,7 +398,7 @@ def calculateHLR(self, size=None, scale=None, centroid=None, flux_frac=0.5):
         offset = galsim.PositionD(0.2, 0.33)
         im = self.drawImage(nx=size, ny=size, scale=scale, offset=offset)
 
-        center = im.trueCenter() + offset + centroid/scale
+        center = im.trueCenter() + offset + centroid / scale
```
rmjarvis: And please revert all of these style changes.  These are not improvements.

tests/test_inclined_exponential.py
```diff
@@ -0,0 +1,258 @@
...
+#    and/or other materials provided with the distribution.
+#
+
+from __future__ import print_function
+import numpy as np
+import os
+import sys
+
+from galsim_test_helpers import *
+from IPython.core.magics.execution import Timer
```
rmjarvis: This causes Travis to fail.  And since you don't use it, I think you can just remove this line.

galsim/base.py
brgillis: Sorry about this; I must have some autoformatting turned on in my editor.

tests/test_inclined_exponential.py
```diff
@@ -0,0 +1,258 @@
...
+        # Now make a test image
+        test_profile = galsim.InclinedExponential(inc_angle*galsim.radians,scale_radius,scale_height,
+                                                  gsparams=gsparams)
+        
+        # Check that the k value at maxK() is below maxk_threshold in both the x and y dimensions
+        kx = test_profile.maxK()
+        ky = test_profile.maxK()
+        
+        kx_value=test_profile.kValue(kx=kx,ky=0.)
+        np.testing.assert_(np.abs(kx_value)<gsparams.maxk_threshold)
```
rmjarvis: I added print statements here to see how kValue(maxk) actually compares with maxk_threshold, and you are using much too large a value of maxk:
```
kvalue(maxk) =  (1.38421755486e-06+0j)
maxk_threshold =  0.001
```
The problem is you are still basing `maxk` off of where kval falls below `kvalue_accuracy`, rather than `maxk_threshold`.  

This is why you found it necessary to impose such a large `maximum_fft_size` in these tests (20,000 in `test_edge_on`!)  If you change that calculation to be based on `maxk_threshold`, you won't need such large ffts.

src/SBInclinedExponential.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+        // Now set up, using this value of cosi
+
+        _r0_cosi = _r0*cosi;
+        _inv_r0_cosi = 1./_r0_cosi;
+
+        _h_tani_over_r = scale_height*std::abs(inclination.sin())*_inv_r0_cosi; // A tiny bit more accurate than using tan of
+                                                                                // the truncated value
+
+        _half_pi_h_tani_over_r = 0.5*M_PI*_h_tani_over_r;
```
rmjarvis: I've been trying to understand why you have to have special handling for cosi ~= 0, and I think the problem is entirely artificial.  

This constant would admittedly be infinity, but the only place you use it is to multiple a scaled version of ky, which is really ky * r0 * cosi.  So that calculation is really h sin(i) ky, which is not problematic near i=Pi/2.

I think this implies that you should instead store `_half_pi_h_sini_over_r` and not scale `ky` by `cosi` when you pass it into `kvalueHelper`.

src/SBInclinedExponential.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+        xdbg<<"_h_tani_over_r = "<<_h_tani_over_r<<std::endl;
+        xdbg<<"_r0_cosi = "<<_r0_cosi<<std::endl;
+
+        // For large k, we clip the result of kValue to 0.
+        // We do this when the correct answer is less than kvalue_accuracy.
+        // (1+k^2 r0^2 cos^2(i))^-1.5 = kvalue_accuracy
+        // The convolution factor is generally a small effect, so we don't include it here
+        _ksq_max = (std::pow(this->gsparams->kvalue_accuracy,-1./1.5) - 1.)/cosi;
```
rmjarvis: I don't think this calculation is right.  First, a straightforward reading of the derivation would imply that you should have cosi**2 here.

But more importantly, this calculation too conservative.  Particularly in the case where cosi ~= 0.  In particular, it's not really appropriate to ignore the convolution factor, since it is significant precisely for ky >> 1, which is the limit you are taking for this calculation.  

I think you can probably do a few iterations of a non-linear solver to find where

(1+ky^2 cosi^2)**-1.5 * (_half_pi_h_sini_over_r * ky) / sinh(_half_pi_h_sini_over_r * ky) ~= epsilon

As I mentioned earlier, you'll want to do this separately for epsilon = kvalue_accuracy and maxk_threshold for two quantities _ksq_max and _maxk.  I think you'll find that you don't need to go out to nearly as high a ky value as this calculation would imply.  Particularly when cosi << 1.

src/SBInclinedExponential.cpp
brgillis: The reason I saw this as problematic wasn't due to the convolution factor. Rather, it has to do with the inclination of the exponential disk. See the maths here for the basis for the calculation here: https://www.dropbox.com/s/z6h14bgd199czsi/Inclined_Exponential.pdf?dl=0 - in particular, look at equation (3), which is before any coordinate transformation. The issue comes from the y^2/cos^2(i) term in the exponential on the right side. In the case where i=pi/2, the assumption that we can integrate across exponential disks breaks down, so instead we have to integrate through a disk, and this doesn't have a closed-form solution that I know of.

The possible problem in Fourier space comes from the fact that as i approaches pi/2, ky*R*cos(i) approaches zero, but the limits for ky approach infinity, resulting in a 0/0 type situation. However, I think the convolution saves us here. Since we multiply by ky/sinh(ky), the k value goes to zero at a non-infinite value. This goes back to your other comment that my calculation of the maximum k was too conservative, since I didn't account for this. So, I think once that issue is resolved, the edge-on case won't be an issue either (it will just take a bit of rewriting the code). I'll have a look at this tomorrow.

rmandelb:
@brgillis - thanks.  It's the first week of class here, which is always brutal.  I will have to wait until next week to go over this.

brgillis:
@rmjarvis - I've got it set up now to calculate _maxk using the solver class (and I've reworked it so we don't need special handling for any inclinations). It might not be as efficient as possible, but the timings from the test are all fine. I still have to fix how _ksq_max is calculated, which I can do the same way, but I wanted to get your input before I do that on if you think this is the best way to go about this, or if it's worth trying to make things a bit more efficient.

rmjarvis:
@brgillis Efficiency-wise, I think that's fine.  I'm glad you found the Solver class.  I was going to recommend you use that, since I doubt there's a really simple non-linear iteration for this, and the Solver class is usually quick enough for general root finding like this.  

You can check to see whether using Brent is significantly quicker with `solver.setMethod(Brent)`.  The default is Bisect, which is very stable, but is usually slower than Brent.  However, I wouldn't expect this calculation to be anywhere close to the tall pole for optimization here, so it probably doesn't matter.

The only change I would make to the implementation is to make the calculation generic w.r.t. epsilon so you can call it twice with epsilon = `maxk_threshold` and `kvalue_accuracy` for the two cases.  Probably have the MaxKFunctor take epsilon as a constructor parameter rather than hard code `pimpl->gsparams->maxk_threshold`.

brgillis:
@rmjarvis Alright, I've implemented the functor in that manner, and I'm using it for both _kmax and _ksq_max now. Brent is indeed faster, so I've switched to that as well. I also adjusted the calculation of _ksq_min slightly to account for the changed formula compared to the basic exponential profile (it's actually only a very slight change in it, though).

It occurred to me that I should check with Lance Miller, who wrote the hankelcode.c file, to make sure he has no objections to it being included in the distribution. I'll post back once I hear from him.

CHANGELOG.md
```diff
@@ -25,7 +25,7 @@ Deprecated Features
 New Features
 ------------
 
-
+* Added new light distribution 'InclinedExponential'
```
rmandelb: This is a little too terse (we are normally this terse in HISTORY.md, but the CHANGELOG.md usually has a bit more information).  Should include a sentence saying what this is meant to represent.  Also include issue number at the end so people know where to look for information.

doc/ref/GalSim_Quick_Reference.tex
```diff
@@ -244,6 +244,11 @@ \subsection{GSObject classes and when to use them}\label{sect:gsobjectclasses}
     with index \texttt{n=1}.  Requires one of the
     following \texttt{\emph{size}} parameters: \texttt{scale\_radius} or \texttt{half\_light\_radius}.
 \item[$\circ$]
+  \texttt{galsim.}\href{http://galsim-developers.github.com/GalSim/classgalsim_1_1base_1_1_inclined_exponential.html}{\texttt{InclinedExponential}}\texttt{(inclination, scale\_radius, scale\_height, flux=1.)}
```
rmandelb: URL does not exist, presumably because we only put up documentation for the last tagged version, which this is not in.  For now, I recommend simply not having a URL in the quick reference (rather than linking to something that doesn't exist).

galsim/inclinedexponential.py
```diff
@@ -0,0 +1,131 @@
...
+import galsim
+from galsim import GSObject
+from . import _galsim
+
+
+class InclinedExponential(GSObject):
+    """A class describing an inclined exponential profile.
+
+    The Inclined Exponential surface brightness profile is characterized by three properties: its
+    inclination angle 'i' (where 0 = face-on and pi/2 = edge-on), its scale radius, and its scale
```
rmandelb: I think this should be `inclination` rather than 'i'.

galsim/inclinedexponential.py
```diff
@@ -0,0 +1,131 @@
...
+    height. The 3D light distribution function is:
+
+    I(R,z) = I_0 / (2h_s) * sech^2 (z/h_s) * exp(-R/R_s)
+
+    where z is the distance along the minor axis, R is the radial distance from the minor axis,
+    R_s is the scale radius of the disk, h_s is the scale height of the disk, and I_0 is the central
+    surface brightness of the face-on disk.
```
rmandelb: And then say that the 2D light distribution function is determined by the input scale radius and scale height, along with the input inclination?

galsim/inclinedexponential.py
```diff
@@ -0,0 +1,131 @@
...
+    Methods
+    -------
+
+    In addition to the usual GSObject methods, InclinedExponential has the following access methods:
+
+        >>> inclination = inclined_exponential_obj.getInclination()
+        >>> r0 = inclined_exponential_obj.getScaleRadius()
+        >>> h0 = inclined_exponential_obj.getScaleHeight()
+    """
+    _req_params = { "inclination" : float, "scale_radius" : float, "scale_height" : float }
```
rmandelb: Has the config documentation for v1.5 been edited to include this new light profile?  Perhaps @rmjarvis can recommend where to start?

galsim/inclinedexponential.py
```diff
@@ -0,0 +1,131 @@
...
+
+    In this implementation, the profile is inclined along the y-axis. This means that it will likely
+    need to be rotated in most circumstances.
+
+    At present, this profile is not enabled for photon-shooting.
+
+    Initialization
+    --------------
+
+    @param inclination      The inclination angle, i
```
rmandelb: Say it should be a galsim.Angle instance so that users don't wonder if it should be degrees, radians, ...

rmandelb: Also, state what limitations should be imposed on input.  For example, if I decide to go crazy and  initialize with ```inclination=90.*galsim.degrees``` (perfectly edge on), I get the following output
```
  File "galsim/inclinedexponential.py", line 77, in __init__
    gsparams))
RuntimeError: Solve error: Too many iterations in bracket()
```
Actually the same is true when I try 80 degrees.  This is something that users should be told up-front.

If you prefer not to include such information in the @param statements, then higher up in the docstring you could give an example of how to initialize an instance of this class (this is not very unusual; many classes have something like this) and discuss such limitations there.

It would also be natural for users to wonder about constraints on ```scale_height``` values.

galsim/inclinedexponential.py
```diff
@@ -0,0 +1,131 @@
...
+    @param scale_radius     The scale radius of the exponential disk.  Typically given in arcsec.
```
rmandelb: May be worth saying explicitly that this is comparable to the ```scale_radius``` of the galsim.Exponential class.

galsim/inclinedexponential.py
brgillis: @rmandelb Hmm, it actually shouldn't have any constraints on input in the current implementation. I'll look into what's causing this and try to fix it.

rmandelb: Sorry for multiple comments on this spot, but I am getting this exception even for an inclination of 60, which seems odd to me.  Should that be happening, and do you see this on your system?

galsim/inclinedexponential.py
```diff
@@ -0,0 +1,131 @@
...
+        # Explicitly check for angle type, so we can give more informative error if eg. a float is passed
+        if not isinstance(inclination, galsim.Angle):
+            raise TypeError("Input inclination should be an Angle")
+
+        GSObject.__init__(self, _galsim.SBInclinedExponential(inclination, scale_radius, scale_height, flux,
+                                                              gsparams))
+        self._gsparams = gsparams
+
+    def getInclination(self):
+        """Return the inclination angle for this profile in radians.
```
rmandelb: That implies you get a float value in radians, whereas really I believe you get a galsim.Angle instance (which you can use to get whatever units you want).

galsim/inclinedexponential.py
```diff
@@ -0,0 +1,131 @@
...
+        """
+        return self.SBProfile.getInclination()
+
+    def getScaleRadius(self):
+        """Return the scale radius for this Sersic profile.
+        """
+        return self.SBProfile.getScaleRadius()
+
+    def getScaleHeight(self):
+        """Return the scale radius for this Sersic profile.
```
rmandelb: "radius" -> "height"

include/galsim/SBInclinedExponential.h
```diff
@@ -0,0 +1,86 @@
...
+        SBInclinedExponential(Angle inclination, double scale_radius, double scale_height, double flux,
+                 const GSParamsPtr& gsparams);
+
+        /// @brief Copy constructor.
+        SBInclinedExponential(const SBInclinedExponential& rhs);
+
+        /// @brief Destructor.
+        ~SBInclinedExponential();
+
+        /// @brief Returns the inclination angle of the profile in radians
```
rmandelb: ...as an Angle instance

include/galsim/SBInclinedExponentialImpl.h
```diff
@@ -0,0 +1,130 @@
...
+        Position<double> centroid() const
+        { return Position<double>(0., 0.); }
+
+        /// @brief Returns the true flux (may be different from the specified flux)
+        double getFlux() const { return _flux; }
+
+        /// @brief photon shooting done by rescaling photons from appropriate `InclinedExponentialInfo`
+        boost::shared_ptr<PhotonArray> shoot(int N, UniformDeviate ud) const;
+
+        /// @brief Returns the inclination angle in radians
```
rmandelb: same comment here as above

galsim/inclinedexponential.py
brgillis: @rmandelb I checked, and I'm not getting this error with 90 degrees (or any inclination I tested) - have you checked to make sure you're on the latest revision of this branch? Past revisions did indeed have issues with the edge-on case.

src/SBInclinedExponential.cpp
```diff
@@ -0,0 +1,357 @@
...
+            _ksq_max = clipk*clipk;
+
+            xdbg << "clipk = " << clipk << std::endl;
+            xdbg << "F(" << clipk << ") = " << kValueHelper(0.,clipk) << std::endl;
+        }
+    }
+
+    double SBInclinedExponential::SBInclinedExponentialImpl::xValue(const Position<double>& p) const
+    {
+        throw std::runtime_error("Real-space expression of SBInclinedExponential NYI.");
```
rmandelb: Not obvious to me that users will parse "NYI".  Recommend spelling it out.

rmandelb: (also comes up for the ```shoot``` method)

galsim/inclinedexponential.py
rmandelb: I just pulled and it says my branch is already up-to-date.  The last commit I see when I do �git log� on my machine is 9261e89e4ff9f2cecc6e4b5d48b1a4d9e5010538 from August 31. This is also the last commit I see on the web interface for this pull request.  So I think I do indeed have all commits that are pushed to the repo.  Is it possible you have some commits locally that have not been pushed?

tests/test_inclined_exponential.py
rmandelb: I see that in the time since Mike made this comment, you included the code and README in the subdirectory.  I would still recommend adding a note here in ```test_inclined_exponential.py``` mentioning that the tests are against reference code provided by Lance Miller, and it can be found in the subdirectory with the test images.

galsim/inclinedexponential.py
brgillis: @rmandelb The same commit is the last one for me. I just went through and reinstalled to be sure, running
    $ scons -c
    $ scons install
To reinstall, and then I tested with:
    $ ipython
    >>> import galsim
    >>> prof = galsim.InclinedExponential(scale_radius=3.0,scale_height=0.3,inclination=90.*galsim.degrees)
    >>> prof.kValue(0,0)
    (1+0j)
Which all worked as expected. Are you doing something different to test?

rmandelb: In response to your latest comment, I checked and found that whether or not I get this error depends not only on the inclination, but also the values of ```scale_radius``` and ```scale_height```.  For example, if I run the unit tests, they pass, and the unit tests do have some edge-on disks.  If I take the unit tests and start increasing the scale height of the disk, then eventually they start to fail when the inclination is edge-on.

galsim/inclinedexponential.py
```diff
@@ -0,0 +1,131 @@
...
+                 (self.scale_height == other.scale_height) and
+                 (self.flux == other.flux) and
+                 (self._gsparams == other._gsparams)))
+
+    def __hash__(self):
+        return hash(("galsim.InclinedExponential", self.inclination, self.scale_radius, self.scale_height, self.flux,
+                     self._gsparams))
+
+    def __repr__(self):
+        return 'galsim.InclinedExponential(i=%r, scale_radius=%r, scale_height=%r, flux=%r, gsparams=%r)' % (
```
rmandelb: I think the ```i=``` should be ```inclination=```.  Same comment applies in ```__str__```

galsim/inclinedexponential.py
rmandelb: And nope, I am testing in the same way you are, just that my disks were fatter than your disks (larger ratio of scale height to radius).

rmandelb:
@brgillis - I am done going over the code itself.

One more thought I wanted to run by you and @rmjarvis - I was thinking it might be nice to show this off in one of our demos.  For example, all of demos 2, 5, and 9 have simple exponential profiles.  We could turn one of the later ones to InclinedExponential profiles?

galsim/inclinedexponential.py
brgillis: @rmandelb Okay, I'm getting that error as well now, with larger scale heights. Realistic scale heights are of order ~10% the scale radius, and I didn't start getting the error until it was over ~60% of the scale radius. I'll see if there's an easy fix for this, but this is at the point where it's hardly disk-like, and so perhaps we should just put in a note about limits on the scale_height.

It might also be worth letting scale_height default to 10% of the scale_radius, so users won't enter it themselves unless they're certain they want to (in which case it's less likely they'll choose too-large values).

rmandelb: Sure, a note about the limits on ```scale_height``` would be reasonable.

I suppose having a default ```scale_height``` of 10% of the scale radius would be fine.

rmjarvis:
> One more thought I wanted to run by you and @rmjarvis - I was thinking it might be nice to show this off in one of our demos. For example, all of demos 2, 5, and 9 have simple exponential profiles. We could turn one of the later ones to InclinedExponential profiles?

I think demo9 would be a good choice for this.  I can help with that.

rmandelb:
Great, thanks Mike!

brgillis:
Alright, I think I've covered all of @rmandelb's comments now, aside from adding it to the demo. Also, Lance got back to me and said he has no problem with his code being included in the repo, so we don't have to worry about that. I did want to ask though if we should license that code (hankelcode.c) as well?

rmjarvis:
> I did want to ask though if we should license that code (hankelcode.c) as well?

That's up to Lance I suppose.  If he would like to put his own license at the top of that file, that would be fine, and we would not need to add the GalSim license to the file in addition.  If he doesn't want to bother with that and is ok with the GalSim license, that's fine too.  But probably a good idea to have some license (or at least copyright notice) at the top of the file.

rmjarvis:
I updated demo9 to use InclinedExponential.  One thing I realized would be helpful was allowing `scale_h_over_r` as an independent parameter in lieu of `scale_height`.  If these are being generated as random variates, then h/r and r are more independent than h and r, so it makes more sense to use these as the two parameters.  This means that h/r defaults to 0.1, rather than h defaulting to 0.1 * r.

include/galsim/SBInclinedExponential.h
```diff
@@ -69,9 +69,12 @@ namespace galsim {
         /// @brief Returns the scale radius r0 of the disk profile
         double getScaleRadius() const;
 
-        /// @brief Returns the scale height r0 of the disk profile
+        /// @brief Returns the scale height h0 of the disk profile
         double getScaleHeight() const;
 
+        /// @brief Returns the scale height over scale radius h0/r0 of the disk profile
+        double getScaleHOverR() const;
```
rmjarvis: I'm not sure why this is useful.  What is the advantage of calculating this in the C++ layer?  It seems to me that the the C++ class is perfectly fine being defined in terms of just h and r.  

I don't even think we need the python-layer `getScaleHOverR` method, but even if you want to keep that, I definitely don't see any advantage to implementing it in C++.

brgillis: Starting with the python-layer method, if we allow the user to supply h/r as an input parameter, we should provide a way for them to easily retrieve it, rather than making them figure out how to get it back from other methods.

The reason for implementing it in the C++ layer is efficiency. Doing it this way means only one function call through to that layer is required, rather than two calls and a division operation. It's only a slight gain though, so it might not be worthwhile.

rmjarvis: I can't imagine any situation where the efficiency of the `getScaleHOverR` is at all important.  In fact, this implementation is probably slower, since you _always_ calculate h/r in the constructor to store the value, even if it is not ever accessed.  Unless users will generally access this multiple times, that's not going to be faster overall.  (Although the slow down is, of course, minuscule...)

I'm also not convinced that the user will ever care about retrieving h/r from an existing object.  But I'll leave the final decision on that up to you.

brgillis:
I've removed getScaleHOverR from the C++ layer now, but left it in the Python layer, since I believe it does make sense to allow the user to retrieve any possible input parameter directly. I also added in a test that it's working properly.

Still haven't heard back from Lance about licensing his code. I'll update once I hear from him.

rmjarvis:
I think Lance's decision about a license for hankelcode4.c is the only thing left on this PR.  Maybe time to ping him again to see what he wants to do for that?  

If he doesn't care, the easiest thing would be to put our standard license text at the top and then keep the comment listing you and him as the authors below that.  But I certainly understand if he would prefer something different instead.

brgillis:
Lance is currently on vacation, and will be for another week I believe. I can talk with him in a week, when he should be back. If there's some reason to rush this though, let me know and I can ping him to see if he catches the message.

rmjarvis:
That's fine.  There's no rush.  :)

brgillis:
Okay, I checked with Lance, and he's fine with the default GalSim license. He also mentioned that it would be useful if this profile could apply a truncation radius to the exponential disk, but I think that can be best handled in a separate issue after finalising this.

rmjarvis:
Great.  Thanks for this Bryan!

Issue #792, opened 2016-08-25
small config bug


rmandelb:
There's a line in galsim/config/noise.py that does this:
```
cn -= galsim.UncorrelatedNoise(rng, im.wcs, current_var)
```

However, the call sequence for galsim.UncorrelatedNoise requires the arguments to be in the order variance, rng, scale, wcs.  So it's breaking with a TypeError as it tries to use the supplied rng as a noise variance.

The fix for this is obvious and I can do it on a branch (+look for other places where UncorrelatedNoise and CorrelatedNoise might be called incorrectly, but I think this might be it).  The harder question is can we design a unit test that would catch this kind of thing?  It's hard to come up with a really fast one, but I'm happy to get your ideas.

rmjarvis:
A TODO item that I have not gotten around to yet is to make more unit tests of the config apparatus.  I've been lazily counting on check_yaml as my pseudo-test suite for this, but of course that only checks things that we have somewhere in a demo.

This particular bug happens when you have galaxies with noise, whiten them, and then add COSMOS-correlated noise to the full image.  So, not something that we have in a demo anywhere.

In any case, it's worth at least adding a test that catches this particular bug.  I did so on this branch: b137e522e1801a  It catches the bug you found and it is fixed by your commit.

rmandelb:
Thanks Mike.  I'm making a PR in a moment.

rmandelb:
Closing #792.

PR #793, opened 2016-08-28
#792

rmandelb:
This PR fixes a minor bug in config when trying to whiten the noise in images based on COSMOS (+ new unit test that would have caught this bug, thanks to @rmjarvis ).

Given that this is a bug fix for functionality that was present in v1.4, should we merge it into releases/1.4 as well?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ Dependency Changes
 Bug Fixes
 ---------
 
-
+- Fixed bug when whitening noise in images based on COSMOS training datasets using the config
+  functionality. (#792)
 
 Deprecated Features
 -------------------
--- a/galsim/config/input.py
+++ b/galsim/config/input.py
@@ -272,6 +272,8 @@ def SetupInputsForImage(config, logger):
                 fields = config['input'][key]
                 if not isinstance(fields, list):
                     fields = [ fields ]
+                if 'input_objs' not in config:
+                    raise RuntimeError("ProcessInput needs to be called before building image")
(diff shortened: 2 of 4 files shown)


rmjarvis:
All looks fine to me (unsurprisingly, since I already looked at it).  I'm also fine with merging this to realeases/1.4 and making a v1.4.2 if you want.

rmandelb:
Thanks Mike.  Shall we try to get one more person to look at this, or do you think it's sufficiently minor that it's not necessary?

Re: v1.4.2, normally I wouldn't bother, but in this case the bug-fix is for some sims we're making for HSC, and I really prefer to have our sims run using tagged versions.  So yes, that would be great.

rmjarvis:
I think it's fine.  Go ahead and merge.

rmandelb:
Merging #793.

rmandelb:
Sorry - I meant to ask should I go ahead and merge to releases/1.4 and tag 1.4.2, or is there anything else that we might want to wait for?

rmjarvis:
I think you can go ahead with 1.4.2.  Minor version updates aren't too onerous.  Especially if you don't care about getting it on fink or updating macports, which I suspect is the case here.  (But correct me if I'm wrong.)

rmjarvis:
Actually, Erin just found another sort of obscure bug in the config parsing, which I fixed on master.  5b28887285fb307147ea3fbaa66fb86974a9fecb  So while you're at it, you could cherry-pick that over to releases/1.4 and include it in 1.4.2.

rmandelb:
Done.

Issue #789, opened 2016-08-16
Change SED units from photons/nm to photons/nm/s/meter^2


jmeyers314:
This is a proposal to change the units of SEDs from photons/nm, which is a joint property of a particular object *and* the telescope aperture / exposure time used to observe it, to photons/nm/s/m^2, which is a property of just the object itself.  I think this will make it less confusing to calculate what the same object would look like using different telescopes / exposure times than in our current setup where you have to actually create distinct ChromaticObjects to get this right.  That said, this may lead to a number of downstream changes, such as under what circumstances a Bandpass requires a `.zeropoint` attribute.  I think we should try it and see what breaks though.

rmandelb:
If we were designing this functionality from scratch again, I would be inclined to say that you are right.  So for me, I think it's a question of how many things would break if you do that...

jmeyers314:
> I think it's a question of how many things would break if you do that...

I think the switch is the right thing to do almost regardless, but I'm certainly going to try and make the transition as seamless as possible.  We may be able to get pretty far just by having default zeropoints that assume an exposure time of 1s and a collecting area of 1m^2.

I'll report here which things start to break, and we can discuss further.

jmeyers314:
I think I got this working now, and it did turn out to be most transparent.  I actually decided to go with photons/nm/cm^2/s instead of photons/nm/m^2/s since this is more cgs-consistent, and share/vega.txt is already in cgs units, for example.

I did run into a bit of trouble with one of the wfirst unit tests comparing the expected WFIRST photon flux given an SDSS-normalized SED.  I had to degrade the tolerance of the test from 10% to 15%.  I think this may have been a case where the previous code was actually not doing the right thing though, and may have even passed by virtue that WFIRST and SDSS have similar collecting areas.  I like the current test better.

Finally, this change has made me aware of some other dimensionally inconsistent things.  For example, since I'm now hiding the collecting area and exposure time in the `Bandpass` (btw, I'm now hiding the collecting area and exposure time in a `Bandpass` object), multiplying a `Bandpass` by an `SED` no longer yields an `SED`  (photons/nm/cm^2/s  != photons/nm).  Similarly, I think some of the logic in the Chromatic deconvolution/autoconvolution/etc doesn't make sense.  For example, ChromaticDeconvolution has a line: `self.SED = lambda w: 1./obj.SED(w)`, which is dimensional nonsense.

[...]

PR #798, opened 2016-09-09
#789 Chromatic Normalization

jmeyers314:
This PR implements a number of changes in the way that chromatic objects and SEDs are used, especially the way they’re normalized.  Previously, SEDs had dimensions of [photons / wavelength-interval], which is a joint property of the astronomical object of interest, the projected telescope collecting area, and the exposure time used to generate an image.  This made it particularly cumbersome to investigate what a particular object might look like through two different telescopes, or with two different exposure times.  The new SED dimensions are [photons / wavelength-interval / area / time], which is independent of the telescope and exposure time.  

To set the area and exposure time, new keyword arguments have been added to drawImage: `area=1., exptime=1.`.  The default values of 1 cm^2 and 1 sec ensure that the change in SED normalization is mostly transparent to users who were not attempting to draw chromatic (or achromatic) objects with particular telescopes or exposure times.  

Adding the exptime and area to drawImage also implies that these be removed from the bandpass zeropoint, where they previously lived.  The new bandpass zeropoint is equal to the magnitude at which the flux of a standard spectrum (could be Vega, AB, or something else) through the bandpass is 1 photon / cm^2 / sec.  (Note that this is a bit different than, say, [HST zeropoints](http://www.stsci.edu/hst/acs/analysis/zeropoints) which I believe are at least approximately the magnitude at which the flux through the bandpass and with collecting area equal to the HST primary mirror area is 1 photon / sec).

While changing the dimensionality of SEDs and associated ChromaticObjects, I noticed a few things that didn’t make sense dimensionally, such as the product of two SED objects was another SED object.  These kinds of dimensional issues are now tracked and exceptions generated when appropriate.

The trickiest new development is that there are effectively now two types of ChromaticObjects: those that are associated with an SED (and hence represent an object with dimensions of [photons / wavelength-interval / area / time / solid-angle]) and those that still have a normalization, but aren’t associated with an SED (i.e., dimensions are just [1 / solid-angle])  Generally speaking, the former correspond to stars or galaxies, and the latter to PSFs.  Along with this distinction are a new set of rules for what types of manipulations between SED’d and non-SED’d ChromaticObjects are possible.  For instance:
- only SED’d chromatic objects can be drawn.  Calling .drawImage(bandpass) on a non-SED’d ChromaticObject will raise an exception.
- at most one SED’d chromatic object can be part of a ChromaticConvolution.
- SED’d ChromaticObjects can be added together, as well as non-SED’d ChromaticObjects, but you can’t add an SED’d ChromaticObject to a non-SED’d ChromaticObject.
- multiplying a non-SED’d ChromaticObject by an SED results in an SED’d ChromaticObject (as a consequence of this, you no longer need to create a delta-fn GSObject to make a chromatic image of a star.  You can simply multiply the chromatic PSF by the SED of the star and draw that.)

This refactor let me deprecate one method and one class: 
- `ChromaticObject.__init__()`, which was only useful before as an intermediate step when applying a chromatic transformation to a GSObject: `chrom_obj = galsim.ChromaticObject(gs_obj).dilate(lambda w:w)`.  Now you can just directly apply the chromatic transformation to the GSObject: `chrom_obj = gs_obj.dilate(lambda w:w)`.  The `ChromaticObject` class still exists though as an abstract base class.
- `galsim.Chromatic`.  Multiplying a GSObject by an SED now just creates a `ChromaticTransformation` that’s functionally equivalent to the old `Chromatic` class.

Finally, the normalization of ChromaticObjects that aren't attached to an SED is done via lambdas, which means that fewer things are pickable than previously.  I'm not sure if this merits more attention or not.

Overall, I'm pretty happy with the end results of this refactor, though I'm not sure that the implementation is as clean as it could be.  Comments appreciated.


rmjarvis:
This smells a bit like we should have two classes that encapsulate this difference, rather than two kinds of chromatic objects.  I'm sure you thought about that kind of implementation, so what are the downsides to doing it that way?  (Other than potentially cumbersome class names: ChromaticObjectWithSED, FluxDensityObject, ... ?)

jmeyers314:
>This smells a bit like we should have two classes that encapsulate this difference, rather than two kinds of chromatic objects. I'm sure you thought about that kind of implementation, so what are the downsides to doing it that way? (Other than potentially cumbersome class names: ChromaticObjectWithSED, FluxDensityObject, ... ?)

Most of the `ChromaticObect`s naturally fall into one category or the other, but not `ChromaticSum`, `ChromaticConvolution`, `ChromaticTransformation`, or `InterpolatedChromaticObject` (and I suppose `ChromaticObject` itself).  I think the biggest drawback to having both Fluxed and UnFluxed ChromaticObjects is having two versions of each of these: (`ChromaticConvolution` & `FluxedChromaticConvolution` and so on), that essentially have the same logic.  If there's a way to get both types of compound objects without a lot of code duplication, I think that would probably be a better implementation.

rmjarvis:
> Finally, the normalization of ChromaticObjects that aren't attached to an SED is done via lambdas, which means that fewer things are pickable than previously. I'm not sure if this merits more attention or not.

[This page](http://stackoverflow.com/questions/25348532/can-python-pickle-lambda-functions) claims `dill` enables pickling of lambda functions.  It's pip-installable, so a low bar for a dependency.  Might be worth looking into.  As we move towards having Chromatic objects run in config, I think it might become useful to have them be picklable.

rmjarvis:
> ChromaticSum, ChromaticConvolution, ChromaticTransformation, or InterpolatedChromaticObject 

I _think_ this is doable if the fluxed versions of these are sub-classes of the unfluxed versions (along with having FluxDensityObject also as a base class).  Most of these have two non-trivial bits of code, the `__init__` function which could mostly be implemented in the base class that doesn't care whether it is fluxed or not, and `drawImage` which is only valid for the fluxed version.  So I think that might actually be rather clean.

OTOH, I haven't done much with multiple inheritance in Python so there might be hidden gotchas that I'm not anticipating.

jmeyers314:
What's a `FluxDensityObject`?

rmjarvis:
> What's a FluxDensityObject?

That was my joking reference to the hard-to-name base class for a chromatic object with an SED.  I don't have a particularly good idea for a name for it.

jmeyers314:
>That was my joking reference to the hard-to-name base class for a chromatic object with an SED. I don't have a particularly good idea for a name for it.

Ah. Okay.  I was leaning towards `class FluxedChromaticObject(ChromaticObject)`.

>Most of these have two non-trivial bits of code, the __init__ function which could mostly be implemented in the base class that doesn't care whether it is fluxed or not.

Unfortunately the base class still needs to care about normalization (even if unfluxed), but I think this is worth a look.  And good point about `drawImage()`.  I'll try it out over the weekend with `ChromaticTransformation` and report back.

rmandelb:
I'm going to start going over this code.  However, a few comments before I start:

* Tests pass on both my systems.

* I'm getting a deprecation warning when I run demo12.py:
```
/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/site-packages/galsim/deprecated/__init__.py:53: GalSimDeprecationWarning: (GalSimDeprecationWarning(...), 'Chromatic has been deprecated since GalSim version 1.5.  Construct products of GSObjects and SEDs through multiplication: `chrom_obj = gsobj * sed.`')
```
I think the demo should show the most up-to-date usage, with perhaps a comment that shows how it used to be done in older GalSim versions.

* The clear distinction between SED-carrying (stars/galaxies) and non-SED-carrying (PSFs) chromatic objects makes sense to me.   Your list of consequences that you mentioned in the pull request also make sense to me. 

* I was wondering if it might be worth soliciting feedback on this PR from people who have heavily used this functionality, which I believe includes David Kirkby (for his deblending sims), and Will Dawson and Michael Schneider (for WFIRST+LSST sims).  It could be they will have some further suggestions on the user interface, for example.

CHANGELOG.md
```diff
@@ -3,7 +3,10 @@ Changes from v1.4 to v1.5
 
 API Changes
 -----------
-
+- `drawImage()` now accepts `exptime` and `area` keywords to indicate the image
+  exposure time and telescope collecting area.  (#789)
+- dimensions of `SED` changed from [photons/wavelength-interval] to
```
rmandelb: We don't want to write long discussions in the CHANGELOG, but I think it would be worth ~2 sentences about the consequences of these things - in particular, the distinction between SED'ed and non-SED'ed chromatic objects, and what they represent.

rmandelb:
Also, I think that the discussion of chromatic objects / bandpasses / SEDs in the quick reference in doc/ref/ may need to be updated.

jmeyers314:
>I'm getting a deprecation warning when I run demo12.py

Oops.  I had meant to catch those.  Should be fixed now.

jmeyers314:
I'm starting to think that maybe the simplest way to distinguish the SED'd from the non-SED'd ChromaticObjects is to give every ChromaticObject an SED, but track the dimensions (really, units) of the SED.  (Similar to @dkirkby 's suggestion of using astropy.units to specify profile size parameters [here](https://github.com/GalSim-developers/GalSim/issues/657)).  Specifically, the `wave_type` and `flux_type` arguments to `SED.__init__` could be, in addition to a string, an `astropy.units.core.UnitBase` instance.

We could then propagate the units when multiplying `SED`s and `Bandpass`es and raise an exception if the resulting dimensions are not either [1] or [photons / wavelength-interval / area / time].  (Maybe also allow dimensions with energy <-> photons or wavelength-interval <-> frequency-interval).  We can still only permit `drawImage(bandpass)` and other restrictions if the dimensions make sense.

Pros:
  - simpler ChromaticObject implementation
  - no need for multiple inheritance or multiple versions of `ChromaticConvolution`.

Cons:
  - astropy becomes mandatory (at least for `ChromaticObject`s)
  - `SED` becomes a bit more of a misnomer (although we're already storing the photon distribution and not the energy distribution, so maybe this is okay.)
  - A more complicated `SED.__init__` API, though I think this may still be backwards compatible with existing options for `wave_type` and `flux_type`.

rmjarvis:
> give every ChromaticObject an SED, but track the dimensions

This sounds fine to me if you think that would be simpler to implement.

jmeyers314:
I think I've mostly got the SED class up and running with astropy units.  It's now possible to make an SED as a function of frequency with units of British-Thermal-Units per fortnight per acre:

```python
import galsim
from astropy import units as u
u.imperial.enable()
sed = galsim.SED('1', wave_type=u.Hz, flux_type=u.imperial.BTU/u.fortnight/u.imperial.ac)
```

Next up is enabling dimensionless SEDs, and then integration into chromatic.

rmjarvis:
Suggest importing into the galsim namespace the ones that we think people would actually use?
```
sed = galsim.SED('1', wave_type=galsim.nm, flux_type=galsim.erg/galsim.s/galsim.cm**2/galsim.nm)
```
Or maybe even give an alias to some of the longer combinations that would be relatively typical.  e.g. leave the existing ones
```
'flambda' is equivalent to u.erg / wave_type
'fnu' is equivalent to u.erg / u.Hz
'fphotons' is equivalent to wave_type**-1
```
You probably already did this bit -- I didn't check -- but maybe add additional aliases for the ones that include /s/cm^2.

Behind the scenes of course, these would all be implemented using the astropy.units stuff that you already did.

dkirkby:
What is the advantage to users of importing some units into the galsim namespace?  It could be confusing to users already familiar with astropy units and requires more typing (`galsim.erg` vs `u.erg`).

The `flambda` and `fnu` aliases might be confusing since they usually carry additional units of `/(u.cm **2 * u.s)`, e.g., in [IRAF](http://iraf.net/irafdocs/formats/standard.php) and [Wikipedia](https://en.wikipedia.org/wiki/AB_magnitude).

rmjarvis:
> It could be confusing to users already familiar with astropy units and requires more typing (galsim.erg vs u.erg).

I'm not sure why it would be confusing, but to your latter point, the comparison is actually
`galsim.erg` vs `astropy.units.erg`.  The `u.erg` possibility is only because in the quoted example, Josh imported `astropy.units as u`, which would not be something I expect most users to want to do, since `u` is a common variable name, so importing anything `as u` is likely to cause problems.

> The flambda and fnu aliases might be confusing since they usually carry additional units of /(u.cm **2 * u.s), e.g., in IRAF and Wikipedia.

`flambda` and `fnu` are already options for `flux_type`, so for backwards compatibility I thought we should probably preserve their current meaning.  Although I'm sympathetic to the fact that the do often have units of /cm^2/s.  Maybe if the default area and exposure time are 1, then switching to this wouldn't break any existing code...  In which case, probably switching to that would be a good idea.

dkirkby:
The reason it might be confusing is that it is not obvious from reading code to what extent galsim.erg is compatible with astropy units or overrides its behavior.  For example, is `galsim.ergs.si` valid?  Its the same argument I would make against using `import numpy as myarrays`.

For better or for worse, `import astropy.units as u` (or something equivalent) is used widely in the code I have seen, as well as in the [astropy docs](http://docs.astropy.org/en/stable/units/).

rmjarvis:
> Its the same argument I would make against using `import numpy as myarrays`.

> For better or for worse, `import astropy.units as u` (or something equivalent) is used widely in the code I have seen, as well as in the astropy docs.

I find it amusing that you think the former import is bad, but the latter is good.  (Or at least acceptable.)  Personally, I hate importing anything `as` anything else.  So I would never advocate `import numpy as myarrays` for the same reason that I don't advocate `import astropy.units as u`.  It's bad enough that `import numpy as np` is so widespread that it has become the almost de facto name of the library, so we're kind of stuck with using that.

The only exception I make as a general rule is `import astropy.io.fits as pyfits`.  I still find it obnoxious that they took a perfectly good package name and bloated the prefix so egregiously.  I think I just keep the `pyfits` name as kind of a protest of their taking over the package.

jmeyers314:
I already setup aliases:

for wave_type:
'nm', 'nanometer', 'nanometers' => `u.nm`
'a', 'ang', 'angstrom', 'angstroms' => `u.AA`

for flux_type:
'flambda' => `u.erg / (u.s * u.cm**2 * wave_type)`
'fphotons => `u.astrophys.photon / (u.s * u.cm**2 * wave_type)`
'fnu' => `u.erg / (u.s * u.cm**2 * u.Hz)`

Since the default area and exposure time are indeed 1sec and 1 cm**2, this is essentially gives backwards compatibility.

Personally, I prefer not adding units to the galsim namespace, mostly because I think the above aliases cover the common cases.

rmjarvis:
> I prefer not adding units to the galsim namespace, mostly because I think the above aliases cover the common cases.

That's fair.  My main goal is to make it easy to do the common things, and I think these aliases are sufficient for that.

CHANGELOG.md
jmeyers314: Thanks @rmandelb.  I added a bit more detail to the CHANGELOG.

jmeyers314:
@rmandelb , @rmjarvis - just wanted to note that I got the `astropy.units` modification here working now, so I'd be happy for you to restart your reviewing.  I should still go through the documentation one more time and also update the quick start guide, but I think the code is in good shape now, and less smelly than before :)

@dkirkby , @mdschneider - Rachel noted that since this PR is a bit of a departure from the previous way we normalized `SED`s and `ChromaticObject`s, and you two are the ones (apart from myself and @sowmyakth) we're most aware of actively using the chromatic code, you may wish to comment on this PR too.  I think things are *mostly* backwards compatible, but you may wish to try checking this branch out and trying it on your projects.

CHANGELOG.md
```diff
@@ -4,24 +4,39 @@ Changes from v1.4 to v1.5
...
+- Added restrictions to `ChromaticObject`s and `SED`s consistent with
+  dimensional analysis.  E.g., only `ChromaticObject`s with dimensionful SEDs
+  can be drawn. (#789)
 - Simplified the return value of galsim.config.ReadConfig. (#580)
 
 
 Dependency Changes
 ------------------
-
+- `astropy` is now a required dependency instead of an optional one. (#789)
```
rmandelb: Required for all GalSim functionality?  Or just chromatic functionality? Either way, I think an update in INSTALL.md is needed to reflect this.

galsim/base.py
```diff
@@ -1316,15 +1465,19 @@ def drawImage(self, image=None, nx=None, ny=None, bounds=None, scale=None, wcs=N
...
         imview = image.view()
         imview.setCenter(0,0)
 
         if method == 'phot':
             try:
-                image.added_flux = prof.SBProfile.drawShoot(
+                added_photons = prof.SBProfile.drawShoot(
                     imview.image, n_photons, uniform_deviate, gain, max_extra_noise,
                     poisson_flux, add_to_image)
+                image.added_flux = added_photons / area / exptime
```
rmandelb: One thing I've been thinking over is that some of these new conventions for SEDs and flux units break some of the assumptions that people often make.  For example, the relationship between ```image.added_flux``` and the actual pixel values in the image is no longer trivial (unless ```area=1``` and ```exptime=1``` as is the default).  And, FWIW, I think different conventions out there might confuse people... e.g., in HST they renormalize the pixel values in many of the science images they distribute to an exptime of 1, so the actual exptime is not related to the pixel values in the way that we have done it here.  (I am not saying that what they do makes sense, just that it could be a cause for confusion.)  I wonder if there is an example one could give in the docstring, or added comments in some demos, that might be helpful?

galsim/chromatic.py
```diff
@@ -19,45 +19,89 @@
...
+
+    >>> chrom_obj = (sed * 3.0) * gsobj
+    >>> chrom_obj2 = sed * (gsobj * 3.0)
+
+    Subclasses that instantiate a ChromaticObject directly also exist, such as ChromaticAtmosphere.
+    Even in this case, however, the underlying implementation always eventually wraps one or more
+    GSObjects.
+
+    ChromaticObjects can generally be sorted into two distinct types: those that represent galaxies
+    or stars and have dimensions of [photons/wavelength-interval/area/time/solid-angle], and those
```
rmandelb: Sorry, why is there both a ```/area``` and also a ```/solid-angle```?

rmandelb: (Note this comes up at a few places elsewhere in the file too.)

galsim/chromatic.py
```diff
@@ -19,45 +19,89 @@
...
+    that represent other types of wavelength dependence besides flux, like chromatic PSFs (these
+    have dimensions of [1/solid-angle]).  The former category of ChromaticObjects will have an SED
+    attribute that itself has its `spectral` attribute set to True, while the latter category of
+    ChromaticObjects will have an SED with its `dimensionless` attribute set to True.  These two
+    classes of ChromaticObjects have different restrictions associated with them.  For example, only
+    ChromaticObjects with a spectral SED can be drawn using chrom_obj.drawImage(bandpass, ...).
```
rmandelb: In light of these restrictions, I was wondering if it might be worthwhile to add one more example to demo12?  Something that has a PSF with simple chromatic dependence that isn't being returned from a function that handles chromaticity under the hood?

galsim/sed.py
```diff
@@ -213,77 +277,210 @@ def __call__(self, wave):
...
+        else:  # ndarray or scalar
+            return self._fast_spec(wave / (1.0 + self.redshift))
+
+    def _call_slow(self, wave):
+        """ Return flux in photons / sec / cm^2 / nm or dimensionless normalization.
+
+        Uses self._spec that has not been pre-transformed for desired units, instead does all unit
+        conversions inside this method.
+
+        @param wave  Wavelength.  If not a Quantity, then assumed units are nanometers.
```
rmandelb: I recommend that when referring to classes that live in astropy, indicate that explicitly.

CHANGELOG.md
jmeyers314: I suppose it's just for the chromatic functionality.  I've updated CHANGELOG.md and INSTALL.md to reflect this.

galsim/chromatic.py
jmeyers314: The area is referring to the collecting area of the telescope.  The solid angle is because we're modeling a surface brightness.  You need to integrate over both the collecting area and the galaxy solid angle (and exposure time and wavelength) to get number of collected photons.

galsim/base.py
jmeyers314: Good point.  I added the following to the drawImage() docstring just after the discussion about `.added_flux`: 

```
        Also, recall that the relationship between flux (in photons/cm^2/s) and the pixel values
        depends on `exptime`, `area`, `gain`, and `method`!  For example:

            >>> obj = galsim.Gaussian(fwhm=1)
            >>> obj.flux
            1.0
            >>> im = obj.drawImage()
            >>> im.added_flux
            0.9999630988657515
            >>> im.array.sum()
            0.99996305
            >>> im = obj.drawImage(exptime=10, area=10)
            >>> im.added_flux
            0.9999630988657525
            >>> im.array.sum()
            99.996315
            >>> im = obj.drawImage(exptime=10, area=10, method='sb', scale=0.5, nx=10, ny=10)
            >>> im.added_flux
            0.9999973790505298
            >>> im.array.sum()
            399.9989
            >>> im = obj.drawImage(exptime=10, area=10, gain=2)
            >>> im.added_flux
            0.9999630988657525
            >>> im.array.sum()
            49.998158
```

I think this, plus the discussion at the end of the different drawImage methods "To convert from flux (photons/cm^2/s) into image values of analog-to-digital units (ADU) ..." may be enough?  Let me know if you think more clarification would be helpful.

jmeyers314:
>This page claims dill enables pickling of lambda functions. It's pip-installable, so a low bar for a dependency. Might be worth looking into. As we move towards having Chromatic objects run in config, I think it might become useful to have them be picklable.

Sure enough, dill was able to pickle the formerly unpicklable chromatic objects.  I added it to the list of GalSim optional dependencies, and also (optionally) use it in `do_pickle` now.

tests/test_bandpass.py
```diff
@@ -309,7 +301,7 @@ def test_zp():
     """Check that the zero points are maintained in an appropriate way when thinning, truncating."""
     # Make a bandpass and set an AB zeropoint.
     bp = galsim.Bandpass(os.path.join(datapath, 'LSST_r.dat'), 'nm')
-    bp = bp.withZeropoint(zeropoint='AB', effective_diameter=6.4, exptime=15)
```
rmjarvis: I don't see any new unit tests to replace the old way of setting diameter and exptime.  There are tests of drawImage with area, exptime for regular GSObjects, but I don't think there are any for Chromatic ones.

Possibly, this could go into the check_chromatic_invariant helper function.

tests/test_chromatic.py
```diff
@@ -275,15 +276,15 @@ def test_chromatic_add():
...
     image = galsim.ImageD(stamp_size, stamp_size, scale=pixel_scale)
     image = final.drawImage(bandpass, image=image)
 
     do_pickle(bulge)
     do_pickle(disk)
-    do_pickle(bdgal)
+    # do_pickle(bdgal)
     do_pickle(chromatic_PSF)
-    do_pickle(final)
+    # do_pickle(final)
```
rmjarvis: Oops.. Did we lose even more picklability for chromatic objects?  I guess there are more use of lambda functions now?  This is going to be a problem when we want to use chromatic objects in a multiprocessing framework.  

Any chance we can recover picklability by deferring the application of lambdas until we actually need them?  Or is it too baked in?

tests/test_sed.py
```diff
@@ -316,23 +338,63 @@ def test_SED_init():
...
 
+    # Check a few valid calls for when fast=False
+    sed = galsim.SED(galsim.LookupTable(foo,foo), wave_type='nm', flux_type='flambda',
+                     fast=False)
+    sed(1.5)
+    sed(1.5*units.nm)
+    sed((1.5*units.nm).to(units.Hz, units.spectral()))  # Frequency
+    sed((1.5*units.nm).to(units.erg, units.spectral()))  # Wavelength as an energy
+    sed((1.5*units.nm).to(units.m**-1, units.spectral()))  # Wavenumber in 1/m
+    sed((1.5*units.nm).to(units.radian / units.m, units.spectral()))  # Wavenumber in rad/m
```
rmjarvis: Shouldn't we be checking the values of these?  Or are we just checking that astropy doesn't raise an exception?

tests/test_sed.py
```diff
@@ -557,19 +619,19 @@ def test_thin():
...
+    # test_SED_roundoff_guard()
+    # test_SED_init()
+    # test_SED_withFlux()
+    # test_SED_withFluxDensity()
+    # test_SED_calculateMagnitude()
+    # test_SED_calculateDCRMomentShifts()
+    # test_SED_calculateSeeingMomentRatio()
+    # test_fnu_vs_flambda()
+    # test_ne()
+    # test_thin()
```
rmjarvis: You forgot to uncomment these.

tests/test_wfirst.py
```diff
@@ -267,14 +267,15 @@ def test_wfirst_bandpass():
...
     # non-trivial.
     for filter_name, filter_ in bp.items():
-        flux = sed.calculateFlux(filter_)
-        count_rate = flux / galsim.wfirst.exptime
+        flux = sed.calculateFlux(filter_)  # photons / cm^2 / s
+        count_rate = flux * galsim.wfirst.collecting_area  # photons / s
+        print(count_rate, reference[filter_name])
         np.testing.assert_allclose(
-            count_rate, reference[filter_name], rtol=0.1,
+            count_rate, reference[filter_name], rtol=0.15,
```
rmjarvis: What changed to worsen this agreement?

jmeyers314:
>Sure enough, dill was able to pickle the formerly unpicklable chromatic objects. I added it to the list of GalSim optional dependencies, and also (optionally) use it in do_pickle now.

Actually, scratch that.  I figured out a way to make cPickle work on the things that were failing...  Turned out there was an unnecessary lambda in `SED.__add__`.

galsim/sed.py
jmeyers314: Sure.  Done.

examples/demo12.py
```diff
@@ -116,21 +117,22 @@ def main(argv):
...
 
-    # Here we create a chromatic version of a de Vaucouleurs profile using the Chromatic class.
-    # This class lets one create chromatic versions of any galsim GSObject class.  The first
-    # argument is the GSObject instance to be chromaticized, and the second argument is the
-    # profile's SED.
+    # Here we create a chromatic version of a de Vaucouleurs profile by multipying a GSObject by an
+    # SED.  This is how one generally constructs _separable_ ChromaticObjects in GalSim, that is,
+    # those objects whose spatial dependence and wavelength dependence factor.  (Note that prior to
+    # version 1.5, this would have constructed an instance of the class `Chromatic`.  In GalSim v1.5
+    # and above, `Chromatic` has been deprecated in favor of the `ChromaticTransformation` class.)
```
jmeyers314: Sure.  Done.

galsim/bandpass.py
```diff
@@ -278,15 +292,12 @@ def __div__(self, other):
         else:
             tp = lambda w: self.func(w) / other
 
-        return Bandpass(tp, wave_type, blue_limit, red_limit, _wave_list=wave_list)
+        # Note that .zeropoint attribute gets reset by __div__
```
jmeyers314: Yep.  No propagation of zeropoints for either multiplication or division.  I've updated both the `Bandpass` and `Bandpass.withZeropoint` docstrings as suggested.

galsim/base.py
```diff
@@ -628,13 +628,27 @@ def expand(self, scale):
...
+            def buildScaleJac(w):
+                s = scale(w)
+                return np.diag([s,s])
+            jac = buildScaleJac
+            if hasattr(self, 'noise'):
+                import warnings
+                warnings.warn("Cannot propagate noise through chromatic transformation.")
+        else:
+            jac = np.diag([scale, scale])
+        new_obj = galsim.Transform(self, jac=jac)
```
jmeyers314: > I'm trying to decide whether all these changes to the GSObject transformations are actually a net improvement.

I mostly just thought they'd be more convenient this way.  But you bring up some good points about potential proneness to bugs, for example.  I'll go ahead and revert this I think.

galsim/base.py
```diff
@@ -1053,27 +1188,29 @@ def drawImage(self, image=None, nx=None, ny=None, bounds=None, scale=None, wcs=N
                         it could be useful if you want to view the surface brightness profile of an
                         object directly, without including the pixel integration.
 
+        To convert from flux (photons/cm^2/s) into image values of analog-to-digital units (ADU), we
+        use optional keywords `area` indicating the telescope collecting area in cm^2, `exptime`
+        indicating the exposure time in s, and `gain` which converts between photons and ADU.
```
jmeyers314: I'm happy to make this change.  One question though, which @rmandelb brought up, is what to do with the `image.added_flux` attribute.  I *think* it makes sense to keep it in the same units as `gsobj.flux`, which means it doesn't scale with `exptime` or `area` (or `gain` for that matter).  But this does mean that image.added_flux won't be close to image.array.sum().

galsim/chromatic.py
```diff
@@ -19,45 +19,89 @@
...
-    especially expand(), dilate() and shift().
-
-    @param gsobj  The GSObject to be chromaticized.
+    A ChromaticObject can be instantiated from an existing GSObject by applying a
+    wavelength-dependent transformation.  E.g.,
+
+    >>> gsobj = galsim.Gaussian(fwhm=1)
+    >>> chrom_obj = gsobj.dilate(lambda wave: 1.01**wave)
+
+    In this and similar cases, the univariate object passed to the transformation method (`dilate`
```
jmeyers314: Sure. Done.

galsim/chromatic.py
```diff
@@ -19,45 +19,89 @@
...
+    >>> gal = bulge + disk
+
+    The SEDs above describe the flux density in photons/nm/cm^2/s of an object, possibly normalized
+    with either the sed.withFlux(bandpass) or sed.withMagnitude(bandpass) methods (see the
+    docstrings in the SED class for details about these and other normalization options).  Note that
+    the `flux` attribute of the multiplied GSObject is accounted for in the newly constructed
+    ChromaticObject's normalization.  I.e., the following are equivalent:
+
+    >>> chrom_obj = (sed * 3.0) * gsobj
+    >>> chrom_obj2 = sed * (gsobj * 3.0)
```
jmeyers314: >There is a units issue here.

Ewww.  You're right.  Is there a way to get out of this with a gsobj.flux_type attribute?  I'm thinking not.  It's always going to be a dimensions problem to want both
```
gal = galsim.Gaussian(fwhm=1)
im = gal.drawImage(exptime=10)
```
and
```
gal = galsim.Gaussian(fwhm=1)
sed = make_me_a_sed()
chrom = gal * sed
```

I changed the paragraph preceding the example to:
```
    The SEDs above describe the flux density in photons/nm/cm^2/s of an object, possibly normalized
    with either the sed.withFlux(bandpass) or sed.withMagnitude(bandpass) methods (see the
    docstrings in the SED class for details about these and other normalization options).  Note that
    for dimensional consistency, in this case, the `flux` attribute of the multiplied GSObject is
    interpreted as being dimensionless instead of in its normal units of [photons/s/cm^2].  The
    photons/s/cm^2 units are (optionally) carried by the SED instead, or even left out entirely if
    the SED is dimensionless itself (see discusison on ChromaticObject dimensions below).  The
    GSObject  `flux` attribute *does* still  contribute to the ChromaticObject normalization,
    though.  For example, the following are equivalent:
```

galsim/chromatic.py
```diff
@@ -68,65 +112,50 @@ class ChromaticObject(object):
...
+             "instead of passing through ChromaticObject().")
+        self.separable = obj.separable
+        self.interpolated = obj.interpolated
+        self.wave_list = obj.wave_list
         self.obj = obj
         if isinstance(obj, galsim.GSObject):
-            self.separable = True
-            self.SED = galsim.SED('1', 'nm', 'flambda')  # uniform flambda = 1
-            self.wave_list = np.array([], dtype=float)
+            self.SED = galsim.SED(obj.flux, 'nm', '1')
```
jmeyers314: >I thought the SED was relative to the obj.flux value?

We decided to change this :) : https://github.com/GalSim-developers/GalSim/issues/789#issuecomment-241883876

Mostly to make `ChromaticObject.calculateFlux()` possible.  I thought about making `ChromaticObject.SED` a property that wrapped `self.evaluateAtWavelength().getFlux()`, but I think that would have been pretty slow compared to propagating the `SED` and keeping it as a `LookupTable` whenever possible.

I see now that I forgot all about adding `ChromaticObject.withFlux`, `withFluxDensity`, `calculateMagnitude` and `withMagnitude`.  I'll get on that.

> take a GSObject * sed

This happens in `__mul__`, not here in `__init__`, and ultimately gets forwarded to `ChromaticTransformation(obj, flux_ratio=sed)`.

galsim/chromatic.py
```diff
@@ -196,7 +225,9 @@ def interpolate(self, waves, oversample_fac=1.):
         whereas those that involve making GSObjects with wavelength-dependent keywords or
         transformations do benefit from it.  Note that the interpolation scheme is simple linear
         interpolation in wavelength, and no extrapolation beyond the originally-provided range of
-        wavelengths is permitted.
+        wavelengths is permitted.  However, if the `use_exact_SED` keyword is set to True, then the
```
jmeyers314: Sure. Done.

galsim/chromatic.py
```diff
@@ -472,18 +518,8 @@ def __div__(self, other):
...
-        ret = cls.__new__(cls)
-        for k, v in iteritems(self.__dict__):
-            if k == 'objlist':
-                # explicitly copy all individual items of objlist, not just the list itself
-                ret.__dict__[k] = [o.copy() for o in v]
-            else:
-                ret.__dict__[k] = copy.copy(v)
-        return ret
+    def __neg__(self):
+        return -1. * self
```
jmeyers314: Done.

galsim/chromatic.py
```diff
@@ -1586,41 +1584,61 @@ def __init__(self, *args, **kwargs):
...
+
+        # If everything ended up in a single SED_dict entry (and self.objlist is empty) then this
+        # ChromaticSum is separable.
+        self.separable = (len(self.objlist) == 0 and len(SED_dict) == 1)
+        if self.separable:
+            the_one_SED = list(SED_dict)[0]
+            self.objlist = SED_dict[the_one_SED]
+            # Since we know that the chromatic objects' SEDs already include all relevant
+            # normalizations, we can just multiply the_one_SED by the number of objects.
+            self.SED = the_one_SED * len(SED_dict[the_one_SED])
```
jmeyers314: >Seems like maybe we lost the ability to identify this object as separable

It comes out separable for me!  Since the SED is the same, both objects get added under the same dictionary key, which then points to a list with two entries.

galsim/base.py
rmjarvis: I don't think I have a strong preference here.  I'm not really sure how useful the `added_flux` return value is.  It stems from the desire to know whether the image size is large enough to capture most of the flux.  So you can do a test like `if image.added_flux < 0.99 * obj.flux: ...`  

I guess this means that it should have the same units as obj.flux.

galsim/chromatic.py
rmjarvis: That's good.  Thanks.

galsim/chromatic.py
rmjarvis: The thing that left me most confused is that
```
galsim.Convolve(gal*sed, psf).drawImage(...)
``
and
```
galsim.Convolve(galsim.ChromaticObject(gal) * sed, psf).drawImage(...)
```
end up producing essentially the same image, but have different SEDs along the way, differing by a factor of gal.flux.  I couldn't figure out why they both ended up with the same drawn flux.  It probably doesn't matter, since the code is correct, but it left me very confused.

File: galsim/chromatic.py (hunk 21 above)
Comment (rmjarvis) [inline galsim/chromatic.py]: Right.  I think this comment stems from my confusion about how seds are normalized.  I suppose they wouldn't get found as separable if they were made as `ChromaticObject(mono_disk) * sed + ChromaticObject(mono_bulge) * sed` because of the aforementioned normalization difference.

Anyway, I'm pretty sure my comment here doesn't really matter at all anyway.  Most uses of ChromaticSum won't have multiple objects with equal SEDs, even disregarding the normalization.

File: galsim/chromatic.py (hunk 18 above)
Comment (jmeyers314) [inline galsim/chromatic.py]: >end up producing essentially the same image, but have different SEDs along the way

Hmmm...  I think that means at least one of them isn't satisfying the intended chromatic invariant.  I'll look into that.

File: galsim/chromatic.py (hunk 17 above)
Comment (jmeyers314) [inline galsim/chromatic.py]: Also realized that if we're bringing back `chrom = ChromaticObject(gsobj)`, we need to decide the dimensions of `chrom`.  I think dimensionless makes the most sense here too, so I added the following to the `ChromaticObject` docstring:

```
    One caveat to creating ChromaticObjects directly from GSObjects like this is that even though
    the source GSObject instance has flux units in photons/s/cm^2, then newly formed ChromaticObject
    will be interpreted as dimensionless, i.e., it will have a dimensionless SED.  See below for
    more discussion about the dimensions of ChromaticObjects.
```

File: galsim/base.py (hunk 14 above)
Comment (jmeyers314) [inline galsim/base.py]: Reverted this.

File: galsim/chromatic.py (hunk 18 above)
Comment (jmeyers314) [inline galsim/chromatic.py]: I think the problem here was that gsobj.SED was being set to 1, when it should have been set to gsobj.flux.  I fixed that and added some tests: https://github.com/GalSim-developers/GalSim/pull/798/commits/82e8eb50ffb39742ba8b5162465acea2869297c9

File: galsim/chromatic.py [hunk 22]
```diff
@@ -792,38 +819,30 @@ class InterpolatedChromaticObject(ChromaticObject):
...
+        self.SED = original.SED
         self.wave_list = original.wave_list
 
         # Don't interpolate an interpolation.  Go back to the original.
-        if isinstance(original, InterpolatedChromaticObject):
-            original = original.original
-        self.original = original
-        self.waves = np.sort(np.array(waves))
-        self.oversample = oversample_fac
+        self.original = original.deinterpolated
```
jmeyers314: >Why go through this rigamarole with the deinterpolated property

At one point I had had some thoughts about making a default `ChromaticObject.deinterpolated` property so you wouldn't need to define it in every subclass, but now I think you're right, it's easier to just use an attribute.  Done.

galsim/chromatic.py
rmjarvis: Thanks.  I added another test there (04aef1c) to satisfy myself about this, which passes, so no problem.  Just wanted to make sure.

galsim/chromatic.py
```diff
@@ -1169,80 +1226,35 @@ def evaluateAtWavelength(self, wave):
...
-                                trapezoidal integration rule automatically.]
-        @param **kwargs         For all other kwarg options, see GSObject.drawImage()
-
-        @returns the drawn Image.
-        """
-        return self.build_obj().drawImage(bandpass, image, integrator, **kwargs)
-
 
 class Chromatic(ChromaticObject):
     """Construct chromatic versions of galsim GSObjects.
```
jmeyers314: Done.

galsim/chromatic.py
jmeyers314: >I suppose they wouldn't get found as separable if they were made as `ChromaticObject(mono_disk) * sed + ChromaticObject(mono_bulge) * sed` 

After the fix to GSObject.SED, this case comes out as separable now too.  Added a test to confirm.

galsim/chromatic.py
```diff
@@ -1672,14 +1690,16 @@ def drawImage(self, bandpass, image=None, integrator='trapezoidal', **kwargs):
 
         @returns the drawn Image.
         """
+        if self.SED.dimensionless:
+            raise ValueError("Can only draw ChromaticObjects with dimensionful SEDs.")
```
jmeyers314: Yup.  I'm going with spectral.  All switched now.

galsim/chromatic.py
```diff
@@ -1747,17 +1768,21 @@ def __init__(self, *args, **kwargs):
         if kwargs:
             raise TypeError("Got unexpected keyword argument(s): %s"%kwargs.keys())
 
+        # Accumulate convolutant .SEDs.
+        self.SED = args[0].SED
+        for obj in args[1:]:
+            self.SED *= obj.SED
```
jmeyers314: Sure.  Done.

This also made me realize that a GSObject being dimensionless is more of the rule than the exception, so I think I may go back and redocument this accordingly.  I.e., the only time GSObjects are dimensionful is in drawImage when exptime != 1 or area != 1.  At other times, e.g. when duck-typing as ChromaticObjects, they should be viewed as dimensionless.

galsim/chromatic.py
```diff
@@ -1769,50 +1794,43 @@ def __init__(self, *args, **kwargs):
...
 
     @staticmethod
-    def _get_effective_prof(sep_SED, insep_profs, bandpass,
-                            iimult, wave_list, wmult, integrator,
-                            gsparams):
-            # Collapse inseparable profiles into one effective profile
-            SED = lambda w: reduce(lambda x,y:x*y, [s(w) for s in sep_SED], 1)
-            insep_obj = galsim.Convolve(insep_profs, gsparams=gsparams)
+    def _get_effective_prof(insep_obj, bandpass, iimult, wmult, integrator, gsparams):
             # Find scale at which to draw effective profile
```
jmeyers314: Thanks.  Fixed.

galsim/sed.py
```diff
@@ -16,28 +16,49 @@
 #    and/or other materials provided with the distribution.
 #
 """@file sed.py
-Simple spectral energy distribution class.  Used by galsim/chromatic.py
+Spectral energy distribution class.  Used by galsim/chromatic.py
 """
 
 import numpy as np
+from astropy import units
```
jmeyers314: Sure.  That is a surprisingly long import...  Done.

galsim/sed.py
```diff
@@ -16,28 +16,49 @@
...
+    normalizations, and may thus also return dimensionless values.  By default, the above wavelength
+    used by __call__ is nanometers, but it's possible to use other units via the astropy.units
+    module (at least, if the SED keyword argument `fast=False`, see below).  For instance,
+
+    >>> sed = galsim.SED(...)
+    >>> from astropy import units as u
+    >>> assert sed(500) == sed(5000 * u.AA)
 
-    SEDs are callable, returning the flux in photons/nm as a function of wavelength in nm.
+    Note that the return value *always* has units photons/nm/cm^2/s.  The python type of the return
```
jmeyers314: Yep.  Thanks for catching that.  Fixed.

galsim/sed.py
```diff
@@ -16,28 +16,49 @@
...
+    value depends on the type of the input wavelength(s).  A scalar input wavelength yields a scalar
+    flux density, a tuple yields a tuple, a list yields a list, and a numpy.ndarray yields a
+    numpy.ndarray.  A scalar astropy.units.Quantity yields a python scalar, and a vector
+    astropy.units.Quantity yields a numpy.ndarray.
 
     SEDs are immutable; all transformative SED methods return *new* SEDs, and leave their
     originating SEDs unaltered.
 
     SEDs have `blue_limit` and `red_limit` attributes, which may be set to `None` in the case that
```
jmeyers314: Yep.  Reworded.

galsim/sed.py
```diff
@@ -47,78 +68,113 @@ class SED(object):
...
+            elif flux_type.lower() == 'fphotons':
+                flux_type = units.astrophys.photon/(units.s * units.cm**2 * self.wave_type)
+            elif flux_type.lower() == 'fnu':
+                flux_type = units.erg / (units.s * units.Hz * units.cm**2)
+            elif flux_type.lower() == '1':
+                flux_type = units.dimensionless_unscaled
+            else:
+                raise ValueError("Unknown flux_type '{0}'".format(flux_type))
+        self.flux_type = flux_type
+        if not (self.dimensionless or self.spectral):
```
jmeyers314: Nice simplification.  I changed it a little bit to avoid calling the check_* functions if possible.  Turns out that the `is_equivalent(..., spectral_density())` significantly slows down the code (at least the unit tests) if you run it for every new SED.  That's why there's a `_spectral` init kwarg if you know in advance that you're creating a spectral SED.

galsim/sed.py
```diff
@@ -47,78 +68,113 @@ class SED(object):
...
+    def __init__(self, spec, wave_type, flux_type, redshift=0., fast=True,
+                 _blue_limit=None, _red_limit=None, _wave_list=None,
+                 _spectral=None, _dimensionless=None):
 
         self._orig_spec = spec  # Save this for pickling
-        self._spec = _spec      # This is orig_spec turned into a function
-
+        # If these are known at construction time, use them.
+        self._spectral = _spectral
+        self._dimensionless = _dimensionless
```
jmeyers314: This was originally because I needed to check for both in the case something compatible with neither was passed as flux_type.  I think your solution below though simplifies this.  I did keep dimensionless as a convenience as suggested.

galsim/sed.py
```diff
@@ -213,77 +277,210 @@ def __call__(self, wave):
...
 
     def __rmul__(self, other):
         return self*other
 
     def __div__(self, other):
-        wave_factor = 1.0 + self.redshift
-        wave_type = 'nm'
-        flux_type = 'fphotons'
+        # Prohibit division of two SEDs as dimensionally inconsistent.
+        # I suppose we _could_ return a python function here, but certainly not another SED.
```
jmeyers314: Changed this to just prohibit division by spectral SED.  Added some tests too.

galsim/sed.py
```diff
@@ -292,47 +489,58 @@ def __add__(self, other):
...
+                   _blue_limit=blue_limit, _red_limit=red_limit,
+                   _spectral=_spectral, _dimensionless=_dimensionless)
 
     def __sub__(self, other):
         # Subtract two SEDs, with the same caveats as adding two SEDs.
         return self.__add__(-1.0 * other)
 
-    def copy(self):
-        import copy
-        return copy.deepcopy(self)
```
jmeyers314: Looks like the deprecated mutable GSObject methods (`setFlux`, `applyRotation`, etc.) depend on copy.  Should we maybe just deprecate GSObject.copy()?  Or is it time to actually drop these (been deprecated since v1.1).

galsim/sed.py
```diff
@@ -292,47 +489,58 @@ def __add__(self, other):
...
+        if isinstance(target_flux_density, units.Quantity):
+            target_flux_density = target_flux_density.to(
+                    _photons, units.spectral_density(wavelength_nm)).value
         factor = target_flux_density / current_flux_density
         return self * factor
 
     def withFlux(self, target_flux, bandpass):
-        """ Return a new SED with flux through the Bandpass `bandpass` set to `target_flux`. Note
+        """ Return a new SED with flux through the Bandpass `bandpass` set to `target_flux`.  Note
         that this normalization is *relative* to the `flux` attribute of the chromaticized GSObject.
```
jmeyers314: I think this is mostly a remnant from the previous version where a `Chromatic` took its normalization (in `evaluateAtWavelength` and `drawImage`) from both `chrom.obj` and `chrom.SED`.  I replaced the docstring here, in `withFluxDensity`, and in `withMagnitude` with the equivalent of:

```
        Return a new SED with flux through the Bandpass `bandpass` set to `target_flux`.  See
        ChromaticObject docstring for information about how SED normalization affects
        ChromaticObject normalization.
```

tests/test_bandpass.py
jmeyers314: Good idea.  Added this to check_chromatic_invariant.

galsim/sed.py
rmjarvis: I'd like to wait to drop 1.x deprecations until 2.0.  Ideally, any code that worked with a 1.x version of GalSim will keep working (although I doubt that is strictly true...).  Then at version 2.0, people have to adapt to all the new conventions.

But I'm fine with moving the copy() functions to deprecated and just point out that they are unnecessary unless you are doing something very weird with the internals of an object.

tests/test_chromatic.py
jmeyers314: I got this particular case back with a little bit more work in `SED.__add__`.  There are still cases that aren't picklable, of course -- anything with a python function as SED spec input for instance.

tests/test_sed.py
jmeyers314: >Shouldn't we be checking the values of these? Or are we just checking that astropy doesn't raise an exception?

I removed these, which were mostly for my own edification.  I don't think there's much added value in checking astropy.

tests/test_sed.py
jmeyers314: Fixed.

tests/test_wfirst.py
jmeyers314: >What changed to worsen this agreement?

TBH I'm not sure.  I think it might actually be a different test, and it's not impossible that the reason it succeeded before was at least partially b/c HST and SDSS have similar collecting areas.

I can understand the new test much better, and I don't think 15% is really that bad.

I did notice looking at this diff that the comment mentioning SDSS exposure time and collecting area is wrong, though, so I removed that.

galsim/sed.py
jmeyers314: Deprecated the immutable copy methods.  I left, for now, the unit tests that depend on these methods, though.

jmeyers314: >I left, for now, the unit tests that depend on these methods, though.

Whoops.  Just realized that this means the warnings get emitted during `scons tests`.  I guess I'll go back and move these tests to test_deprecated or just delete them (some look pretty trivial now that we expect immutability).

rmjarvis: Travis is also failing now because `obj.copy()` is called in the main set of tests, outside of the `warnings.simplefilter("always")` context, so it doesn't raise the warning when you finally get to test_deprecated.  So that should get fixed if you remove the tests of `obj.copy()` outside of test_deprecated.

jmeyers314: Moved all the uses of copy to test_deprecated, except for some that were ostensibly testing __imul__ and __idiv__ which I think don't actually exist anymore, so I just deleted 'em.  Travis is happy again.

galsim/base.py
```diff
@@ -1131,6 +1171,8 @@ def drawImage(self, image=None, nx=None, ny=None, bounds=None, scale=None, wcs=N
                             numpy.float32]
         @param method       Which method to use for rendering the image.  See discussion above
                             for the various options and what they do. [default: 'auto']
+        @param area         Collecting area of telescope in cm^2.  [default: 1.]
```
jmeyers314: I think it's alright as is.  In the long run, I think using the astropy.units framework for more things looks nice (especially the various `scale_unit` kwargs), but I don't think we need to tackle this everywhere right now.

rmjarvis:
I figured out what was wrong with the coveralls stuff and fixed it.  I also added codecov, which is more strict in its reporting, since it accurately checks that we cover both cases on all branches (which we don't, so the percentage is lower than coveralls, which just checks that we hit each line).  Anyway, I think this is all done and ready to merge unless anyone else would like to take a look.

jmeyers314:
@rmandelb , @dkirkby , @mdschneider  - I'll plan to merge this over the weekend unless I hear differently.

dkirkby:
I haven't used the chromatic code in galsim yet, so I won't be much use in reviewing and you shouldn't wait for me.

galsim/chromatic.py
```diff
@@ -19,45 +19,100 @@
...
+        >>> chrom_obj = galsim.ChromaticObject(gsobj).dilate(lambda wave: 1.01**wave)
+
+    In this and similar cases, the argument to the transformation method should be a python callable
+    that accepts wavelength in nanometers and returns whatever type the transformation method
+    normally accepts (so an int or float above).
+
+    One caveat to creating ChromaticObjects directly from GSObjects like this is that even though
+    the source GSObject instance has flux units in photons/s/cm^2, then newly formed ChromaticObject
+    will be interpreted as dimensionless, i.e., it will have a dimensionless SED.  See below for
+    more discussion about the dimensions of ChromaticObjects.
```
rmandelb: Might an example be useful here?

galsim/chromatic.py
```diff
@@ -19,45 +19,100 @@
...
+
+        >>> gsobj = galsim.Gaussian(fwhm=1)
+        >>> chrom_obj = galsim.ChromaticObject(gsobj).dilate(lambda wave: 1.01**wave)
+
+    In this and similar cases, the argument to the transformation method should be a python callable
+    that accepts wavelength in nanometers and returns whatever type the transformation method
+    normally accepts (so an int or float above).
+
+    One caveat to creating ChromaticObjects directly from GSObjects like this is that even though
+    the source GSObject instance has flux units in photons/s/cm^2, then newly formed ChromaticObject
```
jmeyers314: Thanks.  Fixed.

jmeyers314:
Merged after receiving approval from @rmandelb offline...

Issue #790, opened 2016-08-18
boost python 3 on ubuntu


esheldon:
On recent ubuntu, the boost python library is named

    libboost_python-py35.so

however galsim expects 

    libboost_python3.so

There is a bug for this in ubuntu (https://bugs.launchpad.net/ubuntu/+source/boost1.53/+bug/1231609) but it is three years old, so unlikely to be fixed.

I  had to add a sym link.  However on many systems the user won't be able to add the sym link due to privilege restrictions, and compiling boost separately is daunting.

Does it make sense to add a check for this library name?

rmjarvis:
Sure.  Try branch '#790'.  I think this should fix it for you.

esheldon:
This works.

But soon there will be a py36, then a py37...   So maybe it would be better to check for a version that matches the one being used to run the setup.py?

esheldon:
Actually any thing with the pattern `-py3*` would probably work fine.  If one is using anaconda one may not have a perfect match anyway.

rmjarvis:
> Actually any thing with the pattern -py3* would probably work fine. 

I don't know an easy way to have SCons check for libraries using a wildcard.  We could generate the name from the python target (not the one used to run scons, since that doesn't have to match the target python -- remember we're not using setup.py here).

esheldon:
right, this is getting complicated.   Might have to bite the bullet and add a long list of possibilities.

rmjarvis:
I switched it to test for the targeted python version.  I think this should work for the current Ubuntu distros and be relatively future-safe for upcoming python versions.

I should mention that it's also possible to manually add library names for SCons to use with `scons EXTRA_LIBS=boost_python-py35` if a user has some other unexpected name for their boost-python library.  I'm happy with adding this as a standard check, but the `EXTRA_LIBS` option should provide a workaround for whatever your system has if it's not one of our standard checks.

rmjarvis:
Closed via #800

PR #800, opened 2016-09-13
#790 Update name of boost python 3 library in SCons for Ubuntu

rmjarvis:
@esheldon reported in #790 that boost on Ubuntu uses a non-standard name for its Python 3 boost-python library.  This update just adds that name to the search list for when SCons is trying to figure out what the right linking command will be.

Probably there's not much to review here other than for Erin to confirm that it works on his Ubuntu system (which he already did, but then I changed things slightly, so... again please?) and for other people to make sure it doesn't break anything on their systems.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1543,21 +1543,23 @@ BOOST_PYTHON_MODULE(check_bp) {
     if not result:
         ErrorExit('Unable to compile a file with #include "boost/python.hpp"')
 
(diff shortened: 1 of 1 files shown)


jmeyers314:
Everything works for me: OS X {10.10.5, 10.11.6} x python {2.7.12, 3.5.2}.

rmandelb:
All works for me on both systems

PR #801, opened 2016-09-19
#580 Add coveralls support

### LINKED ISSUE #580: unit test coverage
In the course of working on #516 I tackled some of the low-hanging fruit in terms of improved unit test coverage.  However, there were some modules that could use more coverage.  This is a catch-all issue for improved unit test coverage for anyone who is interested (I may work on it myself after catching up on some other things, but anyone is welcome).

real.py, image.py, and bandpass.py are examples of files that could use some better coverage, with some decent-sized code chunks that don't get touched by the tests. But there are some more.  Obviously not highest priority but would be nice to do in the next few months, time permitting.

rmjarvis:
I've added coverage testing and reporting via coveralls to the travis config file.  I also added a program that does the equivalent of `check_yaml` for Travis to run, so it can check those as well, and also count the config lines that are run there for the coverage.

I don't know whether this should close #580 or not.  The coveralls report seems most likely to induce people to expand the coverage of the unit tests than an old issue sitting on the second page of the issues list.  So I'm inclined to close it after this.  But if people think it's useful to have an issue that explicitly requests more unit test coverage, we can leave it open.

### DIFF
--- a/tests/run_examples.py
+++ b/tests/run_examples.py
@@ -289,7 +289,7 @@ def test_demo13():
     import demo13
     import check_diff
     print('Running demo13.py')
-    demo13.main(['nuse=3','ntot=100'])
+    demo13.main(['nuse=3','ntot=100','filters=YJH'])
     # There is no demo13.yaml yet, so all this does is check for syntax errors in demo13.py.
 
 @timer
--- a/tests/run_examples.py
+++ b/tests/run_examples.py
@@ -328,12 +328,14 @@ def test_des():
         config['gal']['items'][0]['gal_type'] = 'parametric'
         config['input']['cosmos_catalog']['file_name'] = '../data/real_galaxy_catalog_23.5_example.fits'
         del config['input']['cosmos_catalog']['sample']
+        config['input']['des_wcs']['bad_ccds'] = range(2,63)  # All but CCD 1
         galsim.config.Process(config, logger=logger)
         input_cosmos = config['input']['cosmos_catalog'] # Save example COSMOS catalog spec.
         config = galsim.config.ReadConfig('blend.yaml', logger=logger)[0]
         galsim.config.Process(config, logger=logger)
         config = galsim.config.ReadConfig('blendset.yaml', logger=logger)[0]
         config['input']['cosmos_catalog'] = input_cosmos
+        config['input']['des_psfex']['file_name']['num'] = 1
         galsim.config.Process(config, logger=logger)
 
     finally:
--- a/tests/run_examples.py
+++ b/tests/run_examples.py
@@ -57,6 +57,23 @@ def in_examples(f):
 
 @timer
 @in_examples
+def test_demo1():
+    """Check that demo1 runs properly.
+    """
+    import demo1
+    import check_diff
+    print('Running demo1.py')
+    demo1.main([])
+    logging.basicConfig(format="%(message)s", level=logging.INFO, stream=sys.stdout)
+    logger = logging.getLogger('galsim')
+    config = galsim.config.ReadConfig('demo1.yaml', logger=logger)[0]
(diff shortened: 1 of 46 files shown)


rmandelb:
Possibly I am doing something wrong (?), but when I go to this branch and try to do the usual ```scons``` / ```scons install``` / ```scons tests```, I get a failure of ```scons tests```:
```
Starting python tests...
bash: /Library/Frameworks/EPD64.framework/Versions/Current/bin/nosetests test*.py: No such file or directory
Nosetests returned error code  127
scons: *** [tests/tests.log] Explicit exit, status 127
```
Perhaps it's not going into the tests/ directory before running this command, and that's why it's not finding any ```test*.py``` files?

rmandelb:
That is very cool.  

> Should we add the coveralls badge (and the travis CI badge, and an arxiv badge) to the top 
> of README.md?

+1

> I don't know whether this should close #580 or not.

Given that it appears way down the list, so it mostly isn't seen, I'd go with closing it.  The coveralls report will indeed be a more useful reminder.

tests/run_examples.py
```diff
@@ -0,0 +1,304 @@
...
+    """
+    import demo13
+    import check_diff
+    print('Running demo13.py')
+    demo13.main([])
+    # There is no demo13.yaml yet, so all this does is check for syntax errors in demo13.py.
+
+
+if __name__ == "__main__":
+    shutil.rmtree('../examples/output')
```
jmeyers314: Should probably make this path with respect to `__file__`, in case someone starts this script from a directory other than $GALSIM_DIR/tests/.

Also, I think this line crashes if the target directory doesn't exist.

jmeyers314:
One of these days I should learn how the config framework works, so I can give that a proper review.... (I did scan it briefly).  But until then, this looks good to me, just the one very minor point above.

tests/run_examples.py
rmjarvis: Good points.  Fixed.

rmjarvis:
OK, I think I'm done futzing with this now.  I added a .coveragerc file to have it ignore various kinds of things that we don't actually care about in terms of unit test coverage.  And for some specific bits of code (special cases for old systems, obsolete code that's still sitting there, warnings about bad user input, etc.) I marked them with a `pragma: no cover` comment to exclude them from the coverage report.

I also added code to run the des and great3 examples, which covers a bit more of the code base, at least for syntax errors, if not actual correctness testing, although I think for the additional bits that these catch, syntax errors are probably the main thing we'd worry about.

After all that, the remaining bits I think are code that ought to be included in the unit tests, and of them, we're currently at 93% coverage.  Not too shabby, IMHO.  You can peruse them [here](https://coveralls.io/github/GalSim-developers/GalSim).  It's pretty cool -- the covered lines are green, and the uncovered lines are pink.

I even felt inspired to add a unit test for the `NFWHalo.getLensing` function that was completely untested.  And in doing so, I discovered a bug in it, so that was useful.  I also found a number of typos in uncovered bits of code (like writing `NotImplemented` rather than `NotImplementedError`), so I fixed them as well.

jmeyers314:
Very nice.

Is it straightforward to cherry-pick some of these onto 1.4 so we can have the badges show up on the default github landing page?

rmjarvis:
> Is it straightforward to cherry-pick some of these onto 1.4

Done.

rmjarvis:
OK to merge this?  Or did anyone want to look through my choices for things to skip coverage on?  

I know @jmeyers314 you looked through the code early on, but then I added a bunch of `pragma: no cover` comments, which I don't know if you care about reviewing.

rmandelb:
I'm fine with merging.

jmeyers314:
Me too.

Issue #770, opened 2016-07-08
AEGIS catalog


jmeyers314:
This issue is to finalize and add the HST F606 and F814 AEGIS catalog that @sowmyakth has been assembling into GalSim (mentioned in #551).  Our goals for this catalog are to 1) be compatible with the existing GalSim COSMOS catalogs, and 2) be a source catalog for ChromaticRealGalaxy #640.  

We're also hoping to add to the repository the python scripts that @sowmyakth has been developing for turning HST images into catalogs (based off of a variety of scripts from @rmandelb), which should be generic enough to be useful for adding other HST surveys down the line (e.g. CANDELS).

First things first though: can @rmandelb or @rmjarvis please add @sowmyakth to the GalSim-developers team?

rmandelb:
Glad to see this!  How is it going?  Need any further help translating / replacing my scripts?

sowmyakth:
I'm almost done. I'm working now on getting the required files for galsim. In the selection.fits why do you save the distance to the nearest replaced pixel and the average around this pixel?

rmandelb:
Because we wanted to be able to select (+remove if necessary) galaxies that were more like to have been blended with another nearby object, or objects with sky subtraction errors or failed blends.  Both of those quantities can be used for that purpose.

sowmyakth:
@rmandelb For the file containing correlated noise field is what was reasoning for naming it acs_I_unrot_sci_20_cf.fits? What is sci_20? 
For AEGIS I took a blank region, got the correlated noise using galsim.CorrelatedNoise and drew the image. Is this the right method?

rmandelb:
Hi Sowmya -

The name of this file has its origins inside the brain of @barnabytprowe :) but you can see more about it here: https://github.com/GalSim-developers/GalSim/pull/358 where he describes the process of producing those files.  The relevant scripts are in the GalSim repo, devel/external/hst/ if you want to check it out and make sure you do something comparable for AEGIS?

For COSMOS, he used more than one blank field, partly to beat down the noise in estimating the noise correlation function, and partly because the correlated noise varies somewhat as a function of position, so he took the average to be representative.  I would think that AEGIS is small enough that you wouldn't need as many blank fields as he used (indeed it would be hard to find so many) but you might want more than one, since the noise correlation function estimated from one field would be noisy.  Can you see how noisy your CF image is?

[...]

Issue #795, opened 2016-09-01
make COSMOSCatalog file treatment more general


rmandelb:
I recently had a reason to make a new version of the 25.2 sample COSMOS catalog with some modifications, and discovered that when I tried to read in and use the new catalog, there were a few issues:

1) I used the same file naming scheme (`real_galaxy_catalog_SAMPLE.fits` and similar patterns for the file with selection criteria) just with a different SAMPLE.  However, the code is not set up to try this naming scheme if you tell it a new value of `sample`.  It'll just complain that it doesn't know how to deal with that sample.  I would like it to try to find files with the default naming scheme for that sample, and only if it can't find them in `dir` should it say it doesn't know what to do for this `sample`.

2) If you give it the `file_name` kwarg, it tries to find the selection file for the default sample (and issues a warning).

I propose to make minor code updates in scene.py to address both of these issues.

rmandelb:
@sowmyakth -

To expand a little on what's needed here:

You can basically trace through how the input ```sample``` kwarg is handled to see what might be needed.  It needs to change so as to allow other ```sample``` values, and check whether files exist that follow our usual naming scheme (if yes, use them; if not, complain).  However, it should not go so far as to tell you to try to download any of the helper files for that new ```sample``` with ```galsim_download_cosmos```; that action should only be taken if ```sample``` WAS in the list of samples that GalSim knows about.

However, there is also some work needed if a filename is used to specify the new sample.  I'm suggesting that scene.py should be told what the default file naming scheme is, and try to parse the filename to figure out the sample.  Then it can look for selection files and/or parametric fits files for that sample (rather than for the default sample).

Does that make sense?

rmjarvis:
Cross reference Issue #731.  Making the COSMOSNoise stuff more generic might be worth doing at the same time.

rmandelb:
Yep.  Though I was also thinking of generalizing #731 not just to COSMOSNoise but also COSMOSCatalog (-> HSTCatalog, with COSMOSCatalog being allowed but deprecated, and it would just call HSTCatalog).

sowmyakth:
[...]

PR #806, opened 2016-09-26
#770 Script to create AEGIS catalog

sowmyakth:
Script to create catalog of galaxies from multi band HST images. The script was developed to create a source catalog for ChromaticRealGalaxy by reducing images from the AEGIS survey, but is generic enough to be used with any multiband HST data.  The method used to reduce the images is similar to the method used for reducing [COSMOS  images](http://adsabs.harvard.edu/cgi-bin/bib_query?arXiv:astro-ph/0702359). 

The script takes HST images in multiple bands as input, identifies objects with SExtractor, classifies those objects as stars and galaxies. Stars are then used to measure the PSF. Galaxies that satisfy certain criteria are selected for the main catalog. Separate catalogs are made for different bands. A postage stamp image is drawn for every galaxy in the catalog in each band. Each galaxy image will also have a postage stamp of its PSF in each band. The output catalog can be opened using the RealGalaxyCatalog module of GalSim.


devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
```diff
@@ -0,0 +1,109 @@
+This README file describes the AEGIS F814W<25.2 training sample, which is a
+randomly ordered sample of galaxies in the AEGIS field with AB magnitude<25.2, in F606W(V) and F814W(I) bands.
+
+NOTE: The columns in the catalog are those expected by galsim.RealGalaxyCatalog.
+However, some parts of that analysis, i.e parametric fits, were not performed.
+Hence those values were manually set and do not have physical meaning.
+
+(1) The catalogs themselves are in AEGIS_galaxy_catalog_[VI]_25.2.fits (where
```
rmjarvis: I didn't see any instructions about where to get this file.  Maybe this is waiting on @joezuntz to put them up at Manchester?  Should also add these instructions to the [wiki page](https://github.com/GalSim-developers/GalSim/wiki/RealGalaxy%20Data).

We could also consider adding this to the `galsim_download_cosmos` (perhaps renaming this program) as another sample.  `-s aegis` perhaps.  I'm ok with doing this on #795 to make sure it is compatible with any API changes you decide to make there.

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
```diff
@@ -0,0 +1,109 @@
...
+NOISE_MEAN: mean value of the pixels in the image that do not contain the galaxy.
+
+NOISE_VARIANCE: variance of the pixel values for pixels in the image that do not contain the galaxy.
+
+
+(2) The files containing galaxy and PSF images are
+AEGIS_galaxy_images_[VI]_25.2_n?.fits and AEGIS_galaxy_PSF_images_[VI]_25.2_n?.fits. The number that goes in place of the "?" ranges from 0 to 29.
+
+
+(3) *Parametric fits have not been performed and the values in the table were manually set.*
```
rmjarvis: Is there any plan to do parametric fits?  I guess this is probably a lot of work, but I think it would be useful, since they are much faster to render than RealGalaxies, so it significantly speeds up simulations that don't need the absolute full fidelity of real galaxies, but want some plausibly accurate distribution of size/shapes.

devel/external/AEGIS/README.md
```diff
@@ -0,0 +1,54 @@
+This README file describes the python code used to process AEGIS HST images in V and I bands to produce multi-band postage stamp images of galaxies. The selection procedure is explained in a separate [document](https://www.overleaf.com/read/krbzsccfdbpm).
```
rmjarvis: This is also the document in external/AEGIS/document, right?  Which one is going to be the definitive version if they get out of sync?  

If it's the Overleaf version, then I think maybe we don't necessarily want it in the GalSim repo as well.  

If it's the GalSim version, then I'd suggest rewording this sentence to mention this version first.  e.g.
"... in a separate document in devel/external/AEGIS/document/main.tex; a rendered version of the document may be viewed at https://www.overleaf.com/read/krbzsccfdbpm"

devel/external/AEGIS/document/apj.bst
```diff
@@ -0,0 +1,1568 @@
+% yahapj.bst -- yet another ApJ bibtex style file
```
rmjarvis: Probably don't need this file in the repo.

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
sowmyakth: Yes, this is the README file for the AEGIS Catalog. I was waiting to see if I need to make any changes in the catalog before uploading it.

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
sowmyakth: I wasn't planning on doing the parametric fits for now. I won't be needing it for color gradient measurements with real chromatic galaxies.

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
```diff
@@ -0,0 +1,109 @@
+This README file describes the AEGIS F814W<25.2 training sample, which is a
+randomly ordered sample of galaxies in the AEGIS field with AB magnitude<25.2, in F606W(V) and F814W(I) bands.
```
rmandelb: The flux limit is imposed in both bands?  I thought this was something we had discussed as a possible source of unexpected selection effects (because it omits certain regions of color space once you're near the flux limit in one band).  And more to the point, imposing selection in both bands means that you can't make catalogs that correspond to each other.  It seems that for many use cases, we will want a cut in one band, with images provided in both bands regardless of the flux in the other band (e.g., so dropouts will not be excluded from the catalog).

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
```diff
@@ -0,0 +1,109 @@
...
+
+NOTE: The columns in the catalog are those expected by galsim.RealGalaxyCatalog.
+However, some parts of that analysis, i.e parametric fits, were not performed.
+Hence those values were manually set and do not have physical meaning.
```
rmandelb: I'm a little surprised by this comment.  The main ```real_galaxy_catalog_SAMPLE.fits``` file format used for galsim.RealGalaxyCatalog() doesn't have any information about parametric fits, so it should be possible to define all columns in a way that is meaningful for this sample.  Which ones were you not able to do?

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
```diff
@@ -0,0 +1,109 @@
...
+
+(1) The catalogs themselves are in AEGIS_galaxy_catalog_[VI]_25.2.fits (where
+[VI] indicates either V or I).  These files can be used to initialize a
```
rmandelb: Need to comment on the correspondence between the V and I catalogs: was dual image mode used for measurements (yes), do the entries in the catalogs correspond to each other (I imagine not?), etc.?  For what it's worth, I was expecting a single catalog with information about both bands, where you'd have the same IDENT appearing twice but with different values of 'BAND' and locations of the images.  What is the motivation for separating the bands into different catalogs?  

Note that using such a multi-band catalog that I had envisioned would require some updates to how COSMOSCatalog and RealGalaxyCatalog are used to make images, but it should be possible for them to be backwards compatible.

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
```diff
@@ -0,0 +1,109 @@
...
+needed for image simulation, it is necessary for proper cross-matching with
+other catalogs.
+
+MAG: magnitude (MAG_AUTO calculated from sextractor).
+
+BAND: Passband for the observed image (F814W/F606W).
+
+WEIGHT: A weight factor to use to account for the fact that larger galaxies are
+preferentially excluded from this catalog due to the proximity to CCD edges.
+(Currently all weights are set to 1.0).
```
rmandelb: Why?  The weights are useful for undoing selection effects related to postage stamp creation.

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
rmandelb: If you don't plan to make parametric fits and only intend for people to use this as a RealGalaxyCatalog, then why make a file for parametric fits at all?  If you want to use a RealGalaxyCatalog (or a COSMOSCatalog without parametric fits enabled) you don't need this file, and it seems better to omit it than to have a fake one.

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
```diff
@@ -0,0 +1,109 @@
...
+
+(3) *Parametric fits have not been performed and the values in the table were manually set.*
+A catalog containing information about the fits to the galaxies in
+AEGIS_galaxy_catalog_[VI]_25.2.fits, using the fitting code from Lackner &
+Gunn (2012, MNRAS, 421, 2277) as described in the GREAT3 challenge handbook.
+The catalog of fit information is AEGIS_galaxy_catalog_I(V)_25.2_fits.fits
+(a FITS table), and the entries are in the same order as in
+AEGIS_galaxy_catalog_I(V)_25.2.fits.  The instructions below tell how to use
+the entries in this file to construct galaxy images; however, if you use the
+COSMOSCatalog class, GalSim will take care of these details for you.  The file
```
rmandelb: This sentence is not true, because the entries are not meaningful.  I would omit this sentence.

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
```diff
@@ -0,0 +1,109 @@
...
+contains the following fields:
+
+IDENT: AEGIS identifier.
+
+MAG_AUTO: magnitude (MAG_AUTO computed by SExtractor).
+
+FLUX_RADIUS: half-light radius for the PSF-convolved object (same as flux_radius from the AEGIS catalog), in units of pixels.
+
+ZPHOT: Photometric redshift from the DEEP2 galaxy redshift catalog Data Realase 4 (Newman et al.
```
rmandelb: Release

devel/external/AEGIS/README.md
```diff
@@ -0,0 +1,54 @@
+This README file describes the python code used to process AEGIS HST images in V and I bands to produce multi-band postage stamp images of galaxies. The selection procedure is explained in a separate document in devel/external/AEGIS/document/main.tex; a rendered version of the document may be viewed at https://www.overleaf.com/read/krbzscc. 
+The final products are similar to the COSMOS Real Galaxy training sample.
+The script can be implemented to analyze any HST field data in multiple filters. It takes HST images in multiple bands, identifies objects with SExtractor, classifies those objects as stars and galaxies. Stars are then used to measure the PSF. Galaxies that satisfy certain criteria (see Section 3.8 in document) are selected for the main catalog.Separate catalogs are made for different bands. A postage stamp image is drawn for every galaxy in the catalog in each band. Each galaxy image will also have a postage stamp of its PSF in each band. The output of the code can be opened using the RealGalaxyCatalog and COSMOSCatalog module of GalSim.
```
rmandelb: Should explain why separate catalogs are made for different bands.  There are pros and cons to this and it's not completely obvious why this way was chosen.

devel/external/AEGIS/README.md
```diff
@@ -0,0 +1,54 @@
...
+* star_galaxy_params: Parameters used to separate galaxies and stars in MU_MAX Vs MAG_AUTO plot(x_div, y_div, slope). x_div gives the maximum magnitude, below which the object is saturated. y_div is the value of surface brightness per pixel for a saturated star. The slope of the line separating stars and galaxies is given by slope (See Fig 2 in document). It is recommended to run the first script get_objects.py, with mock sat_galaxy_params values, on a small region and then compute the separation parameters from the objects measured. The parameters may be different for different bands.
+* gain: Detector gain in e/ADU 
+
+### Scripts: 
+The scripts are run entirely in python. Make sure the following modules are loaded:
+galsim, numpy, astropy, asciidata, subprocess, os, scipy.
+The detection and measurement is performed with [SExtractor](http://www.astromatic.net/software/sextractor)
```
rmandelb: Any restrictions on what sextractor version is needed in order for this to work?

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
sowmyakth: The 25.2 magnitude cut was imposed only on the I band. I'll change the wording in the README file.

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
sowmyakth: It's just parametric fits that I wasn't able to. Also, I only have reshift information of about 3,000 out of the 29,000 galaxies in the catalog

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
sowmyakth: The entries and ```IDENT``` in both catalogs correspond to the same galaxies. I made separate catalogs so that each could be used independently. Also  ```ChromaticRealGalaxies```  takes two separate ```RealGalaxyCatalog```s and an ```ident``` as input to create the chromatic galaxy. @jmeyers314 do you have any comments on this?

sowmyakth:
@rmandelb I made AEGIS catalog with only the magnitude cut with the intention that other cuts can always be placed later. However, about 10% of the galaxies would fail the ```exclusion_level="bad_stamp"```  in ```COSMOSCatalog```. Since the aim is for this catalog to be used only with ```RealGalaxyCatalog```, do I remove those galaxies from the catalog?

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
sowmyakth: I didn't do any size dependent weighting. When you sent me your IDL script, I thought you asked me to ignore the weight part. Did I confuse that with some other weight?

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
jmeyers314: I think I slightly prefer having separate (but aligned) catalogs for V and I.  For one thing, it's backwards compatible with `RealGalaxyCatalog`.

The way I've currently set up `ChromaticRealGalaxy` is to take a list of catalogs as one of its input parameters:

```python
crg = galsim.ChromaticRealGalaxy([Vcat, Icat], index=123)
```

Part of the motivation here was thinking ahead to a CANDELS catalog, where we might want to test the difference between using all 4 bands as input to using only 2 or 3 bands as input.  This is trivial with the above API and independent catalogs.

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
rmandelb: Great, please do make that change.

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
rmandelb: OK - so it seems like ```AEGIS_galaxy_catalog_[VI]_25.2.fits``` are complete, and it's just the other files that might not be complete? If so, I would recommend moving this comment down past (1).  Its current location implies that all the catalogs may be incomplete, but it seems from your github comment that (1) is complete.

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
rmandelb: OK, sounds good.  In that case I think all that's needed is for @sowmyakth to address the first part of my comment - that this file should be updated to clarify the correspondence between catalogs.

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
rmandelb: If you're using DEEP2, then is this really a photometric redshift?  Or should this say "spectroscopic"?

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
rmandelb: Well, I'm not sure what the best thing to do is.  Using COSMOSCatalog lets you apply some cuts in the constructor (on stamp quality etc.).  I don't love the fact that it involves a parametric catalog that is completely meaningless.  So perhaps the answer is we should make it possible to use COSMOSCatalog without parametric fits being available, and then remove this file.

rmandelb:
> However, about 10% of the galaxies would fail the exclusion_level="bad_stamp" in 
> COSMOSCatalog.

The question I would be asking is whether that cut is even relevant for this sample, given the differences in the data.  Have you done some inspection of the properties of these objects, to decide whether it makes sense to be cutting them out?

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
sowmyakth: I reworked some portion of the code and included some selection cuts. 26,517 galaxies  pass the cuts out of the total 28,532 galaxies with I band magnitude <25.2. Previously with `exclusion_level=bad_stamp` I was losing about 10% galaxies. I visually inspected the galaxies and realized that most of these were large galaxies for which HSM adaptive moments failed to converge. I altered the input HSM params and most of these passed, except galaxies with high ellipticities. So I decided to include galaxies with `ELLIPTICITY` >0.75 computed by SExtractor, even if  adaptive moments were not computed; I set `sn_ellip_gauss`= -10 manually.

A galaxy is included in the final catalog if:
1)  `sn_ellip_gauss` > 0 or ELLIPTICITY >0.75, in both filters, and 
2) Closest masked pixel is more than 5 pixels from the center, and
3) If closest masked pixel is closer than 11 pixels, `average_mask_adjacent_pixel_count` must be less than 20% of the brightest pixel.

With these cuts, this is the distribution of galaxies that would make it to the final catalog. The left column is the fraction of galaxies that pass selection cuts as a function of half light radius, ellipticity and position angle (measured by SExtractor), and magnitude. The right column is the histogram of all galaxies(without applying cuts)  as a function of the same parameters.

screen shot 2016-11-01 at 2 19 49 pm

The size dependent selection effect I get is different from COSMOS. Out of the 32 galxies larger than 1.4 arcsec only 6 fail the selection cut.  At this point should any other selection cuts be applied? 

From the COSMOS <25.2 catalog, I plotted the weight and  flux radius for the galaxies. The histogram of the the flux radius seems to resemble the distribution for AEGIS. 

screen shot 2016-11-01 at 2 40 17 pm

@rmandelb  I'm not clear as to how the weight was computed for COSMOS?

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
sowmyakth: That's my mistake.. Will correct it.

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
rmandelb: The weight was computed as follows:

- We took our original catalogs based on sextractor and applied whatever selection criteria were used to decide whether we wanted to make a postage stamp (e.g., if you only have a magnitude cut, this would just be a magnitude cut).  This gives us what we call the "parent sample".
- We then took the galaxies for which we actually made postage stamps.  Some galaxies didn't get postage stamps because the stamp would have overlapped with the edges of the individual COSMOS pointings, which would have led to funny image artifacts.  This is the "postage stamp sample".  Ideally it's the same as the parent sample, but for the nuisance effects that make us choose not to make a postage stamp.
- Plot the histogram of flux radius for the parent sample and the postage stamp sample.  The ratio of these is the weight, i.e., we had a deficit of high flux radius objects and that's why the weight is increased for those.

The reason we only did this for flux radius is that the postage stamp size had an explicit flux radius-dependence, and so we knew that if any parameter was going to be a problem, it would be that one.

rmandelb:
Hi @sowmyakth - just checking in because I haven't seen updates on this in a while (not since we talked on slack).  Is there progress on the final things we discussed?  Are you waiting on me for anything?

sowmyakth:
@rmandelb I'm yet to modify the tex file to include the changes made. But the code is complete. Will you be going through each script file?

rmandelb:
Yes, I will go over the code once the last set of changes/updates we discussed is done.  Please let me know once you're ready.

rmjarvis:
This PR has been dormant for over a year now.  Has the project been abandoned?  I thought it was pretty close to finished.

If it's ready for me to take another look, I'd be happy to.  I was waiting for some word from @sowmyakth that she was done making changes in response to my and Rachel's reviews.

sowmyakth:
Looks like I dropped the ball on that one. Yes I made the changes.  @rmandelb   could you review changes made?  
@joezuntz Could you put this  catalog on the GREAT3 server, with other HST-based catalogs. This is the google drive [link](https://drive.google.com/drive/folders/1McFKROhPMOLJv8rLAUGiTaxcww5vg0Pq?usp=sharing). 

@rmjarvis I'll add the link to the permanent catalog location in the tex file and README once that's finalized, and then this should be good to go.

devel/external/AEGIS/AEGIS_25.2_training_sample_readme.txt
```diff
@@ -0,0 +1,68 @@
...
+flags that the GalSim COSMOSCatalog class may use to impose selection criteria
+on the quality of the postage stamps and/or fits.
+
+IDENT: AEGIS identifier.
+
+dmag: not used
+
+sn_ellip_gauss: SNR of the galaxy measured with an elliptical gaussian filter.
+
+min_mask_dist_pixels: Distance from the center of postage stamp tp the nearest masked pixel.
```
rmandelb: "tp" -> "to"

devel/external/AEGIS/README.md
```diff
@@ -0,0 +1,51 @@
+This README file describes the python code used to process AEGIS HST images in V and I bands to produce multi-band postage stamp images of galaxies. The selection procedure is explained in a separate document in devel/external/AEGIS/document/main.tex; a rendered version of the document may be viewed at https://www.overleaf.com/read/krbzsccfdbpm. 
+The final products are similar to the COSMOS Real Galaxy training sample.
+The script can be implemented to analyze any HST field data in multiple filters. It takes HST images in multiple bands, identifies objects with SExtractor, classifies those objects as stars and galaxies. Stars are then used to measure the PSF. Galaxies that satisfy certain criteria (see Section 3.8 in document) are selected for the main catalog. Separate catalogs are made for different bands. Each catalog can be individually used to initialize a galsim.RealGalaxyCatalog or galsim.COSMOSCatalog instance.  A postage stamp image is drawn for every galaxy in the catalog in each band. Each galaxy image will also have a postage stamp of its PSF in each band. The entries in different filter catalog files correspond to the same galaxies.
```
rmandelb: I recommend saying a tiny bit more about PSF estimation.  For example "Stars are then used to measure the PSF by comparing with precomputed TinyTim PSFs and identifying the focus position that most closely matches this exposure.  Once the stars have been used to identify the focus position, the TinyTim PSF models are used for PSF estimation as a function of CCD position."  (I think that's what you did, right?)

devel/external/AEGIS/add_weights.py
```diff
@@ -0,0 +1,90 @@
...
+the final catalog compared to the parent catalog was measured as a function 
+of several parameters like FLUX_RADIUS, ELLIPTICITY, MAGNITUDE etc., to check
+for selection bias. The distribution was observed to show a dependance on the
+half light radius of the galaxy measured by SExtractor. The weight is defined 
+as the ratio of the number of galaxies in the parent sample to the number in 
+the final catalog, at a given half light radius. Due to the small number of 
+galaxies at high half light radius in the parent sample, this definition  
+breaks down. Thus weight for large galaxies is set to 1, and the weights of
+other galaxies were renormalized. 
+Since the magnitude cut in the parent sample was placed in teh I band, the
```
rmandelb: typo, "teh"

devel/external/AEGIS/document/main.tex
```diff
@@ -0,0 +1,516 @@
...
+
+\subsection{Parametric Fits}
+The COSMOS catalog galaxies were subject to an additional step of fitting for parametric models. This step was not performed here, and the values in the output catalogs corresponding to these parameters were manually set to 999.
+
+\subsection{PSF Postage Stamp}
+The PSF for each galaxy is taken to be the {\tt Tiny Tim} model PSF image whose location in the {\tt Tiny Tim} grid is closest to that galaxy's position in the focal plane, for the best-fit focus offset described in \autoref{ssec:psf}. Square postage stamps ($20 \times 20$ pixels) are drawn for each filter.
+
+\subsection{Additional Catalogs}
+The Extended Groth Strip (EGS) has been extensively studied with other surveys in other bands. 
+We include in our catalog photometry and photometric redshift measurements from other catalogs, for objects in our catalog with a center within $1''$ of an object in the other catalog.
```
rmandelb: I'm pretty sure you have spectroscopic redshifts, not photometric.  If it's DEEP2 then it's spectroscopic (correction needed here and also a few lines below here).

devel/external/AEGIS/document/main.tex
```diff
@@ -0,0 +1,516 @@
...
+\centering\includegraphics[width=1.\linewidth]{noise_map.png}
+\caption{Image of noise map generated from covariance matrices, for I band (left) and V band (right).}
+\label{fig:noise_map}
+\end{figure}
+%----------------------------------
+
+To replace nearby objects with noise, a region that is identical in size and shape to the nearby object in the segmentation map is cut from a random location in the noise map shown in \autoref{fig:noise_map} and used to replace the pixel values for the nearby object in the galaxy postage stamp.  
+
+\subsection{Parametric Fits}
+The COSMOS catalog galaxies were subject to an additional step of fitting for parametric models. This step was not performed here, and the values in the output catalogs corresponding to these parameters were manually set to 999.
```
rmandelb: This is not precisely true as stated.  The COSMOS catalog you have referred to so far is the Leauthaud et al (2007) catalog, unless I am missing some other reference.  But that catalog had no parametric fits.  The fits referenced here were produced for the GREAT3 challenge as described in e.g. the GREAT3 handbook.

sowmyakth: I've now mentioned GREAT3 for the parametric fits, and also added a reference at the beginning that the final products are similar to the great3 datasets.

devel/external/AEGIS/clean_pstamp.py
```diff
@@ -0,0 +1,306 @@
...
+                    oth[str(seg_map[x,y])] = [[x,y]]
+                    oth_segs.append(str(seg_map[x,y]))
+    if seg_map[s[0]/2, s[1]/2] != int(num)+1:
+        check = 1
+    return im, bl, oth, oth_segs, check
+
+def get_stats(arr ,str=None):
+    """Returns mean and stdev of arr """
+    mean = np.mean(arr)
+    t2 = arr - mean
```
rmandelb: variables t2 and t3 in this routine don't seem to be used or needed?

devel/external/AEGIS/clean_pstamp.py
```diff
@@ -0,0 +1,306 @@
...
+    if x0>0:
+        x.append(x0-1)
+    if arr.shape[0]-1>x0:
+        x.append(x0+1)
+    if y0>0:
+        y.append(y0-1)
+    if arr.shape[1]-1>y0:
+        y.append(y0+1)
+    neighb = [arr[i][j] for i in x for j in y]
+    avg=sum(neighb)/len(neighb)
```
rmandelb: np.mean(neighb)?

devel/external/AEGIS/clean_pstamp.py
```diff
@@ -0,0 +1,306 @@
...
+postage stamp of galaxy, corresponding segmentation map, noise file
+
+Identification:
+The segmentation map corresponding to the object is used to divide the
+postage stamp pixels into tho belonging to the main galaxy, other 
+objects and background. The pixels belonging to other objects are replaced
+by pixels from the noise file. 
+
+Replace pixels:
+The other object pixels are replaced with a region of noise map identical
```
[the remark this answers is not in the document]
sowmyakth: What I wanted to imply here was that a region was cutout from the noise map ( same statistical properties) with same region as the other object's segmentation map. I rephrased it now. Is it clear now or should I add additional info?

sowmyakth:
Thanks @joezuntz for uploading the catalog to the GREAT3 server. @rmandelb, @rmjarvis I've added the link to the catalog in the README file and in the tex file. If there are no more suggested changes, then the branch is ready to be merged.

rmjarvis:
Thanks Somya.  Looks good to me.  Thanks for finishing this up.  I think it will be quite useful!

Issue #805, opened 2016-09-21
WCS-related warnings in correlatednoise.py


rmandelb:
Also while working on HSC simulations, I am getting some warnings that I'm not sure I need to worry about (but, if I don't need to worry about them, then I wonder if they should be emitted at all?).

In particular, with real galaxies and noise whitening turned on, I am getting
```
UserWarning: Subtracting two CorrelatedNoise objects with different WCS functions.
The result will have the WCS of the first object.
```
So I locally modified GalSim to spit out the ```__str__()``` of the two WCS's to see why they are different.  The result was as follows:
```
galsim.PixelScale(0.168)
```
for one of them and
```
galsim.OffsetWCS(0.168, galsim.PositionD(x=3712.0, y=4160.0), galsim.PositionD(x=0.0, y=0.0))
```

It seems to me that this is not a problem... in one case it's just using a PixelScale and in the other case, it has a PixelScale with an offset to account for wherever we are in the big image.  Am I correct that this is safe to ignore?

And if I'm correct, then would it make sense to check a little more carefully and not emit this warning in the cases where the two WCS's are "something" and "offset version of something"?

rmjarvis:
It seems reasonable to consider wcs's compatible if they everywhere have the same local jacobian.
```
def compatible_wcs(wcs1, wcs2):
    if wcs1.isUniform():
        return wcs1.jacobian() == wcs2.jacobian()
    else:
        return wcs1 == wcs2.withOrigin(wcs1.origin, wcs1.world_origin)

assert compatible_wcs(galsim.PixelScale(0.23), galsim.PixelScale(0.23))
assert compatible_wcs(galsim.PixelScale(0.23), galsim.OffsetWCS(0.23, galsim.PositionD(12,34)))
assert compatible_wcs(galsim.JacobianWCS(0.2,0.01,-0.02,0.23),
                      galsim.AffineTransform(0.2,0.01,-0.02,0.23,
                                             galsim.PositionD(12,34),
                                             galsim.PositionD(45,54))
                     )
assert compatible_wcs(galsim.PixelScale(0.23),
                      galsim.AffineTransform(0.23,0.0,0.0,0.23,
                                             galsim.PositionD(12,34),
                                             galsim.PositionD(45,54))
                     )

assert not compatible_wcs(galsim.PixelScale(0.23), galsim.PixelScale(0.27))
[...]

Issue #804, opened 2016-09-21
config verbosity for COSMOSCatalog


rmandelb:
I want to recommend that config by default (i.e., for the default verbosity level) say a little more about what it's doing with galaxies drawn from COSMOSCatalogs.

In particular, I have had multiple issues (like #802) where it would have been really helpful if the default verbosity level would issue a single statement at the start of a given file saying (a) exactly what catalog is being used and (b) whether the galaxies are real or parametric.

For example, in #802, I had used a template with parametric galaxies, then had the .yaml file later modify to use real galaxies, except it quietly ignored that.  It took me ages before I realized it was not in fact using real galaxies.  I had to go all the way up to verbosity=3, and then notice that it was calling the galaxy generator for parametric galaxies.  If the default verbosity had a single statement at the start of making each large image saying what COSMOSCatalog was used, and how it's being used, I think that would make debugging simple without adding too much clutter to the outputs.

Thoughts?

rmjarvis:
The place to do this would be in `galsim/config/input_cosmos.py`.  Specifically, the (currently quite short) function `COSMOSLoader.setupImage`.  This is the function that gets run at the start of each large image when there is a cosmos_catalog item to process.  

The current function outputs how many objects are in the catalog total and how many passed the initial cuts, since I thought those were interesting bits of information people might want.  But you could certainly add additional lines there that you think would be useful.

The other option would be to override `getKwargs`.  Currently it is using the base class implementation (in `galsim/config/input.py`), but you could modify that to first run the base class version and then output some logging statement about the file name and options for the COSMOSCatalog that is about to be built.

rmandelb:
OK, will try it out.  Thanks.

rmandelb:
Closing #804.

PR #807, opened 2016-09-27
#804 and #805

rmandelb:
This is a small PR that addresses both #804 and #805 (where work was done on branch #805).

For #805, the solution is exactly as written in the issue.

For #804, there were two changes:
- make config say the basic info about the COSMOS catalog at the default verbosity level
- make it say a bit more information as well

So for example, on master if I run cgc.yaml and tell it to use my version of the catalog sitting in a non-default directory, the output says nothing whatsoever about the catalog that is being used (you don't get that information until verbosity=2).  On branch #805, the output includes the following:
```
Using user-specified COSMOSCatalog: 
  sample = 23.5
  dir = /Users/rmandelb/great3/data-23.5/update_tarball/COSMOS_23.5_training_sample
file 0: COSMOS catalog has 56062 total objects; 21883 passed initial cuts.
Using parametric galaxies.
```

EDIT: based on suggestions from Mike, the above output is what you get at default verbosity, minus the ```file 0``` line.  The ```file 0``` line shows up at the next higher verbosity.


galsim/config/input_cosmos.py
```diff
@@ -21,13 +21,26 @@
...
+                if 'cosmos_catalog' in base['input']:
+                    out_str = ''
+                    if 'sample' in base['input']['cosmos_catalog'][0]:
+                        out_str += '\n  sample = %s'%base['input']['cosmos_catalog'][0]['sample']
+                    if 'dir' in base['input']['cosmos_catalog'][0]:
+                        out_str += '\n  dir = %s'%base['input']['cosmos_catalog'][0]['dir']
+                    if out_str != '':
+                        logger.warning('Using user-specified COSMOSCatalog: %s',out_str)
+            logger.warning("file %d: COSMOS catalog has %d total objects; %d passed initial cuts.",
+                           base['file_num'], cosmos_cat.getNTot(), cosmos_cat.getNObjects())
```
rmjarvis: Minor point, but if you're upgrading this to `warning`, then you should probably remove the `file %d:` prefix.  I used this prefix to help keep track of the (many) `info` and `debug` lines, but I think most of the `warning` logging statements don't use it.

OTOH, perhaps it would make more sense to leave this particular line as `info`.  Are these bits of information about the number of objects in the cosmos catalog things that you think people will want to see at the minimal (default) verbosity?  Seems like it's less important than the extra bits you added.

tests/test_wcs.py
```diff
@@ -1755,6 +1755,75 @@ def test_scamp():
...
+    assert not galsim.wcs.compatible_wcs(uv1,uv3)
+    assert not galsim.wcs.compatible_wcs(uv2,uv3)
+    assert not galsim.wcs.compatible_wcs(uv3,uv1)
+    assert not galsim.wcs.compatible_wcs(uv3,uv2)
+    assert not galsim.wcs.compatible_wcs(uv1,affine)
+    assert not galsim.wcs.compatible_wcs(uv2,affine)
+    assert not galsim.wcs.compatible_wcs(uv3,affine)
+    assert not galsim.wcs.compatible_wcs(affine,uv1)
+    assert not galsim.wcs.compatible_wcs(affine,uv2)
+    assert not galsim.wcs.compatible_wcs(affine,uv3)
```
rmjarvis: The tests you had (which I recognize came from my comment) did not include any tests of the `else` branch of `compatible_wcs`.  They were all uniform WCSs.  So I added a bit more to check some combinations using non-uniform WCSs.

galsim/wcs.py
```diff
@@ -2321,3 +2321,14 @@ def __getstate__(self):
     def __setstate__(self, d):
         self.__dict__ = d
         self._initialize_funcs()
+
+def compatible_wcs(wcs1, wcs2):
```
rmjarvis: I think `galsim.wcs.compatible_wcs` is slightly redundant with two `wcs` parts.  Perhaps if it lives in `galsim.wcs` (which seems appropriate), then it doesn't need the `_wcs` suffix.

galsim/config/input_cosmos.py
rmandelb: > OTOH, perhaps it would make more sense to leave this particular line as info.

Good point.  I just made that change.

tests/test_wcs.py
rmandelb: Thank you!

galsim/wcs.py
rmandelb: I agree.  Done.

rmandelb:
@rmjarvis , do you understand why the coverage check is still pending many hours after the last change I pushed?

rmjarvis:
Apparently, sometimes Coveralls just takes forever.  cf. https://github.com/nickmerwin/node-coveralls/issues/95 for instance.

rmandelb:
Wow.

rmjarvis:
Despite the coveralls issue, I'm happy to go ahead and merge this.

jmeyers314:
Looks good to me too.

rmandelb:
Thanks, both!  I'll go ahead then.

Issue #715, opened 2016-02-23
Unsigned int Images


rmjarvis:
On branch #654, I fixed a number of problems that we had related to reading in unsigned int Images.  We don't have any unsigned int instantiations for the Image class in C++, so it had been recognizing the incompatibility and converting to float64 (our default catchall if we can't handle the dtype of a numpy array).  

However, for badpix masks, this isn't the right thing to do, so now the code uses the corresponding int type instead.  i.e. uint16 -> int16 and uint32 -> int32.  This works ok, but it would be better if we actually supported uint Images correctly.

rmandelb:
A student who recently started working with me, Brent ( @zunyibrt ), would like to do this issue as his starter project.

rmandelb:
@rmjarvis - Brent has added this functionality, and we were hoping for some advice from you about one remaining problem.

The unit tests (old + new ones that he added) pass on our Linux cluster here.  But there is a single test failure on my Mac that looks like this:
```
FAIL: Test that all four FITS reference images are correctly read in by both PyFITS and our Image
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/rmandelb/git/GalSim/tests/galsim_test_helpers.py", line 412, in f2
    result = f(*args, **kwargs)
  File "/Users/rmandelb/git/GalSim/tests/test_image.py", line 652, in test_Image_CubeFITS_IO
    assert test_array.dtype.type == types[i], "%s != %s" % (test_array.dtype.type, types[i])
AssertionError: <type 'numpy.float32'> != <type 'numpy.uint16'>
```

We were wondering if either (a) this is something you've seen before, and you can suggest a specific place in the code where Brent should be able to fix this, or (b) you can suggest an effective way that  we can go about investigating the cause of this problem, given that it's showing up on a Mac (he doesn't have a Mac himself) and everything is fine on the Linux system where he's working?  

Thanks!

rmjarvis:
It's not Mac vs Linux.  It's the astropy version.  I didn't have this error on my Mac with astropy 1.1.1, but when I regressed back to astropy 1.0, it showed up.

It's also not just true for the Cube file.  It's true for all of the pyfits reads of unsigned int images.  It's just that apparently, that one is the only one where we check the type of the incoming pyfits data array.
```
[...]

Issue #799, opened 2016-09-13
Update mid-layer API for drawImage to make it easier to employ different sensor models


rmjarvis:
This issue is to implement the API design that I [presented at the LSST Project and Community Workshop](https://project.lsst.org/meetings/lsst2016/agenda/end-end-galsim-lsst).

Here is the proposed new structure for how `drawImage` will work (from slide 14 of my talk):

```
def drawImage(..., sky=None, surface_ops=(), sensor=None):
    ... # Various setup steps that won't need to change.
    if method == 'phot':
        photon_list = prof.SBProfile.shoot(n_photons, ...)
        if return_photon_image:
            photon_image = SimpleCCD().accumulate(photon_list, image.copy())
    else:
        prof.SBProfile.draw(image.image, ...)
        if sensor is None:
            return
        elif return_photon_image:
            photon_image = image.copy()
        ii = InterpolatedImage(image)
        photon_list = ii.SBProfile.shoot(n_photons, ...)

    photon_list.generateWavelengths(sed * bandpass)
    if sky:
        photon_list.addSkyPhotons(sky, sky_sed * bandpass)
    photon_list.shuffle()

    for surface_op in surface_ops:
        photon_list = surface_op.apply(photon_list)

    if sensor is None:
        sensor = SimpleCCD(gain, ...)
    sensor.accumulate(photon_list, image.image)
```

In this issue, the goal is to maintain current functionality, but have the above hooks in place for people to add non-trivial sensor models.

At the LSST Hack week in November, if this is all in place, we can try to merge it into branch #722 where @craiglagegit has been implementing his sensor model.  His model would be passed in as the new `sensor` parameter to `drawImage`.

rmjarvis:
Merged #818

PR #817, opened 2016-10-20
#715

zunyibrt:
Add support for reading in of unsigned int Images


src/hsm/PSFCorr.cpp
```diff
@@ -1881,6 +1881,54 @@ namespace hsm {
         float sky_var, const char* shear_est, const std::string& recompute_flux,
         double guess_sig_gal, double guess_sig_PSF, double precision,
         galsim::Position<double> guess_centroid, boost::shared_ptr<HSMParams> hsmparams);
+    template CppShapeData EstimateShearView(
+        const BaseImage<uint16_t>& gal_image, const BaseImage<int>& PSF_image,
+        const BaseImage<int>& gal_mask_image,
+        float sky_var, const char* shear_est, const std::string& recompute_flux,
+        double guess_sig_gal, double guess_sig_PSF, double precision,
+        galsim::Position<double> guess_centroid, boost::shared_ptr<HSMParams> hsmparams);
```
rmjarvis: I'm pretty sure none of these can possibly get called currently.  There is code in hsm.py to force the mask image to be an ImageI. So none of these UI, US variants can ever be called.

Honestly I think I'd prefer just forcing the gal and psf images into either float or double if they come in as integer images and having the only four instantiations be DDI, FFI, DFI, FDI, and getting rid of the III instantiation that was already here (along with all these new ones).

The other option is to make sure we have all possible combinations, which would need to be ntype^3 (ie. many more than are currently listed here), which seems unwieldy to say the least.

tests/test_hsm.py
```diff
@@ -201,15 +201,17 @@ def test_shearest_basic():
 def test_shearest_precomputed():
     """Test that we can recover shears the same as before the code was put into GalSim."""
     # loop over real galaxies
-    for index in range(len(file_indices)):
+    for index in range(2,len(file_indices)):
```
rmjarvis: Why the 2 here?

galsim/image.py
```diff
@@ -317,7 +312,7 @@ def __init__(self, *args, **kwargs):
             try:
                 ncol = int(ncol)
                 nrow = int(nrow)
-            except Exception:
+            except:
```
rmjarvis: I know this isn't actually new in this branch, but since you highlighted it by editing it...

It probably makes more sense to just get rid of this try/except block.  The custom TypeError here isn't giving any additional information beyond whatever the system would raise from `ncol = int(ncol)` failing.  So it's probably cleaner to just let the user see the original traceback.

tests/test_hsm.py
```diff
@@ -201,15 +201,17 @@ def test_shearest_basic():
...
         # define input filenames
         img_file = os.path.join(img_dir, gal_file_prefix + str(file_indices[index]) + img_suff)
         psf_file = os.path.join(img_dir, psf_file_prefix + str(file_indices[index]) + img_suff)
 
         # read in information for objects and expected results
-        img = galsim.fits.read(img_file)
+        imgR = galsim.fits.read(img_file)
+        img  = galsim.Image(imgR, dtype = int)
```
rmjarvis: Probably useful to add a quick comment here that the reason you are doing this is that the images on file are unsigned, and the `-= 1000` brings them down to a mean of 0, which for an unsigned image leads to badness.  So cast to int first...

galsim/image.py
rmjarvis: Actually, don't worry about this one.  I ended up addressing it on #799, since I was modifying other aspects of the Image constructor, so I fixed this too.

src/hsm/PSFCorr.cpp
zunyibrt: Having all possible combinations does seem too much. I followed your suggestion and made the edit to force the gal and psf images into being floats or doubles so there are simply 4 instantiations.

tests/test_hsm.py
zunyibrt: Sorry, the 2 should have been removed

rmandelb:
All looks good to me.

Issue #810, opened 2016-10-07
Can we get rid of wmult?


rmjarvis:
I'm working on #799 to reorganize how the draw commands work in order to provide the hooks for the sensor effects we want for LSST, and it occurs to me that the wmult parameter is a bit arcane, and is probably rarely (never?) used by end users.

I think it served a useful purpose when @gbernstein originally put it in, since we sometimes wanted to quickly tell GalSim to use a larger FFT size than the default.  But I think now that we have the gsparams stuff (in particular folding_threshold), that's really the more reliable way to adjust the behavior of stepk, and thus the FFT grid size.  wmult seems like a hack that we don't need anymore.

Any objections to my removing it in the reorg?

rmandelb:
Good idea.

jmeyers314:
I think it's safe to remove.

rmjarvis:
The only place anywhere in the code that used wmult other than just to check that it behaves as expected is a single test in test_interpolated.py that used wmult=5 (!) without any explanation of why it was deemed necessary.  But when I removed it the test still passes.  So presumably the various accuracy improvements we've made over the years in various places have made this parameter largely obsolete.

esheldon:
I use it in code like this

    wmult=1.0
    dim = ii.SBProfile.getGoodImageSize(
        ii.nyquistScale(),
        wmult,
    )

Is this something that would transition naturally in the new code?

-e

On 10/8/16, Mike Jarvis <notifications@github.com> wrote:
> The only place anywhere in the code that used wmult other than just to check
> that it behaves as expected is a single test in test_interpolated.py that
> used wmult=5 (!) without any explanation of why it was deemed necessary.
> But when I removed it the test still passes.  So presumably the various
> accuracy improvements we've made over the years in various places have made
> this parameter largely obsolete.
>
> --
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/GalSim-developers/GalSim/issues/810#issuecomment-252406258


-- 
Erin Scott Sheldon
Brookhaven National Laboratory erin dot sheldon at gmail dot com

rmjarvis:
Well, in your case, wmult=1, so you're not actually using it other than to conform to the requirements of the C++ call signature.  In the new version it would probably be simply:
```
dim = ii.SBProfile.getGoodImageSize(ii.nyquistScale())
```
If you wanted to use some value of wmult **other** than 1, then that would be equivalent to
```
[...]

Issue #722, opened 2016-03-12
Using results of Poisson CCD Simulator to simulate BF effect


arunkannawadi:
Craig Lage at UC Davis has developed a Poisson equation solver that computes the distortion of the pixel edges as the electrons get collected in the CCD pixels. In the DESC Hack day today, we made some progress towards integrating his code (in C++) with the C++ layer of the GalSim. We identified  that the natural way to include this functionality would be by tweaking the photon-shooting method inside the `drawImage` routine.  I will push the code after some cleanup, but this is not completely done yet.

arunkannawadi:
Pinging @craiglagegit so that discussions and updates can happen here.

rmandelb:
Oh right, I forgot you had already opened an issue.  That's convenient.  :)

rmjarvis:
@craiglagegit I made a branch over here called '#722' (which is our usual style of branch naming -- using the issue number) and cherry-picked your code over to here (50fa1b98840bcaa7a9eae149d9f85f8bc0ef38eb).  I also added you as a collaborator.  This will make it a lot easier for us to work with you to get this properly incorporated into GalSim, since we can all edit and push to this branch.

Some things that I think need to be addressed before we merge this into master:
- As you suggested, we want BF to be optional, so we need this to be a separate function, not a replacement for the normal `PhotonArray::addTo` function.
- We'll also want a way to do this when we are doing image generation using FFTs.  Maybe draw the image, then make an InterpolatedImage out of it and shoot that down the rest of the way through the silicon.  
- Probably implement this in `drawImage` to do the necessary steps (differently for `method='phot'` vs other methods).
- You have a random number generator bit in there.  We'll need to get that using the GalSim rng, not the system one, since we want repeatable simulations.
- You currently are reading a file every time into this function, which is bound to be inefficient.  We'll want to have the Silicon constructor happen once and save that object for later.  Probably best to wrap that up into the python layer to make it easier for people to make modifications without having to write new files for any change they want.  Then people could build the Silicon object in python once and pass it into `drawImage` so it wouldn't have to read the file each time.
- We should wrap your bare pointers into shared_ptrs to be a bit safer wrt memory leaks.

@arunkannawadi if you want to start tackling any of these bits, feel free.  I won't have time to for a while, but I'm happy to help with design questions.

craiglagegit:
[...]

PR #818, opened 2016-10-20
#799 Refactor the middle-layer code to make more of the operations available from Python.

rmjarvis:
Issue #799 nominally involved moving the contents of drawShoot, plainDraw, and fourierDraw up to the Python layer so we can more easily incorporate the changes we need for #722.

I did manage to do that, although I didn't implement all of the new API design that I billboarded at the LSST workshop yet.  I may do more before the hack week, but there's a lot here already, so I think it's worth putting this out as a PR so people can have a chance to look at it before then.

As usual, I got a little carried away on the refactoring, so there's a bit more here that I found useful to change in the process of moving these functions to Python.

New features:
- Implemented ImageC as a complex image type.  It can't be written to/read from FITS files, but everything else seems to be working.
- Switched drawKImage to return an ImageC rather than two real images for (re, im).
- Switched InterpolatedKImage to take an ImageC rather than two real images.
- Added the ability to directly take an FFT or inverse FFT of an image without going through the XTable and KTable classes.  This is more efficient, since it doesn't require any extra storage aside from the input and output images.
- Added a `maxSB` method that all GSObjects implement.  When possible it returns the maximum value of the surface brightness at any location in real space.  For all analytic objects and transformations, it's easy to return this exactly.  For compound objects, it can only return an estimate, which is generally an overestimate.  This is fine, since the use case for this is in calculating a good number of photons to shoot, and an overestimate will just use more photons than necessary, which is inefficient, but not inaccurate.
- Added im[x,y] syntax, which seems like we should have had from the beginning, but for whatever reason we never implemented.  im[x,y] = value and value = im[x,y] both work now and are synonyms for `im.setValue(x,y,value)` and either `value = im(x,y)` or `value = im.getValue(x,y)`.

Some deprecations:
- `wmult` cf. #810 
- `image.at(x,y)` is replaced by `image.getValue(x,y)`.  This is very rarely used, since `im(x,y)` and now `im[x,y]` are more convenient.  But the name of the function doesn't really match up with other similar methods (e.g `setValue`), so it seemed like `getValue` would be more intuitive for the rare case when someone needs a named method rather than just use the `()` or `[]` syntax.
- The old syntax with `re, im` images for `drawKImage` and `InterpolatedKImage` still work, but give deprecation warnings.

Various mostly functionality-preserving middle-layer changes:
- Changed all the `fillXValue` and `fillKValue` functions to directly write to images, rather than `tmv::MatrixView`s.
- Got rid of `fillXGrid` and `fillKGrid`, since they are no longer necessary with the ability to FFT images directly.
- Moved `goodFFTSize` to Image.h and in Python made it available as a classmethod of Image.
- Allow Images to have non-unit step.  This lets the real and imag parts of complex images be views rather than require copying.
- Changed mutating ImageView methods to not be `const`.  This was a confusing notation, which isn't actually required, so I switched it.  Although, it doesn't really make any difference.
- Made the memory allocation for Images align on 16 byte boundaries.  This lets ImageAlloc memory be appropriate for use with FFTW without requiring the use of `FFTW_Array`.  The latter is still around in `FFT.h`, but it and `XTable` and `KTable` are not used for the regular draw commands.  They are still used by SBInterpolatedImage and SBInterpolatedKImage, which I'd like to change (and use the new direct image ffts), but I'll put that off to another issue.
- Changed the way images are wrapped when the size of the FT is smaller than the full size needed according to maxk.  We used to make it Hermitian along the edges, then wrap it, and make it Hermitian again.  It turns out there is a subtle bug in doing this that only shows up at a very low level (and only popped in the tests when I added another Spergel instance with nu>0 to the list of gsobjects in test_flip).  Now we only wrap once which handles both the wrapping and the Hermitianity, and this seems to have fixed the problem.
- Changed the dbg stuff to use a Singleton pattern, which obviates the need for some of the boiler-plate code at the top of many .cpp files.
- Added a bunch of unit tests.  All the new functionality of course, but also added a function called `check_basic` which runs through a bunch of simple tests of flux, maxSB, xValue, kValue, etc. which turned out to be handy in checking my refactoring of the fillXImage and fillKImage functions.
- Fixed a minor bug in how uninitialized Images are handled in a couple places.  e.g. they used to not pickle correctly.
- You can now construct an image from a 2D list, like `im = galsim.Image([[1,2,3], [4,5,6], [7,8,9]], dtype=float)`, which saves you the step of converting to a numpy array first.
- You can also now `drawImage` with `nx,ny` or `bounds` and not provide a `scale`, which used to be disallowed.  Now it automatically sets scale to `obj.nyquistScale()`.  Likewise, `drawKImage` with a size but not a scale will use `obj.stepK()`.  This latter one is the more useful feature, it seems to me, but for symmetry, it seemed worth allowing both.


rmjarvis:
Yay!   codecov/patch — 100% of diff hit (target 89.83%)

That was fun.  :)  Even caught a few subtle bugs in the process...

esheldon:
How will one access the complex array at the C level?  I couldn't find docs on
accessing complex numpy arrays at the C level.  Are the two values packed
together, so that one can use PyArray_GETPTR2 to get the real component, then
increment a pointer to get the imaginary?

rmjarvis:
> How will one access the complex array at the C level?

In the C++ layer, it's an array of `std::complex<double>`.  That's just two doubles sequentially in memory: real then imag.  That's the same way numpy stores things for an array of `np.complex128`.  It's also the way FFTW expects things for their complex type `fftw_complex`.

[Here](https://github.com/GalSim-developers/GalSim/blob/%23799/pysrc/NumpyHelper.h#L202) is the line that casts the C++ memory into a complex numpy array.  T = `std::complex<double>`, and `getCode()` will return `NPY_COMPLEX128`.  The [line](https://github.com/GalSim-developers/GalSim/blob/%23799/pysrc/NumpyHelper.h#L122) to go back the other way from numpy data to `std::complex<double>*` is even simpler, since it's just a `reinterpret_cast`.

esheldon:
OK, then I think if one is using the C api (which I do in ngmix) then this translates to using the `PyArray_GETPTR2` macro to get a pointer to a pixel, and then interpreting that as two 64-bit floats packed together as `(re, im)`

jmeyers314:
Hey Mike,
I'm getting a few python3 related errors on this one.  I attempted to fix a few of these, but now I'm not sure what to do about the latest error, which is:

```
  File "test_image.py", line 2954, in <module>
    test_Image_inplace_divide()
  File "/Users/josh/src/GalSim/tests/galsim_test_helpers.py", line 557, in f2
    result = f(*args, **kwargs)
  File "test_image.py", line 1508, in test_Image_inplace_divide
    image1 /= image2
  File "/Users/josh/src/GalSim/galsim/image.py", line 1413, in Image_idiv
    self.array[:,:] /= a
TypeError: ufunc 'true_divide' output (typecode 'd') could not be coerced to provided output parameter (typecode 'h') according to the casting rule ''same_kind''
```

I.e., should we allow in-place division of two integer `Image`s?  I don't have particularly strong feelings either way, though maybe it's good to be consistent with numpy and just prohibit this.

rmjarvis:
Thanks for checking this.  (I should have tried it on Python 3 myself...)

> should we allow in-place division of two integer Images?

I think `image_int /= 2` is worth enabling, and it seems that numpy does allow this in place without any trouble, so that's good.  I can imagine cases where people would expect to be able to divide an integer image by a scalar in place, where the result would just round (or truncate, really) to an integer.

I can't think of a use case for `image_int1 /= image_int2`, but it feels a bit inconsistent to allow in-place division by a scalar, but not allow this.  So I guess I'm inclined to fix it so this works (which I know how to do).

jmeyers314:
>(I should have tried it on Python 3 myself...)

Can we add a python3 travis-ci?

> it seems that numpy does allow this in place without any trouble

Hmm...  Not for me (on python3.5, numpy 1.11.2):

```python
import numpy as np
a = np.array([1, 2, 3])
b = np.array([2, 3, 4])
a /= b  # TypeError: ufunc 'true_divide' output (typecode 'd') could not be coerced to provided output parameter (typecode 'l') according to the casting rule ''same_kind''
a /= 2  # Same error
a //= 2 # This works
a //= b # So does this
```

rmjarvis:
> Hmm... Not for me 

Sorry, you're right.  I was just basing this on the fact that there wasn't an error for im /= 2.  But of course the im /= im test happens first, so that's the one that triggered.  Anyway, the fix I just pushed (8c462562573b0b622d07dd0368d9d9d82dc448b6) fixes both cases if we're happy enough with allowing the im /= im construct for int Images.

rmjarvis:
> Can we add a python3 travis-ci?

I finally managed to make this happen.  There were issues with both boost (of course) and scons (which doesn't support python 3 yet), but I got it all straightened out finally.  So now we have travis builds for 2.6, 2.7, 3.4, and 3.5.

rmjarvis:
FYI, since the change diff had exceeded GitHub's limit, I moved most of the boring C++-layer changes that had no direct impact on the python functionality over to master.  The diff here is now a fair bit more reasonable now.

cwwalter:
> FYI, since the change diff had exceeded GitHub's limit

What does this mean?  I didn't even know there was such a thing...

rmjarvis:
> What does this mean? I didn't even know there was such a thing...

If the changes exceed I think 10,000 lines in total (added plus removed), then GitHub refuses to show them all.  Mostly because most browsers start to slow to a crawl with that large a page.  So the last several files just list as something like "Diff too large, not showing diff for this file".

It's happened to me a few times over the years.  Especially on big refactorings like this one.

rmandelb:
Mike - I just ran unit tests on the linux cluster here and my laptop.  Laptop is fine, but linux cluster gives me some DeprecationWarnings all over the place:
```
/home/rmandelb/software/lib/python2.7/site-packages/galsim/image.py:1421: DeprecationWarning: Implicitly casting between incompatible kinds. In a future numpy release, this will raise an error. Use casting="unsafe" if this is intentional.
  self.array[:,:] /= a
```

I have numpy 1.9.2 on there.

rmjarvis:
> linux cluster gives me some DeprecationWarnings all over the place:

I think I fixed this.

rmandelb:
Yes you did, thanks.  I hope to go over the code itself during hack week, unless you were hoping to get this merged before?

rmjarvis:
No, that's fine.  For the hack week, we'll be working on #722, where I've already merged this branch over.

galsim/base.py
```diff
@@ -1247,8 +1237,8 @@ class DeVaucouleurs(GSObject):
 
     def __init__(self, half_light_radius=None, scale_radius=None, flux=1., trunc=0.,
                  flux_untruncated=False, gsparams=None):
-        GSObject.__init__(self, _galsim.SBDeVaucouleurs(scale_radius, half_light_radius, flux,
-                                                        trunc, flux_untruncated, gsparams))
+        GSObject.__init__(self, _galsim.SBSersic(4, scale_radius, half_light_radius, flux,
```
rmjarvis: First the rationalization:

Because there isn't any implementation advantage to having a C++ SBDeVaucouleurs class.  It was already a subclass of SBSersic, which just hard-coded the n=4 and farmed out to SBSersic **all** of the implementation details.  (Unlike SBExponential, which has a number of efficiency improvements over SBSersic(n=1).)  So it seems more transparent to just use SBSersic with n=4 directly so as not to imply that there is something special about SBDeVaucouleurs that is different from SBSersic(n=4).

Now the real reason I got rid of it and came up with the above rationalization to justify it:

Because there were edge cases of the SBDeVaucouleurs class that weren't being covered, and it was easier to just get rid of it (given that it doesn't really serve any useful purpose) than to add the tests that would cover them.

galsim/chromatic.py
```diff
@@ -2223,12 +2223,13 @@ def evaluateAtWavelength(self, wave):
 class ChromaticFourierSqrtProfile(ChromaticObject):
     """A class for computing the Fourier-space square root of a ChromaticObject.
 
-    The ChromaticFourierSqrt class represents a wavelength-dependent Fourier-space square root of a profile.
+    The ChromaticFourierSqrt class represents a wavelength-dependent Fourier-space square root of
```
rmjarvis: FourierSqrt is the function that people would typically call that in turn constructs either a FourierSqrtProfile or ChromaticFourierSqrtProfile class.  So the user should typically only need to write FourierSqrt for any use case.

This is analogous to Convolve, Convolution and ChromaticConvolution.  or Add, Sum, ChromaticSum.  Just in this case, FourierSqrt is less obviously a verb.

But certainly your first point is correct.  There is no such thing as ChromaticFourierSqrt.

rmandelb:
OK, that�s all good.

galsim/fits.py
```diff
@@ -537,6 +540,10 @@ def write(image, file_name=None, dir=None, hdu_list=None, clobber=True, compress
     """
     from galsim._pyfits import pyfits
 
+    if image.iscomplex:
```
rmjarvis: I propose making a separate issue for this.  Mostly since I'm not sure what the use case is for writing a complex image to a fits file, so I'm not sure what the most appropriate way to write it would be from a user-friendliness standpoint.  For instance another option would be to write the real and imaginary components to separate HDUs, which would let them be easily viewable by ds9.  But that complicates some of the hdu bookkeeping we currently have.

So I think it would be worth making an issue and seeing if we get some ideas about how people would like it to work before choosing an implementation.

galsim/gsobject.py
```diff
@@ -1351,57 +1322,410 @@ def drawImage(self, image=None, nx=None, ny=None, bounds=None, scale=None, wcs=N
...
                 raise
-        else:
-            added_photons = prof.SBProfile.draw(imview.image, gain, wmult)
-            image.added_flux = added_photons / area / exptime
 
-        return image
+            phot_array.scaleFlux(g * thisN / Ntot);
+
+            added_flux += phot_array.addTo(image.image)
+            Nleft -= thisN;
```
rmjarvis: More that I failed to remove the semi-colons.  This was a big copy past from SBProfile.cpp.  Thanks for catching it.

galsim/chromatic.py
rmandelb: Right, got it.

galsim/gsobject.py
```diff
@@ -1412,84 +1736,84 @@ def drawKImage(self, re=None, im=None, nx=None, ny=None, bounds=None, scale=None
...
-                            defined bounds. [default: False]
+                            Note: This requires that `image` be provided and that it has defined
+                            bounds. [default: False]
 
-        @returns the tuple of Image instances, `(re, im)` (created if necessary)
+        @returns an ImageC instance (created if necessary)
         """
-        # Check for obsolete dk parameter
+        if isinstance(image,galsim.Image) and isinstance(nx,galsim.Image):
+            # If the user calls drawK(re,im), then give the proper deprecation below.
```
rmjarvis: This will give a warning, just not in this code block.  It sets re, im, which a few lines down will call depr, which raises a warning.

galsim/fits.py
rmandelb: The use case I had in mind was debugging, mostly, for cases where I sometimes want to see what something looks like in Fourier space.  But anyway I'm fine with having this as a separate issue.

galsim/base.py
rmandelb: Weird, my first (emailed) reply to this showed up on GitHub as a reply to the whole issue rather than associated with this comment.  So I deleted it and am trying again.  All this was to say... it's all good.

galsim/image.py
```diff
@@ -475,21 +574,232 @@ def setSubImage(self, bounds, rhs):
...
+        The input image must have a PixelScale wcs.  The output image will be complex (an ImageC
+        instance) and its scale will be 2pi / (N dx), where dx is the scale of the input image.
+
+        @returns an ImageC instance with the k-space image.
         """
-        self.setSubImage(bounds,rhs)
+        if self.wcs is None:
+            raise ValueError("calculate_fft requires that the scale be set.")
+        if not self.wcs.isPixelScale():
+            raise ValueError("calculate_fft requires that the image has a PixelScale wcs.")
```
rmjarvis: No, because later we'll call `dk = self.scale`, which would raise an exception if it's not either PixelScale or OffsetWCS.  We'd need special code to handle a JacobianWCS with a jacobian that happens to be a multiple of the identity matrix.  (Plus, `wcs.compaible` wouldn't work to determine that, since what would you compare it to?)

galsim/interpolatedimage.py
```diff
@@ -607,13 +607,13 @@ class InterpolatedKImage(GSObject):
     essentially equivalent images when drawn with the `.drawImage()` method:
 
     >>> a = returns_a_GSObject()
-    >>> b = galsim.InterpolatedKImage(*a.drawKImage())
+    >>> b = galsim.InterpolatedKImage(a.drawKImage())
 
-    The real- and imaginary-part Images must have the same data type, same bounds, and same scale.
+    The input `kimage` must have dtype=numpy.complex128, which is also known as an ImageC object.
     The only wcs permitted is a simple PixelScale.  Furthermore, the complex-valued Fourier profile
```
rmjarvis: Yes, shifted is actually fine.  We just need it to have a `scale` attribute.

galsim/fits.py
rmjarvis: FWIW, I think for that use case, you will typically have real-valued k-space images, so you can just do `kimage.real.write('kim.fits')`.

galsim/gsobject.py
rmjarvis: Turns out there were a whole bunch more than just these two.

galsim/image.py
rmandelb: I would compare it to ````self.wcs.local()```.  But anyway, it's a moot point, because 
I was thinking of this line as doing something like ```isinstance(self.wcs, galsim.PixelScale())``` so I thought even an ```OffsetWCS``` wouldn't get through, and I wanted an ```OffsetWCS``` to get through.  Now that I read your comment I realize I was oversimplifying and of course ```isPixelScale()``` will let an ```OffsetWCS``` through, so it's all good.

rmjarvis:

galsim/interpolatedimage.py
```diff
@@ -657,58 +666,83 @@ class InterpolatedKImage(GSObject):
...
-        # Make sure any `wcs`s are `PixelScale`s.
-        if ((real_kimage.wcs is not None
-             and not real_kimage.wcs.isPixelScale())
-            or (imag_kimage.wcs is not None
-                and not imag_kimage.wcs.isPixelScale())):
-            raise ValueError("Real and Imag kimage wcs's must be PixelScale's or None.")
-
-        # Check for Hermitian symmetry properties of real_kimage and imag_kimage
-        shape = real_kimage.array.shape
+        # Make wcs is a PixelScale.
```
jmeyers314: Make sure

include/galsim/SBDeconvolve.h
```diff
@@ -48,6 +48,9 @@ namespace galsim {
         /// @brief Destructor.
         ~SBDeconvolve();
 
+        /// @brief Get the SBProfile being convolved
```
jmeyers314: deconvolved

galsim/compound.py
```diff
@@ -58,7 +58,7 @@ def Add(*args, **kwargs):
     @returns a Sum or ChromaticSum instance as appropriate.
     """
     if len(args) == 0:
-        raise ValueError("At least one ChromaticObject or GSObject must be provided.")
+        raise TypeError("At least one ChromaticObject or GSObject must be provided.")
```
rmjarvis: I think this corresponds to what Python normally does for the wrong number of arguments.  I don't particularly like this convention, since it doesn't seem to be related to types, but there isn't really a better choice AFAICT.

IMO, there should also be a standard exception called ArgumentError or ParameterError.  Something like that.  But these don't exist, and every time one of these is what I think I should use, python convention seems to be to use TypeError instead.

galsim/gsobject.py
```diff
@@ -1351,57 +1322,410 @@ def drawImage(self, image=None, nx=None, ny=None, bounds=None, scale=None, wcs=N
...
+        if n_photons < 0.:
+            raise ValueError("Invalid n_photons < 0.")
+
+        if poisson_flux is None:
+            if n_photons == 0.: poisson_flux = True
+            else: poisson_flux = False
+
+        # Check that either n_photons is set to something or flux is set to something
+        if (n_photons == 0. and self.getFlux() == 1.
+            and area == 1. and exptime == 1.): # pragma: no cover
```
rmjarvis: Fixed.  And added test that would have caught this.  (And removed no cover)  9296ac1

galsim/compound.py
jmeyers314: Okay.  Good to know.

PR #819, opened 2016-10-26
add a new profile RandomWalk

esheldon:
This class generates a set of point sources.  The location of these
is based on a random walk with specified number of steps.

the user can choose the number of point sources, number of steps,
the expected half-light radius of the distribution, and the total
flux in the points.

The points are represented as small galsim.Gaussian objects

### DIFF
--- a/examples/demo4.py
+++ b/examples/demo4.py
@@ -102,7 +102,12 @@ def main(argv):
     disk_frac        = 0.6
     knot_frac        = 0.42
     smooth_disk_frac = 0.18
-    n_knots = 100
+
+    # number of knots of star formation.  To simulate a nice irregular (all the
+    # flux is in knots) we find ~100 is a minimum number needed, but we will
+    # just use 10 here to make the demo run fast.
+
+    n_knots = 10
 
     xsize = 64                      # pixels
     ysize = 64                      # pixels
--- a/examples/demo4.py
+++ b/examples/demo4.py
@@ -104,8 +104,8 @@ def main(argv):
     smooth_disk_frac = (1.0-knot_frac)*disk_frac
     n_knots = 100
 
-    xsize = 48                      # pixels
-    ysize = 48                      # pixels
+    xsize = 64                      # pixels
+    ysize = 64                      # pixels
 
     logger.info('Starting demo script 4 using:')
     logger.info('    - parameters taken from catalog %r',cat_file_name)
--- a/examples/demo4.py
+++ b/examples/demo4.py
@@ -157,6 +157,10 @@ def main(argv):
         disk_hlr = cat.getFloat(k,5)
         disk_e1 = cat.getFloat(k,6)
         disk_e2 = cat.getFloat(k,7)
+        bulge_hlr = cat.getFloat(k,8)
+        bulge_e1 = cat.getFloat(k,9)
+        bulge_e2 = cat.getFloat(k,10)
+
         smooth_disk = galsim.Exponential(flux=smooth_disk_frac, half_light_radius=disk_hlr)
 
         knots = galsim.RandomWalk(n_knots, disk_hlr, flux=knot_frac)
@@ -165,9 +169,8 @@ def main(argv):
         disk = disk.shear(e1=disk_e1, e2=disk_e2)
 
         # the rest of the light goes into the bulge
-        bulge_hlr = cat.getFloat(k,8)
         bulge = galsim.DeVaucouleurs(flux=bulge_frac, half_light_radius=bulge_hlr)
-        bulge = bulge.shear(e1=cat.getFloat(k,9), e2=cat.getFloat(k,10))
+        bulge = bulge.shear(e1=bulge_e1, e2=bulge_e2)
 
         # The flux of an Add object is the sum of the component fluxes.
         # Note that in demo3.py, a similar addition was performed by the binary operator "+".
--- a/examples/demo4.py
+++ b/examples/demo4.py
@@ -25,7 +25,8 @@ This script is our first one to create multiple images.  Typically, you would wa
 to have at least some of its attributes vary when you are drawing multiple images (although 
 not necessarily -- you might just want different noise realization of the same profile).  
 The easiest way to do this is to read in the properties from a catalog, which is what we
-do in this script.  The PSF is a truncated Moffat profile, and the galaxy is bulge plus disk.
+do in this script.  The PSF is a truncated Moffat profile, and the galaxy is bulge plus disk,
+with knots of star formation modeled as point sources distributed according to a random walk.
 Both components get many of their parameters from an input catalog.  We also shift the 
 profile by a fraction of a pixel in each direction so the effect of pixelization varies
 among the images.  Each galaxy has the same applied shear.  The noise is simple Poisson noise.
@@ -36,6 +37,7 @@ New features introduced in this demo:
 - cat = galsim.Catalog(file_name, dir)
 - obj = galsim.Moffat(beta, fwhm, trunc)
 - obj = galsim.DeVaucouleurs(flux, half_light_radius)
+- obj = galsim.RandomWalk(npoints, half_light_radius, flux=)
 - obj = galsim.Add([list of objects])
 - obj = obj.shift(dx,dy)
 - galsim.fits.writeMulti([list of images], file_name)
@@ -57,6 +59,8 @@ def main(argv):
       - Only galaxies.  No stars.
       - PSF is Moffat
       - Each galaxy is bulge plus disk: deVaucouleurs + Exponential.
+      - A fraction of the disk flux is placed into point sources, which can model
+        knots of star formation.
       - The catalog's columns are:
          0 PSF beta (Moffat exponent)
          1 PSF FWHM
@@ -89,19 +93,32 @@ def main(argv):
     gal_flux = 1.e6                 # arbitrary choice, makes nice (not too) noisy images
     gal_g1 = -0.009                 #
     gal_g2 = 0.011                  #
-    xsize = 64                      # pixels
-    ysize = 64                      # pixels
+
+    # the fraction of flux in each component
+    # 40% is in the bulge, 60% in a disk.  70% of that disk light is placed
+    # into point sources distributed as a random walk
+
+    bulge_frac = 0.4
+    disk_frac = 1 - bulge_frac
+    knot_frac = 0.7*disk_frac
+    smooth_disk_frac = (1.0-knot_frac)*disk_frac
+    n_knots = 100
+
+    xsize = 48                      # pixels
+    ysize = 48                      # pixels
 
     logger.info('Starting demo script 4 using:')
     logger.info('    - parameters taken from catalog %r',cat_file_name)
     logger.info('    - Moffat PSF (parameters from catalog)')
     logger.info('    - pixel scale = %.2f',pixel_scale)
     logger.info('    - Bulge + Disc galaxies (parameters from catalog)')
+    logger.info('    - 100 Point sources, distributed as random walk')
     logger.info('    - Applied gravitational shear = (%.3f,%.3f)',gal_g1,gal_g2)
     logger.info('    - Poisson noise (sky level = %.1e).', sky_level)
 
     # Read in the input catalog
     cat = galsim.Catalog(cat_file_name)
+
 
     # save a list of the galaxy images in the "images" list variable:
     images = []
@@ -133,16 +150,28 @@ def main(argv):
         # Take the (e1, e2) shape parameters from the catalog as well.
         psf = psf.shear(e1=cat.getFloat(k,2), e2=cat.getFloat(k,3))
 
-        # Galaxy is a bulge + disk with parameters taken from the catalog:
-        disk = galsim.Exponential(flux=0.6, half_light_radius=cat.getFloat(k,5))
-        disk = disk.shear(e1=cat.getFloat(k,6), e2=cat.getFloat(k,7))
+        # Galaxy is a bulge + disk(+knots) with parameters taken from the catalog:
 
-        bulge = galsim.DeVaucouleurs(flux=0.4, half_light_radius=cat.getFloat(k,8))
+        # put some fraction of the disk light into knots of star formation
+
+        disk_hlr = cat.getFloat(k,5)
+        disk_e1 = cat.getFloat(k,6)
+        disk_e2 = cat.getFloat(k,7)
+        disk = galsim.Exponential(flux=smooth_disk_frac, half_light_radius=disk_hlr)
+        disk = disk.shear(e1=disk_e1, e2=disk_e2)
+
+        knots = galsim.RandomWalk(n_knots, disk_hlr, flux=knot_frac)
+        knots = knots.shear(e1=disk_e1, e2=disk_e2)
+
+        # the rest of the light goes into the bulge
+        bulge_hlr = cat.getFloat(k,8)
+        bulge = galsim.DeVaucouleurs(flux=0.4, half_light_radius=bulge_hlr)
         bulge = bulge.shear(e1=cat.getFloat(k,9), e2=cat.getFloat(k,10))
 
         # The flux of an Add object is the sum of the component fluxes.
         # Note that in demo3.py, a similar addition was performed by the binary operator "+".
-        gal = galsim.Add([disk, bulge])
+        gal = galsim.Add([disk, knots, bulge])
+
         # This flux may be overridden by withFlux.  The relative fluxes of the components
         # remains the same, but the total flux is set to gal_flux.
         gal = gal.withFlux(gal_flux)
--- a/examples/demo4.py
+++ b/examples/demo4.py
@@ -157,11 +157,12 @@ def main(argv):
         disk_hlr = cat.getFloat(k,5)
         disk_e1 = cat.getFloat(k,6)
         disk_e2 = cat.getFloat(k,7)
-        disk = galsim.Exponential(flux=smooth_disk_frac, half_light_radius=disk_hlr)
-        disk = disk.shear(e1=disk_e1, e2=disk_e2)
+        smooth_disk = galsim.Exponential(flux=smooth_disk_frac, half_light_radius=disk_hlr)
 
         knots = galsim.RandomWalk(n_knots, disk_hlr, flux=knot_frac)
-        knots = knots.shear(e1=disk_e1, e2=disk_e2)
+
+        disk = galsim.Add([smooth_disk, knots])
+        disk = disk.shear(e1=disk_e1, e2=disk_e2)
 
         # the rest of the light goes into the bulge
         bulge_hlr = cat.getFloat(k,8)
@@ -170,7 +171,7 @@ def main(argv):
 
         # The flux of an Add object is the sum of the component fluxes.
         # Note that in demo3.py, a similar addition was performed by the binary operator "+".
-        gal = galsim.Add([disk, knots, bulge])
+        gal = galsim.Add([disk, bulge])
 
         # This flux may be overridden by withFlux.  The relative fluxes of the components
         # remains the same, but the total flux is set to gal_flux.
--- a/examples/demo4.py
+++ b/examples/demo4.py
@@ -166,7 +166,7 @@ def main(argv):
 
         # the rest of the light goes into the bulge
         bulge_hlr = cat.getFloat(k,8)
-        bulge = galsim.DeVaucouleurs(flux=0.4, half_light_radius=bulge_hlr)
+        bulge = galsim.DeVaucouleurs(flux=bulge_frac, half_light_radius=bulge_hlr)
         bulge = bulge.shear(e1=cat.getFloat(k,9), e2=cat.getFloat(k,10))
 
         # The flux of an Add object is the sum of the component fluxes.
--- a/examples/demo4.py
+++ b/examples/demo4.py
@@ -163,7 +163,7 @@ def main(argv):
 
         smooth_disk = galsim.Exponential(flux=smooth_disk_frac, half_light_radius=disk_hlr)
 
-        knots = galsim.RandomWalk(n_knots, disk_hlr, flux=knot_frac)
+        knots = galsim.RandomWalk(n_knots, disk_hlr, flux=knot_frac, rng=rng)
 
         disk = galsim.Add([smooth_disk, knots])
         disk = disk.shear(e1=disk_e1, e2=disk_e2)
--- a/examples/demo4.py
+++ b/examples/demo4.py
@@ -98,10 +98,10 @@ def main(argv):
     # 40% is in the bulge, 60% in a disk.  70% of that disk light is placed
     # into point sources distributed as a random walk
 
-    bulge_frac = 0.4
-    disk_frac = 1 - bulge_frac
-    knot_frac = 0.7*disk_frac
-    smooth_disk_frac = (1.0-knot_frac)*disk_frac
+    bulge_frac       = 0.4
+    disk_frac        = 0.6
+    knot_frac        = 0.42
+    smooth_disk_frac = 0.18
     n_knots = 100
 
     xsize = 64                      # pixels
--- a/examples/demo4.yaml
+++ b/examples/demo4.yaml
@@ -21,20 +21,23 @@
 # The fourth YAML configuration file in our tutorial about using Galsim config files
 # (This file is designed to be viewed in a window 100 characters wide.)
 #
-# This script is our first one to create multiple images.  Typically, you would want each
-# object to have at least some of its attributes vary when you are drawing multiple images 
-# (although not necessarily -- you might just want different noise realization of the same 
-# profile).  The easiest way to do this is to read in the properties from a catalog, which is 
-# what we do in this script.  The PSF is a truncated Moffat profile, and the galaxy is bulge
-# plus disk.  Both components get many of their parameters from an input catalog.  We also
-# shift the profile by a fraction of a pixel in each direction so the effect of pixelization 
-# varies among the images.  Each galaxy has the same applied shear.  The noise is simple 
-# Poisson noise.  We write the images out into a multi-extension fits file.
+# This script is our first one to create multiple images.  Typically, you would
+# want each object to have at least some of its attributes vary when you are
+# drawing multiple images (although not necessarily -- you might just want
+# different noise realization of the same profile).  The easiest way to do this
+# is to read in the properties from a catalog, which is what we do in this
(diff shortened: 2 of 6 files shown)


galsim/compound.py
```diff
@@ -902,3 +904,162 @@ def __setstate__(self, d):
...
+    
+    Initialization
+    --------------
+    @param  npoints  Number of point sources to generate.
+    @param  hlr      Half light radius of the distribution of points.  This
+                     is the mean half light radius produced by an 
+                     infinite number of points.  A single instance will be
+                     noisy.
+    @param  flux     Total flux in all point sources.  Default 1
+    @param  nstep    Number of steps in random walk.  Default 40
```
rmjarvis: This doesn't really explain what nstep does.  Suggest something like "Number of random walk steps to take between each of the npoints."  Also, I don't have a feel for what are the pros and cons of increasing or decreasing this value, so perhaps a brief discussion in the intro about this would be warranted.

galsim/compound.py
```diff
@@ -902,3 +904,162 @@ def __setstate__(self, d):
...
+    @param  rng      Optional random number generator. Should be a UniformDeviate.
```
rmjarvis: Typically, we let the `rng` parameters to be any BaseDeviate.  Then this function can convert to a UniformDeviate with `ud = galsim.UniformDeviate(rng)`.

galsim/compound.py
```diff
@@ -902,3 +904,162 @@ def __setstate__(self, d):
...
+        .gaussians
+            The list of galsim.Gaussian objects representing the points
+
+    Notes
+    -----
+
+    - The algorithm is a modified version of that presented in
+      https://arxiv.org/abs/1312.5514v3 : There is no outer cutoff to how far a
+      point can wander.
+    - This class inherits from galsim.Sum
```
rmjarvis: This note is redundant with the above statement in Methods.  Plus, I think the user shouldn't probably care about this implementation detail other than how it relates to which methods are available, so the above statement is probably sufficient.

galsim/compound.py
```diff
@@ -902,3 +904,162 @@ def __setstate__(self, d):
...
+    adding this profile to an Exponential to represent knots of star formation.
+
+    Random walk profiles have a "shape noise" of g~0.05, but can be sheared to
+    give an additional ellipticity, for example to follow that of an associated
+    disk.
+    
+    Initialization
+    --------------
+    @param  npoints  Number of point sources to generate.
+    @param  hlr      Half light radius of the distribution of points.  This
```
rmjarvis: In other classes, we spell this out as half_light_radius, so probably should keep to that convention here.

galsim/compound.py
```diff
@@ -902,3 +904,162 @@ def __setstate__(self, d):
...
+            The list of galsim.Gaussian objects representing the points
+
+    Notes
+    -----
+
+    - The algorithm is a modified version of that presented in
+      https://arxiv.org/abs/1312.5514v3 : There is no outer cutoff to how far a
+      point can wander.
+    - This class inherits from galsim.Sum
+    """
```
rmjarvis: You should add in the lines about `req_params`, etc. that let the config layer know what types each of the parameters are, which ones are required, etc.:
```
    _req_params = { "npoints" : int, "half_light_radius" : float }
    _opt_params = { "flux" : float, "nstep" : int }
    _single_params = []
    _takes_rng = True
```
This bit is all that is required to make this class usable in config files.

galsim/compound.py
```diff
@@ -902,3 +904,162 @@ def __setstate__(self, d):
...
+
+        scale=self._scale
+        npoints=self._npoints
+        nstep=self._nstep
+
+        pts=numpy.zeros( (npoints, 2) )
+
+        rng=self._rng
+
+        for i in xrange(npoints):
```
rmjarvis: For Python 3 compatibility, this should be simply `range`.  If you think the list construction overhead is significant, you can check [here](http://python-future.org/compatible_idioms.html) for the python 2/3 compatible construction, but I suspect it's fine to just use `range`.

Update: the 2/3 compatible thing is to just put `from builtins import range` at the top, so that's pretty easy.

tests/test_randwalk.py
```diff
@@ -0,0 +1,165 @@
...
+    Create two objects, each with three exposures. Save them to a MEDS file.
+    Load the MEDS file. Compare the created objects with the one read by MEDS.
+    """
+
+    # try constructing with mostly defaults
+    npoints=100
+    hlr = 8.0
+    rw=galsim.RandomWalk(npoints, hlr)
+
+    assert rw.npoints==npoints,"expected npoints==%d" % (npoints, rw.npoints)
```
rmjarvis: Your string only has one %, but you have two variables, so this will give a `TypeError: not all arguments converted during string formatting` rather than an `AssertionError` if it got triggered.  There are some more like this below.

tests/test_randwalk.py
```diff
@@ -0,0 +1,165 @@
...
+
+    # try with rng wrong type
+
+    npoints=100
+    hlr = 8.0
+    rng=37
+
+    args=(npoints, hlr)
+    kwargs={'rng':rng}
+    np.testing.assert_raises(TypeError, galsim.RandomWalk, *args, **kwargs)
```
rmjarvis: Technically, `assert_raises` depends on nose, which we don't claim as a dependency of GalSim, so we typically guard all our `assert_raises` tests within a 
```
try:
    ... tests using assert_raises ...
except ImportError:
    pass
```
block to allow the tests to run if people don't have nose installed.

tests/test_randwalk.py
```diff
@@ -0,0 +1,165 @@
...
+    g=rw.gaussians
+    ngauss=len(g)
+    assert ngauss == npoints==npoints,"expected %d gaussians, got %d" % (npoints, ngauss)
+
+@timer
+def test_randwalk_valid_inputs():
+    """
+    Create two objects, each with three exposures. Save them to a MEDS file.
+    Load the MEDS file. Compare the created objects with the one read by MEDS.
+    """
```
rmjarvis: This comment doesn't seem to apply to the below tests.  (Likewise for the other test functions.)

galsim/compound.py
```diff
@@ -902,3 +904,162 @@ def __setstate__(self, d):
...
+    @param  gsparams GSParams for the gaussians representing each point source.
+
+    Methods
+    -------
+
+    This class inherits from galsim.Sum. Additional methods, implemented as
+    read-only properties, are provided as "getters" for the basic parameters:
+
+        .npoints
+        .half_light_radius
```
rmjarvis: I wonder if this should have a slightly different name than just `half_light_radius`.  Other profiles that have this return an exact value, whereas this profile will have a half-light radius different from this due to noise in the random walk.  

Maybe `input_half_light_radius`?  It's a bit verbose, but I'm concerned that people will expect `rw.half_light_radius` to give them a true value for this particular profile, rather than just an input parameter that doesn't quite match the realized value.

galsim/compound.py
```diff
@@ -902,3 +904,162 @@ def __setstate__(self, d):
...
+        if rng is None:
+            rng = galsim.UniformDeviate()
+
+        self._rng=rng
+
+        self._verify()
+
+        # this is the scale factor by which to multiply each step
+        # in order to get the requested half light radius
+        factor = numpy.sqrt(nstep)/2.09
```
rmjarvis: Perhaps a bit more explanation of where this magic number 2.09 comes from?  Is this some integral result?  Or the root of some equation?

galsim/compound.py
esheldon: OK, didn't know about these, will add

galsim/compound.py
esheldon: OK, I typically do this at the top of my files for compatibility
```python
try:
   xrange=xrange
except:
    xrange=range
```
I can just  use `range`

tests/test_randwalk.py
esheldon: whoops; I based my test on the `test_des.py` and forgot to change the comments

esheldon:
Regarding the half light radius:  this is the half light radius of the points, not the final image, which must be convolved with a psf in order to be rendered, due to the tiny `Gaussian` objects being used.

I will add a test on the points themselves

galsim/compound.py
rmjarvis: I don't think if necessarily needs to be exact, but the current doc string says that the input half-light radius "is the mean half light radius produced by an infinite number of points."  So if that's not exactly true, there should be some discussion about how exact it really is.   Like, "In the limit as nstep >> 1, the realized half-light radius will typically be within about 1% of the requested half-light radius."  I don't know if that's true, but we should give some sort of guidance to the user about how close they can expect it to be.

esheldon: That statement is true actually.

But I think I can come up with a quick fudge to get within 1% for nstep > 3

esheldon: I don't think small nstep should be used by the way, and I added a doc saying that I think 40 is a good compromise

esheldon:
I think I've addressed all the requested changes, please have a look.

esheldon:
The current (2016-10-27 14:50 ET) travis failure seems to be in finding tmv, and unrelated to any of the code in this pull request.

rmjarvis:
I noticed that the hlr test is really slow.  It takes 11 seconds currently, which is a lot for a single test.  I have a number of optimization suggestions that can speed it up -- I managed to get it down to 1.7 seconds so far.

But it got me thinking about the part where you take nstep random walk steps.  You say you really want to be in the limit of nstep >> 1, and 40 is a compromise for speed.  But I'm pretty sure we can go all the way to nstep = infinity analytically.  Then the position of the next point relative to the current one is distributed as a Gaussian.  That's basically going to the limit of this being a diffusion problem.

So I think rather than doing nstep random walk steps of size `scale`, we can take a single step where dx and dy are each Gaussian distributed with some `sigma`. (I believe the equivalent sigma is `sigma = sqrt(nstep/2) * scale` using your current calculation.)

This would significantly speed it up, since it would save a bunch of trig calls (now needing none) and rng() calls (just two per point).  And presumably, the results would be better, since we'd be firmly in the limit of nstep >> 1, which is apparently preferable.

rmjarvis:
> The current (2016-10-27 14:50 ET) travis failure seems to be in finding tmv, and unrelated to any of the code in this pull request.

I restarted the build, and it got past that point now.  It looked like Travis had some kind of issues with their Ruby scripts for handling the cache, since none of the cache was being loaded.  Now it is, so I guess they fixed it.

galsim/compound.py
```diff
@@ -902,3 +905,224 @@ def __setstate__(self, d):
...
+    @property
+    def gaussians(self):
+        return self._gaussians
+
+    @property
+    def points(self):
+        return self._points.copy()
+
+    def _get_gaussians(self, points):
+
```
rmjarvis: This function is mostly slow because of the GalSim front end for Gaussian and shift to check for correct inputs, so going directly to the C++ functions is faster.  Plus I used list comprehension to help a bit:
```
return [ galsim._galsim.SBTransform(
                galsim._galsim.SBGaussian(
                    sigma=self._gaussian_sigma,
                    flux=fluxper,
                    gsparams=self._input_gsparams),
                1.0, 0.0, 0.0, 1.0, galsim.PositionD(p[0],p[1]), 1.0, self._input_gsparams)
         for p in points ]
```
Then this is not a list of GSObjects, but rather a list of SBProfiles, so back in `__init__` rather than use `super(RandomWalk, self).__init__(self._gaussians)`, you would need to go directly to an SBAdd object:
```
galsim.GSObject.__init__(self, galsim._galsim.SBAdd(self._gaussians, gsparams))
```

galsim/compound.py
```diff
@@ -902,3 +905,224 @@ def __setstate__(self, d):
...
+            gaussians.append(g)
+
+        return gaussians
+
+    def _get_points(self):
+
+        scale=self._scale
+        npoints=self._npoints
+        nstep=self._nstep
+
```
rmjarvis: There are a number of speed-ups possible in this function, even keeping the current functionality of doing nstep random walk steps, rather than switching to the nstep->inf limit.

First, we can pre-generate a bunch of random values by applying `DeviateNoise` to an image, rather than call `rng()` repeatedly, which suffers from some Python wrapping overhead.
```
rand_values = np.zeros((npoints * nstep,2))
galsim.DeviateNoise(rng).applyTo(galsim.Image(rand_values))
```
Then we can use numpy vectorization to do all the trig and multiplications
```
r_values = scale * rand_values[:,0]
angle_values = 2*np.pi*rand_values[:,1]

dx = r_values * np.cos(angle_values)
dy = r_values * np.sin(angle_values)
```
And finally, it's a bit more efficient to let numpy do the partial sums for you
```
return np.array([ (np.sum(dx[i*nstep : (i+1)*nstep]),
                   np.sum(dy[i*nstep : (i+1)*nstep])) for i in range(npoints) ])
```

esheldon:
First off I want to make it clear that these optimizations are only necessary for the unit tests of the hlr accuracy.  In a real scenario, profiling shows that `drawImage`  dominates over any calculations in `RandomWalk`, which is the reason I didn't bother optimizing it at all.

esheldon:
I think the gaussian approximation is good enough

rmjarvis:
>  In a real scenario, profiling shows that drawImage dominates over any calculations in RandomWalk

Maybe with complicated PSFs where the PSF is dominating the time for `drawImage`, but for just the RandomWalk part of the calculation, the construction time is more than the drawing time, even for the default `nstep=40`.  

Here is a quick script the draws 100 x 100 galaxies consisting of a RandomWalk with `npoints=100` convolved with a simple Gaussian:
```
import galsim
import time

nx = 100
ny = 100

rng = galsim.UniformDeviate(1234)
psf = galsim.Gaussian(sigma=0.5)

stamp_size = 64
image = galsim.Image(100*stamp_size, 100*stamp_size)

t1 = time.time()
rw_list = [ galsim.RandomWalk(100, 2.5, flux=23, rng=rng) for k in range(nx*ny) ]
t2 = time.time()

gal_list = [ galsim.Convolve(rw, psf) for rw in rw_list ]
t3 = time.time()

k = 0
for i in range(nx):
    for j in range(ny):
        b = galsim.BoundsI(i*stamp_size+1, (i+1)*stamp_size, j*stamp_size+1, (j+1)*stamp_size)
        gal_list[k].drawImage(image[b])
        k = k+1
t4 = time.time()

image.write('rw.fits')

print 'Time for making 10,000 RandomWalk objects: ',t2-t1
print 'Time for convolving 10,000 RandomWalk objects: ',t3-t2
print 'Time for drawing 10,000 RandomWalk objects: ',t4-t3
```
The output on my laptop is:
```
Time for making 10,000 RandomWalk objects:  134.190493107
Time for convolving 10,000 RandomWalk objects:  0.139138936996
Time for drawing 10,000 RandomWalk objects:  87.2488839626
```
So I think it's worth making at least some simple optimizations to avoid the python-layer overhead that is causing most of the slow down.

rmjarvis:
BTW, for those who haven't seen this yet, here is a screen shot from the image that the above script produces.  These RandomWalk galaxies really do look a lot like irregular galaxies!
rw

esheldon:
Yeah, I'm already implementing the optimizations. Sorry I should have said that.

esheldon:
More eye candy
test

esheldon:
Now that we are using an analytic distribution, we may want to take this to its logical conclusion and allow drawing from an arbitrary distribution. E.g. drawing from an exponential.

rmjarvis:
> Now that we are using an analytic distribution, we may want to take this to its logical conclusion and allow drawing from other distributions such as Exponential
I think this is fine for now.  Distributions other than Gaussian don't have quite as simple a formulation, since they aren't separable, but we do already have much of the machinery available in GalSim via DistDeviate, which can take an arbitrary target distribution for P(r).  So if you feel like tackling it, feel free.

galsim/compound.py
```diff
@@ -902,3 +905,252 @@ def __setstate__(self, d):
...
+        rep='galsim.RandomWalk(%(npoints)d, %(hlr).16g, flux=%(flux).16g)'
+        rep = rep % dict(
+            npoints=self._npoints,
+            hlr=self._half_light_radius,
+            flux=self._flux,
+        )
+        return rep
+
+    def __repr__(self):
+        return self.__str__()
```
rmjarvis: The repr (but not the str) should include `gsparams` to be fully descriptive.

Also, the str probably doesn't need 16 digits of precision.  We usually use %s for everything in the str and %r for things in the repr.  For POD, that guarantees that the repr gets all the relevant digits of accuracy, and the str has some reasonable, but not excessive, number of digits.

esheldon: The reason I use high precision is that I like the repr to be usable with `eval` to recover the same object.   So I printed to approximately the precision of a double.

rmjarvis: We do too.  In fact, we require it.  (If you add `check_pickle(rw)` to your tests, it will check both picklability and `eval(repr(rw))`.)  But we don't require it of the str version, which we allow to be slightly easier on human eyes.

rmjarvis:
I'm happy with this Erin.  Thanks again.  Anyone else want to take a look?

rmandelb:
I will look this weekend.

galsim/compound.py
esheldon: OK, I'll make str use less

galsim/compound.py
```diff
@@ -902,3 +905,263 @@ def __setstate__(self, d):
...
+    Random walk profiles have "shape noise" that depends on the number of point
+    sources used.  For example, with 100 points the shape noise is g~0.05, and
+    this will decrease as more points are added.  The profile can be sheared to
+    give additional ellipticity, for example to follow that of an associated
+    disk.
+
+    We use the analytic approximation of an infinite number of steps, which is
+    a good approximation even if the desired number of steps were less than 10.
+
+    The requested half light radius (hlr) should be thought of as a rough
```
jmeyers314: Should probably use `half_light_radius` here instead of `hlr`.

galsim/compound.py
```diff
@@ -902,3 +905,263 @@ def __setstate__(self, d):
...
+    --------------
+    @param  npoints                 Number of point sources to generate.
+    @param  half_light_radius       Half light radius of the distribution of
+                                    points.  This is the mean half light
+                                    radius produced by an infinite number of
+                                    points.  A single instance will be noisy.
+    @param  flux                    Optional total flux in all point sources.
+                                    [default: 1]
+    @param  rng                     Optional random number generator. Can be
+                                    any galsim.BaseDeviate.  If not sent or
```
jmeyers314: `If not sent`

typo

galsim/compound.py
```diff
@@ -902,3 +905,263 @@ def __setstate__(self, d):
...
+
+
+    @property
+    def input_half_light_radius(self):
+        """
+        getter for the input half light radius
+        """
+        return self._half_light_radius
+
+    @property
```
jmeyers314: Do we need this?  Doesn't the `GSObject.flux` property already cover this?  (Maybe this version is slightly faster?)

galsim/compound.py
```diff
@@ -902,3 +905,263 @@ def __setstate__(self, d):
...
+        return self._points.copy()
+
+    def _get_gaussians(self, points):
+        """
+        Create galsim.Gaussian objects for each point.
+
+        Highly optimized
+        """
+
+	gaussians = []
```
jmeyers314: Indentation error here prevents running on python3.

jmeyers314: (Mix of tabs and spaces).

galsim/compound.py
```diff
@@ -902,3 +905,263 @@ def __setstate__(self, d):
...
+            The list of galsim.Gaussian objects representing the points
+        .points
+            The array of x,y offsets used to create the point sources
+
+    Notes
+    -----
+
+    - The algorithm is a modified version of that presented in
+
+          https://arxiv.org/abs/1312.5514v3.
```
rmandelb: Super minor point, but I recommend omitting the . at the end of the URL (if someone just copies this whole line and doesn't notice it, they'll get a "can't find this preprint" blah blah error).

galsim/compound.py
esheldon: I used (hlr) with parentheses to show I would be using an abbreviation for this section of docs

galsim/compound.py
esheldon: GSObject.flux does not return the exact input value

galsim/compound.py
esheldon: fixed

jmeyers314:
Thanks @esheldon .  Looks good.

esheldon:
@rmandelb I updated demo4.py and demo4.yaml  but please have a look to see if this is what you meant

examples/demo4.py
```diff
@@ -133,16 +150,29 @@ def main(argv):
...
+        smooth_disk = galsim.Exponential(flux=smooth_disk_frac, half_light_radius=disk_hlr)
+
+        knots = galsim.RandomWalk(n_knots, disk_hlr, flux=knot_frac)
+
+        disk = galsim.Add([smooth_disk, knots])
+        disk = disk.shear(e1=disk_e1, e2=disk_e2)
+
+        # the rest of the light goes into the bulge
+        bulge_hlr = cat.getFloat(k,8)
+        bulge = galsim.DeVaucouleurs(flux=0.4, half_light_radius=bulge_hlr)
```
rmandelb: for consistency with the rest of the updates, maybe call this ```bulge_frac``` instead of 0.4?  That way people can change some of the numbers up top and everything down here gets updated appropriately.

examples/demo4.yaml
```diff
@@ -103,14 +106,32 @@ gal :
...
+
         - 
             type : Exponential
             half_light_radius : { type : Catalog , col : 5 }
             ellip : 
                 type : E1E2 
                 e1 : { type : Catalog , col : 6 }  
                 e2 : { type : Catalog , col : 7 }
-            flux : 0.6
+            flux : 0.18
```
rmandelb: I would think you could use the config eval functionality to define these in terms of each other (i.e., define variables for a bulge fraction and knot fraction, then use those to set the fluxes of each component using the appropriate formula) rather than having them be hard-coded numbers.

rmandelb:
@esheldon - that was exactly what I meant.  I had just two comments on the details.  Thanks.

examples/demo4.py
esheldon: Yes, I meant to, I even defined `bulge_frac` above.  will fix.

examples/demo4.yaml
rmandelb: Well, demo11.yaml gives some examples you can follow (of defining variables and then using them in expresssions later on).  But if we add that to this demo then we have to explain about the eval capabilities here, which would be a fairly big change.  So let me change my recommendation: leave it as-is, but put a comment saying that if you don't want to hard-code such things, see demo11.yaml for an example of how to define variables and then use them for calculations within the config framework.

rmjarvis: I think at this (early) stage in the demo progression, leaving these hard-coded is appropriate, although this demo already has the `some_variables` bit at the top, so you could put bulge_frac and knot_frac at the top if you want.

You could also intentionally omit of the last component flux and point out that when the overall galaxy has a flux specified (as it is here), the fluxes of each component are the fractional flux, and the final component's flux fraction will be automatically set to make the total = 1.

esheldon:
I'm going to keep it simple and hard code them all.  Still waiting for one check to complete, I didn't realize there were tests to make sure the py demos outputs exactly matched the yaml demos.

rmjarvis:
>  I didn't realize there were tests to make sure the py demos outputs exactly matched the yaml demos.

I think you need to update `galsim_default_input.asc`, which is done with the script `make_default_input.py`.  This is used by demo4.py as the input catalog, but it doesn't currently have any knot information in it.

rmjarvis:
Or, never mind.  You don't read in the knot stuff from the catalog.  

You can do the check yourself locally by running check_yaml and check_json in the examples directory.  Or while debugging, just run demo4 by itself and use check_diff to make sure they match up.

esheldon:
I made the obvious things the same, not sure why the images are different.

rmandelb:
I think you are not making the same rng for the RandomWalk?

rmjarvis:
I'll take a look.  (I've done lots of these py/yaml reconciliations...)

examples/demo4.py
```diff
@@ -133,16 +150,32 @@ def main(argv):
...
+        disk_hlr = cat.getFloat(k,5)
+        disk_e1 = cat.getFloat(k,6)
+        disk_e2 = cat.getFloat(k,7)
+        bulge_hlr = cat.getFloat(k,8)
+        bulge_e1 = cat.getFloat(k,9)
+        bulge_e2 = cat.getFloat(k,10)
+
+        smooth_disk = galsim.Exponential(flux=smooth_disk_frac, half_light_radius=disk_hlr)
+
+        knots = galsim.RandomWalk(n_knots, disk_hlr, flux=knot_frac)
```
rmjarvis: Yes, Rachel was right.  The only thing you need is `rng=rng` here (so it's not building a default urandom rng).

However, I think I'd also recommend using somewhat fewer than 100 knots here, since it makes the demo much slower.  (~51 seconds vs ~14 seconds).  Probably the pedagogical impact is achieved with only ~10-20 knots?

esheldon:
Yes, not sending the rng was a bug, thanks.

esheldon:
Examples of npoints=10, 20, 100


test-n10
test-n20
test-n100

examples/demo4.yaml
```diff
@@ -103,14 +106,35 @@ gal :
...
+        # 0.6 of flux is in the total disk.  Further divide this
+        # into a smooth disk and knots of star formation, modeled
+        # as point sources distributed via a random walk
+        #
+        # note the knots get the same hlr as the smooth disk, and
+        # get sheared by the same ellipticity, but the knots will
+        # themselves have additional shape noise
+
+        # see demo11.yaml for a way to specify these flux variable relative to
+        # one another needing hard coding them
```
rmandelb: "needing" -> "without"?

examples/demo4.py
rmandelb: I agree with your statement about needing 100 for an irregular, but in this case there is a bulge+disk component so why not go with 10-20, and have a comment explicitly noting that that's only sufficient because of the existence of those components?  That way you've given that information, but the demo is still nearly as fast as before.

esheldon:
I think I've addressed all the requested changes.

rmandelb:
Yep, all done as far as I'm concerned.

Issue #822, opened 2016-11-07
Random sampling of SED wavelengths


kadrlica:
At the DESC hack week we discussed implementing random sampling of the wavelengths of photons generated from an SED. This would involve adding a function to the `sed.SED` object which would make use of `random.DistDeviate` to sample from the SED function. There would be some convolution by the bandpass, which may occur later on.

kadrlica:
Here are a few examples:

```
import numpy as np
import pylab as plt

import galsim

dirname = "/Users/alex/desc/software/galsim"
sed = galsim.SED(dirname+'/share/CWW_Scd_ext.sed', wave_type='nanometers', flux_type='flambda')
band = galsim.Bandpass(dirname+'/examples/data/LSST_r.dat', 'nm')
sedbp = sed*band

size=1e6
bins=200

plt.figure()
plt.hist(sed.sampleWavelength(size,None),bins=bins,normed=True)
plt.plot(sed.wave_list,sed(sed.wave_list)/sed.calculateFlux(None),'r',lw=1.5)
plt.xlabel("Wavelength (nm)")
plt.ylabel("Normalized Counts")
plt.title("Full SED")
plt.savefig('sample_full.png',bbox_inches='tight')

plt.figure()
plt.hist(sed.sampleWavelength(size,band),bins=bins,normed=True)
plt.plot(sedbp.wave_list,sedbp(sedbp.wave_list)/sedbp.calculateFlux(None),'r',lw=1.5)
plt.xlabel ("Wavelength (nm)")
plt.ylabel ("Normalized Counts")
plt.title("LSST r-band")
plt.savefig('sample_rband.png',bbox_inches='tight')
```

![sample_full](https://cloud.githubusercontent.com/assets/5605123/20157822/2a865df4-a6a5-11e6-86b5-852c00f63db8.png)
![sample_rband](https://cloud.githubusercontent.com/assets/5605123/20157821/2a85bcfa-a6a5-11e6-820b-eb2f48d01a9e.png)

kadrlica:
There is no test covering redshifting the SED. It would be good to add this.

PR #826, opened 2016-11-09
#822 Adding SED wavelength sampling

kadrlica:
Added `SED.SampleWavelength` to randomly sample wavelengths from an SED. Additional changes from Mike to add the `BaseDeviate.generate` function to create arrays of random variates. Some tests for each new functionality.


galsim/sed.py
```diff
@@ -818,6 +818,49 @@ def calculateSeeingMomentRatio(self, bandpass, alpha=-0.2, base_wavelength=500):
             return galsim.integ.int1d(lambda w: weight(w) * kernel(w),
                                       bandpass.blue_limit, bandpass.red_limit) / flux
 
+    def sampleWavelength(self, nphotons, bandpass, seed=0, npoints=256):
+        """ Sample 'nphoton' random wavelengths from the SED (and bandpass).
+
+        @param nphotons  Number of samples (photons) to randomly draw.
+        @param bandpass  Bandpass through which object is being imaged.
+        @param seed      Something that can seed a BaseDeviate: a long int seed or another
```
rmjarvis: This is my fault, since I think I told you to call this parameter `seed`, but to make this more consistent with other functions/classes that need random deviates, we typically call this parameter `rng`, and nominally it should just be an existing BaseDeviate object.  None is still allowed, in which case a default BaseDeviate object is constructed automatically for the user.

galsim/sed.py
```diff
@@ -818,6 +818,49 @@ def calculateSeeingMomentRatio(self, bandpass, alpha=-0.2, base_wavelength=500):
             return galsim.integ.int1d(lambda w: weight(w) * kernel(w),
                                       bandpass.blue_limit, bandpass.red_limit) / flux
 
+    def sampleWavelength(self, nphotons, bandpass, seed=0, npoints=256):
+        """ Sample 'nphoton' random wavelengths from the SED (and bandpass).
```
rmjarvis: Minor point, but the quoted 'nphoton' here is not the actual name of the parameter (nphotons), but I think it might read better to just use plain text here.  Something like "Sample a number of random wavelength values from the SED, possibly as observed through a bandpass."

galsim/sed.py
```diff
@@ -818,6 +818,49 @@ def calculateSeeingMomentRatio(self, bandpass, alpha=-0.2, base_wavelength=500):
...
+        try:
+            dev = self.deviate[key]
+        except KeyError:
+            dev = galsim.DistDeviate(function=fn,x_min=sed.blue_limit,x_max=sed.red_limit,
+                                     npoints=npoints)
+            self.deviate[key] = dev
+
+        # Set the seed explicitly rather than in the constructor
+        ret = np.empty(nphotons)
+        dev.seed(seed)
```
rmjarvis: I believe this only works if `seed` is an integer.  For the use case of `rng` being another BaseDeviate, which is more often how this will be used in practice, this should be `dev.reset(rng)`.  This also works if `rng` is an integer seed, btw.

This implies that there are probably no unit tests currently that use an existing BaseDeviate as the seed object, so we should add some that test that use case (in addition to the existing ones with an integer seed).

tests/test_sed.py
```diff
@@ -544,6 +544,56 @@ def test_SED_calculateSeeingMomentRatio():
...
+    np.testing.assert_equal(len(out),1e3,"Unexpected number of SED samples.")
+
+    np.testing.assert_equal(np.sum(out > sedbp.red_limit),0,
+                            "SED sample outside of function bounds.")
+    np.testing.assert_equal(np.sum(out < sedbp.blue_limit),0,
+                            "SED sample outside of function bounds.")
+
+    out2 = sed.sampleWavelength(1e3,bandpass,seed=seed,npoints=512)
+    np.testing.assert_equal(len(sed.deviate),3,"Unexpected number of SED deviates.")
+    np.testing.assert_almost_equal(out,out2,0,"SED sample npoints not equal.")
```
rmjarvis: This error message here is wrong.  You aren't testing npoints, but rather that the values are close (equal to 0 d.p.). Suggest the error message be "SED samples using different npoints don't match to nearest integer"

tests/test_sed.py
```diff
@@ -544,6 +544,56 @@ def test_SED_calculateSeeingMomentRatio():
...
+    out = sed.sampleWavelength(1e5,None,seed=seed)
+    x,step = np.linspace(sed.blue_limit,sed.red_limit,1e5,retstep=True)
+    c = (x[:-1]+x[1:])/2.
+    cdf = np.cumsum(sed(c)*step)
+    cdf /= cdf.max()
+
+    test = np.linspace(sed.blue_limit,sed.red_limit,100)
+    cdf1 = np.interp(test,c,cdf)
+    cdf2 = np.array([(out <= w).sum() for w in test],dtype=float)/len(out)
+    np.testing.assert_almost_equal(cdf1,cdf2,2,"Sampled SED does not match input.")
```
rmjarvis: You mentioned this yourself on the #822 issue page, but I wanted to second that it would be worth adding a test that uses a redshifted SED, since there is non-trivial code that depends on sed.redshift.

CHANGELOG.md
```diff
@@ -61,6 +61,7 @@ New Features
   inclined to the line of sight at a desired angle. If face-on (inclination =
   0 degrees), this will be identical to the Exponential profile.  (#782)
 - Added support for reading in of unsigned int Images (#715)
+- Added 'generate' function to BaseDeviate and 'sed.sampleWavelength' to draw random wavelengths from an SED. (#822)
```
rmandelb: Should wrap this line (we wrap the CHANGELOG at fewer characters I believe because of how it shows up on GitHub when viewing directly - you can see this in how the other entries are wrapped).

galsim/sed.py
rmandelb: +1

galsim/sed.py
```diff
@@ -818,6 +818,49 @@ def calculateSeeingMomentRatio(self, bandpass, alpha=-0.2, base_wavelength=500):
             return galsim.integ.int1d(lambda w: weight(w) * kernel(w),
                                       bandpass.blue_limit, bandpass.red_limit) / flux
 
+    def sampleWavelength(self, nphotons, bandpass, seed=0, npoints=256):
+        """ Sample 'nphoton' random wavelengths from the SED (and bandpass).
+
+        @param nphotons  Number of samples (photons) to randomly draw.
+        @param bandpass  Bandpass through which object is being imaged.
```
rmandelb: Perhaps mention here that if you don't wish to sample through a bandpass, it's completely legit to pass in 'None'.

galsim/sed.py
rmandelb: You can see an example of how this is done in real.py, in the RealGalaxy constructor.

tests/test_integ.py
```diff
@@ -214,6 +214,25 @@ def test_midpoint_basic():
         result/expected_val, 1.0, decimal=2, verbose=True,
         err_msg='Simple test of midpt() method failed for f(x)=x^2 from 0 to 10')
 
+@timer
+def test_trapz_basic():
```
rmjarvis: Need to add this to the `if __name__ == '__main__'` bit at the bottom.

CHANGELOG.md
```diff
@@ -92,6 +92,8 @@ New Features
   point sources distributed using a random walk.  Uses of this profile include
   representing an "irregular" galaxy, or adding this profile to an Exponential
   to represent knots of star formation. (#819)
+- Added support for reading in of unsigned int Images (#715)
```
rmjarvis: This is a stray addition that already shows up at the top of this section.  (Our ordering scheme is by issue number, so 715 comes before the other items in this list.)

Issue #823, opened 2016-11-07
Give photons plausible incidence angles


rmjarvis:
To properly rain electrons through a detector, they should really have the right distribution of incidence angles, so the different wavelengths that convert to electrons at different depth have the right spatial positions.

This issue is to add a function to give a PhotonArray with a bunch of photons values for the incidence angles given some f/ratio and obscuration fraction.

The assumption is that photons are uniformly distributed within the non-obscured solid angle.

I propose that our "angles" really be dx/dz and dy/dz values giving the slope in x or y relative to the motion in z.  This will likely save us from having trig calls when we actually want to use the angles to compute something.

rmjarvis:
Merged.  #830

Issue #827, opened 2016-11-09
Add hook for running surface-layer operations on photon array


rmjarvis:
At the hack week, @kadrlica has implemented code to sample a random selection of wavelengths from an SED (or more typically an SED * bandpass) cf. #822.  @sethdigel is working on giving the photons appropriate dxdz, dydz values on #823.  

The next step to be able to use these is to add a hook for running through a list of surface-layer operations in drawImage.  Here is the mock up that I presented in my [talk](https://project.lsst.org/meetings/lsst2016/sites/lsst.org.meetings.lsst2016/files/Sensors.pdf) about this (see p. 6 and 11):
```
prof.drawImage(stamp, method='phot', rng=rng,
               sky=sky, # need to have this now
               surface_ops=[fratio, sed, fringe],
                           # all surface-layer operations in order
               sensor=quick_ccd # several options here
)
```
The `fratio` surface_op is what I called the thing that Seth is working on to give appropriate angles based on the f/ratio.  `sed` (which wasn't in the original presentation, but which I added here) represents code to apply Alex's wavelength sampling.  `fringe` would be something that implements @dkirkby's code for simulating fringing.  

There are more options than just these that we can consider adding down the line as well, but the point of this issue is to implement the code in `drawImage` that can run through such a list of surface operations.  i.e. something like this:
```
for op in surface_ops:
    photon_array = op.applyTo(photon_array)
```

rmjarvis:
Done.  PR #831

PR #830, opened 2016-11-10
#823 Implemented assignPhotonAngles function

sethdigel:
The assignPhotonAngles function generates directions (in the form of dx/dz and dy/dz slopes) for an existing photon array.  The generated directions are uniformly distributed in solid angle for inclination angles between the obscuration angle and the FOV size (defined in terms of the f-ratio).  An associated testing script was also implemented.

### DIFF
--- a/galsim/photon_array.py
+++ b/galsim/photon_array.py
@@ -25,7 +25,7 @@ class FRatioAngles(object):
     obscuration.
 
     Assigns arrival directions at the focal plane for photons, drawing from a uniform
-    brightness distribution between the obscuration angle and the angle of the FOV defined
+    brightness distribution between the obscuration angle and the edge of the pupil defined
     by the f/ratio of the telescope.  The angles are expressed in terms of slopes dx/dz
     and dy/dz.
 
@@ -61,8 +61,8 @@ class FRatioAngles(object):
         # The f/ratio is the ratio of the focal length to the diameter of the aperture of
         # the telescope.  The angular radius of the field of view is defined by the 
         # ratio of the radius of the aperture to the focal length
-        fov_angle = np.arctan(0.5 / self.fratio)  # radians
-        obscuration_angle = self.obscuration * fov_angle
+        pupil_angle = np.arctan(0.5 / self.fratio)  # radians
+        obscuration_angle = np.arctan(0.5 * self.obscuration / self.fratio)
 
         # Generate azimuthal angles for the photons
         phi = np.empty(n_photons)
@@ -70,10 +70,10 @@ class FRatioAngles(object):
         phi *= (2 * np.pi)
 
         # Generate inclination angles for the photons, which are uniform in sin(theta) between
-        # the sine of the obscuration angle and the sine of the FOV radius
+        # the sine of the obscuration angle and the sine of the pupil radius
         sintheta = np.empty(n_photons)
         self.ud.generate(sintheta) 
-        sintheta = np.sin(obscuration_angle) + (np.sin(fov_angle) - np.sin(obscuration_angle)) \
+        sintheta = np.sin(obscuration_angle) + (np.sin(pupil_angle) - np.sin(obscuration_angle)) \
             * sintheta
 
         # Assign the directions to the arrays. In this class the convention for the
--- a/galsim/photon_array.py
+++ b/galsim/photon_array.py
@@ -20,57 +20,67 @@ import numpy as np
 from ._galsim import PhotonArray
 import galsim
 
-def assignPhotonAngles(self, fratio, obscuration, seed):
-    """
+class FRatioAngles(object):
+    """A surface-layer operator that assigns photon directions based on the f/ratio and
+    obscuration.
+
     Assigns arrival directions at the focal plane for photons, drawing from a uniform
     brightness distribution between the obscuration angle and the angle of the FOV defined
-    by the f-ratio of the telescope.  The angles are expressed in terms of slopes dx/dz
+    by the f/ratio of the telescope.  The angles are expressed in terms of slopes dx/dz
     and dy/dz.
 
-    @param fratio               The f-ratio of the telescope (1.2 for LSST)
-    @param obscuration_angle    The angular radius of the central obscuration (deg)
-    @param seed                 Random number seed (optional)
+    @param fratio           The f-ratio of the telescope (e.g. 1.2 for LSST)
+    @param obscuration      Linear dimension of central obscuration as fraction of aperture
+                            linear dimension. [0., 1.).  [default: 0.0]
+    @param rng              A random number generator to use or None, in which case an rng
+                            will be automatically constructed for you. [default: None]
     """
+    def __init__(self, fratio, obscuration=0.0, rng=None):
 
-    if obscuration < 0 or obscuration > 1:
-        raise ValueError("The obscuration fraction must be between 0 and 1.")
+        if fratio < 0:
+            raise ValueError("The f-ratio must be positive.")
+        if obscuration < 0 or obscuration >= 1:
+            raise ValueError("The obscuration fraction must be between 0 and 1.")
+        if rng is None:
+            ud = galsim.UniformDeviate()
+        else:
+            ud = galsim.UniformDeviate(rng)
 
-    if fratio < 0:
-        raise ValueError("The f-ratio must be positive.")
-        
-    dxdz = self.getDXDZArray()
-    dydz = self.getDYDZArray()
-    n_photons = len(dxdz)
+        self.fratio = fratio
+        self.obscuration = obscuration
+        self.ud = ud
 
-    fov_angle = np.arctan(0.5 / fratio)  # radians
-    obscuration_angle = obscuration * fov_angle
 
-    if seed is None:
-        ud = galsim.UniformDeviate()
-    else:
-        ud = galsim.UniformDeviate(seed)
+    def applyTo(self, photon_array):
+        """Assign directions to the photons in photon_array."""
 
-    # Generate azimuthal angles for the photons
-    # Set up a loop to fill the array of azimuth angles for now
-    # (The array is initialized below but there's no particular need to do this.)
-    phi = np.zeros(n_photons)
+        dxdz = photon_array.getDXDZArray()
+        dydz = photon_array.getDYDZArray()
+        n_photons = len(dxdz)
 
-    for i in np.arange(n_photons):
-        phi[i] = ud() * 2 * np.pi 
+        fov_angle = np.arctan(0.5 / self.fratio)  # radians
+        obscuration_angle = self.obscuration * fov_angle
 
-    # Generate inclination angles for the photons, which are uniform in sin(theta) between
-    # the sine of the obscuration angle and the sine of the FOV radius
-    sintheta = np.zeros(n_photons)
+        # Generate azimuthal angles for the photons
+        # Set up a loop to fill the array of azimuth angles for now
+        # (The array is initialized below but there's no particular need to do this.)
+        phi = np.zeros(n_photons)
 
-    for i in np.arange(n_photons):
-        sintheta[i] = np.sin(obscuration_angle) + (np.sin(fov_angle) - \
-                      np.sin(obscuration_angle))*ud()
+        for i in np.arange(n_photons):
+            phi[i] = self.ud() * 2 * np.pi
 
-    # Assign the directions to the arrays. In this class the convention for the
-    # zero of phi does not matter but it might if the obscuration dependent on
-    # phi
-    costheta = np.sqrt(1. - np.square(sintheta))
-    dxdz[:] = costheta * np.sin(phi)
-    dydz[:] = costheta * np.cos(phi)
+        # Generate inclination angles for the photons, which are uniform in sin(theta) between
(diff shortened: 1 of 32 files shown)


rmjarvis:
Note: All the C++ layer stuff is also in #827, which includes real unit tests for it all, so probably better to review that part on #831 rather than here.  But Seth needed this functionality so I implemented it here first.  The files photon_array.py and test_photon_array.py are really what this PR is about.

rmjarvis:
@sethdigel I reworked the front end of this code slightly in anticipation of the design in #827.  Now this is a class FRatioAngles that can `applyTo` a photon array, rather than be a method of PhotonArray directly.  That way it will be easier to have several different options for how we assign directions to the photons.

galsim/photon_array.py
```diff
@@ -0,0 +1,86 @@
...
+                            will be automatically constructed for you. [default: None]
+    """
+    def __init__(self, fratio, obscuration=0.0, rng=None):
+
+        if fratio < 0:
+            raise ValueError("The f-ratio must be positive.")
+        if obscuration < 0 or obscuration >= 1:
+            raise ValueError("The obscuration fraction must be between 0 and 1.")
+        if rng is None:
+            ud = galsim.UniformDeviate()
```
rmjarvis: This line is currently uncovered by the tests.  We should have a test that makes FRatioAngles without a specified seed.  Probably also one that uses an existing BaseDeviate rather than an integer seed.  We don't necessarily need to do the full test of the distribution with these, especially the default one, but tests of min/max values are probably worthwhile.

galsim/photon_array.py
```diff
@@ -0,0 +1,86 @@
...
+                            linear dimension. [0., 1.).  [default: 0.0]
+    @param rng              A random number generator to use or None, in which case an rng
+                            will be automatically constructed for you. [default: None]
+    """
+    def __init__(self, fratio, obscuration=0.0, rng=None):
+
+        if fratio < 0:
+            raise ValueError("The f-ratio must be positive.")
+        if obscuration < 0 or obscuration >= 1:
+            raise ValueError("The obscuration fraction must be between 0 and 1.")
```
rmjarvis: Probably worth adding some assert_raises tests to check these lines work properly.  (They aren't required for our coverage report, but they're worth covering anyway.).  Something like the following:
```
try:
    np.testing.assert_raises(ValueError, FRatioAngles, fratio=-0.3)
    np.testing.assert_raises(ValueError, FRatioAngles, fratio=1.2, obscuration=-0.3)
    np.testing.assert_raises(ValueError, FRatioAngles, fratio=1.2, obscuration=1.0)
    np.testing.assert_raises(ValueError, FRatioAngles, fratio=1.2, obscuration=1.9)
except ImportError:
    pass
```
The try/except wrapper is because these tests technically require nose, which we don't have as an official dependency, so we don't require the user to have it installed.

galsim/photon_array.py
```diff
@@ -0,0 +1,86 @@
...
+        fov_angle = np.arctan(0.5 / self.fratio)  # radians
+        obscuration_angle = self.obscuration * fov_angle
+
+        # Generate azimuthal angles for the photons
+        # Set up a loop to fill the array of azimuth angles for now
+        # (The array is initialized below but there's no particular need to do this.)
+        phi = np.zeros(n_photons)
+
+        for i in np.arange(n_photons):
+            phi[i] = self.ud() * 2 * np.pi
```
rmjarvis: This would be more efficient as
```
phi = self.ud.generate(n_photons) * 2 * np.pi
```
The `generate` function is new in Alex's branch, so once that's merged into master, it would be worth merging it into this branch and using it here.

galsim/photon_array.py
```diff
@@ -0,0 +1,86 @@
...
+
+        # Generate inclination angles for the photons, which are uniform in sin(theta) between
+        # the sine of the obscuration angle and the sine of the FOV radius
+        sintheta = np.zeros(n_photons)
+
+        for i in np.arange(n_photons):
+            sintheta[i] = np.sin(obscuration_angle) + (np.sin(fov_angle) - \
+                          np.sin(obscuration_angle))*self.ud()
```
rmjarvis: Likewise, this would be more efficient as
```
sinobs = np.sin(obscuration_angle)
sinfov = np.sin(fov_angle)
sintheta = sinobs + (sinfov - sinobs) * self.ud.generate(n_photons)
```

galsim/photon_array.py
```diff
@@ -0,0 +1,86 @@
...
+
+        # Assign the directions to the arrays. In this class the convention for the
+        # zero of phi does not matter but it might if the obscuration dependent on
```
rmjarvis: typo: either "depends" or "is dependent"

tests/test_photon_array.py
```diff
@@ -0,0 +1,93 @@
...
+    path, filename = os.path.split(__file__)
+    sys.path.append(os.path.abspath(os.path.join(path, "..")))
+    import galsim
+
+@timer
+def test_photon_angles():
+    """Test the photon_array function
+    """
+    # Make a photon array
+    ud = galsim.UniformDeviate()
```
rmjarvis: This should be given a definite seed to make sure tests don't fail randomly at rare intervals.

tests/test_photon_array.py
```diff
@@ -0,0 +1,93 @@
...
+    assigner.applyTo(photon_array)
+
+    dxdz = photon_array.getDXDZArray()
+    dydz = photon_array.getDYDZArray()
+
+    # Compare these slopes with the expected distributions (uniform in azimuth 
+    # over all azimiths and uniform in sin(inclination) over the range of
+    # allowed inclinations
+    phi = np.arctan2(dydz,dxdz)
+    sintheta = np.sqrt(1. - np.square(dxdz) - np.square(dydz))
```
rmjarvis: I don't think this is quite right.  I think tan(theta) = sqrt(dxdz^2 + dydz^2), which leads to sin(theta) = tan(theta) / sqrt(1+tan(theta)^2).  
However, in the unit test, it's probably worth using the more obvious trig formula:
```
tantheta = np.sqrt(np.square(dxdz) + np.square(dydz))
sintheta = np.sin(np.arctan(tantheta))
```
I'm not sure how much this affects the below tests of the distribution.

tests/test_photon_array.py
```diff
@@ -0,0 +1,93 @@
...
+    np.testing.assert_almost_equal(chisqr_phi, 1.05562, 5, \
+        "Distribution in azimuth is not uniform")
+    np.testing.assert_almost_equal(chisqr_sintheta, 0.95920, 5, \
+        "Distribution in sin(inclination) is not uniform")
+
+    # Also check that the values are within the ranges expected
+    np.testing.assert_almost_equal(np.min(phi) + np.pi, 0.000258, 5, \
+        "Azimuth angle range extends to too-small angles")
+    np.testing.assert_almost_equal(np.max(phi) - np.pi, -6.474875e-05, 5, \
+        "Inclination angle range extends to too-large angles")
```
rmjarvis: I think these could just check that -pi < phi < pi.  (Also the message on the second one is wrong.)
```
np.testing.assert_less(phi, np.pi, "Azimuth angle range extends to too-large angles")
np.testing.assert_less(-np.pi, phi, "Azimuth angle range extends to too-small angles")
```
Likewise for the below tests of sintheta.

tests/test_photon_array.py
```diff
@@ -0,0 +1,93 @@
...
+                      sintheta_ref) / sintheta_histo.size
+
+    
+    # In the assert_almost_equal tests below, the expected values are defined
+    # for the particular set of directions generated for the seed value specified
+    # above
+    np.testing.assert_almost_equal(chisqr_phi, 1.05562, 5, \
+        "Distribution in azimuth is not uniform")
+    np.testing.assert_almost_equal(chisqr_sintheta, 0.95920, 5, \
+        "Distribution in sin(inclination) is not uniform")
```
rmjarvis: I think probably what we really care about here is that it is near 1, so maybe test that 0.9 < chisqr < 1.1 in each case.  Rather than test the exact values.

tests/test_photon_array.py
```diff
@@ -0,0 +1,93 @@
...
+    # allowed inclinations
+    phi = np.arctan2(dydz,dxdz)
+    sintheta = np.sqrt(1. - np.square(dxdz) - np.square(dydz))
+
+    phi_histo, phi_bins = np.histogram(phi, bins=100)
+    sintheta_histo, sintheta_bins = np.histogram(sintheta, bins=100)
+
+    phi_ref = phi_histo*0 + float(np.sum(phi_histo))/phi_histo.size
+    sintheta_ref = sintheta_histo*0 + float(np.sum(sintheta_histo)
+                   )/sintheta_histo.size
```
rmjarvis: I think you don't need the `*0` bits.  The below calculations should work if the references are floats rather than arrays.

galsim/photon_array.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+    def applyTo(self, photon_array):
+        """Assign directions to the photons in photon_array."""
+
+        dxdz = photon_array.getDXDZArray()
+        dydz = photon_array.getDYDZArray()
+        n_photons = len(dxdz)
+
+        fov_angle = np.arctan(0.5 / self.fratio)  # radians
```
rmjarvis: Probably comment with a brief summary of the math for why this is the right formula would be appropriate above this line.

tests/test_photon_array.py
sethdigel: Thanks, you're right.  I've fixed this on the generation side as well.  I was treating dxdz and dydz like direction cosines

tests/test_photon_array.py
sethdigel: Thanks.  I've implemented the checks using np.testing.assert_array_less.

galsim/photon_array.py
sethdigel: I've implemented this using self.ud.generate as you suggested.  The syntax is not quite as above because the argument is not n_photons but an array of that size.

sethdigel:
Thanks a lot, Mike.  I've pushed updated versions with changes to address each of your comments.

rmjarvis:
Great.  Looks good Seth.  I added one more thing to the test suite (testing rng=None, ud, or seed), which all passed and got the coverage up to 100%.

Anyone else want to review?

galsim/photon_array.py
```diff
@@ -179,3 +180,68 @@ def applyTo(self, photon_array):
         """Assign wavelengths to the photons sampled from the SED * Bandpass."""
         photon_array.wavelength = self.sed.sampleWavelength(
                 photon_array.size(), self.bandpass, rng=self.rng, npoints=self.npoints)
+
+
+class FRatioAngles(object):
+    """A surface-layer operator that assigns photon directions based on the f/ratio and
+    obscuration.
+
+    Assigns arrival directions at the focal plane for photons, drawing from a uniform
```
jmeyers314: I think "angle of the FOV" isn't quite right.  Maybe "angle of the pupil"?

galsim/photon_array.py
```diff
@@ -179,3 +180,68 @@ def applyTo(self, photon_array):
...
+
+    def applyTo(self, photon_array):
+        """Assign directions to the photons in photon_array."""
+
+        dxdz = photon_array.getDXDZArray()
+        dydz = photon_array.getDYDZArray()
+        n_photons = len(dxdz)
+
+        # The f/ratio is the ratio of the focal length to the diameter of the aperture of
+        # the telescope.  The angular radius of the field of view is defined by the
```
jmeyers314: I think FOV is the wrong term here too.  For LSST, the fratio is 1.234, which leads to a maximum inclination angle of ~22 degrees.  `fov_angle` makes me think of the ~1.75 degree radius of the FOV in sky coordinates though.  So I think something like `pupil_angle` makes more sense.

galsim/photon_array.py
```diff
@@ -179,3 +180,68 @@ def applyTo(self, photon_array):
...
+
+        dxdz = photon_array.getDXDZArray()
+        dydz = photon_array.getDYDZArray()
+        n_photons = len(dxdz)
+
+        # The f/ratio is the ratio of the focal length to the diameter of the aperture of
+        # the telescope.  The angular radius of the field of view is defined by the 
+        # ratio of the radius of the aperture to the focal length
+        fov_angle = np.arctan(0.5 / self.fratio)  # radians
+        obscuration_angle = self.obscuration * fov_angle
```
jmeyers314: I think this is only true under the small angle approximation.  I'm not sure 22 degrees is small enough though for LSST.  Should probably use `np.arctan(0.5 * self.obscuration / self.fratio)` instead.

galsim/photon_array.py
sethdigel: Yes, that is what I should have used.

galsim/photon_array.py
sethdigel: Thanks for this.  I had the mistaken impression that the obscuration fraction was defined as a fraction of the pupil angular radius but I see that in Mike's description it is a fraction of the aperture size.

sethdigel:
I've committed a new version of photon_array.py addressing Josh's comments (and merging in the other updates since yesterday).

jmeyers314:
Thanks Seth.  Looks good to me.

Issue #824, opened 2016-11-07
Add raytracing mode for realistic atmospheric PSFs.


jmeyers314:

jmeyers314:
Here's an example comparing fft mode to photon shooting mode for an optical psf.

![psf_comparison](https://cloud.githubusercontent.com/assets/3650485/20225412/2c211420-a811-11e6-9d35-65a83e4afc53.png)

jmeyers314:
Left is FFT-mode, Right is photon-shooting mode with the refractive approximation.

jmeyers314:
I've been thinking more about how to properly delay computation for atmospheric PSFs and also making an intuitive API for this.  Currently, the way to draw multiple PSFs corresponding to the same atmosphere over the same time interval (but at different field angles) is something like this:

```python
atm = function_to_make_atmosphere()
thetas = function_returning_field_angles_of_interest()
psfs = atm.makePSF(..., theta=thetas)  # returns list of PSFs

for psf, gal in zip(psfs, gals):
    obj = galsim.Convolve(psf, gal)
    obj.drawImage(...)
```

Note that this is different than:
```python
atm = function_to_make_atmosphere()
thetas = function_returning_field_angles_of_interest()
psfs = []
for theta in thetas:
    psfs.append(atm.makePSF(..., theta=theta)

for psf, gal in zip(psfs, gals):
    obj = galsim.Convolve(psf, gal)
    obj.drawImage(...)
```

since the latter increments the clock each time `makePSF` is called.  I think it may be more intuitive, however, to have these two code snippets perform the same computation.  To make the latter efficient, I also think we need to delay the actual computation of the PSF to drawImage (currently all the Fourier optics math occurs at initialization).  For photon-shooting, I think this has more-or-less already been accomplished by pulling `SBProfile.shoot()` into the python layer and overriding `PhaseScreenPSF.shoot()`.  I guess something similar could be accomplished for FFT-mode by pulling `SBProfile.drawK()` into the python layer and adding a custom `PhaseScreenPSF.drawK()`?

[...]

PR #831, opened 2016-11-11
#827 Add surface_ops parmater to drawImage

rmjarvis:
Add surface_ops as a new parameter to drawImage.

This also adds a WavelengthSampler class that calls Alex's new function implemented in #822 in a standard format that we can use for all surface operations.

It also adds some boiler-plate python-layer code and unit tests for PhotonArray, which had previously been primarily a C++-layer class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,6 +94,9 @@ New Features
   to represent knots of star formation. (#819)
 - Added 'generate' function to BaseDeviate and 'sed.sampleWavelength' to draw
   random wavelengths from an SED. (#822)
+- Added `surface_ops` option to `drawImage` function, which applys a list of
+  surface operations to the photon array before accumulating on the image.
+  (#827)
 
 
 New config features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,8 @@ New Features
   point sources distributed using a random walk.  Uses of this profile include
   representing an "irregular" galaxy, or adding this profile to an Exponential
   to represent knots of star formation. (#819)
+- Added 'generate' function to BaseDeviate and 'sed.sampleWavelength' to draw
+  random wavelengths from an SED. (#822)
 
 
 New config features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,7 +94,7 @@ New Features
   to represent knots of star formation. (#819)
 - Added 'generate' function to BaseDeviate and 'sed.sampleWavelength' to draw
   random wavelengths from an SED. (#822)
-- Added `surface_ops` option to `drawImage` function, which applys a list of
+- Added `surface_ops` option to `drawImage` function, which applies a list of
   surface operations to the photon array before accumulating on the image.
   (#827)
 
--- a/galsim/gsobject.py
+++ b/galsim/gsobject.py
@@ -979,7 +979,7 @@ class GSObject(object):
     def drawImage(self, image=None, nx=None, ny=None, bounds=None, scale=None, wcs=None, dtype=None,
                   method='auto', area=1., exptime=1., gain=1., add_to_image=False,
                   use_true_center=True, offset=None, n_photons=0., rng=None, max_extra_noise=0.,
-                  poisson_flux=None, setup_only=False, dx=None, wmult=1.):
+                  poisson_flux=None, surface_ops=(), setup_only=False, dx=None, wmult=1.):
         """Draws an Image of the object.
 
         The drawImage() method is used to draw an Image of the current object using one of several
@@ -1232,6 +1232,9 @@ class GSObject(object):
                             Poisson statistics for `n_photons` samples when photon shooting.
                             [default: True, unless `n_photons` is given, in which case the default
                             is False]
+        @param surface_ops  A list of operators that can modify the photon array that will be
+                            applied in order before accumulating the photons on the sensor.
+                            [default: ()]
         @param setup_only   Don't actually draw anything on the image.  Just make sure the image
                             is set up correctly.  This is used internally by GalSim, but there
                             may be cases where the user will want the same functionality.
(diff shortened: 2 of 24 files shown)


jmeyers314:
Looks good to me.  I fixed a couple typos, that's it.

CHANGELOG.md
```diff
@@ -94,6 +94,9 @@ New Features
   to represent knots of star formation. (#819)
 - Added 'generate' function to BaseDeviate and 'sed.sampleWavelength' to draw
   random wavelengths from an SED. (#822)
+- Added `surface_ops` option to `drawImage` function, which applys a list of
```
rmandelb: applies

galsim/gsobject.py
```diff
@@ -1713,6 +1721,9 @@ def drawPhot(self, image, n_photons=0, rng=None, max_extra_noise=None, poisson_f
 
             phot_array.scaleFlux(g * thisN / Ntot)
 
+            for op in surface_ops:
```
rmandelb: Question: I think that the way you've set this up (as a loop over a bunch of things to be done in order using an ```applyTo``` method) makes sense.  But what I'm wondering about is information flow.  Will some of these processes need access to additional information besides what is flowing through here?  Can we assume that if necessary the PhotonArray class will be modified to carry more info around, or might this need to have some more complex way of passing through the relevant info?  I realize I'm asking you to speculate about potentially arbitrarily complicated sensor effects so possibly the answer is just "I don't know, we'll see if it happens" which I guess is also legit.

galsim/photon_array.py
```diff
@@ -0,0 +1,181 @@
...
+import numpy as np
+
+# Add on more methods in the python layer
+
+PhotonArray.__doc__ = """
+The PhotonArray class encapsulates the concept of a collection of photons incident on
+a detector.
+
+A PhotonArray object is not typically constructed directly by the user.  Rather, it is
+typically constructed as the return value of the `GSObject.shoot` method.
```
rmandelb: SBProfile.shoot?

galsim/photon_array.py
```diff
@@ -0,0 +1,181 @@
...
+Initialization
+--------------
+
+The initialization constructs a PhotonArray to hold N photons, but does not set the values of
+anything yet.  The constructor allocates space for the x,y,flux arrays, since those are always
+needed.  The other arrays are only allocated on demand if the user accesses these attributes.
+
+@param N            The number of photons to store in this PhotonArray.  This value cannot be
+                    changed.
+@param x            Optionally, the inital x values. [default: None]
```
rmandelb: inital-> initial here and in all lines below

galsim/photon_array.py
```diff
@@ -0,0 +1,181 @@
...
+"""
+
+# In python we want the init function to be a bit more functional so we can serialize properly.
+# Save the C++-layer constructor as _PhotonArray_empty_init.
+_PhotonArray_empty_init = PhotonArray.__init__
+
+# Now make the one we want as the python-layer init function.
+def PhotonArray_init(self, N, x=None, y=None, flux=None, dxdz=None, dydz=None, wavelength=None):
+    _PhotonArray_empty_init(self, N)
+    if x is not None: self.x[:] = x
```
rmandelb: Do we need to do any sanity checks of the x,y,etc. values that get passed in?

galsim/photon_array.py
```diff
@@ -0,0 +1,181 @@
...
+be implemented soon, so the above paragraph is a bit aspirational atm.
+
+Attributes
+----------
+
+A PhotonArray instance has the following attributes, each of which is a numpy array:
+
+- x,y           the incidence positions at the top of the detector
+- flux          the flux of the photons
+- dxdz, dydz    the tangent of the inclination angles in each direction
```
rmandelb: How are the x,y directions defined?  (question applies both to the x,y inputs and the dxdz,dydz inputs) I'm thinking of situations where there is a WCS describing the detector geometry, how is the connection between the WCS and the x,y here made?

tests/test_photon_array.py
```diff
@@ -0,0 +1,183 @@
...
+
+    # Check assignment via numpy [:]
+    photon_array.x[:] = 5
+    photon_array.y[:] = 17
+    photon_array.flux[:] = 23
+    np.testing.assert_array_equal(photon_array.x, 5.)
+    np.testing.assert_array_equal(photon_array.y, 17.)
+    np.testing.assert_array_equal(photon_array.flux, 23.)
+
+    # Check assignment via numpy [:]
```
rmandelb: comment wrong?

galsim/gsobject.py
rmjarvis: I was hoping that the constructors for these operators would be able to get any necessary information such as where in the focal plane this postage stamp sits or where the telescope is pointing, etc. outside of drawImage.  Some could be set in the constructor.  And some could be done as setters that would be called before drawImage.  

The config layer has similar hooks for things that need to be done at the start of an image or at the start of a stamp for other kinds of things.  Of course these surface ops aren't in there yet, but I think it will be straightforward to do the same thing for these.  

And hopefully that would cover all use cases, but of course we may need to revisit if something doesn't fit this model.

galsim/photon_array.py
rmjarvis: This anticipates the change Josh is making on #824, where this will be a method of GSObject rather than SBProfile.  So I think the comment is wrong for this branch, but will be correct for the released 1.5.

galsim/gsobject.py
rmandelb: OK.

rmjarvis:
I think I addressed everything here.  And got to 100% coverage too.  :)  Let me know if you have further comments.

rmjarvis:
Ah.  Never mind.  You just approved it.  :)  Thanks.

Issue #834, opened 2016-11-18
drawImage creates an image with all zeros


esheldon:
I'm working on master a6e94d1a782aeb6009f7c1d933106fe358612e75

I've found a situation where drawImage creates a zero image when an interpolated image is convolved with an analytic profile and a full wcs is used.  This does not occur if a simple scale is used.  I've attached a tar ball with code and images that reproduce the error.

[galsim-bug.tar.gz](https://github.com/GalSim-developers/GalSim/files/599783/galsim-bug.tar.gz)

rmjarvis:
Thanks for the report Erin.  I can at least confirm that I get the same behavior.  Also, it was not broken in v1.4 (both paths give similar sums).  

So I'll `git bisect` to figure out what change broke it.  (Probably something I did on #799 if I had to guess.)

rmjarvis:
Turns out the InterpolatedImage stuff was all a red herring.  Here is a stripped down example that also gives the same error:
```
    gal = galsim.Gaussian(sigma=2)
    psf = galsim.Gaussian(sigma=1)
    psf_inv = galsim.Deconvolve(psf)
    newpsf=galsim.Gaussian(sigma=1.1)
    final = galsim.Convolve(gal, psf_inv, newpsf)
    newim = final.drawImage(nx=25,ny=25,
                            wcs=galsim.JacobianWCS(0,0.5,0.5,0),
                            method='no_pixel')
```
I found the problem, which was in the Deconvolve class.  It was iterating over the image wrong when drawing in k-space with non-rectilinear k steps.  And yes, this was my error in doing some of the conversions in #799.

I'm adding a version of your script to the unit tests and should have a PR for the fix shortly.

PR #836, opened 2016-11-18
Fix bug in Deconvolve drawK with non-rectilinear steps

rmjarvis:
This fixes a bug identified by @esheldon in issue #834 that I accidentally introduced in #799.

I added two new unit tests.  The one in test_metacal.py is pretty close to what Erin posted in the bug report.  The other (in test_compound.py) is a bit closer to the proximate source of the error and tests that a transform commutes with a deconvolve.  (Other than the flux scaling, which isn't expected to commute.)

### DIFF
--- a/src/SBDeconvolve.cpp
+++ b/src/SBDeconvolve.cpp
@@ -127,10 +127,10 @@ namespace galsim {
         int skip = im.getNSkip();
         assert(im.getStep() == 1);
 
-        for (int j=0; j<n; ++j,ky0+=dky,ptr+=skip) {
+        for (int j=0; j<n; ++j,kx0+=dkxy,ky0+=dky,ptr+=skip) {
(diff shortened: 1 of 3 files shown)


jmeyers314:
Looks good to me Mike.  My only comment was going to be about the junk*fits files, but I see you already addressed that.

rmjarvis:
Yep.  I did my own code review and noticed them, so I removed them.  :)

rmandelb:
This all looks good to me.

rmjarvis:
@esheldon Could you please try out this branch to make sure it fixes your original error?

esheldon:
Yes i will, probably tomorrow.

esheldon:
This works for me.

rmjarvis:
Great.  Thanks, Erin.

Issue #835, opened 2016-11-18
Add force_stepk/force_maxk/ii_pad_factor kwargs to PhaseScreenPSF and OpticalPSF.


jmeyers314:
`PhaseScreenPSF` currently accepts `_force_stepk` and `_force_maxk` kwargs to forward to `InterpolatedImage`.  `OpticalPSF` is really just a wrapper around `PhaseScreenPSF`, so it should get these kwargs too.

Both `PhaseScreenPSF` and `OpticalPSF` also accept a `pad_factor` kwarg which governs the transformation from pupil space to real space.  It would also be nice to be able to regulate the real space -> Fourier space transformation by forwarding a `pad_factor` kwarg to `InterpolatedImage`.  I *think* these two `pad_factor`s should be distinct, so I'm inclined to call the kwargs that gets forwarded from `PhaseScreenPSF` and `OpticalPSF` to `InterpolatedImage` `ii_pad_factor` or something like that.

PR #837, opened 2016-11-20
#835 force_stepk, force_maxk, and ii_pad_factor

jmeyers314:
This PR adds `_force_stepk` and `_force_maxk` kwargs to `OpticalPSF` by simply forwarding them to `PhaseScreenPSF`, where they were already available.  Additionally, both `PhaseScreenPSF` and `OpticalPSF` get a new kwarg, `ii_pad_factor`, which gets forwarded to the wrapped `InterpolatedImage` as `pad_factor`.  

(Note, we already have a (poorly named) `pad_factor` keyword for `OpticalScreen` and `PhaseScreenPSF`, which adjusts the size of the image to be passed to the `InterpolatedImage` (though it doesn't actually do any padding, it just uses a larger image to reduce FFT aliasing).  We could think about renaming the `PhaseScreenPSF` and `OpticalPSF` `pad_factor` kwarg to something like `bounds_factor`, and then the new `ii_pad_factor` could just be `pad_factor`, but I think it may be hard to safely alert users to this change in kwarg meaning).

There are also a few new tests to cover the new features.

### DIFF
--- a/galsim/phase_psf.py
+++ b/galsim/phase_psf.py
@@ -957,6 +957,9 @@ class PhaseScreenPSF(GSObject):
                                to use.  [default: galsim.Quintic()]
     @param scale_unit          Units to use for the sky coordinates of the output profile.
                                [default: galsim.arcsec]
+    @param ii_pad_factor       Zero-padding factor by which to extend the image of the PSF when
+                               creating the `InterpolatedImage`.  See the `InterpolatedImage`
+                               docstring for more details.  [default: 4.]
     @param suppress_warning    If `pad_factor` is too small, the code will emit a warning telling
                                you its best guess about how high you might want to raise it.
                                However, you can suppress this warning by using
@@ -1005,7 +1008,7 @@ class PhaseScreenPSF(GSObject):
     """
     def __init__(self, screen_list, lam, exptime=0.0, flux=1.0, aper=None,
                  theta=(0.0*galsim.arcmin, 0.0*galsim.arcmin), interpolant=None,
-                 scale_unit=galsim.arcsec, suppress_warning=False, gsparams=None,
+                 scale_unit=galsim.arcsec, suppress_warning=False, ii_pad_factor=4., gsparams=None,
                  _eval_now=True, _bar=None, _force_stepk=None, _force_maxk=None, **kwargs):
         # Hidden `_bar` kwarg can be used with astropy.console.utils.ProgressBar to print out a
         # progress bar during long calculations.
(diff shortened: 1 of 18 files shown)


jmeyers314:
Yay!  100% coverage!  :)

galsim/phase_psf.py
```diff
@@ -1068,15 +1080,18 @@ def __str__(self):
 
     def __repr__(self):
         outstr = ("galsim.PhaseScreenPSF(%r, lam=%r, exptime=%r, flux=%r, aper=%r, theta=%r, " +
-                  "scale_unit=%r, interpolant=%r, gsparams=%r)")
+                  "interpolant=%r, scale_unit=%r, gsparams=%r)")
```
rmandelb: It looks like you've swapped scale unit  and interpolant here, but not in the line below, so they are now inconsistent?

jmeyers314: Ack!  Good catch.  I was attempting to make the repr line up better with the `__init__` arg list, but obviously only half-fixed it.

rmjarvis: Does that mean there's a missing unit test?  `do_pickle(phasescreen_psf)`

rmjarvis:
Other than my concern that the unit tests didn't catch the error in the PhaseScreenPSF repr, I don't have any other comments.

galsim/phase_psf.py
jmeyers314: > Does that mean there's a missing unit test? `do_pickle(phasescreen_psf)`

This took a bit of tinkering to figure out.  We did have some `do_pickle(phasescreen_psf)`s, but none of them triggered an error for two reasons:

1) The existing ones were all for atmospheric phase screens that have an `rng` class member and so the eval(repr()) test gets skipped since we don't expect the BaseDeviates to pass an eval(repr()) roundtrip.
2) When I made a `do_pickle` on an `OpticalScreen` `do_pickle` test, the try/except block designed to catch unevaluatable reprs triggered, skipping the test.

So I rewrote the eval/repr roundtrip test, and got this error to trigger.  Doing that also brought up a bunch of other errors, though, so I went through and fixed all of them.

jmeyers314:
@rmjarvis , @rmandelb , would you mind going through this again since I ended up changing quite a few files (in small ways) to fix the missing unit test?  Thanks.

rmandelb:
Looks like you opened a nice big can of worms with that new test (which means Mike's idea was a good one!).  I just went over the updates and they look good to me.

jmeyers314:
>Looks like you opened a nice big can of worms with that new test

I think I'd call it a medium can of worms.  :)  At any rate, I'm pretty sure the misbehaving try/expect block was my own creation, so this was really just cleaning up my own mess.  Glad it's done now.

Issue #820, opened 2016-10-31
Config with object that `_takes_rng`


esheldon:
I tried to test using a config with `RandomWalk`, which has `_takes_rng=True` set.

I initially created the following configuration
```python
    gal_config = {
        'type':'RandomWalk',
        'npoints':100,
        'half_light_radius':2.0,
    }
    config={
        'gal':gal_config,
    }
```

An exception is raised because the `rng` is missing.  This surprised me, because `rng` is an optional parameter
```python
Traceback (most recent call last):
  File "test_randwalk.py", line 253, in <module>
    test_randwalk_config()
  File "/home/esheldon/git/GalSim/tests/galsim_test_helpers.py", line 456, in f2
    result = f(*args, **kwargs)
  File "test_randwalk.py", line 199, in test_randwalk_config
    rw_from_config = galsim.config.BuildGSObject(config, 'gal')
  File "/home/esheldon/anaconda/lib/python2.7/site-packages/galsim/config/gsobject.py", line 182, in BuildGSObject
    gsobject, safe = _BuildSimple(param, base, ignore, gsparams, logger)
  File "/home/esheldon/anaconda/lib/python2.7/site-packages/galsim/config/gsobject.py", line 256, in _BuildSimple
    raise ValueError("No base['rng'] available for type = %s"%type_name)
ValueError: No base['rng'] available for type = RandomWalk

```

Note `flux` is not being set above, it is optional, yet no error is raised.

This goes away if I instead add the `rng` to the _outer_ config
```python
    gal_config = {
        'type':'RandomWalk',
        'npoints':100,
        'half_light_radius':2.0,
        'flux':np.pi,
    }
    config={
        'gal':gal_config,
        'rng':galsim.BaseDeviate(),
    }
```

So I have two questions:  

1. Should the `rng` be dealt with like other optional parameters?
2. Must the rng always be at the outer config?

rmandelb:
I think that the ```test_config_gsobject.py``` test for ```RealGalaxy```  (another one with ```takes_rng=True```), ```test_realgalaxy```, suggests an answer.  You don't need to provide an ```rng``` for a RealGalaxy/RandomWalk, but config does need to have some ```rng``` set otherwise in the outer config.  Hopefully Mike will comment if I've misunderstood here.

rmjarvis:
Right.  Within the config stuff, we typically want our scripts to be repeatable.  So that means always providing an explicit rng rather than let the object make its own from /dev/urandom.  

[...]

PR #838, opened 2016-11-20
#820 Make it a bit easier to run config functions without going through galsim executable

rmjarvis:
@esheldon reported on issue #820 that when running config functions by hand within python, e.g. `galsim.config.BuildGSObject(config, 'gal')`, it wasn't allowing the rng to be automatic for things that have `_takes_rng = True`.

This is normally not a problem when running config files through the `galsim` executable, since prior to that call, an appropriate rng will have bee constructed.  However, I think it does make sense to allow users to call functions such as this without requiring them to manually make the rng and put it in `config['rng']`.  So that's done here.

However, I also took this opportunity to go through the missing config coverage (which was a lot) and add unit tests to cover the things that weren't being covered by anything else yet.  Along the way, I found and fixed a number of bugs in edge case scenarios and made a few other changes to make it easier to call the config functions manually.  Here is a complete list of changes:
- Setting config['rng'] is no longer required when manually running commands like `galsim.config.BuildGSObject`.
- Changed the name of galsim.config.CalculateNoiseVar to the slightly more verbose, but also more readable CalculateNoiseVariance.
- Allow PoissonNoise without any sky level, in which case only the shot noise in the signal photons contribute to the noise.  Likewise for CCDNoise.
- Let 'None' in the config file mean `None`.
- Require 'max_extra_noise' to be explicitly set for photon shooting if you want it rather than have it default to 0.01.
- Added `--except_abort` option to `galsim` executable to abort execution if a file has an error, rather than just reporting the exception and continuing on (which is still the default behavior).
- Added optional probability parameter 'p' for Random bool values.
- Allow `drawImage` to take a non-uniform WCS when default constructing the image from either `nx`,`ny` or `bounds`.  (Note: this one is not a config change -- it was just a restriction that we had in `drawImage` that isn't actually necessary.)

Also a couple other functionality-neutral changes you'll notice here:
- When people used galsim.utilities for things, there was a mix of using the full `galsim.utilities.blah` vs `from . import utilities` ... `utilities.blah`.  It seemed that the former was more common, so I switched everything to use that style for consistency.
- Josh pointed out during the hack week that `_absdet` in SBTransform` wasn't actually correct.  It's really the total flux scaling.  And there was another parameter called `_fluxScaling`, which is really the amplitude scaling.  So I renamed these `_fluxScaling` and `_ampScaling` respectively to be clearer.  And I kept `_absdet` to actually mean |det(m)|.
- I also made SBConvolve and SBTransform delay some calculations until they are actually needed.  This won't usually be a performance change, since you usually do need them eventually when you draw the object, but for my `test_reject` function that builds a lot of objects and never gets around to drawing them, it dropped the running time from 7.5 seconds to a little over 2 seconds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,3 +118,6 @@ New config features
 - Let 'None' in the config file mean `None`. (#820)
 - Require 'max_extra_noise' to be explicitly set for photon shooting if you
   want it rather than have it default to 0.01.  (#820)
+- Add --except_abort option to galsim executable to abort execution if a file
+  has an error, rather than just reporting the exception and continuing on
+  (which is still the default behavior). (#820)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,7 +90,7 @@ New Features
   doing so are `im.calculate_fft()` and `im.calculate_inverse_fft()`.  There
   is also `im.wrap()` which can be used to wrap an image prior to doing the
   FFT to properly alias the data if necessary. (#799)
-- Added new profile `galsim.RandomWalk`, a class for generating a set of 
+- Added new profile `galsim.RandomWalk`, a class for generating a set of
   point sources distributed using a random walk.  Uses of this profile include
   representing an "irregular" galaxy, or adding this profile to an Exponential
   to represent knots of star formation. (#819)
@@ -98,9 +98,17 @@ New Features
   image from either nx,ny or bounds. (#820)
 - Added 'generate' function to BaseDeviate and 'sed.sampleWavelength' to draw
   random wavelengths from an SED. (#822)
+- Added function assignPhotonAngles to add arrival directions (in the form of
+  dx/dz and dy/dz slopes) to an existing photon array. (#823)
 - Added `surface_ops` option to `drawImage` function, which applies a list of
   surface operations to the photon array before accumulating on the image.
   (#827)
+- Added `ii_pad_factor` kwarg to PhaseScreenPSF and OpticalPSF to control the
+  zero-padding of the underlying InterpolatedImage. (#835)
+- Added galsim.fft module that includes functions that act as drop-in
+  replacements for np.fft functions, but using the C-layer FFTW package.
+  Our functions have more restrictions on the input arrays, but when valid
+  are generally somewhat faster than the numpy functions. (#840)
 
 
 New config features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,6 +118,7 @@ New config features
 - Let 'None' in the config file mean `None`. (#820)
 - Require 'max_extra_noise' to be explicitly set for photon shooting if you
   want it rather than have it default to 0.01.  (#820)
-- Add --except_abort option to galsim executable to abort execution if a file
+- Added --except_abort option to galsim executable to abort execution if a file
   has an error, rather than just reporting the exception and continuing on
   (which is still the default behavior). (#820)
+- Added optional probability parameter 'p' for Random bool values. (#820)
--- a/examples/demo7.yaml
+++ b/examples/demo7.yaml
@@ -197,7 +197,7 @@ stamp :
     # The max_extra_noise parameter indicates how much extra noise per pixel we are
     # willing to tolerate as a fraction of the noise that we will be adding from the
     # sky (cf. the image.noise field).  In this case, we allow an extra 1% larger variance,
-    # but only for photon
+    # but only for photon shooting.  We use None here when the draw_method is fft.
     max_extra_noise :
         type: List
         items: [ None, 0.01 ]
--- a/galsim/config/extra.py
+++ b/galsim/config/extra.py
@@ -220,7 +220,7 @@ def WriteExtraOutputs(config, main_data, logger=None):
         # Do any final processing that needs to happen.
         builder.ensureFinalized(field, config, main_data, logger)
 
-        # Call the write function, possible multiple times to account for IO failures.
+        # Call the write function, possibly multiple times to account for IO failures.
         write_func = builder.writeFile
         args = (file_name,field,config,logger)
         galsim.config.RetryIO(write_func, args, ntries, file_name, logger)
--- a/galsim/config/extra.py
+++ b/galsim/config/extra.py
@@ -190,7 +190,7 @@ def WriteExtraOutputs(config, main_data, logger=None):
         if 'file_name' in field:
             galsim.config.SetDefaultExt(field, '.fits')
             file_name = galsim.config.ParseValue(field,'file_name',config,str)[0]
-        else:
+        else: # pragma: no cover
             # If no file_name, then probably writing to hdu
             continue
         if 'dir' in field:
@@ -251,7 +251,7 @@ def AddExtraOutputHDUs(config, main_data, logger=None):
         field = output[key]
         if 'hdu' in field:
             hdu = galsim.config.ParseValue(field,'hdu',config,int)[0]
-        else:
+        else: # pragma: no cover
             # If no hdu, then probably writing to file
             continue
         if hdu <= 0 or hdu in hdus:
--- a/galsim/config/extra.py
+++ b/galsim/config/extra.py
@@ -268,7 +268,7 @@ def AddExtraOutputHDUs(config, main_data, logger=None):
     first = len(main_data)
     for h in range(first,len(hdus)+first):
         if h not in hdus:
-            raise ValueError("Cannot skip hdus.  Not output found for hdu %d"%h)
+            raise ValueError("Cannot skip hdus.  No output found for hdu %d"%h)
     # Turn hdus into a list (in order)
     hdulist = [ hdus[k] for k in range(first,len(hdus)+first) ]
     return main_data + hdulist
--- a/galsim/config/extra.py
+++ b/galsim/config/extra.py
@@ -201,6 +201,8 @@ def WriteExtraOutputs(config, main_data, logger=None):
         if dir is not None:
             file_name = os.path.join(dir,file_name)
 
+        galsim.config.EnsureDir(file_name)
+
         if noclobber and os.path.isfile(file_name):  # pragma: no cover
             if logger:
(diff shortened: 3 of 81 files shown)


rmandelb:
I'm getting some output from the unit tests on my Mac:
```
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/multiprocessing/process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/site-packages/galsim/config/process.py", line 822, in worker
    ps = pstats.Stats(pr, stream=s).sort_stats(sortby).reverse_order()
  File "/Library/Frameworks/EPD64.framework/Versions/7.2/lib/python2.7/pstats.py", line 233, in sort_stats
    sort_tuple = sort_tuple + sort_arg_defs[word][0]
KeyError: 'tottime'
Process Process-4:
```

and on the Linux cluster, something completely different:
```
ERROR: Test the options in config to whiten images
  File "/opt/python27/lib/python2.7/site-packages/nose-1.3.0-py2.7.egg/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/rmandelb/git/GalSim/tests/galsim_test_helpers.py", line 559, in f2
    result = f(*args, **kwargs)
  File "/home/rmandelb/git/GalSim/tests/test_config_noise.py", line 661, in test_whiten
    im3d, cv3d = galsim.config.BuildStamp(config2)
  File "/home/rmandelb/software/lib/python2.7/site-packages/galsim/config/stamp.py", line 405, in BuildStamp
    im, current_var = builder.addNoise(stamp,config,im,skip,current_var,logger)
  File "/home/rmandelb/software/lib/python2.7/site-packages/galsim/config/stamp.py", line 876, in addNoise
    current_var = galsim.config.AddNoise(base,image,current_var,logger)
  File "/home/rmandelb/software/lib/python2.7/site-packages/galsim/config/noise.py", line 88, in AddNoise
    var = builder.addNoise(noise, config, im, rng, current_var, draw_method, logger)
  File "/home/rmandelb/software/lib/python2.7/site-packages/galsim/config/noise.py", line 326, in addNoise
    "Whitening already added more noise than the requested Poisson noise.")
RuntimeError: Whitening already added more noise than the requested Poisson noise.
From whiten, current_var =  9.81378306774e-05
From BuildStamp, current_var =  9.81378306774e-05
```

rmjarvis:
> RuntimeError: Whitening already added more noise than the requested Poisson noise.

Travis found that one too, so I was already working on it.  I'm not actually sure what the error was, but when I cleaned up that bit of code slightly to not use quite so many temporary numpy arrays, the error went away.  Weird.  cf.1b6060d

> KeyError: 'tottime'

This seems to indicate that you have an old version of pstats.  AFAICT, the name for this used to be 'time' and they switched it to 'tottime', but 'time' is still a valid alias.  So assuming this alias will remain valid going forward, I just switched to that.  cf. 7f7d909c5c

rmandelb:
I confirm these are both fixed, thanks.  I plan to go over the code sometime before Thanksgiving (but likely not until Weds or so).

tests/test_config_value.py
```diff
@@ -980,6 +1100,78 @@ def test_shear_value():
...
+                                 galsim.Shear)
+        config['world_pos'] = galsim.PositionD(6,8)
+        # Still raise an error because no redshift
+        np.testing.assert_raises(ValueError, galsim.config.ParseValue, config, 'nfw', config,
+                                 galsim.Shear)
+        # With this, it should work.
+        config['gal'] = { 'redshift' : 1.3 }
+    except ImportError:
+        pass
+    nfw_halo = galsim.NFWHalo(mass=1.e14, conc=4, redshift=0.3)
```
rmandelb: Sorry to pick on this random example, but there are lots of magic numbers that need to be written identically between the config and the direct cases, so this is another test routine that might benefit from defining them as variables.  This is true in quite a few of the new tests.

CHANGELOG.md
```diff
@@ -106,3 +108,17 @@ New config features
 
 - Output slightly more information about the COSMOSCatalog() (if any) being used
   as the basis of simulations, at the default verbosity level. (#804)
+- Changed the name of galsim.config.CalculateNoiseVar to the slightly more
+  verbose, but also more readable CalculateNoiseVariance. (#820)
+- Setting config['rng'] is no longer required when manually running commands
+  like `galsim.config.BuildGSObject`.  If you do not care about deterministic
+  pseudo-rngs, then it will just use /dev/urandom for you. (#820)
```
rmjarvis: Sure.  I can add that.  Although, it would require that you are running things with a logger, since that's where config warnings are emitted, and I might guess that most people who neglect to specify the rng might also neglect to pass in a logger.

CHANGELOG.md
```diff
@@ -106,3 +108,17 @@ New config features
...
+- Allow PoissonNoise without any sky level, in which case only the shot noise
+  in the signal photons contribute to the noise.  Likewise for CCDNoise. (#820)
```
rmjarvis: I agree, and that's why it wasn't there before.  But it seemed like a silly restriction to not allow using PoissonNoise without a sky.  It seems perfectly legitimate to investigate galaxies where the only noise is from their own photons.

galsim/config/noise.py
```diff
@@ -506,9 +490,9 @@ def addNoiseVariance(self, config, base, im, include_obj_var, logger):
             # The current image at this point should be the noise-free, sky-free image,
             # which is the object variance in each pixel.
             if gain != 1.0:
-                im += base['current_image']/gain
+                im += base['current_noise_image']/gain
```
rmjarvis: It's just an efficiency.  Dividing by 1 is not a trivial operation on a image.  So it's better to check first and skip the calculation if it won't actually do anything.

tests/test_config_image.py
```diff
@@ -31,9 +33,612 @@
...
+    im4a = gal.drawImage(scale=1, method='phot', max_extra_noise=5, rng=ud, poisson_flux=True)
+    im4a.addNoise(galsim.GaussianNoise(sigma=math.sqrt(50), rng=ud))
+    im4b = galsim.config.BuildImage(config)
+    np.testing.assert_array_equal(im4b.array, im4a.array)
+
+@timer
+def test_reject():
+    """Test various ways that objects can be rejected.
+    """
+    # Make a custom function for rejecting COSMOSCatalog objects that use Sersics with n > 2.
```
rmjarvis: Do you think this is more helpful than the example in cgc.yaml?

rmandelb: I don't think it's _more_ helpful.  I just think it's another helpful thing.  They showcase different functionality (i.e., the one in cgc.yaml is based on properties of the drawn image whereas the one in this test lets you really dig into the COSMOSCatalog, and I don't find that trivial/obvious, personally).

galsim/config/noise.py
rmandelb: OK, I wouldn't have guessed that this mattered.

galsim/config/extra.py
```diff
@@ -164,71 +162,70 @@ def WriteExtraOutputs(config, main_data, logger=None):
...
     config['index_key'] = 'file_num'
-    if 'output' in config:
-        output = config['output']
-        if 'retry_io' in output:
-            ntries = galsim.config.ParseValue(config['output'],'retry_io',config,int)[0]
-            # This is how many retries.  Do at least 1, so ntries is 1 more than this.
-            ntries = ntries + 1
-        else:
-            ntries = 1
+    output = config['output']
```
rmandelb: Cool.

CHANGELOG.md
rmandelb: Oh.  Well, that's not terribly satisfying.

examples/demo7.yaml
```diff
@@ -191,6 +191,17 @@ stamp :
...
+    # For bright objects, it can be quite slow to shoot all the photons from the object.
+    # Shooting fewer photons with flux > 1 each will be faster, but will add some extra
+    # Poisson noise due to the smaller number of photons.
+    # The max_extra_noise parameter indicates how much extra noise per pixel we are
+    # willing to tolerate as a fraction of the noise that we will be adding from the
+    # sky (cf. the image.noise field).  In this case, we allow an extra 1% larger variance,
+    # but only for photon
+    max_extra_noise :
+        type: List
+        items: [ None, 0.01 ]
```
rmjarvis: That was what I was starting to write in that sentence that I didn't finish and apparently never came back to.

galsim/config/noise.py
rmjarvis: It likely would be a very minor inefficiency, you're right.  But these little things can add up.  :)

tests/test_config_image.py
rmjarvis: I added something [here](https://github.com/GalSim-developers/GalSim/wiki/Config-Stamp#stamp-types).  The link is to the master version, which doesn't actually have this code yet, but it will once this is merged.

tests/test_config_value.py
rmjarvis: OK. I did these two, but I didn't try to comprehensively go through the tests to look for examples where doing this would be helpful.  If you notice any others that you think are egregious, let me know.  (Personally, I don't much mind magic numbers in tests -- in the main code is where I try to avoid them.)

rmjarvis:

rmjarvis:
Any objection to merging this?

jmeyers314:
>Any objection to merging this?

I bet you were waiting to hear from me on this...  Let me give it a quick once-over.

rmjarvis:
:)  Yep.  Thanks.

jmeyers314:
It was a quick scan, but looks good to me.  If nothing else, I learned of the existence of `pragma: no branch` and `RealGalaxyOriginal` :).

rmjarvis:
Great.  Thanks, Josh.

Issue #833, opened 2016-11-17
FindAdaptiveMoments output depends on ordering of input image array


jmeyers314:
While comparing galsim optical PSFs to Aaron Roodman's donutlib optical PSFs, I noticed that visually consistent images were giving inconsistent HSM measurements.  I believe the problem is related to the ordering (FORTRAN vs. C) of the numpy arrays wrapped by galsim.Image.  For example:

```python
import numpy as np
import galsim

gal = galsim.Gaussian(sigma=0.5).shear(g1=0.1)
img = gal.drawImage(nx=32, ny=32, scale=0.2)
print(galsim.hsm.FindAdaptiveMom(img).observed_shape)
# galsim.Shear(g1=0.0986737815706,g2=-7.749890904e-18)

# Transpose the image, which should just flip the sign of g1.
# Note, though, that this changes the ndarray from C-ordering to FORTRAN-ordering.
img2 = galsim.Image(img.array.T, scale=img.scale)
print(galsim.hsm.FindAdaptiveMom(img2).observed_shape)
# galsim.Shear(g1=-3.65148834574e-16,g2=-0.609263416273)  #  huh?

# Restore C-ordering without changing the logical contents of the ndarray
img3 = galsim.Image(np.ascontiguousarray(img2.array), scale=img2.scale)
print(galsim.hsm.FindAdaptiveMom(img3).observed_shape)
# galsim.Shear(g1=-0.0986737815706,g2=-3.6084271443e-18)  # That's better

# Verify that FORTRAN-ordering was inducing the problem.
img4 = galsim.Image(np.asfortranarray(img3.array), scale=img3.scale)
print(galsim.hsm.FindAdaptiveMom(img4).observed_shape)
# galsim.Shear(g1=-3.65148834574e-16,g2=-0.609263416273)
```

This may not be a problem for images created by drawImage(), but clearly can become a problem for images created directly from an ndarray.

rmjarvis:
I suspect that's new from my changes in #799.  I think before, if the array wasn't C-contiguous, we recast it (ie. allocated new memory) as a C-contiguous array.  In #799, I added code to allow the Image class to work with non-unit stepped arrays (mostly so cimage.real and cimage.imag didn't require copying), but probably there is something in the HSM code that implicitly assumes C-contiguousness.

So I'll take a look and try to fix it.  The above is a good start for a unit test for this.

PR #839, opened 2016-11-21
#833 Fix uses of Image when underlying array is not C-contiguous

rmjarvis:
In #799, I updated the Image class to work with arbitrary steps between the values in both directions, i.e. not require the underlying array to be C-contiguous.  However, I missed some uses that depended on the C-contiguousness of the array.  Since that's usually what we have, it didn't trigger any errors in the tests.

Anyway, Josh ran across one, which he reported in #833.  I fixed that and added unit tests that would have caught the problem.  There were also problems with the addNoise implementations, so I fixed that and added tests there as well.

SBInterpolatedImage serialize also used the error-prone iterators, although it always uses a C-contiguous image, so it wasn't ever a problem.  But I switched it anyway and removed the Image iterator methods.

### DIFF
--- a/tests/test_hsm.py
+++ b/tests/test_hsm.py
@@ -803,24 +803,30 @@ def test_noncontiguous():
     img = gal.drawImage(nx=64, ny=64, scale=0.2)
     meas_shape1 = galsim.hsm.FindAdaptiveMom(img).observed_shape
     print(meas_shape1)
-    np.testing.assert_almost_equal(meas_shape1.g1, 0.1, decimal=3)
-    np.testing.assert_almost_equal(meas_shape1.g2, 0.2, decimal=3)
+    np.testing.assert_almost_equal(meas_shape1.g1, 0.1, decimal=3,
+                                   err_msg="HSM measured wrong shear on normal image")
+    np.testing.assert_almost_equal(meas_shape1.g2, 0.2, decimal=3,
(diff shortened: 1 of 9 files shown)


tests/test_hsm.py
```diff
@@ -791,6 +791,37 @@ def test_ksb_sig():
...
+
+    This test was inspired by a bug report in issue #833.
+    """
+    gal = galsim.Gaussian(sigma=1.3).shear(g1=0.1, g2=0.2)
+
+    # First a normal C-contiguous image as build by drawImage.
+    img = gal.drawImage(nx=64, ny=64, scale=0.2)
+    meas_shape1 = galsim.hsm.FindAdaptiveMom(img).observed_shape
+    print(meas_shape1)
+    np.testing.assert_almost_equal(meas_shape1.g1, 0.1, decimal=3)
```
rmandelb: Put a useful ```err_msg``` on all of these?

rmjarvis:
> this isn't in CHANGELOG, but I assume that was by design since the updates are just fixing bugs that happened after the last tagged release.

Yes.  This wasn't broken in 1.4.  So no need to count it as a bug fix as far as the CHANGELOG is concerned.

tests/test_hsm.py
rmandelb: Thanks.

jmeyers314:
@rmjarvis , this one looks good to me too.

PR #840, opened 2016-11-24
Optimizations for OpticalPSF code

rmjarvis:
This is basically all speed-related optimizations.  

Josh was comparing the timing of @aaronroodman's donutlib code to the equivalent functionality in GalSim's phase_psf and found that GalSim was significantly slower.  He made some changes in #835 to try to get the code to at least do the same thing in terms of padding and such.  But even with an apples-to-apples comparison, GalSim was still more than a factor of 2 slower.

Anyway, he and I have been working on some optimizations.  I think we're probably done so here they are.  Some highlights:

- Caching some of the calculations related to Zernikes that are independent of the wavefront and just a function of jmax.
- Faster horner implementation that beats np.polyval for the typical inputs we have in the phase_screen code by more than a factor of 3.
- Use `np.dot` for doing the sum of aberration coefficients times their corresponding wavefronts as a matrix-vector product to avoid the explicit python loop.
- New galsim.fft package that has some functions which can serve as drop in replacements for the same-named np.fft functions, although with some more restrictive requirements on the input array and fewer options.  e.g. `np.fft.fft2(a)` -> `galsim.fft.fft2(a)` is equivalent, but typically ~20-30% faster.
- Added a number of leading-underscore functions that avoid the python kruft in some of our regular functions or class constructors.  The most significant of these is `_Image(array, bounds, wcs)` which only accepts those arguments, not the panoply of options the regular constructor accepts, and doesn't do any sanity checks on the inputs.  But since we do that combination pretty often, and we already know the inputs are sane, it's a nice time savings.  (This is a standard trick that Erin uses a lot in his code.)
- Lots of micro-optimizations like using `max` rather than `np.max` for scalars, avoiding copies in `astype()` when possible, etc.

The upshot is that the code is now basically as fast as donutlib.  And some of this effort has of course impacted other uses as well.  On my laptop, the test suites are now about 5% faster, so that's more than just the OpticalPSF stuff being sped up.

Note: I only inserted the galsim.fft functions into the phase psf code.  The PowerSpectrum and CorrelatedNoise classes are other purveyors of np.fft.  However, they currently allow odd-sized inputs, which I didn't implement for galsim.fft.  I may get around to updating the code to allow that (there are annoying bookkeeping issues to deal with), but a better solution would likely be to pad out the fft arrays to a "good fft size".  However, I suspect there are subtleties to that, which I didn't feel like tackling, so I put them off to a later issue.

### DIFF
--- a/galsim/fft.py
+++ b/galsim/fft.py
@@ -97,14 +97,14 @@ def fft2(a, shift_in=False, shift_out=False):
             # This only returns kx >= 0.  Fill out the full image.
             kar = np.empty( (M,N), dtype=np.complex128)
             if shift_out:
-                kar[:,N/2:N] = rkar[:,0:N/2]
-                kar[0,0:N/2] = rkar[0,N/2:0:-1].conjugate()
-                kar[1:M/2,0:N/2] = rkar[M-1:M/2:-1,N/2:0:-1].conjugate()
-                kar[M/2:M,0:N/2] = rkar[M/2:0:-1,N/2:0:-1].conjugate()
+                kar[:,No2:N] = rkar[:,0:No2]
+                kar[0,0:No2] = rkar[0,No2:0:-1].conjugate()
+                kar[1:Mo2,0:No2] = rkar[M-1:Mo2:-1,No2:0:-1].conjugate()
+                kar[Mo2:M,0:No2] = rkar[Mo2:0:-1,No2:0:-1].conjugate()
             else:
-                kar[:,0:N/2] = rkar[:,0:N/2]
-                kar[0,N/2:N] = rkar[0,N/2:0:-1].conjugate()
-                kar[1:M,N/2:N] = rkar[M-1:0:-1,N/2:0:-1].conjugate()
+                kar[:,0:No2] = rkar[:,0:No2]
+                kar[0,No2:N] = rkar[0,No2:0:-1].conjugate()
+                kar[1:M,No2:N] = rkar[M-1:0:-1,No2:0:-1].conjugate()
     return kar
 
 
--- a/galsim/fft.py
+++ b/galsim/fft.py
@@ -90,23 +90,23 @@ def fft2(a, shift_in=False, shift_out=False):
     else:
         a = a.astype(np.float64, copy=False)
         xim = galsim._galsim.ConstImageViewD(a, -No2, -Mo2)
-        if False:
-            # This works, but it's a bit slower.
-            kar = xim.cfft(shift_in=shift_in, shift_out=shift_out).array
+
+        # This works, but it's a bit slower.
+        #kar = xim.cfft(shift_in=shift_in, shift_out=shift_out).array
+
+        # Faster to start with rfft2 version
+        rkar = xim.rfft(shift_in=shift_in, shift_out=shift_out).array
+        # This only returns kx >= 0.  Fill out the full image.
+        kar = np.empty( (M,N), dtype=np.complex128)
+        if shift_out:
+            kar[:,No2:N] = rkar[:,0:No2]
+            kar[0,0:No2] = rkar[0,No2:0:-1].conjugate()
+            kar[1:Mo2,0:No2] = rkar[M-1:Mo2:-1,No2:0:-1].conjugate()
+            kar[Mo2:M,0:No2] = rkar[Mo2:0:-1,No2:0:-1].conjugate()
         else:
-            # Start with rfft2 version
-            rkar = xim.rfft(shift_in=shift_in, shift_out=shift_out).array
-            # This only returns kx >= 0.  Fill out the full image.
-            kar = np.empty( (M,N), dtype=np.complex128)
-            if shift_out:
-                kar[:,No2:N] = rkar[:,0:No2]
-                kar[0,0:No2] = rkar[0,No2:0:-1].conjugate()
-                kar[1:Mo2,0:No2] = rkar[M-1:Mo2:-1,No2:0:-1].conjugate()
-                kar[Mo2:M,0:No2] = rkar[Mo2:0:-1,No2:0:-1].conjugate()
-            else:
-                kar[:,0:No2] = rkar[:,0:No2]
-                kar[0,No2:N] = rkar[0,No2:0:-1].conjugate()
-                kar[1:M,No2:N] = rkar[M-1:0:-1,No2:0:-1].conjugate()
+            kar[:,0:No2] = rkar[:,0:No2]
+            kar[0,No2:N] = rkar[0,No2:0:-1].conjugate()
+            kar[1:M,No2:N] = rkar[M-1:0:-1,No2:0:-1].conjugate()
     return kar
 
 
--- a/galsim/fft.py
+++ b/galsim/fft.py
@@ -82,10 +82,10 @@ def fft2(a, shift_in=False, shift_out=False):
         raise ValueError("Input array must have even sizes")
 
     if a.dtype.kind == 'c':
-        a = a.astype(np.complex128)
+        a = a.astype(np.complex128, copy=False)
         xim = galsim._galsim.ConstImageViewC(a, -No2, -Mo2)
     else:
-        a = a.astype(np.float64)
+        a = a.astype(np.float64, copy=False)
         xim = galsim._galsim.ConstImageViewD(a, -No2, -Mo2)
     return xim.cfft(shift_in=shift_in, shift_out=shift_out).array
 
@@ -142,10 +142,10 @@ def ifft2(a, shift_in=False, shift_out=False):
         raise ValueError("Input array must have even sizes")
 
     if a.dtype.kind == 'c':
-        a = a.astype(np.complex128)
+        a = a.astype(np.complex128, copy=False)
         xim = galsim._galsim.ConstImageViewC(a, -No2, -Mo2)
     else:
-        a = a.astype(np.float64)
+        a = a.astype(np.float64, copy=False)
         xim = galsim._galsim.ConstImageViewD(a, -No2, -Mo2)
     return xim.cfft(inverse=True, shift_in=shift_in, shift_out=shift_out).array
 
@@ -194,7 +194,7 @@ def rfft2(a, shift_in=False, shift_out=False):
     if M != Mo2*2 or N != No2*2:
         raise ValueError("Input array must have even sizes")
 
-    a = a.astype(np.float64)
+    a = a.astype(np.float64, copy=False)
     xim = galsim._galsim.ConstImageViewD(a, -No2, -Mo2)
     return xim.rfft(shift_in=shift_in, shift_out=shift_out).array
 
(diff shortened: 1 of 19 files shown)


rmandelb:
It looks like the majority of the improvements are in the code for aberrations, right?  The reason I ask is that i merged this into #556 and started playing with it on there, but the calculations I was doing without aberrations did not seem much faster, and then looking at the code I think that might make sense... but I wanted to make sure I wasn't doing something wrong...

galsim/fft.py
```diff
@@ -0,0 +1,267 @@
...
+        >>> ka1 = numpy.fft.fft2(numpy.fft.fftshift(a))
+        >>> ka2 = galsim.fft.fft2(a, shift_in=True)
+
+    If shift_out is True, then this is equivalent to applying numpy.fft.fftshift to the output.
+
+        >>> ka1 = numpy.fft.fftshift(numpy.fft.fft2(a))
+        >>> ka2 = galsim.fft.fft2(a, shift_out=True)
+
+    @param a            The input array to be transformed
+    @param shift_in     Whether to shift the input array so that the center is moved to (0,0).
```
rmandelb: In this routine and all others, need to mention the default value of these kwargs in the docstring.

galsim/fft.py
```diff
@@ -0,0 +1,267 @@
...
+    @param shift_out    Whether to shift the output array so that the center is moved to (0,0).
+
+    @returns a complex numpy array
+    """
+    s = a.shape
+    if len(s) != 2:
+        raise ValueError("Input array must be 2D.")
```
rmandelb: For this ValueError and the one below (+ in the other routines in this file), it would be valuable to give a little more diagnostic info -- like what is ```len(s)```, or (for the one below), what is ```(M, N)```.

galsim/gsobject.py
```diff
@@ -674,7 +674,12 @@ def dilate(self, scale):
 
         @returns the dilated object.
         """
-        return self.expand(scale) * (1./scale**2)  # conserve flux
+        # equivalent to self.expand(scale) * (1./scale**2)
+        new_obj = galsim.Transform(self, jac=[scale, 0., 0., scale], flux_ratio=scale**-2)
+
+        if hasattr(self, 'noise'):
```
rmandelb: Were we just kind of missing this all along?  Ugh.  If so, is there a unit test that would have caught this that we could add?

galsim/phase_screens.py
```diff
@@ -448,11 +455,18 @@ def _zern_rho_coefs(n, m):
...
-
-def _zern_coef_array(n, m, eps=0., shape=None, annular=False):
+def _zern_coef_array(n, m, eps, shape, annular):
     """Assemble coefficient array for evaluating Zernike (n, m) as the real part of a
     bivariate polynomial in abs(rho)^2 and rho, where rho is a complex array indicating position on
     a unit disc.
+
+    @param n        Zernike radial coefficient
+    @param m        Zernike azimuthal coefficient
+    @param eps      Linear obscuration fraction.
```
rmandelb: Any reason not to call this ```obscuration``` like we do elsewhere?  (question applies to routines later on in this file too)

galsim/phase_screens.py
```diff
@@ -468,6 +482,30 @@ def _zern_coef_array(n, m, eps=0., shape=None, annular=False):
...
     return out
 
+def __noll_coef_array(jmax, eps, annular):
+    """Assemble coefficient array for evaluating Zernike (n, m) as the real part of a
+    bivariate polynomial in abs(rho)^2 and rho, where rho is a complex array indicating position on
+    a unit disc.
+
+    @param jmax     Maximum Noll coefficient
+    @param eps      Linear obscuration fraction.
+    @param shape    Output array shape
```
rmandelb: Shape doesn't seem to actually be a parameter of this routine?

galsim/lensing_ps.py
```diff
@@ -1617,7 +1617,7 @@ def __call__(self, gd):
         if self.amplitude_E is not None:
             r1 = galsim.utilities.rand_arr(self.amplitude_E.shape, gd)
             r2 = galsim.utilities.rand_arr(self.amplitude_E.shape, gd)
-            E_k = np.empty((self.ny,self.nx)).astype(type(1.+1.j))
+            E_k = np.empty((self.ny,self.nx), dtype=type(1.+1.j))
```
rmandelb: What is the value in doing it like this as opposed to just ```dtype=np.complex```?

rmjarvis:
The bulk of this was definitely related to the aberrations.  The other improvements are more at the margins, dropping a few percent here and there at most.  Some of them were in tight loops, so they could matter more, but I doubt any of them would cause a big noticeable difference without actually timing things.

galsim/table.py
```diff
@@ -195,20 +195,20 @@ def __call__(self, x):
             if dimen > 2:
                 raise ValueError("Arrays with dimension larger than 2 not allowed!")
             elif dimen == 2:
-                f = np.empty_like(x.ravel()).astype(float)
-                self.table.interpMany(x.astype(float).ravel(),f)
+                f = np.empty_like(x.ravel(), dtype=float)
```
rmandelb: Whenever you do these ```empty_like``` (or ```zeros_like``` or ```ones_like```) I thought the rule is that it gives it the type of the argument.  Is that right?  If so, then given that you've already enforced ```x``` and ```f``` to have type float, perhaps all these ```dtype``` kwargs (here and later) are unnecessary?

I just did a quick check and found that having the ```dtype``` kwarg seemed to increase the time by 60%, so if it's unnecessary I imagine there are a bunch of places throughout the code where it could helpfully be removed.  (If you want I can post a code snippet...)

rmjarvis:
To add more context, the first three items in my list above were definitely the most important speed ups.  They together get around a factor of two, but they are all only relevant for aberrations.  

The fft switch was the next most important, and that was 20-30% improvement for that single step, but since that's a smallish fraction of the total, it still only nets maybe 5-10% percent speedup on the overall calculation.

Everything else together gets another 5-10%.  That was when I decided I was too far along the path of diminishing returns, and stopped working on it.

tests/test_draw.py
```diff
@@ -1114,6 +1114,154 @@ def test_fft():
             im2_real.array, im2_alt_real.array, 9,
             "inverse_fft produce a different array than obj2.drawImage(nx,ny,method='sb')")
 
+@timer
+def test_np_fft():
```
rmandelb: Looks like this new test doesn't get added to the list in the ```if __name__ == "__main__":``` at the bottom.

galsim/table.py
rmandelb: *And by "a code snippet" I mean code that shows where I got the 60% statement from.

rmandelb:
Thanks for the explanation, Mike.

galsim/fft.py
```diff
@@ -0,0 +1,267 @@
...
+
+    For valid inputs, the result is equivalent to numpy.fft.ifft2(a), but usually faster.
+
+        >>> a1 = numpy.fft.ifft2(ka)
+        >>> a2 = galsim.fft.ifft2(ka)
+
+    Restrictions on this version vs the numpy version:
+
+        - The array must be 2-dimensional.
+        - It must be square.
```
rmjarvis: Thanks.  That was the case for my first pass at these functions, but then I enabled non-square inputs and forgot to update the doc.

galsim/table.py
rmjarvis: Interesting.  Maybe.  (Although I only get about 25% faster.)  But either is much faster than using `astype`.
```
$ python -m timeit -s 'import numpy; a = numpy.random.random((128,128))' 'x = numpy.empty_like(a, dtype=float)'
1000000 loops, best of 3: 0.908 usec per loop
$ python -m timeit -s 'import numpy; a = numpy.random.random((128,128))' 'x = numpy.empty_like(a)'
1000000 loops, best of 3: 0.677 usec per loop
 python -m timeit -s 'import numpy; a = numpy.random.random((128,128))' 'x = numpy.empty_like(a).astype(float)'
100000 loops, best of 3: 7.7 usec per loop
```

galsim/fft.py
```diff
@@ -0,0 +1,267 @@
...
+        raise ValueError("Input array must have even sizes")
+
+    if a.dtype.kind == 'c':
+        a = a.astype(np.complex128, copy=False)
+        xim = galsim._galsim.ConstImageViewC(a, -No2, -Mo2)
+        kar = xim.cfft(shift_in=shift_in, shift_out=shift_out).array
+    else:
+        a = a.astype(np.float64, copy=False)
+        xim = galsim._galsim.ConstImageViewD(a, -No2, -Mo2)
+        if False:
```
rmjarvis: No, I'll remove it.  It was useful for testing to flip between them to see the speed difference.  The rfft2 version is a good 30% or so faster IIRC.

galsim/lensing_ps.py
rmandelb: Sounds good.

Also... I was playing around and noticed that ```empty_like``` seems to be faster than ```empty``` but ```zeros_like``` seems slower than ```zeros```, which kind of confuses me.  Perhaps my test cases are too contrived, or this is system-dependent; is there a reason this would make sense?  If it's right, then that might motivate changing around some of our ```empty``` vs. ```empty_like```, but have you ever seen this?

galsim/table.py
rmandelb: I just updated my test case to use 2d arrays (I was using 1d) and the improvement shrank from 60% to 20% on my system.  So for most realistic use cases, the improvement won't be that much.  But I suspect that there are quite a few places where we could take advantage of this.

galsim/phase_screens.py
rmjarvis: I think the annular Zernike paper calls this parameter epsilon, so at some point in the chain it probably makes sense to switch the name to eps, since it's probably clearer at that point for comparison to the paper formulae.  But I switched it here to obscuration, since that is what we've been calling it for most of these functions.

rmjarvis:

rmandelb:
Thanks, Mike.  One further and minor question (though I've approved the changes in my review since they are nominally done): perhaps we should have some unit tests for whether the inputs to the new FFT routines fulfill the requirements (i.e., some ```assert_raises``` -> ValueError tests)?

galsim/phase_screens.py
jmeyers314: Yep.  Mahajan calls this epsilon, which is why I went with `eps` in the lower-level functions.

I'll have a chance to look through the rest of this (and the other pending PRs) tomorrow...

rmjarvis:
>  some assert_raises -> ValueError tests

Good idea, and indeed they had bugs, since the `s` being printed is a tuple, it was interpreted as two values rather than a single value to convert to a string.  So that's fixed now.

jmeyers314:
@rmjarvis , I fixed one tiny typo.  Other than that, looks good to me!

rmjarvis:
Thanks Josh.

Issue #811, opened 2016-10-10
Add a truncation radius option to InclinedExponential profiles


brgillis:
Following up on #782, it would be useful if we could make the profiles a bit more realistic by including an option for a truncation radius (observations have shown that the disc seems to be truncated at approximately 4.5 times the scale radius).

I think Lance's code (the hankelcode.c file included in the tests as reference) already does this, so hopefully it won't be too hard to implement.

rmandelb:
I think we should use the same conventions for defining truncation radii at the python layer as we use for truncated Sersic profiles, to avoid confusion/inconsistency.  (That�s not to say we can�t take advantage of what Lance has done, just that the bit we expose to the user should define the relevant kwargs in a way that is consistent to how truncation is done elsewhere in GalSim.)

brgillis:
I have some time to look into this now. Unfortunately, it looks like Lance's code doesn't have an easy solution we can reapply here; it just brute-forces the Fourier transform. I haven't been able to figure out any easy or clever way to do this, so I think the only possibility is to go in a similar direction as with the Sersic profile, except with an extra operation in Fourier space to handle the fact of inclination.

This probably means that it would be more efficient to separate this into a different profile. And seeing as the exponential profile is a special case of a Sersic profile, perhaps it could be generalized a bit. My current thought is to make a new profile, called InclinedSersic, which will handle the truncation and inclination of a Sersic profile, and can be used as a truncated, inclined exponential profile if desired.

Before I go ahead with this, I'd like to check if you have any thoughts on this or alternative recommendations, @rmandelb and @rmjarvis?

rmjarvis:
Sure.  Sounds fine to me.

rmjarvis:
Merged via PR #843

PR #843, opened 2016-12-12
#811 - InclinedSersic profile

brgillis:
Finished setting up the InclinedSersic profile (which is a generalization of InclinedExponential, allowing for n != 1 and a truncation radius). Some tests seem to be failing, but those all seem to be failing on master as well.


rmandelb:
According to Travis, all tests are currently passing on the master branch:
https://travis-ci.org/GalSim-developers/GalSim/branches

So the fact that checks are failing on this branch indicates an issue on this branch.  At least one of the test failures is because of an error like this:
```
No such file or directory: './inclined_exponential_images/galaxy_1.0_0.0_3.0_1.0_20_0.0.fits'
```

Is it possible that you did not add one of the test images to version control?  I didn't see that among the list of new files, unless I missed it.

brgillis:
Added that file in, and waiting on Travis now.

The tests were definitely failing for me on master on my own machine... I'll try again with a fresh install to be sure, and raise an issue if it persists.

rmandelb:
OK, yes, please do check those unit test failures on master.  In general we like the tests to be passing for everyone on master (though we don't check on tons of systems until it's time to tag a new version - so things can slip through the cracks until then).

brgillis:
Alright, Travis says all checks are passing here, so I think this branch should be good to go.

I found that the tests are failing on master even when I test on a different machine, so I'll open an issue about it. The first couple I looked at seemed to be failures to match known results, which wouldn't be surprising to be system-dependent.

brgillis:
Any other comments on this? If not, I'll go ahead and merge.

rmandelb:
Nobody has reviewed the code.  You need two code reviews.

brgillis:
Ah, okay, thanks.

tests/test_inclined_exponential.py
```diff
@@ -126,211 +281,223 @@ def test_edge_on():
...
+    for mode in ("InclinedExponential","InclinedSersic"):
+
+        # First test that some different initializations that should be equivalent:
+        same_gals = [get_prof(mode,0.1*galsim.radians, 3.0),
+                get_prof(mode,0.1*galsim.radians, 3.0, 0.3),  # default h/r = 0.1
+                get_prof(mode,0.1*galsim.radians, 3.0, scale_height=0.3),
+                get_prof(mode,0.1*galsim.radians, 3.0, scale_h_over_r=0.1),
+                get_prof(mode,0.1*galsim.radians, 3.0, flux=1.0),  # default flux=1
+                get_prof(mode,-0.1*galsim.radians, 3.0),  # negative i is equivalent
+                get_prof(mode,(np.pi--0.1)*galsim.radians, 3.0),  # also pi-theta
```
rmjarvis: Looks like this was an error previously, but this was supposed to be `np.pi-0.1` I think.

tests/test_inclined_exponential.py
```diff
@@ -29,51 +33,155 @@
...
 image_scale_heights = ("0.3", "0.5", "0.5", "0.5", "1.0", "0.5")
 image_pos_angles = ("0.0", "0.0", "0.0", "0.0", "-0.2", "-0.2")
+
+# InclinedSersic-only test cases
+sersic_image_ns = ("1.0","1.5","2.0","1.5","2.5","2.5","1.5","1.5","2.5","2.5")
+sersic_image_inc_angles = ("0.0","0.0","0.0","0.0","0.0","0.0","0.1","0.1","0.1","0.1")
+sersic_image_scale_radii = ("3.0","3.0","3.0","3.0","3.0","3.0","2.0","2.0","2.0","2.0")
+sersic_image_scale_heights = ("1.0","1.0","1.0","1.0","1.0","1.0","1.0","1.0","1.0","1.0")
+sersic_image_trunc_factors = ("20","20","20","4.5","20","4.5","20","4.5","20","4.5")
+sersic_image_pos_angles = ("0.0","0.0","0.0","0.0","0.0","0.0","-0.2","-0.2","-0.2","-0.2")
```
rmjarvis: These values don't have a huge variety in values.  Like all those with `inc_angle` != 0 have the same `inc_angle` and `pos_angle`.  Also, there are no `trunc=0` cases and all the scale heights are equal (and 1 even, which is generally a number to be avoided, since it's easy to miss bugs when 1's are involved).  And finally, I think you use flux=1 for these, which also makes it easy to miss bugs related to the flux normalization.

tests/test_inclined_exponential.py
```diff
@@ -126,211 +281,223 @@ def test_edge_on():
...
+    """ Performs various sanity checks on a set of InclinedExponential and InclinedSersic profiles. """
+    
+    for mode in ("InclinedExponential","InclinedSersic"):
+
+        print('flux, inc_angle, scale_radius, scale_height, pos_angle')
+        for flux, inc_angle, scale_radius, scale_height, pos_angle in zip(fluxes,
+                                                                          image_inc_angles,
+                                                                          image_scale_radii,
+                                                                          image_scale_heights,
+                                                                          image_pos_angles):
```
rmjarvis: Should also test the sersic options for InclinedSersic.  Otherwise, you aren't getting any sanity tests for n!=1, trunc!=0.

tests/test_inclined_exponential.py
```diff
@@ -126,211 +281,223 @@ def test_edge_on():
...
-    test_exponential()
-    test_edge_on()
     test_sanity()
     test_k_limits()
     test_eq_ne()
     test_pickle()
+    test_exponential()
+    test_sersic()
+    test_edge_on()
+    test_regression()
```
rmjarvis: Extremely minor point, but we usually keep these in the same order that they appear in the file.

galsim/inclinedsersic.py
```diff
@@ -0,0 +1,216 @@
...
+    """A class describing an inclined sersic profile. This class is general, and so for certain
+    special cases, more specialized classes will be more efficient. For the case where n==1
+    with no truncation, the InclinedExponential class will be much more efficient. For the case
+    where the inclination angle is zero (face-on), the Sersic class will be slightly more efficient.
+
+    The Inclined Sersic surface brightness profile is characterized by four properties: its
+    Sersic index `n', its inclination angle (where 0 degrees = face-on and 90 degrees = edge-on),
+    its scale radius, and its scale height. The 3D light distribution function is:
+
+        I(R,z) = I_0 / (2h_s) * sech^2 (z/h_s) * exp[-b*(r/r_s)^{1/n}]
```
arunkannawadi: Trivial, but it would be good to use r or R on both sides of the equation here.

galsim/inclinedsersic.py
```diff
@@ -0,0 +1,216 @@
...
+    In addition to the usual GSObject methods, InclinedSersic has the following access methods:
+
+        >>> n = inclined_sersic_obj.getN()
+        >>> inclination = inclined_sersic_obj.getInclination()
+        >>> r0 = inclined_sersic_obj.getScaleRadius()
+        >>> h0 = inclined_sersic_obj.getScaleHeight()
+        >>> hlr = inclined_sersic_obj.getHalfLightRadius()
+    """
+    _req_params = { "inclination" : galsim.Angle, "n" : float }
+    _opt_params = { "scale_height" : float, "scale_h_over_r" : float, "flux" : float }
```
arunkannawadi: Shouldn't only one of `scale_height` and `scale_h_over_r` be specifiable? This currently takes in both without complaining but retains the input `scale_height`. I notice this even with the `InclinedExponential` class which has already been merged on to the master, I guess.

galsim/inclinedsersic.py
```diff
@@ -0,0 +1,216 @@
...
+    _single_params = [ { "scale_radius" : float , "half_light_radius" : float } ]
+    _takes_rng = False
+
+    def __init__(self, n, inclination, half_light_radius=None, scale_radius=None, scale_height=None,
+                 scale_h_over_r=0.1, flux=1., trunc=0., flux_untruncated=False, gsparams=None):
+
+        if scale_height is None:
+            scale_height = scale_h_over_r * scale_radius
```
arunkannawadi: Need to get the `scale_radius` parameter if the object was instantiated with `half_light_radius`.

CHANGELOG.md
```diff
@@ -70,7 +70,7 @@ New Features
 - `SED`s can now be constructed with flexible units via the `astropy.units`
   module. (#789).
 - Added new surface brightness profile, 'InclinedExponential'. This represents
-  the 2D projection of the 3D profile:
+  the 2D projection the 3D profile:
```
arunkannawadi: Why was the word 'of' removed?

galsim/inclinedsersic.py
```diff
@@ -0,0 +1,216 @@
...
+    `scale_radius` or `half_light_radius`.  Exactly one of these two is required. Similarly,
+    at most one of `scale_height' and `scale_h_over_r' is required; if neither is given, the
+    default of scale_h_over_r = 0.1 will be used.
+
+    Initialization
+    --------------
+
+    @param n                The Sersic index, n.
+    @param inclination      The inclination angle, which must be a galsim.Angle instance
+    @param scale_radius     The scale radius of the exponential disk.  Typically given in arcsec.
```
arunkannawadi: This is probably not the best place to comment, but I am a little late to the party. I find it odd that `InclinedExponential` does not take in `half_light_radius` as a parameter where as `InclinedSersic`, `Exponential` and `Sersic` do.

tests/test_inclined_exponential.py
```diff
@@ -29,51 +33,155 @@
...
-        test_profile = galsim.InclinedExponential(inc_angle*galsim.radians, scale_radius,
-                                                  scale_height)
-        check_basic(test_profile, "InclinedExponential")
+        test_profile = galsim.Sersic(n=sersic_n,scale_radius=scale_radius,trunc=trunc_factor*scale_radius)
+        check_basic(test_profile, mode)
 
         # Rotate it by the position angle
         test_profile = test_profile.rotate(pos_angle*galsim.radians)
-
+
```
arunkannawadi: Trailing whitespace

galsim/compound.py
```diff
@@ -322,18 +322,26 @@ def __init__(self, *args, **kwargs):
...
-                msg = """
-                Doing convolution of 2 objects, both with hard edges.
-                This might be more accurate and/or faster using real_space=True"""
-            else:
-                msg = """
-                Doing convolution where all objects have hard edges.
-                There might be some inaccuracies due to ringing in k-space."""
-            warnings.warn(msg)
+            
+            # If any aren't analytic, we don't have a choice here, so suppress the warning
```
brgillis: I could make this change, but it still leaves open the possibility that an issue could arise from a convolution of three or more profiles. Eg., convolving:

A - Not real-space analytic
B - Hard edges, real-space analytic
C - Hard edges, real-space analytic

This isn't likely to arise (it would require eg. a PSF with hard edges), but it's possible. In fact, I think I did something almost like this myself once for other work, except A was real-space analytic so problems didn't arise. But if it hadn't have been, a warning would have been raised no matter what I tried to do. So, I think it is proper to put this check here.

CHANGELOG.md
brgillis: This actually arose because I noticed that "of" was missing, and accidentally fixed it on the master branch instead of this one. I thought I'd reverted that and made the change here, but it appears that didn't take for some reason.

galsim/inclinedsersic.py
brgillis: I'm working from the definition Lance Miller sent me of this type of profile, and he only has sech^2(z/hs). Unless there's some physical motivation to believe it's otherwise, I think it's best to leave it as this. This cuspiness issue would really only come up if it's perfectly edge-on, with a high n (and it looks like Lance never used n higher than 2 in his code), and with good enough resolution. And even then... that might just be how it is in reality. For now, I'm inclined to go with it as-is and trust that Lance has it right for what he wants to use it.

galsim/inclinedsersic.py
brgillis: I've put in a check for this now

galsim/inclinedsersic.py
brgillis: Ah yes. Since we don't get that until the constructor of the Impl, I had to handle this by delaying calculation of scale_height until that point as well.

CHANGELOG.md
brgillis: Per my comment to rmjarvis, this was due to me mistakenly committing a fix of this to the master instead of this branch.

galsim/inclinedsersic.py
brgillis: You're right, I should add that in too. This branch isn't officially about that class though - is it okay to update it here, or should I do it in a new issue/branch?

galsim/inclinedsersic.py
brgillis: Done.

tests/test_inclined_exponential.py
brgillis: Makes sense. I re-sorted them so they'd run in the order of the most fundamental checks first. I can re-sort the actual functions in the file to match this. This will unfortunately make spotted the changes in diffs more difficult, though, so I'll wait until after Rachel has had a chance to review the code before doing so.

galsim/compound.py
rmjarvis: I think you'll find that your proposed scenario would not emit a warning.  Only if all three have hard edges would there be a warning, and that would be one that lets the user know about possible ringing, not telling them they should use real space (since you can't with 3).

My point is really that the hard_edges attribute should imply analytic_x, in which case the current logic in this function is (I believe) correct.

galsim/inclinedsersic.py
rmjarvis: I suppose it's fine as the definition of this class.  I'm just skeptical that it is actually what galaxies look like near the center.  It doesn't seem particularly physical.

brgillis: I think it isn't much an issue for the requested use case as Lance always includes an additional bulge component (a standard Sersic profile) which is cuspy near the centre; this profile isn't intended to represent the whole galaxy, but only the disk component.

galsim/compound.py
brgillis: Just tested it, and you're correct here. I've switched to just setting the profile to state it doesn't have hard edges, with a comment to note it might have to be changed if it's made x-analytic in the future

rmandelb:
@brgillis - I was going to go over this today, but travis is indicating failures now after the latest changes (the code doesn't compile, according to the log which you can access [here](https://travis-ci.org/GalSim-developers/GalSim/jobs/184286092)).  I will wait until that is fixed...

brgillis:
@rmandelb - I fixed the compilation error. I suspect some tests are going to show errors now due to the FFT required being too large (since with n>1, I was using the same scale radius, which results in a much larger half-light radius), so I'll have to adjust them. It's probably still fine if you want to review the code while I do that, or at least the parts outside of the tests.

rmandelb:
Travis claims it still can't compile.

brgillis:
Ah, my bad, looks like I missed committing changes to one file. Hopefully it'll compile now

src/SBInclinedSersic.cpp
```diff
@@ -0,0 +1,470 @@
...
+        }
+    }
+
+    double SBInclinedSersic::SBInclinedSersicImpl::maxK() const
+    {
+        return _maxk * _inv_r0;
+    }
+    double SBInclinedSersic::SBInclinedSersicImpl::stepK() const
+    {
+        double stepk = _info->stepK() * _inv_re;
```
rmjarvis: I'm pretty sure this is wrong.  It should be `_inv_r0`, not `_inv_re`.

brgillis: The problem I was having with _inv_r0 was that it was excluding too much flux outside of 1/stepK (even in the face-on case, when it matches the Sersic profile - it looks like that profile was never tested for this), so it actually needs to be somewhat smaller than what the SersicInfo class is giving. I'm looking into this a bit more now and update soon; it might be that the class's estimate of stepK is what needs to be corrected

brgillis: Update: Yes, when I do that (pushing to illustrate, though not all test galaxies are in yet so the tests will fail), I get this error from the test_k_limits() function:

AssertionError: Too much flux lost due to folding.
Folding threshold = 0.005
Total flux = 1.0
Contained flux = 0.994454
Lost = 0.00554609298706

This is also the case if I use a standard Sersic profile instead, so it seems the issue is with the calculations performed by the SersicInfo class.

rmjarvis: That's because your test for this isn't really valid.  You're only including pixels whose center is < rmax.  This leaves off some of the flux at r<rmax in pixels whose center has r>=rmax.  And you're using big chunky pixels here (scale=1), so it's rather a large effect.

If you use smaller pixels and set your mask to (r + sqrt(2)*scale) < rmax, I predict it will pass.

brgillis: Ah yes, you're correct. I've updated the test to handle this better

src/SBInclinedSersic.cpp
```diff
@@ -0,0 +1,469 @@
...
+        _xnorm = _shootnorm * _inv_r0_sq;
+        dbg<<"norms = "<<_xnorm<<", "<<_shootnorm<<std::endl;
+
+        // For small k, we can use up to quartic in the taylor expansion of both terms
+        // in the calculation.
+        // This is acceptable when the next term is less than kvalue_accuracy.
+        // (35/16 + 31/15120 pi/2*h*sin(i)/r) * (k^2*r^2)^3 = kvalue_accuracy
+        // This is a bit conservative, note, assuming kx = 0
+        _ksq_min = std::pow(this->gsparams->kvalue_accuracy /
+                            (35./16. + 31./15120.*_half_pi_h_sini_over_r), 1./3.);
```
brgillis: Yes, it is. I've put in an adjustment for this now, though it comes with the caveat that it doesn't affect the ksq_min used internally by the SersicInfo class. This isn't much of an issue, as the adjustment needed to it would be quite small (about 1 part in 1000, I believe). It could be fixed, but that would require changes to the SersicInfo class which would mean we'd need to cache it for different half_pi_h_sini_over_r values as well, which is hardly ideal.

One possibility is to just be a little more conservative inside the SersicInfo class and increase its ksq_min slightly to account for this, if it's worth bothering with at all.

rmjarvis: I don't understand this comment at all.  The SersicInfo class isn't doing anything with h.  It's only returning the part of the kValue for the radial direction.  Not anything to do with the perpendicular direction.  So why would you need to change what its ksq_min is?

Indeed, I suspect my initial comment was wrong.  Isn't this ksq_min just relevant for `res_conv`, which is the ky/sinh(ky) factor?  And that has the same functional form in both InclinedExponential and InclinedSersic, so the ksq_min would be the same, no?

brgillis: It might help to look at the kValueHelper function in SBInclinedExponential.cpp. In the inclined exponential case, _ksq_min is used as the limit for k in both res_base and res_conv, so when calculating _ksq_min, we need to take this into account so that the sum of errors introduced through this simplification won't exceed kvalue_accuracy.

In the inclined sersic case, res_base is calculated by the SersicInfo class, which has its own internal ksq_min (determined in a similar way). If we were to calculate ksq_min for the SBInclinedSersic class as if it were only relevant for res_conv, then we'd be allowing for res_conv to have error in it of up to kvalue_accuracy. But at the same time, res_base (calculated from the SersicInfo class) also might have error up to kvalue_accuracy, and so res_base*res_conv could have error up to double(ish) kvalue_accuracy, which we don't want.

galsim/inclinedsersic.py
```diff
@@ -0,0 +1,230 @@
...
+from . import _galsim
+
+
+class InclinedSersic(GSObject):
+    """A class describing an inclined sersic profile. This class is general, and so for certain
+    special cases, more specialized classes will be more efficient. For the case where n==1
+    with no truncation, the InclinedExponential class will be much more efficient. For the case
+    where the inclination angle is zero (face-on), the Sersic class will be slightly more efficient.
+
+    The Inclined Sersic surface brightness profile is characterized by four properties: its
```
rmandelb: "Inclined Sersic" -> "InclinedSersic"

galsim/inclinedsersic.py
```diff
@@ -0,0 +1,230 @@
...
+
+    The allowed range of values for the `n` parameter is 0.3 <= n <= 6.2.  An exception will be
+    thrown if you provide a value outside that range, matching the range of the Sersic profile.
+
+    This class shares the caching of Hankel transformations with the Sersic class; see that
+    class for documentation on efficiency considerations with regards to caching.
+
+    A profile can be initialized using one (and only one) of two possible size parameters:
+    `scale_radius` or `half_light_radius`.  Exactly one of these two is required. Similarly,
+    at most one of `scale_height' and `scale_h_over_r' is required; if neither is given, the
```
rmandelb: Should clarify both here and below (in the parameter list) that even if the user chooses to specify the half-light radius, ```scale_h_over_r``` still uses scale radius. Unless I've misunderstood that?

galsim/inclinedsersic.py
```diff
@@ -0,0 +1,230 @@
...
+    default of scale_h_over_r = 0.1 will be used.
+
+    Initialization
+    --------------
+
+    @param n                  The Sersic index, n.
+    @param inclination        The inclination angle, which must be a galsim.Angle instance
+    @param scale_radius       The scale radius of the disk.  Typically given in arcsec.
+                              This can be compared to the 'scale_radius' parameter of the
+                              galsim.Sersic class, and in the face-on case, the same same scale
```
rmandelb: "same same"

galsim/inclinedsersic.py
rmandelb: The check you've added here is good, but as Arun pointed out, you should add a check for this in InclinedExponential as well.

rmandelb: Also, was it a deliberate choice to not enable users to set ```trunc``` or ```flux_untruncated``` via config?  Or should this be added?

galsim/inclinedsersic.py
```diff
@@ -0,0 +1,230 @@
...
+    _single_params = [ { "scale_radius" : float , "half_light_radius" : float } ]
+    _takes_rng = False
+
+    def __init__(self, n, inclination, half_light_radius=None, scale_radius=None, scale_height=None,
+                 scale_h_over_r=None, flux=1., trunc=0., flux_untruncated=False, gsparams=None):
+
+        # Check that the scale/half-light radius is valid
+        if scale_radius is not None:
+            if not scale_radius > 0.:
+                raise Exception("scale_radius must be > zero.")
```
rmandelb: I think we usually call these kinds of errors due to bad inputs (here and below) ValueErrors rather than more generic Exceptions.

tests/test_inclined_exponential.py
```diff
@@ -30,173 +35,340 @@
...
+both_fluxes = ("1.0", "10.0", "0.1", "1.0", "1.0", "1.0")
+both_inc_angles = ("0.0", "1.3", "0.2", "0.01", "0.1", "0.78")
+both_scale_radii = ("3.0", "3.0", "3.0", "3.0", "2.0", "2.0")
+both_scale_heights = ("0.3", "0.5", "0.5", "0.5", "1.0", "0.5")
+both_pos_angles = ("0.0", "0.0", "0.0", "0.0", "-0.2", "-0.2")
+
+# InclinedSersic-only test cases
+sersic_fluxes = ("1.0", "2.0", "1.5", "1.1", "10.0", "1.0e6", "1.0e-6", "2.3e4",)
+sersic_ns = ("1.0", "1.5", "2.0", "2.5", "1.0", "1.5", "2.0", "2.5",)
+sersic_inc_angles = ("0.0", "0.0", "0.0", "0.0", "0.1", "0.2", "0.3", "0.5",)
```
rmandelb: I presume these are radians.  Is there any reason to restrict to such a limited range?  You aren't getting particularly close to pi/2.

tests/test_inclined_exponential.py
```diff
@@ -16,10 +16,15 @@
 #    and/or other materials provided with the distribution.
 #
 
+"""This file contains tests for the InclinedExponential and InclinedSersic classes.
```
rmandelb: Unless I missed it, it doesn't seem that there are tests of initializing using ```half_light_radius``` or ```flux_untruncated```.

tests/test_inclined_exponential.py
```diff
@@ -16,10 +16,15 @@
...
+"""
+
```
rmandelb: Also, I think you should add tests for both inclined profiles to ```test_config_gsobject.py``` to make sure that the config initialization works properly.

src/SBInclinedSersic.cpp
rmjarvis: I see where you're coming from here, but I think you misunderstand the intent of `kvalue_accuracy`.  We don't make particularly strong claims that the final kvalues are accurate to this level.  Rather we state in the doc about `kvalue_accuracy`:

>  Whenever there is some kind of approximation to be made in the calculation of a Fourier space value, the error in the approximation is constrained to be no more than this value times the total flux.

If we tried to make a stronger claim that the final k values won't have errors more than this much, then you'd always need to know what else is being convolved along with any given profile and take everything into account in any approximation being made.  This is clearly infeasible.

So I think it's fine if there are two places here where you make an approximation, each of which might have errors at this level.  I don't think it's necessary or appropriate to try to account for both approximations at once in the value of ksq_min that you use here.

tests/test_inclined_exponential.py
brgillis: I can add in a test for the edge-on case. The hold-up here is with generating galaxies with Lance's code. It's pretty inefficient, and I have to go to a much larger size to properly simulating galaxies with n > 1 (since the half-light radius is much higher if the scale radius is kept the same). I think there's also an issue with his code due to the fact that it handles pixel integration through subsampling and summing up subsampled pixels, which means it's likely a bit off near the centre.

All this is to say, it might take a fair bit more computer time to get comparison models which we can trust to percent-level accuracy. I can run some over the holidays, but it would be good to make sure the set of parameters is good enough before doing so, so I don't have to repeat it. How about I keep what's here, but triple all the inclinations to cover a better range (up to 1.5), and add in one more with inclination of pi/2? Should that cover it?

galsim/inclinedsersic.py
brgillis: Yes, I can add it to InclinedExponential too. Is it fine to do this in this branch, or should I open a new issue and branch to handle that?

Missing trunc and flux_untruncated was a mistake; I'll fix that once I have my work computer out.

tests/test_inclined_exponential.py
rmjarvis: It's ok to have the regression tests only cover the range where it is reasonable to run Lance's code (and where it doesn't take forever).  The problem is really that this same list is also used for the sanity tests, which cover a bit more in terms of the correctness of the functionality.  I propose just adding more elements to this set that are only run for the sanity tests, not the regression tests.

galsim/inclinedsersic.py
rmjarvis: Yes, ok to do this on this branch.

galsim/inclinedsersic.py
rmandelb: Doing it here is fine.

tests/test_inclined_exponential.py
brgillis: Done.

galsim/inclinedsersic.py
brgillis: Added now.

include/galsim/SBInclinedSersic.h
```diff
@@ -0,0 +1,150 @@
...
+     * by sech^2(z/Hs), where Hs is the scale height of the disk and z is the distance along the
+     * minor axis. The inclination angle determines how elliptical the profile appears.
+     *
+     * Note that the position angle is always zero. A profile with a different position angle can be
+     * obtained through the rotate() method of the corresponding Python class.
+     *
+     * If the inclination will always be zero (face-on), the SBSersic class can instead be used
+     * as a slightly faster alternative. If no truncation radius will be applied and n=1, the
+     * SBInclinedExponential class can be used as a much faster alternative.
+     */
```
brgillis: Done, simplified the comment now.

brgillis:
Okay, I believe I've addressed all changes and comments for this now (and I expect the checks to pass for this commit). Are there any comments, or changes requested which I missed?

One thing I wanted to check on was the maxSB method. For this profile, I don't believe the maximum surface brightness is analytic except for the face-on and edge-on cases, so I programmed this in with a rough interpolation between them. Since this is only used for photon-shooting as far as I can tell, which isn't enabled for this profile, this is only an issue if the user calls this method. Should I perhaps have it output a warning when the method is called?

Also, should we add this profile to any of the examples, so users will see that it exists?

include/galsim/SBInclinedSersic.h
```diff
@@ -0,0 +1,127 @@
...
+         * @brief Constructor.
+         *
+         * @param[in] n                 Sersic index.
+         * @param[in] inclination       Inclination of the disk relative to line of sight, where
+         *                              0 = face-on and pi/2 = edge-on.
+         * @param[in] size              Size specification.
+         * @param[in] rType             Kind of size being specified (HALF_LIGHT_RADIUS or
+         *                              SCALE_RADIUS).
+         * @param[in] height            Height specification.
+         * @param[in] rType             Kind of height being specified (SCALE_H_OVER_R or
```
rmjarvis: hType

tests/test_config_gsobject.py
```diff
@@ -45,36 +47,36 @@ def test_gaussian():
                    'dilate' : 3, 'ellip' : galsim.Shear(e1=0.3),
                    'rotate' : 12 * galsim.degrees,
                    'magnify' : 1.03, 'shear' : galsim.Shear(g1=0.03, g2=-0.05),
-                   'shift' : { 'type' : 'XY', 'x' : 0.7, 'y' : -1.2 }
+                   'shift' : { 'type' : 'XY', 'x' : 0.7, 'y' :-1.2 }
```
rmjarvis: Why??  

It looks to me like you used some robot style formatter on this file.  Please don't.  As you can see, it does stupid things like remove the space between the colon and the number if the number starts with a minus sign.  This makes this less legible, not more.

Please revert all changes to this file other than the functions you added.

tests/test_config_gsobject.py
```diff
@@ -515,13 +517,165 @@ def test_devaucouleurs():
...
+        # For the maximum_fft_size test, we need to do things a little differently
+        # We lower maximum_fft_size below the size that SBProfile wants this to be,
+        # and we check to make sure an exception is thrown.  Of course, this isn't how you
+        # would normally use maximum_fft_size.  Normally, you would raise it when the default
+        # is too small.  But to construct the test that way would require a lot of memory
+        # and would be rather slow.
+        gal6a = galsim.config.BuildGSObject(config, 'gal6')[0]
+        gal6b = galsim.InclinedExponential(inclination=0.7 * galsim.radians, half_light_radius=1, flux=50)
+        np.testing.assert_raises(RuntimeError, gsobject_compare, gal6a, gal6b,
+                                 conv=galsim.Gaussian(sigma=1))
```
rmjarvis: The above gal5c test is sufficient for this.  You don't need to also do the maximum_fft_size test as well.  We did that for Sersic to make sure that *something* tested maximum_fft_size.  But here, we only need to test that the GSParams option is being parsed.  And gal5c tests that already.

tests/test_config_gsobject.py
```diff
@@ -515,13 +517,165 @@ def test_devaucouleurs():
...
+        # For the maximum_fft_size test, we need to do things a little differently
+        # We lower maximum_fft_size below the size that SBProfile wants this to be,
+        # and we check to make sure an exception is thrown.  Of course, this isn't how you
+        # would normally use maximum_fft_size.  Normally, you would raise it when the default
+        # is too small.  But to construct the test that way would require a lot of memory
+        # and would be rather slow.
+        gal6a = galsim.config.BuildGSObject(config, 'gal6')[0]
+        gal6b = galsim.InclinedSersic(n=0.7, inclination=0.7 * galsim.radians, half_light_radius=1, flux=50)
+        np.testing.assert_raises(RuntimeError, gsobject_compare, gal6a, gal6b,
+                                 conv=galsim.Gaussian(sigma=1))
```
rmjarvis: ditto

tests/test_inclined_exponential.py
```diff
@@ -30,155 +38,319 @@
...
+    ("2.0", "1.5", "0.0" , "3.0", "0.3", "0.0", "0.0"),
+    ("1.5", "2.0", "0.0" , "3.0", "0.3", "0.0", "0.0"),
+    ("1.1", "0.5", "0.0" , "3.0", "0.3", "0.0", "0.0"),
+    ("10.0", "1.0", "0.1" , "1.9", "0.3", "4.5", "7.3"),
+    ("1.0e6", "1.5", "0.2" , "2.1", "0.2", "3.9", "-0.9"),
+    ("1.0e-6", "2.0", "0.3" , "3.4", "0.1", "5.0", "0.6"),
+    ("2.3e4", "0.5", "0.5" , "1.8", "0.5", "2.5", "0.3"),)
+
+# Parameter sets used for regression tests of Sersic profiles
+# TODO - generate these and add them in
```
rmjarvis: Is this still TODO?

tests/test_inclined_exponential.py
```diff
@@ -30,155 +38,319 @@
...
+
+# Parameter sets valid only for the inclined Sersic case
+inclined_sersic_test_parameters = (
+    ("2.0", "1.5", "0.0" , "3.0", "0.3", "0.0", "0.0"),
+    ("1.5", "2.0", "0.0" , "3.0", "0.3", "0.0", "0.0"),
+    ("1.1", "0.5", "0.0" , "3.0", "0.3", "0.0", "0.0"),
+    ("10.0", "1.0", "0.1" , "1.9", "0.3", "4.5", "7.3"),
+    ("1.0e6", "1.5", "0.2" , "2.1", "0.2", "3.9", "-0.9"),
+    ("1.0e-6", "2.0", "0.3" , "3.4", "0.1", "5.0", "0.6"),
+    ("2.3e4", "0.5", "0.5" , "1.8", "0.5", "2.5", "0.3"),)
```
rmjarvis: You have special code for cosi < 0.01 and cosi > 0.95.  It doesn't look like you're testing that with these inclination values (other than the cosi = 1 case).  Should add cases that make sure to hit those branches.

tests/test_inclined_exponential.py
```diff
@@ -30,155 +38,319 @@
...
         # Compare to the example - Due to the different fourier transforms used, some offset is
         # expected, so we just compare in the core to two decimal places
 
-        image_core = image.array[image_ny//2-2:image_ny//2+3, image_nx//2-2:image_nx//2+3]
-        test_image_core = test_image.array[image_ny//2-2:image_ny//2+3, image_nx//2-2:image_nx//2+3]
-
-        ratio_core = image_core / test_image_core
-
-        # galsim.fits.write(test_image,"test_"+image_filename,image_dir)
+        image_core = image.array[nx // 2 - 2:nx // 2 + 3, ny // 2 - 2:ny // 2 + 3]
```
rmjarvis: I know this is PEP8 compliant, but this would be much more readable with spaces around the colons, and not around the //, +, -.

tests/test_config_gsobject.py
brgillis: Ah sorry. One of my computers had auto-format on save turned on

tests/test_inclined_exponential.py
brgillis: These are added now.

src/SBInclinedExponential.cpp
```diff
@@ -220,15 +220,11 @@ namespace galsim {
...
-        // Empirically, it is vaguely linearish in sqrt(cosi), so we use that for
-        // the interpolation.  It's accurate to ~10-20% for moderate values of h0/r0.
-        double sc = sqrt(std::abs(_cosi));
-        maxsb *= (_h0 * sc + _r0 * (1.-sc)) / _h0;
-        return std::abs(maxsb);
+        // Maximum surface brightness isn't analytic, and is only needed for photon-shooting,
+        // so this is disabled.
+        throw std::runtime_error(
+            "The maxSB method of SBInclinedExponential is not yet implemented");
+        return 0;
```
rmjarvis: Why did you change this?  The existing implementation was fine.  `maxSB` only needs to be a rough estimate (and overestimates are fine), which this was.  Even if it's not used currently (since, as you point out, it is only typically used for photon shooting), I don't see why the existing working code was disabled.

brgillis: I got thinking about it again once I noticed that it was behaving even more poorly than expected for the new edge-on Sersic test case. It looks like the method is exposed to the user, so I didn't want them to falsely assume it would be accurate. I can put it back in for this and the InclinedSersic class though if you want (and perhaps have it output a warning?).

rmjarvis: No need for a warning.  The documentation about it already warns the user that it is not always terribly accurate:
```
        Some profiles will return the exact peak SB, typically equal to the value of
        obj.xValue(obj.centroid()).  However, not all profiles (e.g. Convolution) know how to
        calculate this value without just drawing the image and checking what the maximum value is.
        Clearly, this would be inefficient, so in these cases, some kind of estimate is returned,
        which will generally be conservative on the high side.

        This routine is mainly used by the photon shooting process, where an overestimate of
        the maximum surface brightness is acceptable.
```

rmjarvis: Also, for edge on Sersic, you should be able to make it exact (as it is analytic in that case).  For the InclinedExponential I interpolated between the exact values for face-on and edge-on profiles.  The tricky bit was figuring out which variable to use for the interpolation.  sqrt(cos(i)) turned out to be reasonably ok, and an overestimate over most of the range (0,1).

brgillis: I put this back in now. After some more testing, the most linear interpolation I could find was sqrt(cos(i)) -> ln(maxSB). At the end of the method, I multiply by a constant (1.56 for InclinedSersic, 1.33 for InclinedExponential) which is just large enough to ensure that this will never underestimate the maximum surface brightness. However, this means it's no longer exact for the face-on and edge-on cases. Do you think this is fine, or should I remove this overestimation factor?

rmjarvis: Either way.  I'd probably be more inclined to leave off the extra factor for simplicity sake.  A factor of 2 isn't really a big deal for how this function is used and these are better than that already.  But OK if you'd prefer to leave them in.

brgillis: Alright, in that case I agree that it makes more sense to go without this factor, so I took it out.

Issue #841, opened 2016-12-06
versions of dependencies


rmandelb:
@rmjarvis - can we give any further guidance in INSTALL.md regarding the necessary versions of GalSim's various dependencies?  This came up in the Euclid context, but I think it is kind of useful more generally.

We mention required versions for python, scons, FFTW, and TMV in INSTALL.md.  Can we add anything useful about required versions for NumPy, boost, pyfits or astropy, future, pyyaml, or the optional dependencies?

rmandelb:
@brgillis - I opened an issue for this to help fill in that info, JFYI.

rmjarvis:
I think we're compatible with rather old versions of numpy, boost, and pyfits/astropy, since we've had lots of systems over the years that had various old versions of these and added workaround for various bugs we discovered in them.  However, we probably don't want to guarantee compatibility with them, since I'm not sure which old versions are still semi-regularly tested.

For a "minimum requirements" listing, it would probably be safer to take whatever the current versions used by Travis and put a statement like "GalSim has been tested to work with the following versions, but it is likely that most versions released within the last several years will also work."

rmandelb:
Thanks.  That's a good way to deal with it.  While this is just a documentation change, I still want to do it on a branch and make a PR just so you can easily view the changes and make sure I'm not missing something.  Will do it in a sec.

rmandelb:
Actually... is it just me or do we not have a good way to get Travis to tell us which NumPy, future, or pyyaml version it's using?  I was looking at the logs, like https://travis-ci.org/GalSim-developers/GalSim/jobs/183439328 , but they don't seem to say the versions for those.

The logs do give versions for boost and astropy.  I can modify the build process to spit out the versions used and pull that info from the next build, but is there another way to get this information?

rmjarvis:
Numpy seems to be pre-installed, so we don't get the report from pip.  I'm pretty sure it's 1.6.1 though.  I could add a print statement to SCons that will tell us.

The rest though are reported here:
```
Successfully installed argparse-1.4.0 astropy-1.1.1 codecov-2.0.5 coverage-4.2 coveralls-1.1 cycler-0.10.0 docopt-0.6.2 future-0.16.0 matplotlib-1.5.2 pyparsing-2.1.10 python-dateutil-2.6.0 pytz-2016.10 pyyaml-3.12 requests-2.12.3 scipy-0.17.0 starlink-pyast-3.9.0
```

rmandelb:
[...]

PR #844, opened 2016-12-13
#841 - INSTALL.md updates only

rmandelb:
This PR just has edits to INSTALL.md to clarify versions of dependencies and a minimal set of dependencies.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -66,12 +66,15 @@ The GalSim package also requires
   versions will also work.
 
 * the future module, which is used to ease compatibility between Python 2
-  and Python 3.
+  and Python 3.  Currently GalSim is regularly tested to ensure 
+  it works with version 0.16.0 of this module, but other versions may work.
 
 * the PyYAML package for parsing YAML files (http://pyyaml.org/wiki/PyYAML)
   Note: PyYAML is only technically required if you are using the `galsim`
   executable for parsing YAML config files.  Users who will only use GalSim
   in Python (or use only JSON config files) may skip this dependency.
+  Currently GalSim is regularly tested to ensure it works with version 3.12
+  of this package, but other versions may work.
 
 * Optional dependency: PyAst WCS package.  This is a really nice front end
   for the Starlink AST astrometry code.  It seems to support pretty much
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -47,7 +47,7 @@ bandpasses and SEDs can be enabled if you have the pandas module (but the code
 will work, albeit more slowly, without this module).
 
 There are alternative modules to enable some of the GalSim functionality, such
-as pyfits, starlink, and so on (listed below as optional), but the above list is
+as pyfits and so on (listed below as optional), but the above list is
 a minimal set of dependencies to get started.  The sections below give a lot
 more details about how to obtain these dependencies; many are available from
 sources like pip or easy_install, rather than having to be installed from
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -5,6 +5,8 @@ System requirements: GalSim currently only supports Linux and Mac OSX.
 
 Table of Contents:
 
+0. [Overall summary](#0-overall-summary)
+
 1. [Software required before building GalSim](#1-software-required-before-building-galsim)
 
 2. [Installing the GalSim Python package](#2-installing-the-galsim-python-package)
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -17,6 +19,33 @@ Table of Contents:
...
+==================
+
+While the sections below detail how to install GalSim including its required and
+optional dependencies, this section gives a brief summary.  A minimal
+installation of GalSim requires Python (versions 2.6, 2.7, 3.4, or 4.5) along
+with the following dependencies: (a) SCons (versions 2.0 or 2.1 recommended),
+(b) NumPy (version 1.11 recommended), (c) astropy (version 1.1.1 recommended),
+(d) the future module (version 0.16.0 recommended), (e) FFTW (version >=3, often
+called fftw3), (f) TMV (version >=0.72), and (g) boost (version 1.61
+recommended).
```
rmjarvis: I think this might be more helpful as a bulleted list, rather than prose.  e.g.

The following list of dependencies include a canonical version number that is known to work.  In most cases, other recent versions will also work:

* Python (2.7, 3.4, 3.5)
* SCons (2.1.0)
* Boost (1.61)
* FFTW (3.3)
* TMV (0.73)
* NumPy (1.11)
* AstroPy (1.1.1)
* Future (0.16)

rmandelb: Sure.  Updating in a sec.

INSTALL.md
```diff
@@ -17,6 +19,33 @@ Table of Contents:
...
+
+A few optional dependencies provide additional functionality, but GalSim can
+otherwise be compiled and used without them.  To get WCS functionality with
+astropy, scipy is required (since it is a dependency of astropy WCS tools).  To
```
rmjarvis: For anyone who cares about good WCS functionality, I'd definitely recommend starlink, rather than astropy.  It's really far superior, so if we say anything about WCS here, I think that should be the one to mention.

rmandelb: I'm fine with your counterproposal.

rmandelb: which is to say, I'm doing it now

INSTALL.md
```diff
@@ -17,6 +19,41 @@ Table of Contents:
...
+to GalSim, but for users with more complicated WCS needs, we recommend
+installing starlink-pyast. Astropy's WCS package is also supported, but note
+that it requires scipy as an additional dependency.  To use yaml for config
+parsing, the pyyaml module is needed.  Faster text file parsing for reading in
+bandpasses and SEDs can be enabled if you have the pandas module (but the code
+will work, albeit more slowly, without this module).
+
+There are alternative modules to enable some of the GalSim functionality, such
+as pyfits and so on (listed below as optional), but the above list is
+a minimal set of dependencies to get started.  The sections below give a lot
```
rmjarvis: Maybe remove this first sentence, since pyfits is kind of the only thing left that wasn't mentioned above.  (I think.)  And that's probably irrelevant for most users now I think.

rmandelb:
done

rmandelb:
Thanks Mike.  @jmeyers314 , want to take a look?  It's just documentation, no code.

jmeyers314:
Sure.  I'll look right now.

INSTALL.md
```diff
@@ -17,6 +19,38 @@ Table of Contents:
...
+0. Overall summary
+==================
+
+While the sections below detail how to install GalSim including its required and
+optional dependencies, this section gives a brief summary.  A minimal
+installation of GalSim requires the following dependencies. This dependency list
+includes a canonical version number that is known to work. In most cases, other
+recent versions will also work:
+
+- Python (2.7, 3.4, 3.5)
```
jmeyers314: Just in time for 3.6, which I believe gets released on Friday :)

INSTALL.md
```diff
@@ -5,6 +5,8 @@ System requirements: GalSim currently only supports Linux and Mac OSX.
 
 Table of Contents:
 
+0. [Overall summary](#0-overall-summary)
```
jmeyers314: Looks like github markdown renumbered the list to start with 1., even though you made a 0.  The headings for the subsections are still 0-6 though.  https://github.com/GalSim-developers/GalSim/blob/%23841/INSTALL.md

INSTALL.md
```diff
@@ -247,8 +288,10 @@ v) Boost C++ (http://www.boost.org)
 -----------------------------------
 
 GalSim makes use of some of the Boost C++ libraries, and these parts of Boost
-must be installed. It is particularly important that your installed Boost
-library links to the same version of Python which which you will be using
+must be installed. Currently GalSim is regularly tested to ensure it works with
+Boost version 1.61, but it is likely that most versions released within the 
+last several years will also work. It is particularly important that your installed 
+Boost library links to the same version of Python which which you will be using
```
jmeyers314: typo

jmeyers314: (with which)

jmeyers314:
Looks good to me, @rmandelb .  Just the two small changes listed above.

INSTALL.md
rmjarvis: Mmm.  This is a classic markdown [bug](http://stackoverflow.com/questions/15078393/begin-ordered-list-from-0-in-markdown)-that-they-think-is-a-[feature](http://daringfireball.net/projects/markdown/syntax#list).  They finally fixed this on [StackOverflow](http://meta.stackexchange.com/questions/137077/cant-start-a-numbered-list-on-a-number-other-than-1), but [GitHub still forces you](https://docs.gitlab.com/ee/user/markdown.html#lists) to start lists at 1.

The workarounds that I know about are
0. put a `\` before the dot: `0\.`, etc. 
0. use a right paren instead: `0)`.  Or `(0)` perhaps.
0. renumber everything starting with 1.
0. use [custom html code](http://stackoverflow.com/a/15078486/1332281)

rmjarvis: Update.  Apparently Github did finally fix it for comments.  (I used `0.` for each of the above list items.)  But apparently not for the file markdown parsing (based on Josh's view of this file that he posted).

rmandelb: Thanks for listing all those alternatives!  Since the section headings themselves are fine, I went the lazy route of just converting to ```0)``` and so on.  Now the list itself looks fine to me in markdown format.   Possibly this is a bit weird but I wanted to indicate it's a summary before all the detailed bits, so unless you object, let's just go with this...

rmandelb:
Thanks for going over this.  Merging #844.

Issue #845, opened 2016-12-13
Various tests failing on master branch


brgillis:
Various tests are failing for me on the master branch. I've tested on two different machines, and get the same results. I've attached a log of the tests. For some reason, these aren't showing as failures with Travis, though.

[testlog.txt](https://github.com/GalSim-developers/GalSim/files/649643/testlog.txt)

rmandelb:
What numpy version?
Generally a discrepancy between your system and Travis means different versions of some dependency.

brgillis:
The machine I used for this log is version 1.7.1, and the other it failed on is 1.8.2

rmandelb:
Hmmm, I think numpy version 1.7 is 3 years old or so.  It wouldn�t surprise me if none of our machines used for testing has that version.

rmjarvis:
OK, I can reproduce this by rolling back to old versions of a bunch of things.  I'll take a look at whether the tests ought to be changed to allow these numerical differences or if we should just require a newer numpy version than this.

rmjarvis:
@brgillis Try branch #845.  If it works for you, I'll make a PR.

brgillis:
@rmjarvis Everything passes now for me (on numpy v1.7.1).

rmjarvis:
Closed via #846

PR #846, opened 2016-12-13
#845 Backwards compatibility with numpy 1.7

rmjarvis:
@brgillis reported in #845 that some tests were failing.  Turns out they were all due to numpy 1.7, which predates some accuracy improvements numpy made (I think in 1.9), especially in functions like `np.sum` which now accumulate in double precision, even if the array is single precision.  In numpy 1.7, the accumulator used whatever type the array was by default, although you can specify a different type using `dtype=float`.

This PR mostly amounts to adding that kwarg to many of our `np.sum(...)` calls.  It doesn't hurt anything on new numpy versions, and it makes things more accurate on old ones.

There were still a few places where the tests disagreed at the 5th decimal place, so there may be a couple more places where the numpy accuracy changed that I didn't find.  But for those, I just changed the test to only test 4 decimal places instead of 5.  I think that's acceptable.

I branched this off of #841, so feel free to wait to review this until that one is merged in.

### DIFF
--- a/galsim/config/extra_psf.py
+++ b/galsim/config/extra_psf.py
@@ -53,7 +53,7 @@ def DrawPSFStamp(psf, config, base, bounds, offset, method, logger):
 
         sn_target = galsim.config.ParseValue(config, 'signal_to_noise', base, float)[0]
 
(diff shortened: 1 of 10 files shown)


jmeyers314:
Looks good to me too.

PR #847, opened 2016-12-16
#824 geometric optics photon-shooting

jmeyers314:
I think this is almost ready.  PR-ing now to check the coverage report.  (There's no other way to do that, right?)


jmeyers314:
Okay, I think this one is ready for review now.

The main goal was to implement the geometric optics approximation to Fourier optics, which enables something like a 10^4-10^6 speed-up in photon-shooting optical and/or atmospheric PSFs (depending on source brightness and exposure time).  This approximation is now the default when drawing these objects using photon-shooting (and when drawing transformations, convolutions, etc. involving these objects).

I also made a few changes to the atmospheric API that I think make it easier to use.  For example, previously, the following would make two PSFs over two different time intervals:

```python
atm = galsim.Atmosphere(...)
psf1 = atm.makePSF(exptime=30.0, ...)
psf2 = atm.makePSF(exptime=30.0, ...)
```

I.e., `psf1` is exposing between t=0 and t=30 seconds, and `psf2` is exposing between t=30 and t=60 seconds.  After this PR, the above snippet makes both psfs expose from 0 to 30 seconds.  If you *want* 30 to 60 seconds, then you would do:

```python
psf3 = atm.makePSF(t0=30.0, exptime=30.0, ...)
```

Also new in this PR, the actual calculation for `psf1` and `psf2` above is delayed until either `psf1.drawImage()` or `psf2.drawImage()` is called (or `drawImage` for some object composed with `psf1` or `psf2` is called).  If drawing using Fourier methods, then drawing `psf1` will compute both `psf1` and `psf2`, which can sometimes be faster (and shouldn't be slower) when a boiling atmosphere is being used.

A few more changes:

- Changed signature of `wavefront()` method from `aper` to `u, v, t`.  I.e., instead of querying using an `Aperture` and using the "current time" of the screen, you query the locations and time of interest directly.
- Added `wavefront_gradient()` methods to screens and `PhaseScreenList`.
- Added `gradient` method to LookupTable2D.
- `OpticalScreen` now requires a `diam` be specified.  This is due to an API change for the `wavefront` method.
- Clock management methods (`advance()`, `reset()`) of phase screens and `PhaseScreenList` have been deprecated in favor of using the `t0` and `exptime` arguments when creating `PhaseScreenPSF`s.  (or when querying `wavefront()` or `wavefront_gradient()`.
- Deprecated the ability to return a list of PSFs from a single call to `makePSF`, since the delayed computation mentioned above makes it just as fast to call `makePSF` separately for each desired PSF.

Finally, I also added a new script: `examples/fft_vs_geom_movie.py`, which is a cleaned up and merged version of the scripts I used to make the movies below: 

- https://www.youtube.com/watch?v=l8d0jLVxBow
- https://www.youtube.com/watch?v=H2LVXN7XT3c

examples/fft_vs_geom_movie.py
```diff
@@ -0,0 +1,348 @@
...
+evolving set of Zernike aberrations and a frozen-flow Kolmogorov atmospheric phase screen.
+
+To visualize only the Zernike aberrations, add `--nlayers 0` as a command line argument.
+
+To visualize only atmospheric effects, add `--sigma 0` as a command line argument.
+
+Some suggested command lines, appropriate for LSST:
+
+Optics only simulation:
+  $ python fft_vs_geom_movie.py --diam 8.36 --obscuration 0.61 --sigma 0.05 --nlayers 0
```
jmeyers314: Oh, interesting...   This command line works fine for me with python3.5, but fails (in a different way) under python2.7:

```
[py2 josh@sr13-97df2a3081 src/GalSimPy2/examples (#824|✔)]$ python fft_vs_geom_movie.py --diam 8.36 --obscuration 0.61 --size 4.0 --nx 32
Adding layer at altitude  0.00 km with velocity ( 0.08, -4.13) m/s, and r0_500 0.255 m.
Adding layer at altitude  3.09 km with velocity ( 6.61, -6.62) m/s, and r0_500 0.618 m.
Adding layer at altitude  6.18 km with velocity ( 0.86, 16.40) m/s, and r0_500 1.300 m.
Adding layer at altitude  9.28 km with velocity ( 0.28,  1.58) m/s, and r0_500 1.365 m.
Adding layer at altitude 12.37 km with velocity (10.01,  9.02) m/s, and r0_500 0.979 m.
Adding layer at altitude 15.46 km with velocity (-8.78, -16.90) m/s, and r0_500 1.945 m.
|>-------------------------------------------------------------|   1 /100  ( 1.00%) ETA  1m48s
  File "fft_vs_geom_movie.py", line 348, in <module>
    make_movie(args)
  File "fft_vs_geom_movie.py", line 235, in make_movie
    writer.grab_frame(facecolor=fig.get_facecolor())
  File "/Users/josh/src/lsstsw/miniconda/lib/python2.7/site-packages/matplotlib/animation.py", line 244, in grab_frame
    'with --verbose-debug.'.format(e, out, err))
IOError: Error saving animation to file (cause: [Errno 32] Broken pipe) Stdout:  StdError: . It may help to re-run with --verbose-debug.
```

My guess is also matplotlib; I'll google around to see if I can learn anything.  In the meantime, what version are you running?  I'm on 1.5.3 for both python2.7 and python3.5

rmandelb: 1.5.2

jmeyers314: Ummm, actually....   It's more mundane than that.  I just had to create an 'examples/output' directory and things started working for me in py2.7.  

Can you check if this directory exists for you?  

I can add some logic to create the output directory if it doesn't exist...

rmandelb: Ha!  Sadly that won't help me, because I already had this directory.

jmeyers314: That's too bad.  I'll continue googling in that case.

rmandelb: I didn't mean to foist my problem on you - I was mainly asking to establish whether it really is just my system-dependent issue (vs. misuse of script, or a generic python 2.7 issue).  Thanks for offering but I will look into it myself. :)

jmeyers314: Okay.  I have one more suggestion and then I'll leave the rest to you :)  Do you have ffmpeg installed?  (I think you should be able to check with `which ffmpeg`).  I think the matplotlib animation stuff in this script requires that.

rmandelb:
That could indeed be the problem, because I do not have ffmpeg on this cluster.  Maybe list it as a requirement in docstring?

tests/test_phase_psf.py
```diff
@@ -200,23 +224,15 @@ def test_phase_screen_list():
 
     # Check some actual derived PSFs too, not just phase screens.  Use a small pupil_plane_size and
     # relatively large pupil_plane_scale to speed up the unit test.
-    atm.advance_by(1.0)
-    do_pickle(atm)
-    atm.reset()
-    kwargs = dict(exptime=0.06, diam=1.0, lam=1000.0)
+    atm._reset()
+    assert atm[0]._time == 0.0, "Wrong time for AtmosphericScreen"
+    kwargs = dict(exptime=0.05, time_step=0.025, diam=1.0, lam=1000.0)
```
rmjarvis: Probably better to use a non-default time_step here.  

I also don't particularly like using 1.0 for any variable in tests.  Too much chance to miss bugs related to multiplication, so I'd prefer using a different value for diam here.

tests/test_phase_psf.py
```diff
@@ -254,67 +282,68 @@ def test_phase_psf_reset():
     # Test frozen AtmosphericScreen first
     atm = galsim.Atmosphere(screen_size=30.0, altitude=10.0, speed=0.1, alpha=1.0, rng=rng)
     aper = galsim.Aperture(diam=1.0, lam=500.0)
-    wf1 = atm.wavefront(aper)
-    atm.advance()
-    wf2 = atm.wavefront(aper)
+    wf1 = atm.wavefront(aper.u, aper.v, None)
+    atm._seek(1.0)
+    wf2 = atm.wavefront(aper.u, aper.v, None)
```
rmjarvis: I think you say somewhere that public functions should return equal results for two objects that evaluate as equal.  If I've followed everything correctly, then this test seems to violate that condition, right?  `atm` is a nominally immutable object, but here it gives different values depending on an internal state.

I think this means that `wavefront` should not be a public API method.  Is there any reason an end user would want to call this directly rather than use the more user-friendly methods like `makePSF`?

tests/test_phase_psf.py
```diff
@@ -399,57 +436,149 @@ def test_ne():
...
+    nlayers = 6
+    time_step = 0.025  # s
+    screen_size = 102.4  # m
+    screen_scale = 0.1  # m
+    max_speed = 20  # m/s
+
+    rng = galsim.BaseDeviate(seed)
+    u = galsim.UniformDeviate(rng)
+    Ellerbroek_alts = [0.0, 2.58, 5.16, 7.73, 12.89, 15.46]  # km
+    Ellerbroek_weights = [0.652, 0.172, 0.055, 0.025, 0.074, 0.022]
```
rmjarvis: Helpful to also add a quick comment saying who Ellerbroek is and why we're using his atmospheric model.

tests/test_phase_psf.py
```diff
@@ -399,57 +436,149 @@ def test_ne():
...
+        second_moment_tolerance = 1.5
+
+    psfs = [atm.makePSF(lam, diam=diam, theta=th, exptime=exptime, aper=aper) for th in thetas]
+    shoot_moments = []
+    fft_moments = []
+
+    # At the moment, Ixx and Iyy (but not Ixy) are systematically smaller in phase gradient shooting
+    # mode than in FFT mode.  For now, I'm willing to accept this, but we should revisit it once we
+    # get the "second kick" approximation implemented.
+    offset = [0.5, 0.5, 0.0]
```
rmjarvis: I guess maybe this test is the one that is more related to the runs of fft_vs_geom_movie.py.  But won't the second kick only add to the size?  So does this square with the third plots I posted where geometric optics had a larger size than the fft?

tests/test_table.py
```diff
@@ -360,6 +360,105 @@ def f(x_, y_):
...
+    test_dfdx, test_dfdy = tab2d.gradient(newxx+3*tab2d.xperiod, newyy)
+    np.testing.assert_array_almost_equal(ref_dfdx, test_dfdx)
+    np.testing.assert_array_almost_equal(ref_dfdy, test_dfdy)
+
+    test_dfdx, test_dfdy = tab2d.gradient(newxx, newyy+13*tab2d.yperiod)
+    np.testing.assert_array_almost_equal(ref_dfdx, test_dfdx)
+    np.testing.assert_array_almost_equal(ref_dfdy, test_dfdy)
+
+    # Check that edge_mode='constant' behaves as expected.
+    tab2d = galsim.LookupTable2D(x, y, z, edge_mode='constant')
```
rmjarvis: Shouldn't this have a few asserts after it?  This doesn't seem to test anything other than "constant" is a valid value for `edge_mode`.

examples/fft_vs_geom_movie.py
jmeyers314: I added the check for the output directory existence and mentioned in the docstring and argparse description that ffmpeg is required for this script.

examples/fft_vs_geom_movie.py
```diff
@@ -0,0 +1,348 @@
...
+    else:
+        atm = galsim.PhaseScreenList()
+
+    # Setup Fourier and geometric apertures
+    fft_aper = galsim.Aperture(args.diam, args.lam, obscuration=args.obscuration,
+                               pad_factor=args.pad_factor, oversampling=args.oversampling,
+                               nstruts=args.nstruts, strut_thick=args.strut_thick,
+                               strut_angle=args.strut_angle*galsim.degrees)
+    geom_aper = galsim.Aperture(args.diam, args.lam, obscuration=args.obscuration,
+                                pad_factor=args.geom_oversampling, oversampling=0.5,
```
jmeyers314: Good question, and worth elaborating somewhere, maybe in the docstrings for `PhaseScreenPSF` and `OpticalPSF`?

The reason is that `oversample=0.5` produces a (square) pupil plane array that precisely circumscribes the (circular) aperture.  (`oversample=1.0` produces a square twice this size, which is just large enough for two copies of the aperture to cleanly "slide" past each with periodic boundary conditions, which is what you need for no aliasing in Fourier optics.)  Setting `oversample` larger than 0.5 adds points to the pupil array that are never sampled when collecting phase gradients, and setting `oversample` smaller starts to clip pieces of the aperture that we do want to sample.

examples/fft_vs_geom_movie.py
```diff
@@ -0,0 +1,348 @@
...
+
+    # Generate an atmosphere.  Same procedure as in psf_wf_movie.py
+    Ellerbroek_alts = [0.0, 2.58, 5.16, 7.73, 12.89, 15.46]  # km
+    Ellerbroek_weights = [0.652, 0.172, 0.055, 0.025, 0.074, 0.022]
+    Ellerbroek_interp = galsim.LookupTable(Ellerbroek_alts, Ellerbroek_weights,
+                                           interpolant='linear')
+    alts = np.max(Ellerbroek_alts)*np.arange(args.nlayers)/(args.nlayers-1)
+    weights = Ellerbroek_interp(alts)  # interpolate the weights
+    weights /= sum(weights)  # and renormalize
+    max_speed = 20  # Pick (an arbitrary) maximum wind speed in m/s.
```
jmeyers314: I went ahead and made `max_speed` a command line argument, like most of the other keyword arguments.  I also added some text to the file docstring indicating that you can run `python fft_vs_geom_movie.py --help` to see all the options and defaults.

examples/fft_vs_geom_movie.py
```diff
@@ -0,0 +1,348 @@
...
+    ak = np.sqrt(2*pk)
+    phi = np.random.uniform(size=(args.n, args.jmax))
+    zk = ak[:, None]*np.exp(2j*np.pi*phi)
+    aberrations = args.n/2*np.fft.ifft(zk, axis=0).real
+    measured_std = np.mean(np.std(aberrations, axis=0))
+    aberrations *= args.sigma/measured_std
+    aberrations -= np.mean(aberrations, axis=0)
+
+    # Generate an atmosphere.  Same procedure as in psf_wf_movie.py
+    Ellerbroek_alts = [0.0, 2.58, 5.16, 7.73, 12.89, 15.46]  # km
```
jmeyers314: Added comment:
```
    # For the atmosphere screens, we first estimates weights, so that the turbulence is dominated by
    # the lower layers consistent with direct measurements.  The specific values we use are from
    # SCIDAR measurements on Cerro Pachon as part of the 1998 Gemini site selection process
    # (Ellerbroek 2002, JOSA Vol 19 No 9).
```

galsim/phase_psf.py
```diff
@@ -709,97 +686,140 @@ def __ne__(self, other): return not self == other
...
-
-    def advance_by(self, dt):
-        """Advance each phase screen in list by specified amount of time.
-
-        @param dt  Amount of time in seconds by which to update the screens.
-        @returns   The actual amount of time updated, which can potentially (though not necessarily)
-                   differ from `dt` when `dt` is not a multiple of self.time_step.
-        """
+    def _seek(self, t):
+        """Seek all layers to time t."""
```
jmeyers314: Changed to "Set all layers' internal clocks to time t."

galsim/phase_psf.py
```diff
@@ -709,97 +686,140 @@ def __ne__(self, other): return not self == other
...
+            return np.sum([layer._wavefront_gradient(u, v, t, theta) for layer in self], axis=0)
+        else:
+            return self._layers[0]._wavefront_gradient(u, v, t, theta)
+
+    def makePSF(self, lam, **kwargs):
+        """Create a PSF from the current PhaseScreenList.
 
         @param lam                 Wavelength in nanometers at which to compute PSF.
+        @param t0                  Time at which to start exposure in seconds.  [default: 0.0]
         @param exptime             Time in seconds overwhich to accumulate evolving instantaneous
```
jmeyers314: Done.

galsim/phase_psf.py
```diff
@@ -23,6 +23,9 @@
 
 where x, y are focal plane coordinates and u, v are pupil plane coordinates.
 
+Alternatively, drawing using photon-shooting can use a fast geometric optics approximation to
+Fourier optics.
```
jmeyers314: I rewrote the beginning of this file docstring:

```
Utilities for creating PSFs from phase screens.

For PSFs drawn using real-space or Fourier methods, these utilities essentially evaluate the Fourier
optics diffraction equation:

PSF(x, y) = int( |FT(aperture(u, v) * exp(i * phase(u, v, x, y, t)))|^2, dt)

where x, y are focal plane coordinates and u, v are pupil plane coordinates.

For PSFs drawn with method='phot', an often significantly faster geometric approximation is used 
instead.  To use photon-shooting without this approximation, set `geometric_shooting=False` when 
creating the PSF.
```

galsim/phase_psf.py
```diff
@@ -817,79 +837,58 @@ def makePSF(self, lam, **kwargs):
...
-                                   obscuration and strut parameters.  Note that if the image is saved as
-                                   unsigned integers, you will get a warning about conversion to floats,
-                                   which is harmless. [default: None]
-        @param pupil_angle         If `pupil_plane_im` is not None, rotation angle for the pupil plane
-                                   (positive in the counter-clockwise direction).  Must be an Angle
-                                   instance. [default: 0. * galsim.degrees]
+        @param pupil_plane_im      The GalSim.Image, NumPy array, or name of file containing the
+                                   pupil plane image, to be used instead of generating one based on
+                                   the obscuration and strut parameters.  Note that if the image is
+                                   saved as unsigned integers, you will get a warning about
```
jmeyers314: Looks like this is no longer a problem.  I tried:

```python
import galsim
import numpy as np

psf = galsim.OpticalPSF(lam=500.0, diam=1.0)
img = galsim.Image(psf._psf.aper.illuminated, scale=psf._psf.aper.pupil_plane_scale, dtype=np.uint16)
img.write("illuminated.fits")
aper = galsim.Aperture(lam=500.0, diam=1.0, pupil_plane_im="illuminated.fits")
psf = galsim.OpticalPSF(lam=500.0, diam=1.0, pupil_plane_im="illuminated.fits")
```

Neither of the last 2 lines raised any warnings, so I'm removing this bit from the appropriate docstrings.

galsim/compound.py
```diff
@@ -1159,3 +1322,7 @@ def __repr__(self):
             gsparams=repr(self._input_gsparams),
         )
         return rep
+
+    def _prepareDraw(self):
+        # RandomWalk never wraps a PhaseScreenPSF, so no need to prepare anything.
+        pass
```
jmeyers314: It's because `RandomWalk` derives from `Sum`, which does have an override.  So this de-overrides that override.

galsim/phase_screens.py
```diff
@@ -704,14 +792,16 @@ def __repr__(self):
...
+                and self.diam == other.diam
                 and np.array_equal(self.aberrations*self.lam_0, other.aberrations*other.lam_0)
                 and self.annular_zernike == other.annular_zernike)
 
     def __ne__(self, other): return not self == other
 
     # This screen is immutable, so make a hash for it.
     def __hash__(self):
-        return hash(("galsim.AtmosphericScreen", tuple((self.aberrations*self.lam_0).ravel())))
+        return hash(("galsim.AtmosphericScreen", self.diam, self.obscuration, self.annular_zernike,
```
jmeyers314: Thanks.  It's good to be consistent I suppose.  Fixed.

tests/test_base.py
```diff
@@ -118,6 +118,8 @@ def test_gaussian():
 
     # Test photon shooting.
     do_shoot(gauss,myImg,"Gaussian")
+    # Verify that shoot with rng=None still runs
+    gauss.shoot(100, rng=None)
```
jmeyers314: Yep.  Done.

galsim/compound.py
```diff
@@ -645,6 +744,33 @@ def __str__(self):
...
+
+        @param n_photons    The number of photons to use for photon shooting.
+        @param rng          If provided, a random number generator to use for photon shooting,
+                            which may be any kind of BaseDeviate object.  If `rng` is None, one
+                            will be automatically created, using the time as a seed.
+                            [default: None]
+        @returns PhotonArray.
+        """
+        # Setup the rng if not provided one.
+        if rng is None:
```
jmeyers314: Sure.  Actually, is there any reason why we shouldn't make it so this whole snippet:
```python
    if rng is None:
        ud = galsim.UniformDeviate()
    elif isinstance(rng, galsim.BaseDeviate):
        ud = galsim.UniformDeviate(rng)
    else:
        raise TypeError("The rng provided is not a BaseDeviate")
```

is equivalent to just

```python
ud = galsim.UniformDeviate(rng)
```
??

examples/fft_vs_geom_movie.py
```diff
@@ -0,0 +1,348 @@
...
+
+To visualize only atmospheric effects, add `--sigma 0` as a command line argument.
+
+Some suggested command lines, appropriate for LSST:
+
+Optics only simulation:
+  $ python fft_vs_geom_movie.py --diam 8.36 --obscuration 0.61 --sigma 0.05 --nlayers 0
+
+Atmosphere only simulation:
+  $ python fft_vs_geom_movie.py --diam 8.36 --obscuration 0.61 --sigma 0.0 --nlayers 6 --size 3.0
```
jmeyers314: Yeah, that seems reasonable.  It's easy enough to come up with command line options that make things run faster.  Is it easy to add ffmpeg to travis?

galsim/compound.py
rmjarvis: > why we shouldn't make it so this whole snippet equivalent to `ud = galsim.UniformDeviate(rng)`

+1 for this suggestion.  But the way we do the wrapping currently, it's not a trivial change.  If you think you know what to do to implement this, go for it.  Otherwise, I can give it a go.

examples/fft_vs_geom_movie.py
rmjarvis: I think you probably just need to add `ffmpeg` to the apt-get installation line.  But see [this page](http://askubuntu.com/questions/691109/how-do-i-install-ffmpeg-and-codecs) if that doesn't work.

tests/test_phase_psf.py
```diff
@@ -19,7 +19,7 @@
 from __future__ import print_function
 import os
 import numpy as np
-from galsim_test_helpers import *
+from galsim_test_helpers import timer, do_shoot, do_pickle, all_obj_diff, getmoments
```
jmeyers314: Yeah, I'm pretty ambivalent to this too, but pyflakes isn't.  So importing everything by name here is just to make it easier to verify that pyflakes isn't suggesting anything legitimate.

tests/test_phase_psf.py
jmeyers314: Sure, changed this to `kwargs = dict(exptime=0.05, time_step=0.01, diam=1.1, lam=1000.0)`

doc/ref/GalSim_Quick_Reference.tex
```diff
@@ -298,9 +298,9 @@ \subsection{GSObject classes and when to use them}\label{sect:gsobjectclasses}
 
 \subsection{Atmosphere}
 GalSim can also simulate PSFs generated by turbulent layers in the atmosphere over a wide field of
-view.  (Though note that this machinery is much slower than other parts of GalSim.) The syntax for
-creating this type of \texttt{GSObject} is a little bit different than for those above.  First, some
-number \texttt{AtmosphericScreen}s should be assembled into a \texttt{PhaseScreenList}:
+view.  The syntax for creating this type of \texttt{GSObject} is a little bit different than for
+those above.  First, some number \texttt{AtmosphericScreen}s should be assembled into a
```
jmeyers314: Done.

galsim/compound.py
jmeyers314: I've been having some fun trying this out.  My thought is to use `bp::make_constructor` and essentially do runtime type introspection in the pointed to construct function. E.g., for BaseDeviate the construct function and wrapping would look like:

```c++
        static boost::shared_ptr<BaseDeviate> construct(const bp::object &seed) {
            if (seed.ptr() == Py_None) {
                return boost::shared_ptr<BaseDeviate>(new BaseDeviate(0l));
            }
            bp::extract<long> xl(seed);
            if (xl.check()) {
                return boost::shared_ptr<BaseDeviate>(new BaseDeviate(xl()));
            }
            bp::extract<const BaseDeviate&> xbd(seed);
            if (xbd.check()) {
                return boost::shared_ptr<BaseDeviate>(new BaseDeviate(xbd()));
            }
            bp::extract<std::string> xs(seed);
            if (xs.check()) {
                return boost::shared_ptr<BaseDeviate>(new BaseDeviate(xs()));
            }
            throw std::runtime_error("Cannot construct BaseDeviate from given seed.");
        }

        static void wrap() {
            bp::class_<BaseDeviateCallBack>
                pyBaseDeviate("BaseDeviate", "", bp::no_init);
            pyBaseDeviate
                // .def(bp::init<long>(bp::arg("seed")=0))
                // .def(bp::init<const BaseDeviate&>(bp::arg("seed")))
                // .def(bp::init<std::string>(bp::arg("seed")))
                .def("__init__",
                    bp::make_constructor(
                        &construct, bp::default_call_policies(),
                        (bp::arg("seed")=bp::object())
                    ))
                ...
```

This mostly works, except for the virtual override of `_val` in `DistDeviate`, which I'm guessing is the tricky bit you were eluding to above.  I found [this stackoverflow answer](http://stackoverflow.com/questions/19254715/virtual-override-in-boost-python-objects-created-by-static-method-or-factory) which looks relevant, but I thought I'd ask if I'm missing something simpler before tackling that.

jmeyers314: Okay, I found (and [committed](https://github.com/GalSim-developers/GalSim/pull/847/commits/c5629f99130b06bcfd170eafe19e3dc5fa6fa3af)) what I think is a better solution: add a `.def("__init__" ...)` using `make_constructor` as the lowest priority (so first in order) initializer that captures any python object as the seed, check if it's `None`, and throw if not.  This solves the virtual override problem for `DistDeviate`.

I'm getting a new (old) error though now in `test_config_image.py:test_njobs()` similar(identical?) to that found [here](https://github.com/GalSim-developers/GalSim/pull/714#issuecomment-190443809).  Basically, you get different results if you split up a config job into separate processes than if you run them through the same process.  Looking at the debug output, it seems like the random seeds are consistent, but as before, the number of cosmos gals passing cuts is not consistent.  Thoughts?

rmjarvis: That's a nice solution.  Not quite the approach I was thinking of (I was going to make the constructor from another `BaseDeviate` take a pointer rather than `const BaseDeviate&` and have it default to NULL, aka None in python), but this is probably a nicer way to do it.

As for the new bug, it seems that I have been relying on `BaseDeviate(None)` raising an exception, in kind of a subtle way.  I'll have to think a bit to see what the best solution is to not rely on that, since i do think it should behave as you've now implemented it.

rmjarvis: I fixed it.  It's kind of a bandaid hack that is probably telling me I should redesign some of how ProcessInput works, but I think it's fine for now.

jmeyers314: Thanks.  Looks like that error is still showing up though, I'm guessing because constructing a `BaseDeviate` from a `str` is allowed (I think that's how the serialization of `BaseDeviate` works).  Maybe use a custom Sentinel object (or even `object()`) instead?

rmjarvis: Sorry.  I could have sworn it worked on my machine when I tried it.  I must have messed up something with my test though, since of course it doesn't work.  Anyway, object() seems to fit the bill nicely.  I think it should work now.

galsim/compound.py
```diff
@@ -175,6 +174,69 @@ def __str__(self):
...
+        for i, obj in enumerate(self.obj_list):
+            thisAbsoluteFlux = obj.SBProfile.getPositiveFlux() + obj.SBProfile.getNegativeFlux()
+
+            # How many photons to shoot from this summand?
+            thisN = remainingN  # All of what's left, if this is the last summand...
+            if i < len(self.obj_list)-1:
+                # otherwise, allocate a randomized fraction of the remaining photons to summand.
+                bd = galsim.BinomialDeviate(ud, remainingN, thisAbsoluteFlux/remainingAbsoluteFlux)
+                thisN = bd()
+            if thisN > 0:
```
rmjarvis: I noticed the False version of this branch isn't covered.  Probably worth adding tests of shooting 0,1,2 to the `do_shoot` function to make sure they run without errors, even though the results won't be particularly checkable for accuracy.

galsim/compound.py
```diff
@@ -175,6 +174,69 @@ def __str__(self):
...
+                thisPA.scaleFlux(fluxPerPhoton*thisN/thisAbsoluteFlux)
+                result.assignAt(istart, thisPA)
+                istart += thisN
+            remainingN -= thisN
+            remainingAbsoluteFlux -= thisAbsoluteFlux
+        assert remainingN == 0
+        assert np.isclose(remainingAbsoluteFlux, 0.0)
+
+        # This process produces correlated photons, so mark the resulting array as such.
+        if len(self.obj_list) > 1:
```
rmjarvis: The False version of this branch is also not covered.  Probably ought to add a `do_shoot` call for one of the single-object Convolve objects.  (We test that construction for pickling and basic tests, but not shooting.)

examples/fft_vs_geom_movie.py
```diff
@@ -0,0 +1,348 @@
...
+from distutils.version import LooseVersion
+if LooseVersion(matplotlib.__version__) < LooseVersion('1.2'):
+    raise RuntimeError("This demo requires matplotlib version 1.2 or greater!")
+
+try:
+    from astropy.utils.console import ProgressBar
+except ImportError:
+    raise ImportError("This demo requires astropy!")
+
+def simple_moments(img):
```
jmeyers314: Added `galsim.utilities.unweighted_moments` and `galsim.utilities.unweighted_shape`, and went through the old tests to make sure these are being used instead of the old `getmoments` helper function (which I removed).

While writing these, I noticed that it wasn't possible to add (subtract) a `PositionD` to (from) a `PositionI`.  I couldn't think of a good reason not to allow this, so I went ahead and implemented type promotion for Positions too.

galsim/compound.py
jmeyers314: Sure.  Done.

galsim/compound.py
jmeyers314: Also done.

tests/test_table.py
jmeyers314: Yes, thanks.  I added some asserts.

galsim/phase_screens.py
```diff
@@ -41,7 +41,7 @@ class AtmosphericScreen(object):
     @param altitude      Altitude of phase screen in km.  This is with respect to the telescope, not
                          sea-level.  [default: 0.0]
     @param time_step     Interval to use when advancing the screen in time in seconds.
-                         [default: 0.03]
+                         [default: 0.025]
```
jmeyers314: No, there wasn't a lot of thought behind this change.  I just wanted a time step that evenly divided 1.0 second.  I'm happy to set this back to 0.03 if you prefer.

I agree that the right thing to do would involve comparing to the asymptotic results, but unfortunately I think the convergence is likely to depend on a number of additional factors like the simulated wind speeds, exposure time, number of layers, telescope aperture, r0, ...  It might be useful to try some typical values for DES, LSST, HSC though.

galsim/phase_psf.py
```diff
@@ -1051,67 +1040,126 @@ def __init__(self, screen_list, lam, exptime=0.0, flux=1.0, aper=None,
...
+    def _prepareDraw(self):
+        # Trigger delayed computation of all pending PSFs.
+        self._screen_list._prepareDraw()
+
+    # A few items which need the InterpolatedImage to have been prepared before accessing.
+    def maxK(self):
+        """Returns value of k beyond which aliasing can be neglected.
+        """
+        self._prepareDraw()
+        return self.SBProfile.maxK()
```
jmeyers314: Good point.  Done.

galsim/phase_screens.py
rmjarvis: I don't actually prefer either one.  :)  Maybe add a sentence to the doc saying that 0.025 is fairly arbitrary, although probably in the right ball park for most purposes.  For careful studies, one would want to check that the results do not change significantly when going to smaller time steps.  Something like that.

examples/fft_vs_geom_movie.py
jmeyers314: Okay, I think I've got the movie scripts working under run_examples.py now.

tests/test_phase_psf.py
jmeyers314: Added:
```
    # Use atmospheric weights from 1998 Gemini site selection process as something reasonably
    # realistic.  (Ellerbroek 2002, JOSA Vol 19 No 9).
```

tests/test_phase_psf.py
jmeyers314: Good catch.  I purposely left the possibility of using `None` for the `t` argument undocumented, so this might have been okay, but it was easy enough to switch the test to use `_wavefront` instead of `wavefront` and remove the special case logic out of the latter.  So I think the public API methods should be okay now.

As for whether to make wavefront public or not, I think it's nice to include for its own sake, even if it's not simple to convert a wavefront result into a PSF.

Looking through this also made me realize that I should document the fact that the `t` and `theta` arguments to `wavefront` and `wavefront_gradient` don't do anything at all for `OpticalScreen`s, so I also edited those docstrings.

galsim/phase_screens.py
jmeyers314: Alright.  Thanks for pointing this out, as it alerted me to another change I had forgotten to document.  Previously, setting the time_step was done only to the phase screens themselves.  With the new changes, there's still a time_step for the phase screens that's  used when boiling, but there's also a completely independent time_step used when creating the PSF.  I.e., boiling updates and PSF time integration steps no longer need to be synchronized, and in fact individual layers may have different boiling time_steps.

So I went through and documented that fact, and also added requirements that the screen time_step only be set (different from the new default of `None`) if boiling is turned on.

I did add something like your suggested text to the time_step documentation for makePSF and PhaseScreenPSF.

galsim/table.py
```diff
@@ -486,9 +486,66 @@ def __call__(self, x, y):
...
 
+    def _gradient_raise(self, x, y):
+        if not self._inbounds(x, y):
+            raise ValueError("Extrapolating beyond input range.")
+
+        try:
+            xx = float(x)
+            yy = float(y)
+            return self.table.gradient(xx, yy)
+        except:
```
rmjarvis: Probably my error here, but this should be `except TypeError` probably.

galsim/utilities.py
```diff
@@ -1016,7 +1016,7 @@ def lod_to_dol(lod, N=None):
             except TypeError:  # Value is not list-like, so broadcast it in its entirety.
                 out[k] = v
             except Exception:
```
rmjarvis: Something I think we're probably not so good at in GalSim currently is that whenever you catch generic excdption via either `except Exception:` or simply `except:`, you should first catch `KeybordInterrupt` and reraise.  Otherwise, Ctrl-C doesn't always interrupt the program the way you'd want.

galsim/utilities.py
```diff
@@ -1209,6 +1209,47 @@ def generate():
...
+    The ellipticities are:
+        e1 = (Mxx-Myy) / rsqr
+        e2 = 2*Mxy / rsqr
+
+    @param arg   Either a galsim.Image or the output of unweighted_moments(image).
+    @returns  Dict with entries for [rsqr, e1, e2]
+    """
+    try:
+        arg = unweighted_moments(arg)
+    except:
```
rmjarvis: Another generic exception that should either be given a real type or include a `KeyboardInterrupt` case.  Or possibly just use 
```
if isinstance(arg, galsim.Image): 
    arg = unweighted_moments(arg)
```

galsim/table.py
jmeyers314: Fixed.

galsim/utilities.py
jmeyers314: Okay.  Good to keep in mind.  Done.

galsim/utilities.py
jmeyers314: Sure.  Switched this to use `isinstance`.

examples/fft_vs_geom_movie.py
```diff
@@ -0,0 +1,348 @@
...
+    axes[2].scatter(fft_mom[:, 4], geom_mom[:, 4])
+    axes[0].set_title("Ixx")
+    axes[1].set_title("Iyy")
+    axes[2].set_title("Ixy")
+    for ax in axes:
+        ax.set_xlabel("Fourier Optics")
+        ax.set_ylabel("Geometric Optics")
+        symmetrize_axis(ax)
+    fig.tight_layout()
+    fig.savefig(args.out+"2ndMoment.png", dpi=300)
```
jmeyers314: So I've thought about these plots a bit more, and I think they're consistent with missing an additional convolution by an isotropic kernel, which is what the second kick will be once we implement it.  I.e., the first moments are all spot-on, as are the Ixy moments, and the Ixx and Iyy moments are too small (by a fixed amount) without the second kick.

As for the last set of plots, I think the problem here is just that the image size is too small (0.6 arcsec by default).  If you run `python fft_vs_geom_movie.py --diam 8.36 --obscuration 0.61 --nlayer 6 --sigma 0.05 --size 3` instead, then you get the following figure:

fft_vs_geom_2ndmoment

The other piece these comparisons are missing is that they're all instantaneous PSFs, so I'll try running a few time-integrated simulations to see what happens there.

I've also realized that this approximation is probably really only appropriate for atmospheric phase screen PSFs, and not pure optical PSFs.  I.e., using the geometric approximation for shooting WFIRST or Euclid photons is probably not sufficiently accurate, since the high frequency modes don't get averaged down over time the way they do for an atmospheric PSF.  So I propose to make `geometric_shooting=True` the default for `PhaseScreenPSF`, which is where atmospheric PSFs originate, but use `geometric_shooting=False` for the `OpticalPSF` class.  An interesting question then is:  What is the impact of using `geometric_shooting=True` for the optical part of a combined atmosphere+optics PSF?  I'll test that out some too.

galsim/phase_psf.py
```diff
@@ -607,41 +611,12 @@ class PhaseScreenList(object):
...
-
     Methods
     -------
     makePSF()          Obtain a PSF from this set of phase screens.  See PhaseScreenPSF docstring
                        for more details.
-    advance()          Advance each phase screen in list by self.time_step.
-    advance_by()       Advance each phase screen in list by specified amount.
-    reset()            Reset each phase screen to t=0.
     wavefront()        Compute the cumulative wavefront due to all screens.
+    wavefront_gradient()   Compute the cumulative wavefront gradient due to all screens.
```
jmeyers314: Sure.  I added your text to CHANGELOG

tests/test_optics.py
```diff
@@ -946,6 +950,51 @@ def zern_rho_coefs(n, m):
...
+        np.testing.assert_allclose(
+            shoot_moments.moments_centroid.x, fft_moments.moments_centroid.x, rtol=0, atol=0.025,
+            err_msg="")
+        np.testing.assert_allclose(
+            shoot_moments.moments_centroid.y, fft_moments.moments_centroid.y, rtol=0, atol=0.025,
+            err_msg="")
+        # 1% size tolerance
+        np.testing.assert_allclose(
+            shoot_moments.moments_sigma, fft_moments.moments_sigma, rtol=0.01, atol=0,
+            err_msg="")
```
jmeyers314: There's a convolution here by a small (but non-vanishing) Kolmogorov that mitigates some of the differences that are more clearly visible in the movie script.  I think this is okay though if we decide that geometric shooting optics PSFs only makes sense when there's also an atmospheric component present.

tests/test_phase_psf.py
jmeyers314: Same issue as above: the size of the image needs to be large enough to accommodate the PSF.  Then the calculated second moments make sense.

examples/fft_vs_geom_movie.py
rmjarvis: Sounds like a good plan about having the different defaults.  Maybe also add some discussion in the docs (and probably also fftw_vs_geom_movie.py) that for pure optics wavefronts, the geometric approximation isn't a very good one, so you should generally use ffts.  But if you also have an atmospheric part, then you're probably ok to use the geometric mode for both.

Also, I don't really understand the 2nd kick stuff.  Do you know how to estimate the size of the missing isotropic kernel?  Can we just mock it up by convolving what you have with a `Gaussian`?  Or is the issue that that you don't yet know what sigma to use a priori?

jmeyers314: > Do you know how to estimate the size of the missing isotropic kernel?

Roughly, I can play with this in Mathematica a bit to see if I can get a size estimate.  The details are that the second kick PSF is the Fourier transform of the exponential of an integral over the turbulence power spectrum.  I suspect we'll have to add a new `GSObject` for this, though maybe it'll be a private one since I don't see how it would be useful outside of the geometric optics approximation.

jmeyers314: Actually, I'm pretty sure it's the same expression that leads to the Kolmogorov profile. The difference is just the range of k being integrated over.

jmeyers314: Here are some results for 100 different random seeds for 30 second integrations:

centroid

2ndmoment

It's pretty remarkable how little scatter there is.  Just the offset in Mxx and Myy to figure out.

rmjarvis: Do you want to try to figure it out as part of this PR?  Or did you want to close this one out and make another issue for adding the second kick?  I'm fine either way, but I do think it would be nice to have the second kick working before the next released version if you think that's feasible.

jmeyers314: Hmmm....  I should be able to start tackling this over the weekend, so let's wait until then at least.  I agree that the second kick is important to include before the next release.

rmjarvis:
Rachel, OK if we merge this now?  And have the second kick stuff as a new issue?

rmandelb:
Apologies for the delay.  Yes, you can go ahead.  Since we're all in the same place, perhaps we could chat about the status of other open PRs in the coming days?

Issue #842, opened 2016-12-06
issue building with gcc 5 and Anaconda


esheldon:
I'm getting an error building for ubuntu 16.04.  I'm using the latest anaconda as of this data
```
Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
ImportError: /home/esheldon/anaconda3/bin/../lib/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by .sconf_temp/conftest_31_mod/check_tmv.so)
```

Here is the scons output
```
[esheldon@forest GalSim] scons TMV_LINK=/home/esheldon/local/share/tmv/tmv-link TMV_DIR=/home/esheldon/local
scons: Reading SConscript files ...
SCons is version 2.4.1 using python version 2.7.12
Python is from /usr/include/python2.7
Using the following (non-default) scons options:
   TMV_DIR = /home/esheldon/local
   TMV_LINK = /home/esheldon/local/share/tmv/tmv-link
These can be edited directly in the file gs_scons.conf.
Type scons -h for a full list of available options.
Using python =  /home/esheldon/anaconda3/bin/python
Using default PYPREFIX =  /home/esheldon/anaconda3/lib/python3.5/site-packages
Using compiler: /usr/bin/g++
compiler version: 5.4.0
Determined that a good number of jobs = 8
Checking for C++ header file fftw3.h... yes
Checking for correct FFTW linkage... yes
Checking for boost header files... yes
Boost version is 1.58.0
Checking for C++ header file TMV.h... yes
TMV version is 0.73
Using TMV_LINK file: /home/esheldon/local/share/tmv/tmv-link
     -L/home/esheldon/local/lib -ltmv -lblas -lgfortran -lpthread
Checking for correct TMV linkage... (this may take a little while)
Checking for correct TMV linkage... yes
Checking if we can build against Python... yes
Building for python version 3.5
Checking if we can build module using TMV...
Unable to build a python loadable module that uses tmv
```

esheldon:
the name of this issue is probably a misnomer, since the problem might actually be related to anaconda shipping a libstdc++ and scons finding that instead of the system one

rmjarvis:
Yeah, probably has more to do with anaconda than Ubuntu.  Anaconda have been notoriously problematic for this kind of thing.

Aside: Your TMV_LINK specification is redundant.  You only need to specify TMV_DIR.

rmjarvis:
This site might have some helpful advice:
[...]

PR #848, opened 2016-12-17
#842 Fix a bug in Image.subImage that can cause seg faults on some machines

rmjarvis:
@esheldon [reported](https://github.com/rmjarvis/Piff/issues/50) a bug running Piff unit tests that I eventually tracked down to a bug in GalSim that apparently only shows up for particular systems (Ubuntu Anaconda in his case).

The bug shows up if a suitably large image is loaded from a fits file, and subimages are taken from that image and stored.  Then if the main image goes out of scope and the file is closed, then accessing those stored subimages can cause seg faults.

I'm pretty sure the underlying issue is how we do our wrapping of the numpy arrays and how we set the owner of them, but I didn't fix it there, since I'm planning to rewrite all of that stuff relatively soon anyway as part of the no-boost migration.

Rather, I changed the `Image.subImage` call to make the subarray in python and then make an Image of that.  This means numpy is in charge of persisting the memory, and they do that correctly, so all works fine.

There's also one tiny little edit that I did for the #842 effort of tracking down installation problems Erin was having on that same machine.  It turns out it wasn't super helpful in the end, since Erin ended up switching to miniconda, but I figured I'd leave it in.  Erin was getting linking errors about `throw_out_of_range_fmt` not being found trying to cross-compile across gcc versions.  It turned out that we only have [one place in the code](https://github.com/rmjarvis) that uses that (implicitly in a substr call), and it was easy to get rid of.  So maybe this will help someone else down the line not run into that problem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ Bug Fixes
 - Fixed bug when whitening noise in images based on COSMOS training datasets
   using the config functionality. (#792)
 - Fixed some handling of images with undefined bounds. (#799)
+- Fixed bug in image.subImage that could cause seg faults in some cases. (#848)
 
 
 Deprecated Features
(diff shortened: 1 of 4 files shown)


jmeyers314:
Wow, that's annoyingly specific. 

Code looks good to me.

rmjarvis:
OK.  Thanks both.  Merging now.

Issue #851, opened 2016-12-22
Use basestring instead of str


jmeyers314:
Experienced a tricky issue trying to run through the examples yesterday on an officemate's newly-installed copy of GalSim:  

`Bandpass`, (via demo12) failed to recognize that one of the filter filenames was, in fact, a string.  The issue turned out to be that his home directory as returned by `os.path.split(__file__)` happened to be returned as a `unicode` instance instead of a `str` instance.  

(Actually, it was weirder than that, as it only came back as `unicode` when running through `ipython`, but came back as `str` when run through `python` (2.7).  Took a long time to debug that!) 

The unicode caused the `Bandpass.py` expression: `isinstance(self._orig_tp, str)` to return False, and so Bandpass fell through to initialize `self._tp` as a lambda expression with the unicode as part of the expression.  

The error message was not helpful.  (I believe it was something about blue_limit and/or red_limit not being defined).

I think part of the solution is to add `from past.builtins import basestring`, and then check against basestring instead of str.

We should also try to figure out a more helpful error message, and look through the rest of the code to see if there are any more places where we should replace str with basestring.

rmjarvis:
> from past.builtins import basestring

That's fine with me.  We used to use basestring in a number of places, but in the conversion to python 3 support, I switched them to str and didn't see any problems.  I guess cases like this are probably why people recommend using basestring instead.  All my file systems apparently use regular ascii encoding, so it never showed up.

I suspect any place we do `isinstance(blah, str)` probably ought to switch to using `past.builtins.basestring`.

PR #852, opened 2016-12-24
#851 str -> basestring

jmeyers314:
I went through and replaced a subset of the `isinstance(variable, str)` predicates with `isinstance(variable, basestring)`.  (It didn't seem necessary to replace all of them, like the ones where we check if `scale_unit` is a string, for example.)  

I also improved the eval string sanity check in `SED` and `Bandpass` to check that the inferred lambdas yield a float.

Finally, I made a number of small cleanups suggested by pylint on the files I looked through for cases of `isinstance(var, str)`.

### DIFF
--- a/galsim/bandpass.py
+++ b/galsim/bandpass.py
@@ -212,12 +212,13 @@ class Bandpass(object):
                     test_wave = 700
                 try:
                     self._tp = galsim.utilities.math_eval('lambda wave : ' + self._orig_tp)
-                    self._tp(test_wave)
+                    from numbers import Real
+                    assert isinstance(self._tp(test_wave), Real)
                 except Exception as e:
                     raise ValueError(
                         "String throughput must either be a valid filename or something that "+
                         "can eval to a function of wave.\n" +
-                        "Input provided: {0}\n".format(self._orig_tp) +
+                        "Input provided: {0!r}\n".format(self._orig_tp) +
                         "Caught error: {0}".format(e))
         else:
             self._tp = self._orig_tp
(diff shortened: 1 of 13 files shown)


galsim/interpolatedimage.py
```diff
@@ -525,7 +526,7 @@ def buildNoisePadImage(self, noise_pad_size, noise_pad, rng):
                 # Make sure that we are using a specified RNG by resetting that in this cached
                 # CorrelatedNoise instance, otherwise preserve the cached RNG
                 noise = noise.copy(rng=rng)
-        elif isinstance(noise_pad, str):
+        elif isinstance(noise_pad, basestring):
```
rmjarvis: This isn't covered, but adding this coverage isn't something I want to tackle right now.  (Several other branches here also need coverage.)  I think it's fine to leave this one line uncovered for this PR, unless someone feels inspired to add some more variants to `test_corr_padding` in test_interpolatedimage.py.

galsim/bandpass.py
```diff
@@ -211,12 +212,13 @@ def _initialize_tp(self):
                     test_wave = 700
                 try:
                     self._tp = galsim.utilities.math_eval('lambda wave : ' + self._orig_tp)
-                    self._tp(test_wave)
+                    from numbers import Real
+                    assert isinstance(self._tp(test_wave), Real)
```
jmeyers314: Sure.  Thanks for the clarification.  I've switched this `assert`, and the one in sed.py to `isinstance` checks.

Issue #809, opened 2016-10-04
Switch from SCons to setup.py


rmjarvis:
GalSim is currently kind of a pain to install on some systems.  Most python packages can be installed with pip.  It would be nice to switch GalSim to being able to do that, so `pip install galsim` would go install all the appropriate dependencies along with GalSim. 

The main hurdle as far as making this possible is our boost-python dependency, which [can be hard to install](https://github.com/GalSim-developers/GalSim/wiki/Installation%20FAQ#why-is-scons-unable-to-run-a-program-built-with-boostpython) correctly.  I believe that if we want GalSim to be pip-installable, we need to get rid of our boost-python stuff.  This is a big job.

I'd like to replace all the boost-python wrapping with cffi, which I use in TreeCorr to wrap its C++ code.  It is very straightforward if all the connections between Python and C++ are through C functions.  In Python, cffi makes calling the C function extremely easy.  Then in C++ files we have the C functions call out to the regular C++ functions that use full C++ classes to do whatever they need to do.  This isn't currently how GalSim's middle layer is structured, but I think it would be doable to convert to that.

Here are a list of proposals to effect the conversion (going through the list of files in pysrc):
- Angle.cpp: Make Angle and AngleUnit pure python.  We can enforce all the type checking in the python layer and always pass radians down into C++.  If desired, we could keep a separate Angle class in C++ that the C function could convert back into, but I doubt that will generally be required.
- Bessel.cpp: These are already just functions, so that's easy to switch.
- Bounds.cpp: The essence of this can be converted to a C struct, which cffi can handle.  All the Python-layer functionality can be implemented in pure Python, and this struct can pass down into C with an image.
- CDModel.cpp: This is already just a function.  Easy to wrap.
- CorrelatedNoise.cpp: The only thing here is a deprecated function.  We can just remove it, since I expect the pip-installable version can be considered version 2.0, and all 1.x deprecated functions removed.
- HSM.cpp: HSMParams and ShapeData can be turned into C structs.  FindAdaptiveMom and EstimateShear are already just functions, so easy.
- Image.cpp: This is the first difficult one.  I think we can reorg some of the code so that Images are always constructed in Python, and we use a numpy array for the underlying data.  Then these can be passed down into C++ via `double*` pointers and the shape information.  We may want to keep a C++-layer Image class as well, which the middle layer can reconstitute before passing it on to C++ functions that work on it.
[...]

PR #853, opened 2016-12-24
#809 Simplify wrapping of Image class

rmjarvis:
This is the first of what will likely be quite a few PRs for #809 -- the effort to get rid of boost (especially boost python) from GalSim and migrate to a setup.py installation.  I want to do this in stages, partly to keep the change diff manageable, but also to get some feedback on some of the design decisions.

So, here I'm mostly focused on the Image class.  That was one of our more complicated classes to wrap with all the memory management between the two layers.  So the main thing I did was to have the python layer solely responsible for memory management.  All images that are ever exposed to python now need to have their arrays originally constructed in python.  This means numpy is responsible for all memory management.  Images that are just used within C++ can be constructed in C++ still.  But they cannot be returned back up to python.

The Image class now stores `array` and `bounds` directly, rather than an `image`.  Now `image` is a property that constructs a C++ Image type when needed as an ImageView or ConstImageView.

I also directly implemented in python a few of the simpler methods that are easy to do with the numpy array directly, rather than requiring a C++ implementation.

The upshot is that I was able to vastly simplify the code in pysrc/Image.cpp.  Now the only things that are wrapped are the constructors for ImageView and ConstImageView (ImageAlloc is only available in C++), the fft functions, invertImage, and wrapImage.  i.e. the things that significantly benefit from being implemented in C++ rather than python.

This will be helpful regardless of whether we choose to go with cffi or pybind11 for our new wrapping scheme.  It's definitely required for cffi.  And I started trying a pybind11 conversion, and the Image class had the most problems, where our boost-python implementation didn't translate so easily to pybind11, so this will be helpful there as well.  But I'll defer actually switching to either one to a later PR.  Here, we still use Boost Ptyhon -- it's just not as complicated.

The only functionality loss I think is that the C++ image classes are no longer picklable.  But that's not really necessary, since they are constructed on demand as needed, rather than kept around.  So there's not really any need to pickle them.  Or even have legible strings or reprs.

Well, with one exception.  SBInterpolatedImage and SBInterpolatedKImage had relied on them being picklable.  I managed to fix them, so they are still picklable now.  But I think I like this "make C++ objects on demand when needed" thing, so I may see if I can convert all the GSObjects to work the same way.  Then we won't need all the current hacks to get SBProfile pickling to work properly.  But that will be a separate PR...


rmjarvis:
I forgot to mention -- This PR is to merge into the noboost branch, not master.  This is where I'm planning to stage the eventual 2.0 version.  So these changes (including getting rid of some of the deprecated features) won't mess up anything for our next mainstream release. (1.5)

galsim/hsm.py
```diff
@@ -441,14 +440,14 @@ def EstimateShear(gal_image, PSF_image, weight=None, badpix=None, sky_var=0.0,
     @returns a ShapeData object containing the results of shape measurement.
     """
     # prepare inputs to C++ routines: ImageView for galaxy, PSF, and weight map
```
rmandelb: Some outdated comments in this file (`_convertImage` doesn't return an ImageView after the changes in this PR).  I've flagged this one but I see more elsewhere.

galsim/image.py
```diff
@@ -902,14 +923,16 @@ def good_fft_size(cls, input_size):
...
+        return _galsim.goodFFTSize(int(input_size))
 
     def copyFrom(self, rhs):
         """Copy the contents of another image
         """
         if self.isconst:
             raise ValueError("Cannot modify the values of an immutable Image")
-        self.image.copyFrom(rhs.image)
+        if self.bounds.numpyShape() != rhs.bounds.numpyShape():
+            raise ValueError("Trying to copy images that are not the same shape")
```
rmandelb: Give the shapes so the user can more easily debug?

galsim/image.py
```diff
@@ -1171,30 +1214,58 @@ def addValue(self, *args, **kwargs):
...
+        """
+        self._array[y-self.ymin, x-self.xmin] += value
 
     def fill(self, value):
         """Set all pixel values to the given `value`
         """
         if self.isconst:
             raise ValueError("Cannot modify the values of an immutable Image")
-        self.image.fill(value)
+        if not self.bounds.isDefined(): return
```
rmandelb: Shouldn't there be a warning or something?  (Similar to what happens in `invertSelf` when bounds are not defined)

src/Image.inst
```diff
@@ -0,0 +1,33 @@
+/* -*- c++ -*-
+ * Copyright (c) 2012-2016 by the GalSim developers team on GitHub
```
rmandelb: It's 2017.

rmjarvis:
Thanks Rachel.  I've addressed all your comments.

Issue #787, opened 2016-08-10
weighting galaxy postage stamps for RealGalaxy / COSMOSCatalog


rmandelb:
As described in http://adsabs.harvard.edu/abs/2012MNRAS.420.1518M , the COSMOS galaxy postage stamps were cut out in a way that creates a slight bias against large galaxies.   This trend is a weak one for most of the magnitude range that people use in practice (it's really severe only for the very bright, rare objects).  But still, it would be good to have an automated way to probabilistically reweight the sample to remove this bias.

I anticipated that we might want to do such reweighting, so the catalogs contain the weight factor you'd have to use to correct for this; it's stored as a float between 0-1 and for RealGalaxyCatalog you can access it as

```
rgc = galsim.RealGalaxyCatalog(...)
weight = rgc.weight
```

This issue is for two code changes:

1) The minor change is that I'd like to expose the weights for the COSMOSCatalog class to.  Right now you have to do

```
cc = galsim.COSMOSCatalog()
rgc = galsim.RealGalaxyCatalog(...)
weight = rgc.weight[cc.orig_index]
```

to account for the fact that cuts get applied on COSMOSCatalog but not on RealGalaxyCatalog.  This is admittedly only two extra lines of code, but still, there's no reason not to expose the weight in COSMOSCatalog directly and make it do that bookkeeping internally (and not load the catalog twice!).

2) If you want to use the weights, you'd have to do something like choose an index in the catalog, then use a random number generator like this:

```
ud = galsim.UniformDeviate()
if ud() < weight[ind]:
    # use this galaxy
else:
    # skip this iteration
```

I think it would be useful to have a helper function that returns a set of random indices into the catalog without / with the weights taken into account.  This could be a method of COSMOSCatalog and of RealGalaxy called `rand_indices` with a `weight=True` or `False` kwarg.  If `weight=False` then it would literally just permute the indices of the catalog and return however many of them the user requested.  If `weight=True` then it would also use the weights to remove the size bias.

I think (1) is uncontroversial, but thoughts on (2) are welcome.

I plan to do this after tagging v1.4.

jmeyers314:
Should `blah = RealGalaxy(rgc, random=True)` also respect the weights?

rmandelb:
Hmmm, it should probably have a kwarg to decide how to behave, but i'm not sure what the default should be.

rmandelb:
@rmjarvis - I could use some feedback from you on a code design issue related to config functionality.

[...]

Issue #825, opened 2016-11-08
Implement random choice method of BaseDeviate


jmeyers314:
Pick items from a list, with replacement.  Bonus feature: pick items congruently from multiple lists.

rmandelb:
@jmeyers314 - this is all done with #787 except I didn't make it a method of BaseDeviate.  Should I close anyway, or is that something you're still interested in?

jmeyers314:
Feel free to close.  I think this addresses everything I had in mind.

rmandelb:
Closing #825.

PR #854, opened 2016-12-27
#787

rmandelb:
This PR is for functionality that enables the use of "weights" in the RealGalaxyCatalog and COSMOSCatalog.  When galaxies were selected for inclusion in the catalog, there was a mild galaxy size-dependent trend (because we excluded objects that had pixels too near the edges of the CCD, and this will happen for large galaxies that have centroids in a larger fraction of the CCD area).  Now it is possible to randomly choose a galaxy in a way that removes that bias in the postage stamp creation process, and indeed, using those weights is the default behavior.

I made demo11 use this new behavior.

Also, I added a bunch of unit tests but I suspect it's insufficient, especially in the new config functionality.  I will use the coverage report after submitting the PR to add a few more unit tests, but otherwise this is done.

### DIFF
--- a/galsim/config/input_cosmos.py
+++ b/galsim/config/input_cosmos.py
@@ -98,7 +98,8 @@ def _BuildCOSMOSGalaxy(config, base, ignore, gsparams, logger):
             kwargs['rng'] = galsim.config.check_for_rng(base, logger, 'COSMOSGalaxy')
 
     # NB. Even though index is officially optional, it will always be present, either because it was
-    #     set by a call to selectRandomIndices or because it was specified explicitly.
+    #     set by a call to selectRandomIndices, explicitly by the user, or due to the call to
+    #     SetDefaultIndex.
     index = kwargs['index']
     if index >= cosmos_cat.getNObjects():
         raise IndexError(
--- a/galsim/config/input_cosmos.py
+++ b/galsim/config/input_cosmos.py
@@ -79,9 +79,17 @@ def _BuildCOSMOSGalaxy(config, base, ignore, gsparams, logger):
     if gsparams: kwargs['gsparams'] = galsim.GSParams(**gsparams)
 
     if 'index' not in kwargs:
-        make_kwargs = {}
-        make_kwargs['rng'] = galsim.config.check_for_rng(base, logger, 'COSMOSGalaxy')
-        kwargs['index'] = cosmos_cat.selectRandomIndices(1, **make_kwargs)
+        make_kwargs = {'_n_rng_calls': True}
+        rng = galsim.config.check_for_rng(base, logger, 'COSMOSGalaxy')
+        make_kwargs['rng'] = rng
+        kwargs['index'], n_rng_calls = cosmos_cat.selectRandomIndices(1, **make_kwargs)
+
+        # Make sure this process gives consistent results regardless of the number of processes
+        # being used.
+        if not isinstance(cosmos_cat, galsim.COSMOSCatalog) and rng is not None:
+            # Then cosmos_cat is really a proxy, which means the rng was pickled, so we need to
+            # discard the same number of random calls from the one in the config dict.
+            rng.discard(n_rng_calls)
 
     if 'gal_type' in kwargs and kwargs['gal_type'] == 'real':
         if 'make_kwargs' in locals():
--- a/galsim/config/input_cosmos.py
+++ b/galsim/config/input_cosmos.py
@@ -78,11 +78,10 @@ def _BuildCOSMOSGalaxy(config, base, ignore, gsparams, logger):
         ignore = ignore)
     if gsparams: kwargs['gsparams'] = galsim.GSParams(**gsparams)
 
+    rng = None
     if 'index' not in kwargs:
-        make_kwargs = {'_n_rng_calls': True}
         rng = galsim.config.check_for_rng(base, logger, 'COSMOSGalaxy')
-        make_kwargs['rng'] = rng
-        kwargs['index'], n_rng_calls = cosmos_cat.selectRandomIndices(1, **make_kwargs)
+        kwargs['index'], n_rng_calls = cosmos_cat.selectRandomIndices(1, rng=rng, _n_rng_calls=True)
 
         # Make sure this process gives consistent results regardless of the number of processes
         # being used.
@@ -92,10 +91,9 @@ def _BuildCOSMOSGalaxy(config, base, ignore, gsparams, logger):
             rng.discard(int(n_rng_calls))
 
     if 'gal_type' in kwargs and kwargs['gal_type'] == 'real':
-        if 'make_kwargs' in locals():
-            kwargs['rng'] = make_kwargs.get('rng',None)
-        else:
-            kwargs['rng'] = galsim.config.check_for_rng(base, logger, 'COSMOSGalaxy')
+        if rng is None:
+            rng = galsim.config.check_for_rng(base, logger, 'COSMOSGalaxy')
+        kwargs['rng'] = rng
 
     # NB. Even though index is officially optional, it will always be present, either because it was
     #     set by a call to selectRandomIndices, explicitly by the user, or due to the call to
--- a/galsim/config/input_cosmos.py
+++ b/galsim/config/input_cosmos.py
@@ -89,7 +89,7 @@ def _BuildCOSMOSGalaxy(config, base, ignore, gsparams, logger):
         if not isinstance(cosmos_cat, galsim.COSMOSCatalog) and rng is not None:
             # Then cosmos_cat is really a proxy, which means the rng was pickled, so we need to
             # discard the same number of random calls from the one in the config dict.
-            rng.discard(n_rng_calls)
+            rng.discard(int(n_rng_calls))
 
     if 'gal_type' in kwargs and kwargs['gal_type'] == 'real':
         if 'make_kwargs' in locals():
--- a/galsim/config/input_cosmos.py
+++ b/galsim/config/input_cosmos.py
@@ -66,9 +66,6 @@ def _BuildCOSMOSGalaxy(config, base, ignore, gsparams, logger):
 
     ignore = ignore + ['num']
 
-    # Special: if index is Sequence or Random, and max isn't set, set it to nobjects-1.
-    galsim.config.SetDefaultIndex(config, cosmos_cat.getNObjects())
-
     kwargs, safe = galsim.config.GetAllParams(config, base,
         req = galsim.COSMOSCatalog.makeGalaxy._req_params,
         opt = galsim.COSMOSCatalog.makeGalaxy._opt_params,
@@ -76,11 +73,19 @@ def _BuildCOSMOSGalaxy(config, base, ignore, gsparams, logger):
         ignore = ignore)
     if gsparams: kwargs['gsparams'] = galsim.GSParams(**gsparams)
 
-    if 'gal_type' in kwargs and kwargs['gal_type'] == 'real':
-        kwargs['rng'] = galsim.config.check_for_rng(base, logger, 'COSMOSGalaxy')
+    if 'index' not in kwargs:
+        make_kwargs = {}
+        make_kwargs['rng'] = galsim.config.check_for_rng(base, logger, 'COSMOSGalaxy')
+        kwargs['index'] = cosmos_cat.selectRandomIndices(1, **make_kwargs)
 
-    # NB. Even though index is officially optional, the call to SetDefaultIndex above
-    #     means that it will always be present.
+    if 'gal_type' in kwargs and kwargs['gal_type'] == 'real':
+        if 'make_kwargs' in locals():
+            kwargs['rng'] = make_kwargs.get('rng',None)
+        else:
+            kwargs['rng'] = galsim.config.check_for_rng(base, logger, 'COSMOSGalaxy')
+
+    # NB. Even though index is officially optional, it will always be present, either because it was
+    #     set by a call to selectRandomIndices or because it was specified explicitly.
     index = kwargs['index']
     if index >= cosmos_cat.getNObjects():
         raise IndexError(
--- a/galsim/config/input_cosmos.py
+++ b/galsim/config/input_cosmos.py
@@ -78,6 +78,14 @@ def _BuildCOSMOSGalaxy(config, base, ignore, gsparams, logger):
         ignore = ignore)
     if gsparams: kwargs['gsparams'] = galsim.GSParams(**gsparams)
 
+    # Deal with defaults for gal_type, if it wasn't specified:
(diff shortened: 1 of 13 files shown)


rmandelb:
I appear to have broken the build on python 3 (not too surprising, since I don't have a python 3 system to test on myself, but I will look into the details tomorrow).  And as expected, a bunch of the diff does not have unit tests; looking at the report, I have a pretty good idea how to improve it significantly...

galsim/scene.py
```diff
@@ -568,6 +581,86 @@ def makeGalaxy(self, index=None, gal_type=None, chromatic=False, noise_pad_size=
...
+                test_vals[mask] = new_test_vals
+                if _n_rng_calls:
+                    n_rng_calls += 2*n_fail
+                # Finally, update the test array used to determine whether any galaxies failed.
+                mask = test_vals > self.real_cat.weight[self.orig_index[index]]
+        else:
+            import warnings
+            warnings.warn('Selecting random object without correcting for catalog-level '
+                          'selection effects (requires existence of real catalog with '
+                          'weights in addition to parametric one).')
```
rmjarvis: This comment reads a little funny.  Mostly because it makes grammatical sense if you drop the parentheses, but means exactly the opposite of what you want.  Maybe add "Such correction" to the start of the parenthetical clause.

tests/test_config_gsobject.py
```diff
@@ -652,11 +653,17 @@ def test_realgalaxy():
     assert "No base['rng'] available" in cl.output
 
     config['obj_num'] = 5
-    gal1a = galsim.config.BuildGSObject(config, 'gal6')[0]
-    gal1b = galsim.RealGalaxy(real_cat, index=0).original_gal
+    gal6a = galsim.config.BuildGSObject(config, 'gal6')[0]
+    gal6b = galsim.RealGalaxy(real_cat, index=0).original_gal
     # The convolution here
     gsobject_compare(gal1a, gal1b, conv=conv)
```
rmjarvis: You forgot to update this to gal6 as well.

tests/test_config_gsobject.py
```diff
@@ -595,6 +595,7 @@ def test_realgalaxy():
                  },
         'gal5' : { 'type' : 'RealGalaxy' , 'index' : 23, 'noise_pad_size' : 10 },
         'gal6' : { 'type' : 'RealGalaxyOriginal' },
+        'gal7' : { 'type' : 'RealGalaxy' , 'random' : True}
```
rmjarvis: I guess I don't have any tests here of config builds of COSMOSGalaxy.  Probably worth adding a few, including the new random index feature.  (Looks like I use COSMOSGalaxy in a few tests in `test_config_image.py`, but they don't really test for accuracy.)

galsim/config/input_cosmos.py
```diff
@@ -76,11 +78,28 @@ def _BuildCOSMOSGalaxy(config, base, ignore, gsparams, logger):
...
+        rng = galsim.config.check_for_rng(base, logger, 'COSMOSGalaxy')
+        make_kwargs['rng'] = rng
+        kwargs['index'], n_rng_calls = cosmos_cat.selectRandomIndices(1, **make_kwargs)
+
+        # Make sure this process gives consistent results regardless of the number of processes
+        # being used.
+        if not isinstance(cosmos_cat, galsim.COSMOSCatalog) and rng is not None:
+            # Then cosmos_cat is really a proxy, which means the rng was pickled, so we need to
+            # discard the same number of random calls from the one in the config dict.
+            rng.discard(int(n_rng_calls))
```
rmandelb: Thanks.  That would have been very hard for me to figure out without a python3 system!

galsim/config/input_cosmos.py
```diff
@@ -76,11 +78,28 @@ def _BuildCOSMOSGalaxy(config, base, ignore, gsparams, logger):
         ignore = ignore)
     if gsparams: kwargs['gsparams'] = galsim.GSParams(**gsparams)
 
+    if 'index' not in kwargs:
+        make_kwargs = {'_n_rng_calls': True}
+        rng = galsim.config.check_for_rng(base, logger, 'COSMOSGalaxy')
+        make_kwargs['rng'] = rng
+        kwargs['index'], n_rng_calls = cosmos_cat.selectRandomIndices(1, **make_kwargs)
```
rmandelb: Sure.  Done.

galsim/real.py
```diff
@@ -457,8 +466,17 @@ def __init__(self, file_name=None, sample=None, image_dir=None, dir=None, preloa
...
         self.band = self.cat.field('band') # bandpass in which apparent mag is measured, e.g., F814W
-        self.weight = self.cat.field('weight') # weight factor to account for size-dependent
-                                               # probability
+        # Add the weight factor to the catalog only if it behaves properly.  In particular, it
+        # should be a float between 0 and 1 (so that random selections of indices can use it to
+        # remove any selection effects in the catalog creation process).  If the weight does not
+        # behave like this, then the catalog simply won't have a weight column.
+        weight = self.cat.field('weight')
+        if np.min(weight) >= 0. and np.max(weight) <= 1. and \
+                not np.any(np.isnan(weight)) and not np.any(np.isinf(weight)):
```
rmandelb: We don't for the existing catalogs that I've produced, but if we want the code to work for others that people might produce, I thought we should include this check.  We could optionally have a comment that nans and infs will cause problems and users who supply their own catalogs should make sure to check for them themselves when producing the catalogs.

galsim/scene.py
```diff
@@ -568,6 +581,86 @@ def makeGalaxy(self, index=None, gal_type=None, chromatic=False, noise_pad_size=
...
             return gal_list[0]
 
+    def selectRandomIndices(self, n_random, rng=None, _n_rng_calls=False):
+        """
+        Routine to select random indices out of the catalog.  This routine does a weighted random
+        selection with replacement (i.e., there is no guarantee of uniqueness of the selected
+        indices).  Weighting uses the weight factors available in the catalog, if any; these weights
+        are typically meant to remove any selection effects in the catalog creation process.
+
+        @param n_random     Number of random indices to return.
```
rmandelb: Sure.  Done.

galsim/scene.py
```diff
@@ -568,6 +581,86 @@ def makeGalaxy(self, index=None, gal_type=None, chromatic=False, noise_pad_size=
...
+        @param rng          A random number generator to use for selecting a random galaxy
+                            (may be any kind of BaseDeviate or None). [default: None]
+        @returns A NumPy array containing the randomly-selected indices.
```
rmandelb: Fixed.

galsim/scene.py
rmandelb: Changed it into two sentences for clarity.

tests/test_config_gsobject.py
rmandelb: Yes, that's exactly what I was thinking of doing - a set of new tests for COSMOSGalaxy.  Will do shortly and post when done adding unit tests.

tests/test_config_gsobject.py
rmandelb: Ugh, thanks.  Fixed.

rmandelb:
OK... I added a bunch of unit tests.  The status is currently pending for the latest commit, but everything passed (coverage of the diff etc.) on the previous commit, so it should be all good now.  And as usual, this process led to a few changes in the code itself as the unit tests caught things that were missing.

I'm wondering if we should deprecate the ```simReal``` routine (and remove it in the next version).  It is not well-unit-tested, not used in examples, and I'm not sure how often people actually use it.  Frankly I thought I would use it myself and then I haven't really done so; I've been doing most of my sims via config.  Thoughts?

galsim/real.py
rmjarvis: I'm not sure this whole if statement is required.  Currently, codecov tells us that we only ever do the `True` branch here.  So I'd suggest just always dividing by the max weight, rather than doing this test.  And just wherever we document how users can make their own catalog (Do we have such instructions?) we say that they should be float values > 0 and when used, they will be normalized by the maximum weight value.

rmandelb: The "if" statement is not required for the current catalogs, which were produced with weights that go according to the rules. That's why codecov says we always do the "true" branch.  I could make one that would fail this check for testing purposes, but it sounds like you are suggesting I remove the check, so I will go with that.

We don't currently have separate instructions for producing a catalog.  I tried to make the README sufficiently detailed that it can basically serve as instructions.  The documentation on Sowmya's branch could become a more official set of instructions.

rmandelb:
Hmmm, travis failed in a very strange way that doesn't have anything to do with compilation or unit test failures from what I can see, unless I am misreading the output.  Does anybody understand what happened there?

rmjarvis:
> Frankly I thought I would use it myself and then I haven't really done so; I've been doing most of my sims via config. Thoughts?

Fine with me.

rmandelb:
And am I right in thinking I should deprecate it (not remove entirely just yet)?  Or is this a circumstance where just removing it is OK?  I'm not sure there's any circumstance where that is OK, actually...

rmjarvis:
> Hmmm, travis failed in a very strange way that doesn't have anything to do with compilation or unit test failures from what I can see, unless I am misreading the output. Does anybody understand what happened there?

I think related to the recent update to coverage 4.3.  Dropping back to 4.1 (what I have on my machinee -- but 4.2 would probably have also worked) fixed it.  I filed a [bug report](https://bitbucket.org/ned/coveragepy/issues/543/coveragemisccoverageexception-include-and) with coverage.

rmjarvis:
> And am I right in thinking I should deprecate it (not remove entirely just yet)?

Yes.  I think deprecating it is the thing to do.  I don't know if anyone uses this function, but it gives them a chance to complain if they do.

rmjarvis:
You can also add `# pragma: no cover` after it when you do and your coverage diff should go up to 100%.  😁

rmandelb:
OK, I took a stab at this.  It behaves properly in the sense that if I call simReal it does its business will giving the appropriate deprecation warning, but I wasn't sure whether I should also be moving the routine wholesale into galsim/depr/ (which I did not do - for now, it's still in real.py with the deprecation warning).

Also wasn't sure if I put the ```#pragma: no cover``` in the right place, but I will find out once the coverage test is complete :)

galsim/real.py
```diff
@@ -731,6 +728,11 @@ def simReal(real_galaxy, target_PSF, target_pixel_scale, g1=0.0, g2=0.0, rotatio
 
     @return a simulated galaxy image.
     """
+    # pragma: no cover
```
rmjarvis: This goes at the end of the `def` line.  After the colon.

rmandelb: Thank you.

rmjarvis:
> it behaves properly in the sense that if I call simReal it does its business while giving the appropriate deprecation warning,

Also, should have a test in `test_deprecated` that checks this.  I usually just copy over any current tests that use the deprecated feature and change the deprecated calls to look like (in your case)
```
sim_image = check_dep(galsim.simReal,
                            rg, targ_PSF, tps,
                            g1 = targ_applied_shear1, g2 = targ_applied_shear2,
                            rand_rotate = False, target_flux = fake_gal_flux)
```

rmandelb:
Done.

galsim/scene.py
```diff
@@ -568,6 +581,87 @@ def makeGalaxy(self, index=None, gal_type=None, chromatic=False, noise_pad_size=
         else:
             return gal_list[0]
 
+    def selectRandomIndex(self, n_random=1, rng=None, _n_rng_calls=False):
```
jmeyers314: This method reminds me of #825.  I think that issue is actually solved simply by something like

```python
result = np.zeros(n_random)
ud.generate(result)
return (result*n_choices).astype(int)
```

but it might be nice to also include the possibility of weights.  So I think it would nice to turn the logic here into a utility and mark #825 as completed too.

galsim/real.py
```diff
@@ -179,8 +178,15 @@ def __init__(self, real_galaxy_catalog, index=None, id=None, random=False,
                     raise AttributeError('Too many methods for selecting a galaxy!')
                 use_index = real_galaxy_catalog.getIndexForID(id)
             elif random is True:
-                uniform_deviate = galsim.UniformDeviate(self.rng)
-                use_index = int(real_galaxy_catalog.nobjects * uniform_deviate())
+                ud = galsim.UniformDeviate(self.rng)
```
rmandelb: Fixed.

galsim/scene.py
rmandelb: Good catch.  Is what I did just now in https://github.com/GalSim-developers/GalSim/pull/854/commits/958b8e98fabf46e8d4c6d4a0c0651e7bb64c7fd5 what you had in mind?  Or did you also have in mind to add a method of BaseDeviate that calls this utility?

galsim/utilities.py
```diff
@@ -1208,3 +1208,84 @@ def generate():
             c *= b_over_a * (n-i)/(i+1)
             yield c
     return np.fromiter(generate(), float, n+1)
+
+def rand_with_replacement(n, rng, n_choices, weight=None, _n_rng_calls=False):
```
jmeyers314: Looks good.  My only thought is that we might want to reorder the call signature to `n, n_choices, rng`, so that once #824 is merged and `None` becomes a valid seed for `UniformDeviate` we can transparently make `rng=None` the default here.

rmandelb: Sure.  Swapped order.

rmandelb:
All comments to date have been addressed.  I will give it until Wednesday morning before merging; if there are any further comments or objections, please let me know before then.

tests/test_utilities.py
```diff
@@ -460,6 +460,44 @@ def test_python_LRU_Cache():
...
+    except ImportError:
+        print("The assert_raises tests require nose")
+
+    # Make sure results come out the same whether we use _n_rng_calls or not.
+    result_1 = galsim.utilities.rand_with_replacement(n=10, n_choices=100,
+                                                      rng=galsim.BaseDeviate(314159))
+    result_2, _ = galsim.utilities.rand_with_replacement(n=10, n_choices=100,
+                                                         rng=galsim.BaseDeviate(314159),
+                                                         _n_rng_calls=True)
+    assert np.all(result_1==result_2),"Using _n_rng_calls results in different random numbers"
```
rmjarvis: Should repeat this test when there are non-trivial weights, since right now you don't run the code with weights, but `_n_rng_calls=False`.

rmandelb:
Sorry, for some reason I thought that case was covered by the tests that happen via COSMOSCatalog().  Added it just now.

rmandelb:
Merging.

rmandelb:
@rmjarvis - do you want to incorporate this into the noboost branch now that it's on master?

rmjarvis:
Sure.  Merged.

PR #857, opened 2017-01-03
#809b Move Angle class from C++ to Python

rmjarvis:
This is the second installment in issue #809, simplifying the python/C++ interface layer.

The Angle class mostly doesn't do anything useful in the C++ layer.  We pretty much only use it for type checking in the python layer.  By the time things get down to C++, we've almost always processed things into something other than an Angle already.

This isn't particularly a problem or anything, but in the interest of simplifying the python-wrapping interface, I reimplemented the Angle and AngleUnit classes in pure python.  The only thing it calls down to C++ for is the `sincos` function to calculate both sin and cos in a single call.  

And in fact, that function is made simpler in the C++ layer by extracting it from the Angle class, so I think that's a win as well.

Finally, I discovered that you can't set the `__doc__` attribute of pybind11-wrapped classes from python, which we do in a number of places currently for the boost python classes, including Angle and AngleUnit.  So there will probably be a few more of these conversions to move the primary implementation of some classes up to python with regular doc strings and just call down to C++ for the heavy computational bits.


rmjarvis:
Update: This now uses the LSSTDESC.Coord repo for Angle, CelestialCoord, and related functionality.  In fact, the previous version of this branch was the source code for what I ported over to that repository with the expectation of making it a dependency of GalSim.  So now that's what this PR is.

The coverage drop for coveralls is just due to the fact that this removed a little more than 1% of the code in GalSim.  So even though there isn't any new lack of coverage, the denominator decreased, making the percentage covered decrease.  I temporarily changed the coveralls threshold to 2% (from 1%) to allow this on this branch.

rmandelb:
The changes on here are complete?  I was going to tackle this one tomorrow, if so (aiming for a rate of one review per day for the next little while).

rmjarvis:
Yep.  All set.  Thanks.

CHANGELOG.md
```diff
@@ -1,12 +1,20 @@
 Changes from v1.5 to v2.0
 =========================
 
+Dependency Changes
+------------------
+
+- Added LSSTDESC.Coord. (#809b)
```
rmandelb: Maybe say why?

CHANGELOG.md
```diff
@@ -1,12 +1,20 @@
...
 -----------
 
-
-Dependency Changes
-------------------
+- There were some minor API changes to the Angle and CelestialCoord classes we made when we
+  moved it over into LSSTDESC.Coord.  Some were to sever (weak) ties to other GalSim classes and
+  some were just deemed API improvements.  Probably the most relevant of these for most
+  users will be that Angle.rad is now a property rather than a function.  So the value of an
+  angle in radians is just `theta.rad` not `theta.rad()`. (#809b)
```
rmandelb: Huh, I could see that being a pain.  No way to be backwards compatible, or is the philosophy here that going to v2 is a big enough change that people should expect instantaneous API changes anyway?

examples/demo11.py
```diff
@@ -209,10 +209,12 @@ def main(argv):
     theta = 0.17 * galsim.degrees
     # ( dudx  dudy ) = ( cos(theta)  -sin(theta) ) * pixel_scale
     # ( dvdx  dvdy )   ( sin(theta)   cos(theta) )
-    dudx = math.cos(theta.rad()) * pixel_scale
-    dudy = -math.sin(theta.rad()) * pixel_scale
-    dvdx = math.sin(theta.rad()) * pixel_scale
-    dvdy = math.cos(theta.rad()) * pixel_scale
+    # Aside: You can call numpy trig functions on Angle objects directly, rather than getting
```
rmandelb: That's cool.  Question: later on I see exactly one place (galsim/config/value.py) where you do `theta.cos()`.  Is there any effective difference (speed or whatever) compared to doing `numpy.cos(theta)`?  Also, may be worth mentioning both options here?

rmjarvis:
> module compiled against API version a but this version of numpy is 9

[This](https://stackoverflow.com/questions/33859531/runtimeerror-module-compiled-against-api-version-a-but-this-version-of-numpy-is) suggests that upgrading your numpy might fix it:
```
pip install -U numpy
```

OTOH, someone reported that it didn't fix it for them.  So you might need to investigate further.  You might need to find all the old versions of numpy on your system and delete them all.

CHANGELOG.md
rmjarvis: Major versions are where API changes are allowed.  We'd like to keep them to a minimum of course, but if there are any API improvements that we want to make, this is the time to do it.

Personally, I much prefer this API , making rad a property.  It feels much more Pythonic to me, and I think it was a mistake to have made it a function originally.  (Stemming from my lack of experience in Python at the time I originally wrote that class.)

examples/demo11.py
rmjarvis: There is no difference.  

In fact, the numpy implementation for trig acting on objects looks for the corresponding method and calls that.  So `numpy.cos(theta)` just calls `theta.cos()` behind the scenes.  Doing the latter directly is (very) slightly more efficient, since it doesn't include the time for numpy to do that check.  Other than that, they are completely equivalent.

rmjarvis:
> I imagine that the fact that we have such good unit test coverage must make doing this kind of thing correctly so much easier!

Indeed.  This was a point I made in my DE School lesson on unit testing.  Good unit tests do make refactoring code much easier to tell that you aren't breaking anything, so much safer.

I addressed the couple points you made.  Thanks!

rmandelb:
The numpy thing is stupid.  After investigating, I realized that the issue is that it was using an older NumPy that was not from the Canopy distribution (though I am using the Canopy python, and its site-packages directory is the very first thing in my PYTHONPATH).  So that's the real source of the problem.  I bet if it was using the one from Canopy then it would be fine (actually I'm kind of confused about why it works at all right now).

And that also makes me wonder if it's using the wrong astropy and other things.

examples/demo11.py
rmandelb: Great, thanks.

rmandelb:
Also I am confused, I submitted my review but it still says "Review required".

rmjarvis:
> I submitted my review but it still says "Review required".

I think you probably accidentally reviewed as a Comment rather than Request Changes.  I've done that before too.

rmandelb:
I probably did, but it says "rmandelb reviewed 2 hours ago" (with the little eye picture next to it) so I thought that was a review.  Whatever.  I will try again.

PR #858, opened 2017-01-11
Fix galsim_download_cosmos py3 compatibility

jmeyers314:
Quick PR to fix the galsim_download_cosmos script to be python3 compatible (I verified that python2.7 still works).

Note, we also use urllib2 in getLSSTBandpasses.py and getSEDs.py, but I'm willing to postpone addressing those until #640 is ready, where I've reorganized some of the Bandpass/SED stuff (should be soon, I promise!).


jmeyers314:
Sure.  I'd be happy to code up a "download the example" command-line option.  Do we have a suitable tarball ready somewhere?

rmjarvis:
Yeah, we don't want to have travis download the full cosmos catalog each time.  

One option would be to cache the resulting download, so it would download it once, but then later runs would just recognize that it already existed and wouldn't download again.  Then if we make any edits to the `galsim_download_cosmos` script we'd want to manually flush the cache (which you can do on the Travis site).

Or the other option is what you suggested, Josh, which is to have a special download file that would really only be relevant for Travis testing, which is rather small, so would be a quick download.

I think I'm slightly more in favor of the first option.  But either is ok with me.

rmandelb:
Josh, we don't have a suitable example tarball, but the ```real_galaxy_*``` files in examples/data/ are the right size for this kind of thing (100 galaxies).  If you tar them up we could ask Joe to put them on the GREAT3 server?

jmeyers314:
>One option would be to cache the resulting download, so it would download it once, but then later runs would just recognize that it already existed and wouldn't download again.

I was just looking into this and found the following snippet in the travis-ci cache instructions:

`Large files that are quick to install but slow to download do not benefit from caching, as they take as long to download from the cache as from the original source`

We could still try it and see, of course.  Presumably the travis cache server is faster than the GREAT3 server.

rmjarvis:
> Presumably the travis cache server is faster than the GREAT3 server.

Much faster, I expect.  There are some other things in there already that violate their suggestion (e.g. the DES catalogs for the des tests) and it made a big difference to cache them.

jmeyers314:
Well, I managed to make a ~13 GB cache ...  which times out on download with the default settings. :(

Since we're just trying to get the script to run though, and not actually use the downloaded cosmos data, I suppose it would be enough to cache a single file from the dataset instead of the whole thing.

Even better though, would be to run `galsim_download_cosmos` only if it's actually changed.  I think I'll look into that for a while...

rmjarvis:
Mm.  13 GB is a lot.  Let's not do that then.

I was going to suggest that we add another sample option (`test` maybe) that would be a much smaller tarball than 23.5 and 25.2.  But looking through the code for this, there are a lot of if statements that run different parts of the code for different use cases.  e.g. installing into an alternate directory or saving the tarball after unpacking, etc.  Some of which interact with the user, which we can't test on travis.

So I'm wondering if this would be better served by us just doing our best to try out different options on various systems to try to make sure it works, and then rely on user bug reports if we miss something.

jmeyers314:
I think I got the "only run galsim_download_cosmos if it changed" version of .travis.yml working now.  (Still running, but looking good...).  It certainly doesn't cover every user option, as you say, but I think it's lightweight most of the time (when galsim_download_cosmos.py doesn't change), and it at least covers some of the script when it does change.

jmeyers314:
Ack!  Spoke too soon.  The py3.6 test just timed out b/c it had to assemble the complete cache (boost, pip, des, cosmos) all from scratch.  We can probably get this to work though by adding a few things to the cache at a time.

rmjarvis:
Or don't cache the cosmos download, right?  If we're only testing it when it changed, then we don't need to cache the resulting downloaded file.

jmeyers314:
Right.  I'm actually not caching the cosmos (I mistyped), but I'm still downloading it.

rmjarvis:
I see.  Right.  Yeah, should work if you make the cache first and then add the download test.  This seems like a good solution to me.  I didn't know about that TRAVIS_COMMIT_RANGE variable.  That's pretty nice.

jmeyers314:
Okay.  I think it's all working now.  As I'm sure you saw, @rmjarvis , I went ahead and added python 3.6 travis testing to this branch too.  I'll plan to merge Friday if I don't hear anything more.

rmjarvis:
There seem to be a lot of warnings in Python 3 that don't happen in Python 2.  More still in 3.6.

Python 3.4 and 3.5 have a number of:
```
/home/travis/virtualenv/python3.4.4/lib/python3.4/site-packages/galsim/scene.py:605: UserWarning: Selecting random object without correcting for catalog-level selection effects.  This correction requires the existence of real catalog with valid weights in addition to parametric one.
  warnings.warn('Selecting random object without correcting for catalog-level '
```
which I don't understand, since that's our warning, and it doesn't happen in 2.7.  They also warn about an unclosed file, which is probably worth fixing:
```
/home/travis/build/GalSim-developers/GalSim/tests/galsim_test_helpers.py:550: ResourceWarning: unclosed file <_io.FileIO name='des_data/DECam_00154912_12_psfcat.psf' mode='rb' closefd=True>
```

Then Python 3.6 adds a bunch of:
```
/home/travis/virtualenv/python3.6.0/lib/python3.6/importlib/_bootstrap.py:205: ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
  return f(*args, **kwds)
```
which I don't understand at all.  And these:
```
/opt/python/3.6.0/lib/python3.6/subprocess.py:761: ResourceWarning: subprocess 32202 is still running
  ResourceWarning, source=self)
/opt/python/3.6.0/lib/python3.6/subprocess.py:761: ResourceWarning: subprocess 32209 is still running
  ResourceWarning, source=self)
```
which is also very strange.  How did we leave a subprocess running?  Doesn't make any sense to me.

You said you tried running Python 3.6 on your system.  Do you also see these when you run the tests?  If so, maybe you can try to investigate a bit to see if there are easy fixes for any of them.

(Aside: the matplotlib warning is something I consider to be a bug in matplotlib, although they consider it a feature.  I think it should go away with the next release of starlink, which is the package that erroneously depends on matplotlib, but I think Tim fixed that for the next version.)

jmeyers314:
Sure.  I can take a look at those.

Do you happen to know how to get a full traceback out of nose on warnings?  (I tried google but haven't seen anything useful yet.)

rmjarvis:
No, but if you do `nosetests -v` that should at least tell you which function they happen in.  Then you can run that file manually to get more detail.

jmeyers314:
>Python 3.4 and 3.5 have a number of:
>/home/travis/virtualenv/python3.4.4/lib/python3.4/site-packages/galsim/scene.py:605: UserWarning: Selecting random object without correcting for catalog-level selection effects.  This correction requires the existence of real catalog with valid weights in addition to parametric one.
  warnings.warn('Selecting random object without correcting for catalog-level '

This one was easy, py2.7 caught the first of these warnings in a `with warnings.simple_filter()` block, which then made all the subsequent identical warnings disappear.  py3 also caught the first warning, but didn't disappear the subsequent ones.  So I extended the range of the `with` block.

>/home/travis/build/GalSim-developers/GalSim/tests/galsim_test_helpers.py:550: ResourceWarning: unclosed file <_io.FileIO name='des_data/DECam_00154912_12_psfcat.psf' mode='rb' closefd=True>

Fixed.

 >/home/travis/virtualenv/python3.6.0/lib/python3.6/importlib/_bootstrap.py:205: ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
  return f(*args, **kwds)

This one is trickier, must be related to https://bugs.python.org/issue25791.  My best guess is that it's actually a scipy problem, since it goes away if I delete both `test_detectors.py` and `test_table.py`, but remains if either one of those is present.  We also use scipy in test_bessel, but that one doesn't seem to trigger this.

>/opt/python/3.6.0/lib/python3.6/subprocess.py:761: ResourceWarning: subprocess 32202 is still running
  ResourceWarning, source=self)
/opt/python/3.6.0/lib/python3.6/subprocess.py:761: ResourceWarning: subprocess 32209 is still running
  ResourceWarning, source=self)

Still working on this one...

tests/test_des.py
```diff
@@ -564,6 +564,7 @@ def test_psf():
     # Also check the functionality where the file is already open.
     hdu_list = pyfits.open(os.path.join(data_dir,psfex_file))
     psfex = galsim.des.DES_PSFEx(hdu_list[1])
+    hdu_list.close()
```
rmjarvis: Probably slightly better would be
```
with pyfits.open(os.path.join(data_dir,psfex_file)) as hdu_list:
    psfex = galsim.des.DES_PSFEx(hdu_list[1])
```

jmeyers314:
>/opt/python/3.6.0/lib/python3.6/subprocess.py:761: ResourceWarning: subprocess 32202 is still running
ResourceWarning, source=self)
/opt/python/3.6.0/lib/python3.6/subprocess.py:761: ResourceWarning: subprocess 32209 is still running
ResourceWarning, source=self)

This one is weird.  It seems to be occurring during test_image.py, but only shows up if I run all the tests, not if I try just testing test_image.py.

rmjarvis:
Which function in test_image.py?

jmeyers314:
I think test_Image_CubeFITS_IO.  Here's part of the output of `nosetests test_* -vvv --processes=1`:

```
...
Test that reading FITS files with an invalid data type succeeds by converting the ... ok
nose.plugins.multiprocess: DEBUG: Ran 1 tests (total: 461)
nose.selector: DEBUG: Test name /Users/josh/src/GalSim/tests/test_image.py:test_Image_CubeFITS_IO resolved to file /Users/josh/src/GalSim/tests/test_image.py, module None, call test_Image_CubeFITS_IO
nose.plugins.multiprocess: DEBUG: Waiting for results (457/881 tasks), next timeout=60.000s
nose.selector: DEBUG: Final resolution of test name /Users/josh/src/GalSim/tests/test_image.py:test_Image_CubeFITS_IO: file /Users/josh/src/GalSim/tests/test_image.py module test_image call test_Image_CubeFITS_IO
nose.importer: DEBUG: Import test_image from /Users/josh/src/GalSim/tests
nose.importer: DEBUG: Add path /Users/josh/src/GalSim/tests
nose.importer: DEBUG: find module part test_image (test_image) in ['/Users/josh/src/GalSim/tests']
nose.importer: DEBUG: sys.modules has test_image as <module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'>
nose.importer: DEBUG: module already loaded? mod: /Users/josh/src/GalSim/tests new: /Users/josh/src/GalSim/tests
nose.loader: DEBUG: load from test_Image_CubeFITS_IO (<module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'>)
nose.selector: DEBUG: Test name test_Image_CubeFITS_IO resolved to file None, module test_Image_CubeFITS_IO, call None
nose.selector: DEBUG: Final resolution of test name test_Image_CubeFITS_IO: file None module test_Image_CubeFITS_IO call None
nose.loader: DEBUG: parent <module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'> obj <function test_Image_CubeFITS_IO at 0x10d0067b8> module <module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'>
nose.suite: DEBUG: Create suite for <nose.suite.ContextList object at 0x1117e1da0>
nose.suite: DEBUG: tests <nose.suite.ContextList object at 0x1117e1da0> context <module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'>
nose.suite: DEBUG: Context suite for <nose.suite.ContextList object at 0x1117e1da0> (<module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'>) (4588445256)
nose.suite: DEBUG: suite <nose.plugins.multiprocess.NoSharedFixtureContextSuite context=test_image> has context test_image
nose.suite: DEBUG: get ancestry <module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'>
nose.suite: DEBUG: Create suite for [<nose.plugins.multiprocess.NoSharedFixtureContextSuite context=test_image>]
nose.suite: DEBUG: tests [<nose.plugins.multiprocess.NoSharedFixtureContextSuite context=test_image>] context None
nose.suite: DEBUG: wrap [<nose.plugins.multiprocess.NoSharedFixtureContextSuite context=test_image>]
nose.suite: DEBUG: wrapping <nose.plugins.multiprocess.NoSharedFixtureContextSuite context=test_image>
nose.suite: DEBUG: Context suite for [<nose.plugins.multiprocess.NoSharedFixtureContextSuite context=test_image>] (<module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'>) (4588443912)
nose.suite: DEBUG: suite <nose.plugins.multiprocess.NoSharedFixtureContextSuite context=test_image> has context test_image
nose.suite: DEBUG: get ancestry <module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'>
nose.plugins.multiprocess: DEBUG: Worker 0 Test is /Users/josh/src/GalSim/tests/test_image.py:test_Image_CubeFITS_IO (<nose.plugins.multiprocess.NoSharedFixtureContextSuite context=test_image>)
nose.plugins.multiprocess: DEBUG: suite 4588443912 (<nose.plugins.multiprocess.NoSharedFixtureContextSuite context=test_image>) run called, tests: <generator object _get_wrapped_tests at 0x110252d58>
nose.suite: DEBUG: suite 4588443912 setUp called, tests: <generator object _get_wrapped_tests at 0x110252d58>
nose.suite: DEBUG: tests in 4588443912?
nose.suite: DEBUG: ancestor <module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'> may need setup
nose.suite: DEBUG: completed suite setup
nose.suite: DEBUG: precache is [<nose.plugins.multiprocess.NoSharedFixtureContextSuite context=test_image>]
nose.plugins.multiprocess: DEBUG: suite 4588445256 (<nose.plugins.multiprocess.NoSharedFixtureContextSuite context=test_image>) run called, tests: <generator object _get_wrapped_tests at 0x10ee680a0>
nose.suite: DEBUG: suite 4588445256 setUp called, tests: <generator object _get_wrapped_tests at 0x10ee680a0>
nose.suite: DEBUG: tests in 4588445256?
nose.suite: DEBUG: ancestor <module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'> may need setup
nose.suite: DEBUG: completed suite setup
nose.suite: DEBUG: precache is [test_image.test_Image_CubeFITS_IO]
/Users/josh/src/lsstsw3/miniconda/lib/python3.6/subprocess.py:761: ResourceWarning: subprocess 78354 is still running
  ResourceWarning, source=self)
/Users/josh/src/lsstsw3/miniconda/lib/python3.6/subprocess.py:761: ResourceWarning: subprocess 78361 is still running
  ResourceWarning, source=self)
nose.suite: DEBUG: context teardown
nose.suite: DEBUG: ancestor <module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'> may need teardown
nose.suite: DEBUG: ancestor <module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'> already torn down
nose.suite: DEBUG: ancestor <module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'> may need teardown
nose.suite: DEBUG: ancestor <module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'> already torn down
nose.suite: DEBUG: context teardown
nose.suite: DEBUG: ancestor <module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'> may need teardown
nose.suite: DEBUG: ancestor <module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'> already torn down
nose.suite: DEBUG: ancestor <module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'> may need teardown
nose.suite: DEBUG: ancestor <module 'test_image' from '/Users/josh/src/GalSim/tests/test_image.py'> already torn down
nose.plugins.multiprocess: DEBUG: Results received for worker 0, /Users/josh/src/GalSim/tests/test_image.py:test_Image_CubeFITS_IO(), new tasks: 0
nose.plugins.multiprocess: DEBUG: batch result is ('Test that all six FITS reference images are correctly read in by both PyFITS and our Image ... ok\n', 1, [], [], {<class 'nose.plugins.deprecated.DeprecatedTest'>: ([], 'DEPRECATED', False), <class 'unittest.case.SkipTest'>: ([], 'SKIP', False)})
Test that all six FITS reference images are correctly read in by both PyFITS and our Image ... ok
nose.plugins.multiprocess: DEBUG: Ran 1 tests (total: 462)
nose.loader: DEBUG: load from /Users/josh/src/GalSim/tests/test_image.py:test_Image_FITS_IO (None)
...
```

rmjarvis:
I guess there are subprocesses in fits.py for the gzip and bzip2 stuff.  Could be one of those.  Maybe have them print out their pid so you can see which ones correspond to the pids that are still running.  On visual inspection of those bits of code though, I don't see how any of them could have been kept running.  Nor why the CubeFitsIO would be any different in this respect from the other two FitsIO test functions.

jmeyers314:
Looks like it's the subprocess calls in `_WriteFile.gzip_call()` and `_WriteFile.bzip2_call()` that are the issue.

rmjarvis:
That's kind of what I figured it had to be, but I don't really understand why those would keep running.  `p.communicate` is [supposed to](http://python.readthedocs.io/en/latest/library/subprocess.html#popen-objects) "Interact with process: Send data to stdin. Read data from stdout and stderr, until end-of-file is reached. Wait for process to terminate."

So I would have thought that after the `communicate` call, the subprocess should have finished running (i.e. terminated).  I guess maybe try explicitly calling `p.terminate()` or `p.kill()` after the with statement to see if that stops it?  Or failing that, maybe wrap the whole thing (in `__call__`) with a warnings filter to ignore `ResourceWarning`.

Or I'm ok with just ignoring it and hoping the Python 3.6 people eventually address this apparent bug in subprocess.

jmeyers314:
Looks like Popen has had an `__exit__` method since 3.2:  

https://hg.python.org/cpython/file/3.6/Lib/subprocess.py#l741

Probably mimicking that would work.  (And in the distant future, if we ever drop python 2 support, we can just use Popen as a context manager :) )

rmjarvis:
But look at the `__del__` method in that link you gave.  That's where the `ResourceWarning` is emitted and it only happens if `returncode is None`.  But we already assert that `returncode == 0`.  So I don't get it.  How does that branch get triggered?

jmeyers314:
>How does that branch get triggered?

Yeah, I have no idea.  I checked that using the context manager syntax in 3.6 does indeed eliminate the warning, but I haven't been able to eliminate the warning using similar code to the `__exit__` method injected directly into `_WriteFile.gzip_call()` or `_WriteFile.bzip2_call()`.  Heisenbug.

I think I'm also happy to leave this as is for now.  Ready for merge in that case?

rmjarvis:
Yes.  Good to merge as far as I'm concerned.

Issue #861, opened 2017-01-31
Test errors on master


brgillis:
A few tests are failing or throwing errors for me on master, when run on my own computer. I've attached the log of running these tests: [test_failures.txt](https://github.com/GalSim-developers/GalSim/files/742485/test_failures.txt)

I'm using Python v2.7.9, numpy v1.8.2, and astropy v1.0.3 (let me know if any other dependency versions will help).

rmjarvis:
Try branch '#861'.  I don't think any of the changes I made will have fixed the problem (although one of them could conceivably have fixed it).  But since I can't reproduce what you're seeing, it's hard to say.  I did add some output that should help us diagnose what is going on though, so send that if it still fails.

brgillis:
Here's the updated output from this branch: [test_failures.txt](https://github.com/GalSim-developers/GalSim/files/746674/test_failures.txt)

rmjarvis:
So, for some reason, your version of numpy is not respecting the command `np.set_printoptions(threshold=np.inf)`, which is supposed to prevent it from doing all those `...` in the array reprs.  I'm not sure that figuring out a workaround for this is the best use of my time.  

Any chance you can just upgrade your numpy version?  It's pretty easy to do.  `pip install -U numpy`.  Add `--user` if you don't have root privileges.

brgillis:
I tried upgrading numpy (to 1.12.0), but it didn't solve this issue. I dug into it a bit, and it seems that this issue with np.set_printoptions only comes up after astropy is imported - it seems something astropy is doing overrides any printoptions changes by numpy. Once I upgraded astropy (to 1.3), this part started working properly, but I got some other failures now (along with plenty of deprecation warnings): 
[test_failures.txt](https://github.com/GalSim-developers/GalSim/files/748507/test_failures.txt)

jmeyers314:
@brgillis , did you re-pull master in the last 24 hours?  (or re-merge master into this branch, I suppose?)  I think I addressed all the "floats cannot be ints" deprecations when I merged the py3urllib branch yesterday.

rmjarvis:
@jmeyers314 I merged that in to this branch, so that's taken care of.

Of course, as usual, the latest version of astropy is kind of a disaster.  Working on putting out those fires now...

(e.g. [clobber is now overwrite](https://github.com/astropy/astropy/pull/5171) WTF was that for?!!!)

rmjarvis:
[...]

PR #862, opened 2017-02-03
#861 Updates for latest numpy and astropy versions

rmjarvis:
Numpy version 1.12 changed how `np.roll` works slightly, which broke one of our tests.  To make sure `galsim.utilities.roll2d` always returns a C-contiguous array, we now have to coerce it.

And astropy version 1.3 changes the name of the `clobber` kwarg for `writeto` to `overwrite`.  I have no idea what compelled them to make this change, but I can't be bothered to support both APIs, so I just implement the file removal ourselves in the two places that had used `clobber`.

### DIFF
--- a/tests/galsim_test_helpers.py
+++ b/tests/galsim_test_helpers.py
@@ -383,7 +383,7 @@ def do_pickle(obj1, func = lambda x : x, irreprable=False):
         # precision for the eval string to exactly reproduce the original object, and start
         # truncating the output for relatively small size arrays.  So we temporarily bump up the
         # precision and truncation threshold for testing.
-        with galsim.utilities.printoptions(precision=18, threshold=1e6):
+        with galsim.utilities.printoptions(precision=18, threshold=np.inf):
             obj5 = eval(repr(obj1))
         f5 = func(obj5)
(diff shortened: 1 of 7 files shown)


tests/galsim_test_helpers.py
```diff
@@ -435,14 +435,16 @@ def do_pickle(obj1, func = lambda x : x, irreprable=False):
             # Special case: can't change size of LVector or PhotonArray without changing array
             if classname in ['LVector', 'PhotonArray'] and i == 0:
                 continue
-            with galsim.utilities.printoptions(precision=18, threshold=1e6):
+            with galsim.utilities.printoptions(precision=18, threshold=np.inf):
                 try:
                     if classname in galsim._galsim.__dict__:
-                        obj6 = eval('galsim._galsim.' + classname + repr(tuple(newargs)))
+                        eval_str = 'galsim._galsim.' + classname + repr(tuple(newargs))
```
jmeyers314: Why this change?  Were we catching the wrong exception?  If so, then maybe move the `try:` down below the `if-else`.

rmjarvis: No.  The previous code was fine when everything worked.  I just wanted to be able to print the `eval_str` below when this failed in order to help figure out what the problem was with the `eval`.

PR #863, opened 2017-02-10
#809c Sever ties to system boost library from random deviate implementation

rmjarvis:
Here is the third installment in my continuing effort to remove our boost dependence.  It includes the following changes:

1. This one mostly focuses on the boost::random stuff that we have for implementing our BaseDeviates.  We have copied the key boost files from their version 1.48.0 into `include/galsim/boost_1_48_0` to make sure our random deviates are stable and don't depend on people's local boost version.  However, those files include calls out to other boost header files for various configuration stuff.  So here, I basically cut all those ties to the rest of boost, so our boost::random code doesn't require a local boost library.

     Some of what I cut out are related to parts of the random code that we don't use, so there are `#ifdef USE_BOOST` blocks that keep the original code intact, but which we don't compile unless you use `scons USE_BOOST=true`.  Other bits that I removed seem to be special cases for dealing with older compilers or systems.  But the current version works for all of my systems, so hopefully this is relatively portable.

2. I reimplemented the Deviate classes a bit to put the primary class definitions in python and just use the C++ classes as back-end implementations.  This also let me remove the numpy handling from pysrc/Random.cpp, which is a plus, since that's hard to deal with in pybind11.

3. I completely reimplemented the Noise classes in Python.  We never call them or use them directly from C++, so there was no real purpose to having them in C++ and wrapped.  The few places where C++ makes an efficiency difference are now implemented using the `rng.generate` function along with a couple other related functions.

4. Another boost dependency was `tuple.hpp`, which we had used for the LRUCache class.  We don't really need a heavy-duty, massively templated tuple class for this use case, so rather than try to copy over their tuple.hpp and try to sever it's (many) tentacles into the rest of boost, I just wrote our own very simple Tuple class that we can use instead.


jmeyers314:
I haven't read any code yet, just your description above.  This made me wonder though, why not just use std::tuple?  Too early to commit to c++11?

Side note:  I've been playing around with pybind11 and like it quite a lot.  I vote we head that direction.

rmjarvis:
> This made me wonder though, why not just use std::tuple? Too early to commit to c++11?
I don't think `std::tuple` is supported by gcc until 4.7, so I'd rather not depend on that.  Too many systems still have 4.4.

> I've been playing around with pybind11 and like it quite a lot. I vote we head that direction.
I agree.  I actually prefer cffi overall, since it's more portable than pybind11, but I'm ok with moving to pybind11 for GalSim and leave the current boost::python as a backup option for people who still have gcc 4.4.  It's not *too* hard to support both in parallel.  Especially after the kinds of cleanup I'm currently doing on these branches.

rmandelb:
@rmjarvis - is this ready for a review now?  I was going to do it next (Monday).

rmjarvis:
Yep.  Ready to go.  And d now too.

include/galsim/LRUCache.h
```diff
@@ -22,11 +22,66 @@
...
+                second = rhs.second;
+                third = rhs.third;
+                fourth = rhs.fourth;
+            }
+            return *this;
+        }
+
+        bool operator<(const Tuple& rhs) const
+        {
+            return (
```
rmandelb: I'm sure this is right, but ouch, that's a bit painful to read.

include/galsim/boost1_48_0/random/detail/seed.hpp
```diff
@@ -13,21 +13,33 @@
...
+#endif
 
 // Sun seems to have trouble with the use of SFINAE for the
 // templated constructor.  So does Borland.
 #if !defined(BOOST_NO_SFINAE) && !defined(__SUNPRO_CC) && !defined(__BORLANDC__)
 
+#ifdef USE_BOOST
+// MJ: These are in the original file, but we don't actually need this functionality,
+// so rather than include all the explosion of header files they depend on, I slightly
+// edited the code below to no do the disable_if check.
```
rmandelb: "to not do"?

examples/demo7.py
```diff
@@ -183,9 +183,9 @@ def main(argv):
     # A Sersic profile may be truncated if desired.
     # The units for this are expected to be arcsec (or specifically -- whatever units
     # you are using for all the size values as defined by the pixel_scale).
-    bulge = galsim.Sersic(half_light_radius = 0.7, n = 3.2, trunc = 8.5, gsparams=gsparams)
-    disk = galsim.Sersic(half_light_radius = 1.2, n = 1.5, gsparams=gsparams)
-    gal5 = 0.4*bulge + 0.6*disk  # Net half-light radius is only approximate for this one.
+    bulge = galsim.Sersic(flux=0.4, half_light_radius=0.7, n=3.2, trunc=8.5, gsparams=gsparams)
+    disk = galsim.Sersic(flux=0.6, half_light_radius=1.2, n=1.5, gsparams=gsparams)
+    gal5 = bulge + disk  # Net half-light radius is only approximate for this one.
```
rmjarvis: There was a disagreement between demo7.py and demo7.yaml on exactly 1 fft image out of 100, and on only 1 pixel on that image, which was off by exactly 1 ADU.  It turned out that for that pixel the rounding error from rendering the galaxies with flux = 1 and then multiplying by 0.4 and 0.6 was slightly different than rendering directly with flux = 0.4 and 0.6.  Then when the two numbers were used as the mean values for the Poisson deviate noise, they produced values that differed by 1.

I'm not sure precisely which change caused it to break on this branch when it worked on master, but there's probably some slight difference in the order of operations somewhere that changed.  I do at least understand why the old way wasn't necessarily numerically identical between the yaml and py versions.

galsim/config/noise.py
```diff
@@ -345,10 +345,13 @@ def addNoise(self, config, base, im, rng, current_var, draw_method, logger):
                 # back off.
                 im -= total_sky
         else:
-            im += extra_sky
             # Do the normal PoissonNoise calculation.
-            im.addNoise(galsim.PoissonNoise(rng))
-            im -= extra_sky
+            if isinstance(total_sky, galsim.Image):
```
rmjarvis: Just a (minor) efficiency improvement.  If the sky is just a number (which is a typical use case), then the whole calculation can be done in one pass over the pixels, rather than adding the sky, generating noise, and subtracting, which is 3 passes.

include/galsim/LRUCache.h
rmjarvis: It's the unrolled version of the [boost implementation](http://www.boost.org/doc/libs/1_62_0/boost/tuple/tuple_comparison.hpp), hard coded for 4 template arguments.  Hard to say whether boost's template magic makes the code easier or harder to follow.  The lt function is shorter, but you need to understand how the head and tail things work, which is slightly tricky.

examples/demo7.py
rmandelb: Thanks for explaining!

rmjarvis:
I'll plan to merge this along with #809 and #809b this Friday in case anyone else would like to take a look.

Issue #866, opened 2017-02-15
assert failure in galsim::SBProfile::drawShoot function


jchiang87:
Running the `.drawImage` method, I am occasionally getting the following failure:
```
$ python drawImage_test_case.py 
galsim.__version__: 1.4.1

python: src/SBProfile.cpp:1005: double galsim::SBProfile::drawShoot(galsim::ImageView<T>, double, galsim::UniformDeviate, double, double, bool, bool) const [with T = float]: Assertion `!add_to_image' failed.
Aborted (core dumped)
$
```
The `drawImage_test_case.py` script and pickled inputs are attached.
[drawImage_assert_fail.zip](https://github.com/GalSim-developers/GalSim/files/778149/drawImage_assert_fail.zip)

rmandelb:
Hmmmm.  I reproduce this on my  system for v1.4.1 and v1.4.2, but not the latest version on the master branch.

rmjarvis:
Yep.  I figured out the problem.  And why it works on the master branch.  PR soon, just writing up a unit test for it now.

rmandelb:
Cool.  If Jim needs a tagged version, should we make v1.4.3?

jchiang87:
This will be rolled into the LSST sims distribution, so a tagged version would be great.  Thanks!
Pinging @danielsf

rmjarvis:
Yep.  I was planning on making the PR into releases/1.4 with the aim of rolling a 1.4.3 with the fix.  Then we can cherry-pick the unit test over to master.

danielsf:
Excellent.  I will package 1.4.3 up for LSST when it comes out,@jchiang87.  Thanks, @rmjarvis.

rmjarvis:
v1.4.3 is tagged with this fix.

https://github.com/GalSim-developers/GalSim/releases

PR #867, opened 2017-02-15
Fix photon shooting bug reported in #866

rmjarvis:
@jchiang87 reported an occasional assert failure in issue #866.  It turns out that when the flux to be shot was 100001, it would shoot the first batch of 100000 and be left with 1 photon, which could with numerical rounding sometimes be 1 - epsilon.  This led to it going down a branch that I had thought was not possible, so it hit an assertion failure.

To address this, I both fixed the logic of that branch to use `N < 0.5` rather than `N < 1`, which is probably more appropriate.  And I also just don't do anything if `add_to_image == true` rather than assert that it should be `false` in case there is some other case that could hit this branch, so we don't have the possibility of an assert failure here.

Once merged into releases/1.4, this will be released as v1.4.3, so I included the version changes here as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 branches:
   only:
     - master
+    - releases/1.4
 language: python
 compiler:
   - g++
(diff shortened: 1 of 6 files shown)


jmeyers314:
Looks good to me too.

PR #869, opened 2017-02-18
#722

craiglagegit:
This version contains the code to implement the brighter-fatter effect, with each photon having a specific wavelength and angles as drawn from a distribution.  I have tested it and believe it is working OK.  The file examples/brighter-fatter/GalSim_17Feb17.pdf shows the results of the B-F slopes.  The slopes are larger than the measurements, and this discrepancy is not understood at this time.  The magnitude is easily adjusted, and I did this last October, but chose not to do that this time.  However, I think we should still merge this into the master branch while we study the reason for the larger slopes.  The tests in tests/test_sensor.py also seem to run OK.


rmandelb:
This branch is well behind master.  It would be worth updating and making sure that the merge conflicts that GitHub is warning about can be addressed.

rmjarvis:
@craiglagegit I made some edits to the code to make it match GalSim style a bit more.  Things like 100 character lines, how we format the default values in doc strings.  That kind of thing.  

I also moved the stuff you had in `examples/brighter-fatter` to `devel/brighter-fatter`.  I think we'll want to make a single example for people to see how to use the new sensor stuff (maybe as part of the existing `lsst.py` example), but there was a bit too much in that directory for it to be useful as an example to end users, so `devel` seems like a better home for that.  But if you think there is something specific in there that would serve as a good example code for people to look at, let me know, and we can move it back.

In the C++ layer, I also took the opportunity to switch over the Polygon implementation to use std::vector rather than bare pointers and arrays, so no more new and delete calls anywhere.  I don't think you had any memory leaks, but it always feels a bit nicer to let the std library handle all that, so you don't have to worry about it.  This also ended up speeding things up significantly (roughly x2), so that's a nice bonus.  I made a few further tweaks, so the slow unit test, `test_bf_slopes`, which used to take 6.3 seconds on my laptop is now down to 2.6 seconds.

I have some ideas about how to optimize a bit further, but I haven't tried to code it up yet.  In particular, `insidePixel` is still a pretty tall pole.  I think we can use the fact that the Polygons are always nearly square to keep track of an inscribed and circumscribed square that will tell us whether a point is in or out pretty accurately most of the time faster than the current function does it (with the crossing number algorithm).  Then if the point is between the two squares, revert back to the current algorithm.

Another thought I had was to transpose the loops in `updatePixelDistortions` (another fairly tall pole, unsurprisingly).  The inner loop has a check if the pixel flux < 100.  If that happens pretty often, it might be faster to swap them so the `continue` happens in the outer loop instead.  I'm not sure how common this branch happens though, so it might not actually help much in practice.

Anyway, feel free to take a look at my changes.  I also have some specific line comments and questions that I'll try to get to this week.

rmandelb:
Nice.  I will try to go over this during this week as well.

src/Silicon.cpp
```diff
@@ -0,0 +1,506 @@
...
+            double x0 = photons.getX(i); // in pixels
+            double y0 = photons.getY(i); // in pixels
+
+            double zconv;  // This is the reverse of depth. zconv is how far above the substrate.
+            if (photons.hasAllocatedAngles()) {
+                double dxdz = photons.getDXDZ(i);
+                double dydz = photons.getDYDZ(i);
+                double dz = depth[i];
+                x0 += dxdz * dz / 10.0; // in pixels
+                y0 += dydz * dz / 10.0; // in pixels
```
rmjarvis: The 10.0 here seems to be a unit conversion from microns to pixels.  So shouldn't this use `_PixelSize` then?

src/Silicon.cpp
```diff
@@ -0,0 +1,506 @@
...
+                zconv = 100.0 - dz; // Conversion depth in microns
```
rmjarvis: I assume from this (and the 95 above) that the height of the silicon is 100 microns.  I think this should also be an input parameter, but I couldn't find it in you config file.  Isn't this something you need to specify for your Poisson model?  Or do you just have it hard-coded to 100 microns, rather than using a variable?

I note here that the calculation of `diff_step` in the Python layer also hard codes the thickness as 100 microns.

craiglagegit:
Just a status update.  I'm back from vacation, and I've pulled down Mike's changes and run them.  I pushed a minor fix to the plotting code in devel/brighter-fatter to get it to run, and the B-F slopes are about the same as before.  I'll work through Mike's requested changes over the next few days.  I'm also working with Kristen McKee and Simona Murgia at UCI to try to get to the bottom of the B-F slope difference between GalSim and my Poisson_CCD code (about 30%).  I've included an option in GalSim to write out the PhotonArray photon list to a file which we can then read into Poisson_CCD to see if the electrons end up in the same pixels or not.

rmjarvis:
Good.  I made a few more algorithmic tweaks.  (Basically the ones I suggested earlier.)  And the last test is now down to 1.7 seconds on my laptop.  Which means that the accumulate function is only taking about 1/2 the time now.  (Shooting the photons in the first place is most of the rest of the time.)  So I'm pretty happy with that now.

I might have messed up your PhotonArray writing option.  I moved it to the PhotonArray class (`photons.write(file_name)`) and have it write out to a fits file rather than ascii both for speed and to avoid loss of precision.  The file can be read back into a PhotonArray object with `photons = galsim.PhotonArray.read(file_name)`.

But if this messes up your ability to read it back into your code, let me know, and I can add an ASCII option for the write function to output in the format you had before.

src/Silicon.cpp
```diff
@@ -0,0 +1,506 @@
...
+                for (int i=nxCenter-_qDist; i<nxCenter+_qDist+1; i++) {
+                    for (int j=nyCenter-_qDist; j<nyCenter+_qDist+1; j++) {
+                        int chargei = ix + i - nxCenter;
+                        int chargej = iy + j - nyCenter;
+                        if ((chargei < minx) || (chargei > maxx) ||
+                            (chargej < miny) || (chargej > maxy))
+                            continue;
+
+                        double charge = target(chargei,chargej);
+                        if (charge < 100.0) // Don't bother if less than 100 electrons
```
rmjarvis: Note, as part of the changes I just made, I switched this to just `if (charge == 0.0) continue`.  At least for the runs we are doing now, the difference between 100 and 0 wasn't significant, so I just removed the approximation entirely.

craiglagegit:
Mike, I'm a little lost.  I tried to make all of the changes you asked for (see my notes below), and got it all working, but when I tried to push it, the changes you made in Silicon.cpp are conflicting.  I tried to use the merge tool to resolve the differences, but there were things I didn't understand, especially your addition of the inverse pixel size.  Any advice on what to do next?  Can you see the changes I tried to push?  

Notes on what I was trying to do:

I think I've addressed all of your change requests.  A couple of comments:

(1) I've removed all of the alternate directories containing the B-F data, and only kept one named 'lsst_itl', which is the default.  I couldn't find a good way to grab the path to the data, so I currently have the full path specified, which of course won't work except on my machine.  It looks like what others have done was to put data like this into share, which then gets copied into /usr/local/share/galsim.  Is that the way I should do this as well, or is there a better way?

(2) I think we should keep the CollectingPhase=1 option, even though I haven't tested it in GalSim yet. I have these simulations run, and I could add it, but I think it's best to keep it simple for now.  There are no plans that I know of to run the LSST sensors this way.

(3) I removed the unused collXmin, ... as you suggested.  These were from an old version.

(4) I added the parameter minCharge, which can be entered in Python, with a default of 100.

(5) I removed the hard coded 10.0, and 100.0 values and replaced these with _pixelSize and _sensorThickness.  The 100 micron sensor thickness was hard coded into my Poisson_CCD code, but is now a user parameter. This is fixed in the diff_step calculation in sensor.py as well.

(6) On writing out the PhotonArray data, I'll just read your fits file into my simulator. You can leave this as it is.

rmjarvis:
Sorry, I couldn't resist a bit more tweaking for speed.  Sorry to mess you up with it.  The inverse pixel size just computes 1/PixelSize once rather than doing the division each pass.  (Division is several times slower than multiplication, so this saves some calculation time in the loop.)

The steps to resolve it are:
1. `git pull`  (You already did this, I think.)  It tells you there are conflicts.  I think probably just in Silicon.cpp.
2. Edit Silicon.cpp and look for where there are `<<<<` and `>>>>>` lines.
It will show both the version you had before the merge and what the changes were that it was trying to merge that it couldn't figure out how to resolve.  Decide which parts you want to keep in the merged version and delete the rest.  Usually this involves taking some bits from each part, but not necessarily.
3. Git `commit -a`.  This will finish the merge commit that it had started and paused.  So the commit message should already be correct for the merge.
4. Install and run the tests to make sure the merge didn't break anything.
5. `git push`.

And if you have trouble, feel free to ping me with questions, or even to push it broken, and I can fix it up.

craiglagegit:
OK, you must have reverted to the earlier version, because when I tried a git pull this morning it told me it was up to date.  So then it let me push my changes.  So now I think you can see my changes, and you will resolve the conflicts.  Correct?

craiglagegit:
Mike,

Another (probably dumb) question. How should I access your photonArray.PhotonArray_write method?  I can't see a way to get access to the photonArray object in the testing code, like test_sensor.py or bf_plots.py.  It looks like the photonArray object is created in drawImage and then lost.  Should I add an option in sensor.py to allow you to call the PhotonArray_write method in sensor.accumulate?  What am I missing?

I need to get the capability to write the photonArray working again, beacuse the UCI folks are stuck until we can do that.

Thanks,
Craig

rmjarvis:
@craiglagegit I fixed up the merge.  But you'll need to do a hard reset now rather than a regular git pull:
```
git fetch
git reset --hard origin/#722
```
Also, I got rid of the min_charge change you made, since we don't actually need that in the c++ layer anymore.  I had rearranged the part of the code that used it, so it's not actually helpful anymore.

I still need to fix the directory specification to not use your local path, but I'll do that next.  And make sure there is a way to access the photon array from a `drawImage` call.

rmjarvis:
OK, I added a option to `drawImage` that will return the photon array in `image.photons` if you set `drawImage(..., save_photons=True)`.  Then you can write out the photon array as shown in `test_photon_io()` in test_photon_array.py.

galsim/sensor.py
```diff
@@ -0,0 +1,182 @@
...
+
+    @param dir              The name of the directory which contains the sensor information,
+                            presumably calculated from the Poisson_CCD simulator.  This directory
+                            may be specified either as an absolute path or as a subdirectory of
+                            share_dir/sensors/, where share_dir is `galsim.meta_data.share_dir`.
+                            It must contain, at a minimum, the *.cfg file used to simulate the
+                            pixel distortions, and the *_Vertices.dat file which carries the
+                            distorted pixel information.  [default: 'lsst_itl']
+    @param strength         Set the strength of the brighter-fatter effect relative to the
+                            amount in specified by the Poisson simulation results.  [default: 1]
```
rmjarvis: I switched the num_elec parameter to a strength parameter, since you have the reference number of electrons stored in the config file now, so the only purpose left for this was to change the strength of the effect.  I think it will be easier for users to just enter `strength=2` if they want to double the strength of the effect relative to the simulator results.  This is equivalent to setting `num_elec = config['CollectedCharge_0_0'] / 2` in the old scheme.

src/Silicon.cpp
```diff
@@ -0,0 +1,525 @@
...
+                y0 += dydz * dz_pixel; // dy in pixels
+            }
+            // This is the reverse of depth. zconv is how far above the substrate the e- converts.
+            double zconv = _sensorThickness - dz;
+            if (zconv < 0.0) continue; // Throw photon away if it hits the bottom
+            // TODO: Do something more realistic if it hits the bottom.
+
+            // Now we add in a displacement due to diffusion
+            if (_diffStep != 0.) {
+                double diffStep = std::max(0.0, diffStep_pixel_z * (zconv - _pixelSize));
```
rmjarvis: I'm not sure about this line.  This used to be `zconv - 10.0`, and you switched the 10.0 to `_pixelSize`.  But why would you be subtracting the pixel size here?  I thought the 10.0 was just some kind of buffer to keep it from diffusing too much if zconv was close to 0.  

Like maybe once it got 10 microns from the surface, the electric fields we strong enough that it doesn't really diffuse anymore.  Or something like that anyway.  But I don't see how the pixel size would be relevant.  

So could you double check the math on this line?  Maybe the 10.0 here should be a different input parameter?  (Perhaps one that we don't have named yet?)

craiglagegit:
OK, I corrected the overzealous search/replace error you pointed out in the diff step calculation, removed the duplicate sensor data in devel, and updated the bf_plots stuff in devel/brighter-fatter so it writes out the photon_file.  FYI, I added a conversion file in Poisson_CCD to convert the fits file to a text file that I can read into Poisson_CCD.  As far as I am concerned we are good to go.  I will keep working to try to get to the bottom of the BF slope discrepancy, but after all of these changes it still looks the same as it did after the SLAC DESC hack day.

rmjarvis:
Thanks Craig.  Looks good to me.  I think @rmandelb wanted to review it too, so we can wait for her before merging.

rmandelb:
@rmjarvis - I am not sure that you should wait for me.  I was selected to sit on a jury on Wednesday and the case is still not over, so I've effectively been unable to work for 3 days with at least one more to go.  When it's over, I will have to go do a bunch of other time-critical things that I was supposed to do in the past few days.  So at best, I could get to this in a week.  You might want to just go ahead...

rmjarvis:
OK.  Let's plan to merge this on Wednesday evening then to give a few more days for anyone else who might want to take a look.  I don't think there's a big rush, since there are still a few other issues we're planning to get into v1.5.

galsim/gsobject.py
```diff
@@ -1238,10 +1239,17 @@ def drawImage(self, image=None, nx=None, ny=None, bounds=None, scale=None, wcs=N
         @param surface_ops  A list of operators that can modify the photon array that will be
                             applied in order before accumulating the photons on the sensor.
                             [default: ()]
+        @param sensor       An optional Sensor instance, which will be used to accumulate the
```
jmeyers314: I think it would be worth mentioning in the drawImage docstring (above the param section somewhere) something about how the Sensor mechanics work when using FFT methods.  From the code I gather that the procedure is to first draw an image, sample photons from that image as if it were a probability distribution function, and then pass those photons to the sensor code.  (Did I get that right?)

Also, are there any combinations of `sensor != None` and `method != 'phot'` that we should warn on?  I presume only one of `sensor=SiliconSensor, method='fft'` and `sensor=SiliconSensor, method='no_pixel'` would be reasonable?

tests/test_sensor.py
```diff
@@ -0,0 +1,480 @@
...
+    # We'll draw the same object using SiliconSensor, Sensor, and the default (sensor=None)
+    im1 = galsim.ImageD(64, 64, scale=0.3)  # Will use sensor=silicon
+    im2 = galsim.ImageD(64, 64, scale=0.3)  # Will use sensor=simple
+    im3 = galsim.ImageD(64, 64, scale=0.3)  # Will use sensor=None
+
+    rng1 = galsim.BaseDeviate(5678)
+    rng2 = galsim.BaseDeviate(5678)
+    rng3 = galsim.BaseDeviate(5678)
+
+    # TODO: This file needs a better location (and probably name).
```
jmeyers314: stale comment?

galsim/gsobject.py
rmjarvis: You're correct about the mechanics of using the sensor with FFT methods.  I think they are accurate in the limit of the intermediate image having infinitely small pixels, regardless of `no_pixel` or not.  

But your comment reminded me that right now the intermediate image has the same size pixels as the final image.  (Which means it's not quite correct for either 'no_pixel' or 'fft' currently.)  I'd meant to add a parameter that would let the user specify how much smaller the intermediate image pixels should be.  Probably with a default of 2 or 3 maybe.

And I'll add a comment about all of that in the doc string.

tests/test_sensor.py
rmjarvis: Yep.  Thanks.

galsim/gsobject.py
rmjarvis: Josh, I finally got around to doing this.  I added Image methods `bin` and `subsample` which are kind of converse operations to bin pixels together or split them up into subpixels.  And now using fft drawing methods with a sensor defaults to subsampling 3x3.  I also added a bunch of text to the `drawImage` doc string about this and the other new sensor stuff.  Please take a look the recent few commits.

jmeyers314: Thanks Mike,  Looks good to me.  I fixed some typos and an killed an unused variable.  Assuming that gets through travis, I think this is good to merge.

rmjarvis:
Great.  Thanks, Josh!

Issue #865, opened 2017-02-15
Add color term to WCS classes


rmjarvis:
@gbernstein has astrometric solutions for DES that include a color term.  I'm working on wrapping his solutions into GalSim WCS classes, but we don't currently have any way to deal with the color dependence in GalSim.

I'm proposing to add an optional color term to the WCS classes that would be ignored for all the current WCS classes, but which would provide an appropriate hook to work with Gary's solutions.  Mostly, I just need to have a `color` object that defaults to `None`, which would be passed through the front end methods like `toWorld` down into the implementation methods like `_radec`.

Basically, you would be able to do
```
sky_pos = wcs.toWorld(image_pos, color=gminusi)
```
and if the wcs class supports it, it would use the color value appropriately.

Gary's code uses float values of g-i for the color, but I think since we are just passing along the value to the back-end WCS class, we don't need to enforce that anywhere.  So some other WCS class might use a more complicated "color" object that encapsulates the information in whatever way it needs.

rmjarvis:
Merged via #872

PR #872, opened 2017-03-12
#865 Various features I needed for recent projects I've been working on.

rmjarvis:
This PR includes a number of features that I needed for two projects I've been working on lately.

The first project was to wrap @gbernstein's DECam WCS solution into a GalSim WCS, which we want for DES.  This was my lighthouse project a few weeks ago before the DESC meeting.  This PR doesn't include that effort (I'll make a PR [here](https://github.com/gbernstein/pixmappy) soon for that), but it does include the GalSim hooks that I needed to make that possible.

Most of the rest of the changes were precipitated by @NiallMac's and my hack day project at the LSST DESC meeting to make a config output module to [build a full focal plane](https://github.com/esheldon/galsim_extra/issues/2) with a coherent PSF pattern across multiple chips.  We kept running into issues that were best solved by making edits to GalSim.  So I did them on the same branch.  And then kept working on them after the hack day, fleshing out unit tests and such.  I'm now finally done (I think).

Here are summaries of the various feature additions.

1. I added the ability for a GalSim WCS be color dependent.  This was the main thing I needed for wrapping Gary's WCS solutions, since they include a color term.  The change mostly just involved adding an optional `color` terms to many of the WCS methods.  I also added an option `uses_color` to the `UVFunction` WCS type, so it is the one GalSim class that can be color dependent.  Mostly so I could write useful unit tests.

2. The biggest thing Niall and I needed for our hack was a way to have some properties (e.g. galaxy locations and shapes) stay constant across multiple images while having other properties change for each image.  We sort of kludged something together at the time, but never really got it working.  

    It took a while, but I think I finally came up with a good solution -- allow multiple random number generators that change their seed with different cadences.  So for instance one rng can be used for the PSF power spectrum that only increments after each focal plane.  So each image will use the same PSF PS realization.  But a second rng can be used for the noise properties in the image, so they don't all have the same noise.

    This is also the same basic problem that I had with getting the Great3 mutli-epoch simulations done as config files.  I think this multiple rng solution will work there as well.  One rng can repeat its seed sequence for each epoch to get the same galaxy properties.  But a second one can be used for the PSF properties, so each image of the galaxy gets a different PSF.

3. For the PSF power spectrum, I have the functional form from Heymans et al (k^2 + 1/L0^2)^-11/6.  But it wasn't obvious what the right normalization was.  So I added an option to give a variance to the PowerSpectrum object to renormalize the result to have a given variance (in |g|, but it's equal to the variance of kappa too if there is only E mode power).

4. Added the ability to wrap angles with respect to an arbitrary center.  It turned out that this was helpful for finding the range in RA around an arbitrary pointing to properly deal with the possibility of crossing RA = 360 degrees.

5. Added a CelestialCoord constructor from (x,y,z).  Also, a public method to access (x,y,z) for a given CelestialCoord.  These were useful to find the central point of a number of images.  (Get all their x,y,z positions, average normally, and then go back to RA, Dec.)

6. Added a way to specify the `world_pos` as RA, Dec in the config layer.  And in general, added CelestialCoord as a new valid value_type in the config layer (for which RADec is now the only valid type, but someone could now write a custom module to generate celestial coords in some other way if they wanted).

7. Added an `index` option to the power_spectrum input type in the config layer to only rebuild the grid when index increments.  This way you can have `index` be a sequence that increments after completing each focal plane for instance, rather than each image as had been hard coded.  This was easy to implement in a serial run.  But with multi-processing, it was a bit harder to get working.  I ended up using the multiple rng feature to make it multiprocessing-safe.  It makes its own private rng that increments according to the given index and uses that.  This way even if multiple processes need to update the grid, they will all be doing it with the same rng and everything is deterministic.

8. Don't warn if objects are outside the allowed range for the power spectrum realization if they are also off the current image.  This way you can have galaxies that cover a larger area than the current image, but might then be outside the range of the current power spectrum realization.  It had been giving warnings for these galaxies in my unit test, which was simulating a multi-epoch kind of thing with moderate dithering.  This was kind of annoying, so I now have it only pass through the warning if the galaxy will end up being drawn on the current image (i.e. it would probably be a bug for them not to be in the range of the PS grid).  If they are outside the current image, it just logs the information as a debug message rather than a warning.

9. Added ngrid and center options to the power spectrum input object in the config layer.  It still defaults to figuring those out automatically for you based on the current image, but now you can tell it to use a different region for the grid if you want (e.g. the full focal plane, centered at the focal plane center).

10. Skip the drawImage step in the config layer if the drawn postage stamp will fall completely off of the main image.  This is purely for efficiency so you can have a field of galaxies across the full focal plane and just run through them all for each image.  Since most of the running time is in the draw step, we don't want to draw those galaxies N times when we usually throw away the ones that are elsewhere in the focal plane.  But we do want to draw the ones that straddle a chip edge 2 times to make sure the corresponding parts get properly added to both images that they overlap.

11. Added the ability to parse a wcs field in a location other than image.wcs.  I wanted to have a custom module that would include some normal wcs parsing further down in the dict, but the exiting WCS parsing function was hard-coded to look in `config['image']['wcs']` only.

12. Fixed a couple minor bugs along the way.  One in the `Shear.__eq__` method that was not checking that the "other" object was in fact a Shear at all.  I ran into this during debugging.  So I added a unit test to check for such things as part of the `do_pickle` checks.  Turns out there were a few others with the same error.  

13. The other bug I ran across was in the PowerSpectrum initialization, where it used k=0 in its check that the function properly returns numpy arrays.  But one of my unit tests used a power law that gave annoying division by zero warnings.  For the actual usage of the function, we guard against k=0 for this reason, but we forgot about that when we added this sanity check.


rmandelb:
Unit tests pass on both of my machines.  (Did not look at code yet, sorry...)

rmjarvis:
I've made some more updates to this branch, but I think I'm done now.  In particular, the showcase for many of the features here is in [PR into the galxim_extra repo](https://github.com/esheldon/galsim_extra/pull/4) in case you want to check that out.

Notable additional features and improvements:
- Allow a template string to look like `:field` which uses another field within the current configuration dict as the template.  (It had been permissible to write `file_name:field` to take from a different file.)
- Made the output.nfiles information an override-able method `getNFiles` so custom output types can determine this programmatically if they want.
- Fixed a bug in Transformation.shoot if the wcs has a reflection.  (The flux was coming out negative in this case due to a missing `abs`.)
- Added `skip` as an item in the stamp field, rather than just in gal or psf.  It works the same way, but for some reasons to skip, it makes more sense to put it there, so I wanted to allow that.
- Allow custom modules to add their own valid `index_key`s by making `valid_index_keys` a public list, rather than hard-coded within the function that uses it.
- Made the way the extra outputs get handled OutputBuilder methods, so they can be overridden by custom modules if it makes sense to do so.
- Various minor bug fixes that turned up along the way.

I also made some changes that sped up a few things that were taking a long time for my FocalPlane run.
- Ported the CD, PV and AB calculations in GSFitsWCS into C++, since the numpy code was rather slow and used a huge amount of memory to compute a sky image.
- Sort of "compile" aspects of the config processing to not have to parse the same strings over and over again running through all the objects.  e.g. Eval statements now eval a `lambda` expression, which then gets called with the new variables each time.  This is a lot faster than eval-ing again for each object.  Similarly, Current types store the location in the dict where it should evaluate rather than parse the key name each time.

CHANGELOG.md
```diff
@@ -152,6 +157,16 @@ New Features
   `obj._drawKImage`, `image._view`, `image._shift`.  These are appropriate
   for advanced users who are optimizing a tight loop and find that the normal
   Python checks are taking a significant amount of time. (#840, #873)
+- Added a hook to the WCS classes to allow them to vary with color, although
+  most of our current WCS classes are not able to use this feature.  The only
+  one that can is UVFunction, which may now optionally have a color term
+  if you set `uses_color=True`. (#865)
```
rmandelb: Not enabled for config yet though?

galsim/gsobject.py
```diff
@@ -1610,7 +1610,7 @@ def drawFFT_makeKImage(self, image, wmult=1.):
 
         if Nk > self.gsparams.maximum_fft_size:
             raise RuntimeError(
-                "drawFFT requires an FFT that is too large: %s."%Nk +
+                "drawFFT requires an FFT that is too large: %s. "%Nk +
```
rmandelb: Why %s and not %d?  (I realize this is not new, but just wondering)

galsim/image.py
```diff
@@ -506,6 +506,15 @@ def getYMin(self): return self.image.getYMin()
...
     def getBounds(self): return self.image.getBounds()
 
+    def getOuterBounds(self):
+        """Get the bounds of the outer edge of the pixels.
+
+        Equivalent to galsim.BoundsD(im.xmin-0.5, im.xmax+0.5, im.ymin-0.5, im.ymax+0.5)
+
+        @returns a BoundsD instance
+        """
+        return galsim.BoundsD(self.xmin-0.5, self.xmin+0.5, self.ymin-0.5, self.ymin+0.5)
```
rmandelb: I think the second ```xmin``` and second ```ymin``` should be ```xmax``` and ```ymax```?
(Also, need unit test for this, if I'm right that this is a bug.)

galsim/lensing_ps.py
```diff
@@ -423,6 +423,14 @@ def buildGrid(self, grid_spacing=None, ngrid=None, rng=None, interpolant=None,
                                 does not modify the internally-stored power spectrum, just the
                                 shears generated for this particular call to buildGrid().
                                 [default: "hard"]
+        @param variance         Optionally renormalize the variance of the output shears to a
+                                given value.  This is useful if you know the functional form of
+                                the power spectrum you want, but not the normalization.  This lets
+                                you set the normalization separately.  The resulting shears should
+                                have var(g1) + var(g2) ~= variance.  If only e_power_function is
+                                given, then this is also the variance of kappa.  Otherwise, the
+                                variance of kappa may be smaller than the specified variance.
```
rmandelb: Warn about using this for too small of a grid?  (normalization will effectively be applied to a very noisy measure of the variance)

galsim/shear.py
```diff
@@ -312,7 +312,7 @@ def __add__(self, other):
     # order of operations: shear by -other._shear, then by self._shear
     def __sub__(self, other): return self + (-other)
 
-    def __eq__(self, other): return self._g == other._g
+    def __eq__(self, other): return isinstance(other, Shear) and self._g == other._g
```
rmandelb: Heh.  Indeed.

galsim/table.py
```diff
@@ -564,7 +564,7 @@ def __eq__(self, other):
                 self.edge_mode == other.edge_mode)
 
     def __ne__(self, other):
-        return not self.__eq__(self, other)
+        return not self.__eq__(other)
```
rmandelb: Did we not have a unit test for this?

galsim/angle.py
```diff
@@ -173,6 +173,15 @@ def get_angle_unit(unit):
 This could be appropriate before testing for the equality of two angles for example, or
 calculating the difference between them.
 
+If you want to wrap to a different range than [-pi, pi), you can set the `center` argument
+to be the desired center of the the range.  e.g. for return values to fall in [0, 2pi),
+you would call
+
+    >>> theta = theta.wrap(center=math.pi)
+
+@param center   The center point of the wrapped range. [default: 0]
```
rmandelb: Could we allow this to also take an Angle object rather than requiring a number in radians?

galsim/config/input_powerspectrum.py
```diff
@@ -41,9 +43,45 @@ def getKwargs(self, config, base, logger):
...
+            image_num = base.get('image_num',None)
+            file_num = base.get('file_num',None)
+            base['obj_num'] = base['image_num'] = base['file_num'] = 0
+            if isinstance(base['image']['random_seed'],list):
+                first_seed = galsim.config.ParseValue(base['image']['random_seed'], 0, base, int)[0]
+            else:
+                first_seed = galsim.config.ParseValue(base['image'], 'random_seed', base, int)[0]
+            rs = base['image']['random_seed']
+            if not isinstance(rs, list): rs = [rs]
+            first_seed += 31415  # An arbitrary offset to avoid the chance of unwanted correlations
```
rmjarvis: I don't think it's a problem for it to be hard-coded.  I just needs to be something arbitrary and non-zero.  Presumably small numbers like 1 or 2 might be problematic, but I think anything largish like this will make this an independent random number sequence.  Especially since the cadence of this sequence (typically indexed on image_num or file_num) is different from the normal rng indexing cadence (indexed on obj_num).

If the user wants, they can override this whole section by providing their own rng specification (with rng_num and making random_seed a list).  So we're not forcing the user to use this hard-coded value.  Just making it easy for people to not have to think about it too hard.

galsim/config/output.py
```diff
@@ -467,22 +500,47 @@ def getNImages(self, config, base, file_num):
         """
         return 1
 
-    def writeFile(self, data, file_name):
+    def canAddHdus(self):
+        """Returns whether it is permissible to add extra HDUs to the end of the data list.
+
+        In the base class, this returns True.
+        """
+        return True
```
rmjarvis: The DataCube output type cannot add additional HDUs to the end, since it collapses all the images into a single 3d array.  One could probably write a version of that which smartly handled extra weight and badpix hdus in the data list and made data cubes for them as well, but the current implementation doesn't do so.

galsim/config/input_powerspectrum.py
```diff
@@ -124,25 +189,41 @@ def _GenerateFromPowerSpectrumShear(config, base, value_type):
...
+    try:
         shear = galsim.Shear(g1=g1,g2=g2)
     except KeyboardInterrupt:
         raise
     except Exception as e:
-        import warnings
-        warnings.warn("Warning: PowerSpectrum shear is invalid -- probably strong lensing!  " +
-                      "Using shear = 0.")
+        if logger:
+            logger.warning('obj %d: Warning: PowerSpectrum shear (g1=%f, g2=%f) is invalid. '%(
```
rmjarvis: Actually, I think this behavior goes counter to how we deal with such things elsewhere in the config layer.  We only send information to a logger object.  So if the user wants to get these kinds of messages, they provide a logger.  Otherwise, it should be silent about this.  So I switched it to just use the `logger.warning` call.   b2fbbc917

galsim/config/output.py
```diff
@@ -467,22 +500,47 @@ def getNImages(self, config, base, file_num):
...
+        """If appropriate, add any extra output items that go into HDUs to the data list.
+
+        @param config           The configuration dict for the output field.
+        @param data             The data to write.  Usually a list of images.
+        @param logger           If given, a logger object to log progress.
+
+        @returns data (possibly updated with additional items)
+        """
+        if self.canAddHdus():
+            data = galsim.config.AddExtraOutputHDUs(config, data, logger)
```
rmjarvis: Yes.  In fact, I think it should raise an exception.  Added.  5bd542acf

galsim/config/output_datacube.py
```diff
@@ -97,13 +98,16 @@ def getNImages(self, config, base, file_num):
             raise AttributeError("Attribute output.nimages is required for output.type = MultiFits")
         return galsim.config.ParseValue(config,'nimages',base,int)[0]
 
-    def writeFile(self, data, file_name):
+    def writeFile(self, data, file_name, config, base, logger):
```
rmjarvis: Yes.  For more consistency on the call syntax compared to other similar functions.  And so subclasses can use them.  I didn't actually need them for my FocalPlane output type, but when writing that class, I realized that some of the signatures were probably insufficiently flexible.

CHANGELOG.md
rmjarvis: True.  Added that caveat.

galsim/gsobject.py
rmjarvis: I have a tendency to use `%s` for everything, since it always works correctly regardless of what type the variable is.  Unless I want to do something like `%04d` or similar.  But `%d` would be equivalent here.

galsim/image.py
rmjarvis: Yes!  Thank you.  Also added a unit test that would have caught the error.

galsim/lensing_ps.py
rmjarvis: Hmm.  That seems rather subjective.  How small is too small?  

Also, I think the variance that results from this procedure will be correct by construction.  Just the measured variance may not match the expectation of the variance for that power spectrum input.  But it will measure whatever variance actually results from the gaussian field and scale the signal accordingly to give it the specified variance.  

So I guess I'm not even sure what phrasing you would want for the warning.  I'm pretty sure the parameter does work exactly as advertised.

galsim/table.py
rmjarvis: Didn't used to.  But we do now.  I added tests of the `!=` operator in the `do_pickle` tests that we run for everything.

galsim/angle.py
rmjarvis: I went with "instead" rather than "also".  Good idea.

rmjarvis:
Thanks Rachel.  I think I've addressed all your comments.

rmandelb:
Yep, looks good, thanks.

jmeyers314:
Hey Mike, started looking at this.  I'll try to get through it within a couple of days.

One thing I noticed already is that the warnings you suppressed (your item 8 in the first PR message) are not being suppressed in python3.

galsim/config/stamp_ring.py
```diff
@@ -67,7 +67,7 @@ def setup(self, config, base, xsize, ysize, ignore, logger):
 
         # Now go on and do the base class setup.
         ignore = ignore + list(req) + list(opt)
-        return super(self.__class__, self).setup(config, base, xsize, ysize, ignore, logger)
+        return super(RingBuilder, self).setup(config, base, xsize, ysize, ignore, logger)
```
rmjarvis: Yes.  If you subclass from RingBuilder (which I did), then `self.__class__` is the subtype, not RingBuidler, and `super(self.__class__, self)` is `RingBuilder`, not `StampBuilder`, which leads to an infinite loop.

I had seen the trick about writing `super(self.__class__, self)` on StackOverflow somewhere, and I thought it was a clever way to make sure you do the right thing.  But it turns out it breaks in this kind of use case, so I think writing out the actual class name is better practice in general.

galsim/config/value.py
```diff
@@ -428,9 +363,7 @@ def _GetAngleValue(param):
         value = float(value)
         unit = galsim.angle.get_angle_unit(unit)
         return galsim.Angle(value, unit)
-    except KeyboardInterrupt:
-        raise
-    except Exception as e:
+    except (TypeError, AttributeError) as e: # pragma: no cover
```
rmjarvis: Because I'm catching the other two by explicit type name now, so there is no need to catch and reraise a KeyboardInterrupt.  The main thing is to make sure that KeyboardInterrupts don't get caught and ignored by things like `except Exception as e:` or just `except:`.  

And I think in general, trying to catch the exceptions that you are expecting by name is better than just `except Exception`, so I've been trying to switch to that whenever possible and not too unwieldy.

galsim/wcs.py
```diff
@@ -214,30 +214,38 @@ def toWorld(self, arg, **kwargs):
         else:
             return self.posToWorld(arg, **kwargs)
 
-    def posToWorld(self, image_pos, **kwargs):
+    def posToWorld(self, image_pos, color=None, **kwargs):
```
rmjarvis: Right, no.  The type of the color is defined by the WCS type that uses it.  Gary's WCS's use a scalar float.  But a full set of several colors would be fine.  Or even some user-defined class if necessary.  I think I don't limit the implementation of that to anything specific.

rmjarvis:
> One thing I noticed already is that the warnings you suppressed (your item 8 in the first PR message) are not being suppressed in python3.

Fixed.  d92f868

Issue #875, opened 2017-03-17
Efficiency in convolution


brgillis:
A colleague of mine is working on a shear estimation algorithm which uses a model-fitting approach, and efficiency is a very big concern - he basically needs each galaxy model to be generated in ~10ms. He's able to do this with his own code, but for the Euclid pipeline we want everyone to be using the same galaxy models if possible, which means using the GalSim implementations of profiles. The problem is that a few efficiency measures in his code are not currently possible with GalSim (at least as far as I can tell). These are:

- Using FFTW_MEASURE to speed up successive FFTs. From my reading of the code, it looks like GalSim always uses FFTW_ESTIMATE.
- Reusing arrays in convolution. It looks to me like GalSim re-allocates arrays for each convolution, which takes up a fair amount of time.

There might be something I'm missing in GalSim that can help with this, so I'll wait for input before talking about my thoughts on implementing these in GalSim in case it's unnecessary.

rmjarvis:
Before tackling specific improvements, the best bet would be to develop a self-contained example code that you want to optimize.  Then we can actually profile it and see what the tall poles are for optimization.  I'm always happy to add efficiency improvements, but without a target use case to profile, my experience is that one tends to focus on things that aren't actually helpful.

Aside: for very fast model fitting using GalSim, my guess is that you will be better off doing the modeling in k-space rather than real space.  Then you don't need any FFTs.  For simple models, GalSim draws the k-space image in <~ 1ms typically.  cf. https://github.com/GalSim-developers/GalSim/issues/873#issuecomment-287159168.  I.e. FFT your data once and generate models in k-space, rather than repeatedly FFT the models to compare to the data.

brgillis:
Here's some sample code for it, comparable to the model generation step of what my colleague (Giuseppe) is doing (not included is the actual comparison of the generated model to the actual image):

`import galsim
from cProfile import runctx

def main():
    """ @TODO main docstring
    """
    
    # Using very low accuracy GSParams here for speed
    gsparams = galsim.GSParams(minimum_fft_size=256,
                               folding_threshold=0.1,
                               kvalue_accuracy=1e-3,
                               stepk_minimum_hlr=2.5,)
    
    # Note - we actually use an interpolated image instead; just putting this in
    # so you can run the code without needing that file
    psf_prof = galsim.OpticalPSF(lam=725, # nm
                                 diam=1.2, # m
                                 defocus=0,
                                 obscuration=0.33,
                                 nstruts=3,
[...]

Issue #873, opened 2017-03-14
Use fast exp, fast inverse sqrt, etc. when possible


rmjarvis:
@esheldon and I have been discussing some ways that the k-space rendering in GalSim could be sped up.  One of the ideas is to use faster exp and invsqrt functions when possible.  We have parameters xvalue_accuracy and kvalue_accuracy that tell how accurate we need to be.  I suspect in many cases, we can use quick and dirty versions of various functions that are faster than the native functions.  

e.g. the famous [Quake inverse sqrt](https://www.beyond3d.com/content/articles/8/) (although the [SSE inverse sqrt is probably faster nowadays](http://stackoverflow.com/questions/15175654/sqrt-vs-rsqrt-vs-sse-mm-rsqrt-ps-benchmark)).

Another thought I had was possibly using SSE commands in the fillKValue functions.  Some of the profile types may be amenable to this.  Particularly Exponential if we're able to use rsqrtps.

Anyway, this issue is basically to profile and optimize the drawK operations for Exponential, Spergel, and maybe a few other classes.

esheldon:
Some examples
```C

#include <stdio.h>
#include <math.h>
#include <xmmintrin.h>


static inline void sse_rsqrt( float * restrict pOut, float * restrict pIn )
{
   __m128 in = _mm_load_ss( pIn );
   _mm_store_ss( pOut, _mm_rsqrt_ss( in ) );
   // compiles to movss, movaps, rsqrtss, mulss, movss
}

// plus one newton raphson step
static inline void sse_rsqrt_nr( float * restrict pOut, float * restrict pIn )
{
    const float threehalfs = 1.5F;
    float y, x2;

    x2 = *pIn * 0.5F;

    __m128 in = _mm_load_ss( pIn );
    _mm_store_ss( &y, _mm_rsqrt_ss( in ) );

    *pOut  = y * ( threehalfs - ( x2 * y * y ) );

    // compiles to movss, movaps, rsqrtss, mulss, movss
}


// straight from wikipedia
static inline float carmack_rsqrt( float number )
{
    long i;
    float x2, y;
    const float threehalfs = 1.5F;

    x2 = number * 0.5F;
    y  = number;
    i  = * ( long * ) &y;                       // evil floating point bit level hacking
    i  = 0x5f3759df - ( i >> 1 );               // what the fuck? 
    y  = * ( float * ) &i;
    y  = y * ( threehalfs - ( x2 * y * y ) );   // 1st iteration
[...]

PR #879, opened 2017-03-26
#873 Various optimizations and efficiency improvements

rmjarvis:
Two issues recently requested optimizations in GalSim when used for model fitting (and hence tight loops with GalSim calls), so I made a number of changes that help speed things up.  The specific scripts that I focused on are @esheldon's use case of [fitting an exponential or Spergel profile in k space](https://github.com/GalSim-developers/GalSim/blob/%23873/devel/time_expon_k.py) (#873) and @brgillis's use case of [fitting a Sersic profile with fixed PSF in real space](https://github.com/GalSim-developers/GalSim/blob/%23873/devel/time_conv.py) (#875).

The former is now about 7x faster than the master version (although Erin didn't get nearly as much speedup for his actual script).  The latter is about 5x faster.  Furthermore, while those scripts were the focus, the result was an overall general speed up in Galsim by around 10%, as evidenced by the time to run the unit tests and the example scripts.  (Both are around 10% faster than on master.)

The changes include:
1. Switching to building the k-space image with float rather than double (i.e float32 rather than float64).  We had been doing everything in double in the C++ layer, even though we often write the results onto an ImageF, so we lose all that precision.  Turns out it's faster, and not really any less accurate to build things with single precision in the first place.
2. Add an ImageCF class, whose dtype is np.complex64.  I should have done this in the first place, rather than only implement a double-precision complex Image type.  The double precision is now called ImageCD, not ImageC.  (Since ImageC was introduced after v1.4, this isn't an API change.)
3. Wrote a bunch of custom SSE code to speed up some calculations.  The normal implementation of the C++ complex type is really terrible in terms of speed (I found this out when writing and optimizing TMV), so it benefits a lot from writing custom code running over arrays of complex values.  I didn't do everything, just the things that mattered for those scripts, but the template is in place to optimize other things similarly if there is a desire for such.
4. Added some more "leading underscore" functions to skip the Python-layer argument checks: _Shear, _shear, _shift, _drawKImage.  They are appropriate to use within tight loops where the argument checking and other setup stuff are not necessary and can take a significant amount of time.
5. Broke up the drawFFT function into 3 parts: setting up the kimage, drawing the k-space image, and then finishing the process to go back to real space.  When the PSF is complicated, but constant, and you want to run many different galaxy models (cf. #875), then this lets you only draw the k-space image of the PSF once rather than every time.
6. Switched to using a fast exponential function from [fmath](https://github.com/herumi/fmath) when possible.
7. Reworked the SBTransform application of the exp(-ik . cen) phase factors for centroid shifts.  In addition to some SSE stuff to avoid the std library complex multiplication, there is also a nice trick for generating a series of phases without so many trig calls.  Just 2 sets of sin/cos, rather than n.  So that helped quite a bit.
8. Avoided some gratuitous array copies and zeroing.  There were places where following the chain of function calls, the net effect was to do essentially im2 = im1.copy(); im2.setZero(); im2.setZero(); im2(i,j) = [values], im1 += im2.  Now, for things like this, it just assigns to im1 directly.

I don't really expect people to review the SSE stuff, so feel free to skim over that (unless you're interested -- as SSE code goes, some of this is pretty straightforward, so it might be a way to learn how these SSE commands work).  

Mostly this branch would be worth making sure that things install and run correctly on your machine, since I did change some compiler directives.  Erin already discovered that the SLAC virtual machines didn't like the directive I originally had to enable SSE2, so there might be other portability issues lurking.


brgillis:
Checked it out, and it installs and passes tests on both my desktop and a virtual machine I work with.

rmjarvis:
Great.  Thanks, Bryan.  If you'd be willing to do a code review too, that would be much appreciated.

@esheldon Would you mind taking a look at the code as well?  Especially the stuff C++ layer stuff with fmath, which you've used before.  I'd be interested to know if you see any places where we could make further efficiency improvements.

I think Josh and Rachel have been a bit backlogged lately, so it would be very helpful for getting this merged sooner if the either or both of you could review this PR.  Feel free to skip over any parts that you don't feel comfortable reviewing.  Thanks!

brgillis:
I'll see if I can find time for it; I'm pretty busy now as well, but I might end up being free before Josh and/or Rachel.

rmandelb:
@rmjarvis - this is really nice (and impressive).  While you're right about my backlog, I did take a little time to make sure everything installs and unit tests pass on both of my machines (Mac laptop and linux cluster).    Everything on my Mac was fine.  On the linux cluster, the compilation and installation seemed to go without a hitch, but then when I ran ```scons tests -j1``` I got the following output:

```
scons: Reading SConscript files ...
SCons is version 2.0.1 using python version 2.7.4
Python is from /opt/python27/include/python2.7
Using the following (non-default) scons options:
   PREFIX = /home/rmandelb/software
   TMV_DIR = /home/rmandelb.proj/sw-shared
These can be edited directly in the file gs_scons.conf.
Type scons -h for a full list of available options.
Using python =  /opt/python27/bin/python
Using PYPREFIX generated from PREFIX =  /home/rmandelb/software/lib/python2.7/site-packages
Using compiler: /usr/bin/g++
compiler version: 4.4.7
Using SSE with flag -msse2
Checking for C++ header file fftw3.h... (cached) yes
Checking for correct FFTW linkage... (cached) yes
Checking for boost header files... (cached) yes
Boost version is 1.50.0
Checking for C++ header file TMV.h... (cached) yes
TMV version is 0.72
Using TMV_LINK file: /home/rmandelb.proj/sw-shared/share/tmv/tmv-link
     -L/home/rmandelb.proj/sw-shared/lib -ltmv -lpthread -fopenmp
Checking for correct TMV linkage... (this may take a little while)
Checking for correct TMV linkage... (cached) yes
Checking if we can build against Python... (cached) yes
Building for python version 2.7
Checking if we can build module using TMV... (cached) yes
Checking if we can build against NumPy... (cached) yes
Numpy version is 1.9.2
Checking for PyFITS... (cached) yes
PyFITS version is 3.0.8
Checking for future... (cached) yes
Future version is 0.15.2
Checking if we can build against Boost.Python... (cached) yes
Checking if C++ exceptions are propagated up to python... (cached) yes
nosetests version: 1.3.0
GalSim version  1.5
scons: done reading SConscript files.
scons: Building targets ...
g++ -o galsim/_galsim.so -fopenmp -Wl,-rpath=/home/rmandelb/software/lib -Wl,-rpath=/home/rmandelb.proj/sw-shared/lib -Wl,-rpath=/home/rmandelb/git/GalSim/lib -Wl,-rpath=/home/rmandelb/software/lib -Wl,-rpath=/home/rmandelb.proj/sw-shared/lib -Wl,-rpath=/home/rmandelb/git/GalSim/lib -shared pysrc/.obj/module.os pysrc/.obj/Angle.os pysrc/.obj/Bounds.os pysrc/.obj/PhotonArray.os pysrc/.obj/Image.os pysrc/.obj/SBProfile.os pysrc/.obj/SBAdd.os pysrc/.obj/SBConvolve.os pysrc/.obj/SBDeconvolve.os pysrc/.obj/SBFourierSqrt.os pysrc/.obj/SBTransform.os pysrc/.obj/SBBox.os pysrc/.obj/SBGaussian.os pysrc/.obj/SBExponential.os pysrc/.obj/SBSersic.os pysrc/.obj/SBMoffat.os pysrc/.obj/SBAiry.os pysrc/.obj/SBShapelet.os pysrc/.obj/SBInterpolatedImage.os pysrc/.obj/SBKolmogorov.os pysrc/.obj/SBSpergel.os pysrc/.obj/SBInclinedExponential.os pysrc/.obj/SBInclinedSersic.os pysrc/.obj/Random.os pysrc/.obj/Noise.os pysrc/.obj/HSM.os pysrc/.obj/Integ.os pysrc/.obj/Table.os pysrc/.obj/Interpolant.os pysrc/.obj/CorrelatedNoise.os pysrc/.obj/Bessel.os pysrc/.obj/CDModel.os pysrc/.obj/Silicon.os -Llib -L/opt/boost-1.50.0/lib -L/home/rmandelb.proj/sw-shared/lib -L/home/rmandelb/software/lib -ltmv_symband -lfftw3 -lpthread -ltmv -lpthread -lgalsim
run_tests(["tests/tests.log"], ["bin/test_main"])
Using nosetests from:  /opt/python27/bin/nosetests
nosetests is version 1.3.0

Starting python tests...
Nosetests terminated by signal  11
scons: *** [tests/tests.log] Explicit exit, status -11

```

rmjarvis:
Hm.  I tried my extended family of systems and couldn't reproduce this anywhere (although I did find what is probably a different error in the Sensor stuff on one of them, which I'll investigate).  So I need to follow up with you a bit.

> Starting python tests...
Nosetests terminated by signal  11
scons: *** [tests/tests.log] Explicit exit, status -11

This kind of error right at the start usually means the `import galsim` isn't working.  Could you check what happens if you try to import galsim in a python session?

If that works, then the next thing to try would be
```
cd tests
nosetests -vs
```
Hopefully that will give a bit more information to tell us where it is seg faulting.

rmandelb:
Right.  Now I'm wondering if I've screwed something up... it's segfaulting when I try to import:
```
  File "/home/rmandelb/software/lib/python2.7/site-packages/galsim/__init__.py", line 94, in <module>
    from .position import PositionI, PositionD
  File "/home/rmandelb/software/lib/python2.7/site-packages/galsim/position.py", line 22, in <module>
    from . import _galsim
ImportError: /home/rmandelb/software/lib/python2.7/site-packages/galsim/_galsim.so: undefined symbol: _ZTIN5boost6python15instance_holderE
```

Normally I get that kind of thing if I haven't cleaned up before compiling, but I did do ```scons -c install``` beforehand, so I'm not sure what's up.

brgillis:
It looks like it's looking for the wrong boost library. It might be using a cached value for it which is no longer valid. Maybe try running scons --no-cache after a clean, and see if that works.

rmjarvis:
Or `rm -rf .scon*` will remove all the cached stuff before running scons.

rmandelb:
That worked, Mike.  Now I can import galsim.  However, twice I had the unit tests abruptly end partway through with 
```
Nosetests terminated by signal  9
scons: *** [tests/tests.log] Explicit exit, status -9
```
after hanging for quite a while.

rmjarvis:
> Nosetests terminated by signal  9

Did you git pull?  I think that might be the bug I fixed.

rmandelb:
Ah, I didn’t notice there were more commits.  that does fix it.

rmjarvis:
> is there (should there be?) a way to configure at compile time to use the fast math stuff or not?

The fmath expd is accurate at the 1.e-15 level.  Maybe that's one significant digit off from the std library version of exp, but I think always good enough for our purposes.  The single-precision versions are accurate to the full single precision, so that's not ever an issue.

rmjarvis:
BTW, if you're wondering how the fmath can be just as accurate as std::exp, std::log but still be faster, I think it's mostly because they don't conform to the IEEE rules about nan's and inf's and such.  So for bad inputs, they probably do something different than what the IEEE rules dictate.  But in GalSim, for bad inputs, I don't think we care much whether the output is perfectly IEEE compliant.

The other advantage is that they can do 2 or 4 at once within the context of SSE commands, which obviously gives a big speed boost in those routines.

rmjarvis:
> kim3 is never used. I tried changing this to kim, which looks close to the intention, but that didn't work with the integer dtypes, so I'll leave it to you to fix :).

Thanks.  It's supposed to be kim3, since we want to compare the round trip to kim, not this, so we don't want to overwrite kim.  The fix was to actually use it in the next line rather than kim.  However, it failed for int types because the 0.5 values became 0, so I doubled all the values to make them integers, so it works now.  Thanks for catching that.

devel/time_conv.py
```diff
@@ -0,0 +1,77 @@
+from __future__ import print_function
```
rmjarvis: Thanks.  I changed the drawFFT function to respect the input bounds rather than automatically recenter.  But I didn't update this script to reflect that.  Fixed now.

galsim/gsobject.py
```diff
@@ -1432,63 +1457,104 @@ def drawImage(self, image=None, nx=None, ny=None, bounds=None, scale=None, wcs=N
...
-        The image is not cleared out before drawing.  So this profile will be added to anything
-        already on the input image.  This corresponds to `add_to_image=True` in the `drawImage`
-        options.  If you don't want to add to the existing image, just call `image.setZero()`
-        before calling `drawReal`.
-
         Note that the image produced by `drawReal` represents the profile sampled at the center
         of each pixel and then multiplied by the pixel area.  That is, the profile is NOT
         integrated over the area of the pixel.  If you want to render a profile integrated over
-        the pixel, you can convolve with a Pixel first and draw that.
+        the pixel, you can convolve with a Pixel first and draw that.  This is equivalent to
```
rmjarvis: Thanks. I switched the order of the last two sentences, which I think gets the antecedent correct.

tests/test_image.py
```diff
@@ -2296,93 +2311,94 @@ def test_complex_image():
...
-    do_pickle(im1_view)
-    do_pickle(im2)
-    do_pickle(im2_view)
-    do_pickle(im3_view)
-    do_pickle(im4_view)
+    for dtype in [np.complex64, np.complex128]:
+        # Some complex modifications to tests in test_Image_basic
+        im1 = galsim.Image(ncol, nrow, dtype=dtype)
+        im1_view = im1.view()
+        im1_cview = im1.view(make_const=True)
```
rmjarvis: It looks like there isn't any use for it in the bulk of the test.  But I added a test that it pickles correctly at the end, which seemed to be missing.

rmjarvis:
Anyone else want to review this?  I'll plan to merge this Wednesday evening unless someone speaks up before then.

esheldon:
I do worry about large arguments for the fmath `expd`.  I recall seeing issues for large arguments when I originally adopted the code for ngmix.

esheldon:
looking at fmath again, it looks like that code has evolved a lot since I hacked it to create my version for ngmix. I has some guards for range that may avoid the issue I saw.

rmjarvis:
> I do worry about large arguments for the fmath expd. I recall seeing issues for large arguments when I originally adopted the code for ngmix.

It looks like the current version of fmath is accurate enough for what we need, even at large arguments.

I added tests of the speed and accuracy of fmath exp options to the `time_invsqrt.cpp` program (since much of the infrastructure is the same as for testing 1/sqrt options).  It seems to be completely fine up to exp(+-50) which is already higher than we ever care about in GalSim.
```
max arg = 49.990002, exp = 5.133126e+21
std::exp      0.2098 msec, max error = 0
fmath::exp    0.2139 msec, max error = 2.09871e-06
fmath::expd   0.47315 msec, max error = 0
fmath::exp_ps 0.1497 msec, max error = 2.09871e-06
```
Note that `fmath::expd` is slower here than normal `std::exp`, but that's just an artifact of the test.  The values here are actually floats, so the `fmath::expd` run includes the conversion to and from double (which is slow).  When swapping `std::exp` for `fmath::expd` in the actual GalSim code where it is relevant, it did speed things up some.

Issue #881, opened 2017-04-04
assert failure in galsim::OneDimensionalDeviate::shoot


jchiang87:
I'm getting an assert failure when I run the attached code (and pickle files):
```
$ python galsim_test_case.py 
galsim.__version__: 1.4.3
python: src/OneDimensionalDeviate.cpp:328: boost::shared_ptr<galsim::PhotonArray> galsim::OneDimensionalDeviate::shoot(int, galsim::UniformDeviate) const: Assertion `N>=0' failed.
Aborted
$
```
In contrast to #866, I'm seeing this much more often: about 10% of the imsim jobs are dying with this error when I try to run using the Kolmogorov PSF implementation that we recently added.

[oneDimDeviate_assert_fail.zip](https://github.com/GalSim-developers/GalSim/files/894289/oneDimDeviate_assert_fail.zip)

rmjarvis:
Well, the problem here is that the number of photons to shoot is overflowing the size of int and wrapping around to negative values.

So this is clearly a bug, and we should catch this in the python layer rather than crash and burn with an assert failure.  But do you really intend to have something with flux = 3.97e9?  Here is the object being drawn:
```
galsim.Convolve(galsim.Kolmogorov(lam_over_r0=0.529987797134), galsim.Gaussian(sigma=0.171131263474)) * 3973131100.34
```

jchiang87:
This is a question for @danielsf perhaps.  The imsim code is simply getting the parameters for each object (stars in this case) from queries to the catsim db and then creating stars with those fluxes.  We aren't applying any magnitude limits on the stars that are simulated, so maybe that's warranted?   Or maybe those fluxes are not realistic even for the brightest stars in those catalogs, and there is a bug elsewhere.

rmjarvis:
BTW, I'll note that this is already fixed on master.  Now it emits a warning that it won't actually draw the object:
```
/Users/Mike/lib/python2.7/site-packages/galsim/gsobject.py:1810: UserWarning: Automatic n_photons calculation did not end up with positive N. (n_photons = -2147483648.0)  No photons will be shot. prof = galsim.Convolve(galsim.Kolmogorov(lam_over_r0=0.529987797134), galsim.Gaussian(sigma=0.171131263474)).expand(6.91359985819).shift(1410.64264865,349.447410634) * 83123613.715  flux = 3973131100.34  poisson_flux = True  max_extra_noise = 0.0  g = -0.540501582698  
[...]

PR #882, opened 2017-04-05
#881 Use normal_distribution in PoissonDeviate when mean is very high

rmjarvis:
@jchiang87 found another bug when doing the LSST imsim run.  It turns out that the boost poisson deviate class that we use maxes out at 2^31 and wraps around to -2^31.

This turned out to be a problem when photon shooting with `poisson_flux=True` and the flux is >~ 2^31 (~2.1e9).  Then the poisson varied flux comes out negative, and it tries to shoot a negative number of photons, which isn't allowed.

This PR fixes this in two ways.  First, when the flux > 2^30, it uses a Gaussian approximation to the Poisson deviate to do the `poisson_flux=True` option.  And second, once it has its final number of photons to shoot, it checks again if N < 0, and if it is, it raises an error (rather than crashing with an assert failure down the line).

This PR is to merge into releases/1.4, which should be tagged 1.4.4.  I'll then make the same changes on master to keep this functionality, although the changes need to happen in a different place there (specifically, this calculation is in the python layer now).

Quick turnaround on the code review requested, so Jim can start a new imsim run with v1.4.4.

### DIFF
--- a/tests/test_random.py
+++ b/tests/test_random.py
@@ -859,7 +859,6 @@ def test_poisson_highmean():
         np.testing.assert_allclose(
                 pn.getSkyLevel(), mean, rtol=1.e-8,
                 err_msg="PoissonNoise getSkyLevel returns wrong value")
-        continue
 
         # Check that the noise model really does produce this variance.
         big_im = galsim.Image(2048,2048,dtype=float)
--- a/tests/test_random.py
+++ b/tests/test_random.py
@@ -772,6 +772,105 @@ def test_poisson():
     do_pickle(pn)
     do_pickle(galsim.DeviateNoise(p))
 
+@timer
+def test_poisson_highmean():
+    """Test Poisson random number generator with high (>2^30) mean (cf. Issue #881)
+
+    It turns out that the boost poisson deviate class that we use maxes out at 2^31 and wraps
+    around to -2^31.  We have code to automatically switch over to using a Gaussian deviate
+    instead if the mean > 2^30 (factor of 2 from the problem to be safe).  Check that this
+    works properly.
+    """
+    mean_vals =[ 2**30 + 50,  # Uses Gaussian
+                 2**30 - 50,  # Uses Poisson
+                 2**30,       # Uses Poisson (highest value of mean that does)
+                 2**31,       # This is where problems happen if not using Gaussian
+                 5.e20,       # Definitely would have problems with normal implementation.
+               ]
+
+    if __name__ == '__main__':
+        nvals = 10000000
+        rtol_var = 1.e-3
+    else:
+        nvals = 100000
+        rtol_var = 1.e-2
+
+    for mean in mean_vals:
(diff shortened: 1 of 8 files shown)


arunkannawadi:
I verified that in the current version, for high flux values, the GaussianDeviate does not give negative values, unlike the PoissonDeviate. I don't see the option to approve the PR, so I do it verbally here.

jchiang87:
Thanks for fixing this so quickly.   Unfortunately, I won't be able to use this work right away, since I would need to run a tagged version of the lsst sims code, and the pybind11 transition is delaying that.   So we may need to impose a magnitude limit on the imsim DC1 runs.

rmjarvis:
OK.  In that case, let me reconsider the proposed fix slightly.  I think it might be a better solution to make our PoissonDeviate class check for mean > 2^30 and switch to a Gaussian approximation internally.  Rather than have this switch in the drawShoot function.  Then any use of PoissonDeviate will do the right thing even if the mean is higher than boost's pd can handle.

rmjarvis:
I just pushed a new fix that I like better.  It moves the fix farther upstream to the PoissonDeviate class itself.  Now whenever it gets a mean > 2^30, it internally uses a `normal_distribution` for its implementation rather than the boost `poisson_distribution`.  

This way we don't have to worry about the same problem cropping up somewhere else with large means being given to PoissonDeviate.  Also, it afforded me the ability to write unit tests for this fix.  (I wasn't so keen on the unit test being that a photon shooting routine runs for 12 minutes rather than crashing.)

tests/test_random.py
```diff
@@ -772,6 +772,105 @@ def test_poisson():
...
+                err_msg='Wrong poisson random number sequence generated using PoissonNoise')
+
+        # Check PoissonNoise variance:
+        np.testing.assert_allclose(
+                pn.getVariance(), mean, rtol=1.e-8,
+                err_msg="PoissonNoise getVariance returns wrong variance")
+        np.testing.assert_allclose(
+                pn.getSkyLevel(), mean, rtol=1.e-8,
+                err_msg="PoissonNoise getSkyLevel returns wrong value")
+        continue
```
jmeyers314: Did you mean to `continue` here?

jmeyers314:
Also, I guess you'll cherry-pick these into master after you fix up 1.4?

tests/test_random.py
rmjarvis: Good catch.  No, I didn't.  I'd added that when debugging something and forgot to remove it.

rmjarvis:
> Also, I guess you'll cherry-pick these into master after you fix up 1.4?

Yes, I'll cherry-pick it onto master once this is merged.

rmjarvis:
@rmandelb Did you have time to take a look?  Or should I go ahead and merge to let Jim start running jobs with it?

rmandelb:
Please go ahead given that Josh has looked at it and I am being slow.

Issue #884, opened 2017-04-26
Preload option in RealGalaxyCatalog only preload HDUs not data


EiffL:
The preload option in RealGalaxyCatalog, and by extension, COSMOSCatalog does open all the files and reads in the HDUs but the data is not actually loaded into RAM. From what I can read in issue #168 this is by design, but I have a use case where I actually need to have the whole catalog loaded in memory (I need to very rapidly produce galaxy images in random order).
 
I can manually force the catalog to preload all the data with the following:
```
# Doesn't actually preload, returns very quickly
cat = galsim.COSMOSCatalog(preload=True)

# Takes a while as it cycles through the files, memory usage steadily increases in the process
for f in cat.real_cat.loaded_files:
    for h in cat.real_cat.loaded_files[f]:
        h.data;
```

But what to do in this situation ? Should there be separate preload and preload_HDU options ? 

If you need to preload the HDUs, most likely you are going to loop through the entire catalog which will progressively load the whole thing in memory anyway, so in terms of memory usage it doesn't change much. But preloading the catalog by sequentially reading all the files is much faster than randomly accessing different HDUs until the whole catalog is loaded.

rmjarvis:
I'm fine with just making the API change to have preload actually load in all the data as you had expected it to do.  (Although I think the doc string describes the current behavior accurately.)

The case for preloading as we currently do it has largely gone away, I think.  Some time ago I did some timing tests with and without preloading, and there was no longer the kind of difference we were seeing back in 2012 when we were working on issue #168.  If you use all the galaxies, the times were the same, and if you didn't, preloading was slower.  So we don't promote `preload` anywhere in the demos, and I don't ever bother with it myself.

So I'd be fine with changing it to actually load in all the data, rather than just the header info, as the word "preload" sounds like it would be doing.

EiffL:
Ok, great. I'm happy to make a pull request. Is the solution I quoted in the issue  working for you or can you think of a smarter way to do this ?

rmjarvis:
Looks right to me.  You can look in the astropy docs to see if there is something simpler, but what you wrote does what you want and looks efficient enough.

rmjarvis:
Merged via #885

PR #885, opened 2017-04-27
Changed behavior of the preload option in RealGalaxyCatalog

EiffL:
As discussed in  issue #884 the  preload option in RealGalaxyCatalog was only preloading the HDUs, which does save some time but does not actually load the catalog in memory as one might think. This pull request changes the behavior of this option so that it actually loads all the fits files in memory for subsequent fast access.


rmandelb:
Should comment on the change in CHANGELOG.

galsim/real.py
```diff
@@ -547,7 +546,11 @@ def preload(self):
                 # when memmap = True.  Anyway, I don't know what the performance implications
                 # are (since I couldn't finish the run with the default memmap=True), but I
                 # don't think there is much impact either way with memory mapping in our case.
-                self.loaded_files[file_name] = pyfits.open(file_name,memmap=False)
+                f = pyfits.open(file_name,memmap=False)
+                self.loaded_files[file_name] = f
+                # Access all the data from all hdus to force PyFits to read the data
+                for hdu in f:
+                    hdu.data;
```
rmjarvis: Very minor point, but you have an extraneous semi-colon here.

PR #889, opened 2017-06-15
Add Python3 support to SCons files

timj:
- Print function
- has_key -> in

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -1,3 +1,4 @@
+from __future__ import print_function
 # vim: set filetype=python et ts=4 sw=4:
 
 # Copyright (c) 2012-2017 by the GalSim developers team on GitHub
@@ -26,9 +27,9 @@ import distutils.sysconfig
 
 from sys import stdout,stderr
 
-print 'SCons is version',SCons.__version__,'using python version',platform.python_version()
+print('SCons is version',SCons.__version__,'using python version',platform.python_version())
 
-print "Python is from", distutils.sysconfig.get_python_inc()
+print("Python is from", distutils.sysconfig.get_python_inc())
 
 # Require SCons version >= 1.1
 # (This is the earliest version I could find to test on.  Probably works with 1.0.)
@@ -162,7 +163,7 @@ def ClearCache():
         shutil.rmtree(".sconf_temp")
 
 def GetMacVersion():
-    print 'Mac version is',platform.mac_ver()[0]
+    print('Mac version is',platform.mac_ver()[0])
     ver = platform.mac_ver()[0].split('.')
     if len(ver) >= 2:
         return ver[:2]
@@ -185,9 +186,9 @@ def ErrorExit(*args, **kwargs):
     out = open("gs_error.txt","wb")
 
     # Start with the error message to output both to the screen and to gs_error.txt:
-    print
+    print()
     for s in args:
-        print s
+        print(s)
         out.write(s + '\n')
     out.write('\n')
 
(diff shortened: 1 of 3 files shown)


esheldon:
This seems good for the future, but note scons does not support python 3 yet so we can't test it http://scons.org/tag/releases.html

timj:
I am currently testing SCons 3 alpha with the LSST DM stack. I can apply this patch locally to LSST version. You can get SCons 3 from https://testpypi.python.org/pypi/scons/3.0.0.alpha.20170614

So far it seems pretty solid (I have been able to build everything up to `afw` without any problems.

timj:
I haven't technically tested this patch on python3 yet because I was waiting for tmv to be fixed in rmjarvis/tmv#25.

dkirkby:
[Scons 3](http://scons.org/scons-300-is-available.html) is now now the default, with support for python 3.  Should that work now for installing GalSim + tmv?

rmjarvis:
Yes, but I think you need the master branch of both.  If you have scons 3, it would be great if you could test whether it does in fact work.

timj:
We've been using scons3 at LSST for a month now building tmv and galsim.

dkirkby:
> Yes, but I think you need the master branch of both. If you have scons 3, it would be great if you 
> could test whether it does in fact work.

I just tested scons 3 with the master branches of tmv & galsim.  Tmv was fine but the GalSim link fails with boost missing symbols:
```
clang++ -o galsim/_galsim.so -bundle pysrc/.obj/module.os pysrc/.obj/Angle.os pysrc/.obj/Bounds.os pysrc/.obj/PhotonArray.os pysrc/.obj/Image.os pysrc/.obj/SBProfile.os pysrc/.obj/SBAdd.os pysrc/.obj/SBConvolve.os pysrc/.obj/SBDeconvolve.os pysrc/.obj/SBFourierSqrt.os pysrc/.obj/SBTransform.os pysrc/.obj/SBBox.os pysrc/.obj/SBGaussian.os pysrc/.obj/SBDeltaFunction.os pysrc/.obj/SBExponential.os pysrc/.obj/SBSersic.os pysrc/.obj/SBMoffat.os pysrc/.obj/SBAiry.os pysrc/.obj/SBShapelet.os pysrc/.obj/SBInterpolatedImage.os pysrc/.obj/SBKolmogorov.os pysrc/.obj/SBSpergel.os pysrc/.obj/SBInclinedExponential.os pysrc/.obj/SBInclinedSersic.os pysrc/.obj/Random.os pysrc/.obj/Noise.os pysrc/.obj/HSM.os pysrc/.obj/Integ.os pysrc/.obj/Table.os pysrc/.obj/Interpolant.os pysrc/.obj/CorrelatedNoise.os pysrc/.obj/Bessel.os pysrc/.obj/CDModel.os pysrc/.obj/Silicon.os pysrc/.obj/RealGalaxy.os pysrc/.obj/WCS.os -Llib -L/Users/david/anaconda3/envs/DESC/lib -L/usr/local/lib -lboost_python3 -lpython3.6m -ltmv_symband -lfftw3 -ltmv -lcblas -lgalsim
Undefined symbols for architecture x86_64:
  "boost::python::objects::function_object(boost::python::objects::py_function const&, std::__1::pair<boost::python::detail::keyword const*, boost::python::detail::keyword const*> const&)", referenced from:
...
```
This doesn't look like a scons problem per se.  I am using the conda-installed boost 1.61.0 in a py3.6 conda environment.

rmjarvis:
@dkirkby I think this error happens when you have a mismatch in the two kinds of C++ standard libraries.  cf. https://stackoverflow.com/questions/18389799/boost-python-undefined-symbols-error-even-though-the-lib-files-are-linked

Perhaps because anaconda is using a different C++ compiler than you are using to install GalSim. @esheldon ran into similar issues with Anaconda using a different C++ compiler than the native one, which led to various c++11 std library problems.  Not sure if it's exactly the same as what you are seeing here.

He posted a suggestion about installing with Anaconda in #914 that might be helpful, although one of the things he suggested is `conda install boost` which I guess you are already doing, so maybe it's not relevant.  But maybe since this seems not to be scons 3 related, we could move further discussion of this over to that issue.

PR #892, opened 2017-07-12
Add cpp coverage options to scons

jmeyers314:
We have pretty good coverage of the GalSim python layer in our unit tests, but I was curious how we did in the c++ layer.  I found some instructions [here](https://emptysqua.re/blog/code-coverage-python-c-extensions/) that almost worked for this.  Once I bumped the compiler optimizations down to -O1 and added a couple more flags, I got the output posted [here](http://slac.stanford.edu/~jmeyers3/galsim_coverage/).  

This PR just adds an option to SConstruct to make it easy to generate this cpp coverage.  Unfortunately, I'm not sure this would be possible to do on travis, given that the use of -O1 slows down the test run time and we already use a significant fraction of the allowed time, especially when we occasionally need to rebuild the cache.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -117,7 +117,7 @@ opts.Add(PathVariable('LD_LIBRARY_PATH',
          'cf. DYLD_LIBRARY_PATH for why this may be useful.',
          '', PathVariable.PathAccept))
 
-opts.Add('NOSETESTS','Name of nosetests executable','')
+opts.Add('PYTEST','Name of pytest executable','')
 opts.Add(BoolVariable('CACHE_LIB','Cache the results of the library checks',True))
 opts.Add(BoolVariable('WITH_PROF',
             'Use the compiler flag -pg to include profiling info for gprof', False))
@@ -669,17 +669,14 @@ def GetCompilerVersion(env):
     env['CXXVERSION'] = version
     env['CXXVERSION_NUMERICAL'] = float(vnum)
 
-def GetNosetestsVersion(env):
-    """Determine the version of nosetests
-    """
+def GetPytestVersion(env):
+    """Determine the version of pytest"""
     import subprocess
-    cmd = env['NOSETESTS'] + ' --version 2>&1'
-    p = subprocess.Popen([cmd],stdout=subprocess.PIPE,shell=True)
-    line = p.stdout.readlines()[0]
-    version = line.split()[2]
-    print('nosetests version:',version)
-    env['NOSETESTSVERSION'] = version
-
+    cmd = 'pytest --version 2>&1'
+    p = subprocess.Popen([cmd], stdout=subprocess.PIPE, shell=True)
+    line = p.stdout.readlines()[0].decode()
+    version = line.split()[4].replace(',', '')
+    env['PYTESTVERSION'] = version
 
 def ExpandPath(path):
     p=os.path.expanduser(path)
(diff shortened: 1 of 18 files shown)


jmeyers314:
> either massively raise the time for timing out (from the current 60) or manually set njobs=1 to avoid the problem.

I'm experimenting with setting njobs=1, which seems to work fine for getting the tests to pass.

I'm discovering that I'm having trouble getting reasonable coverage reports out of nosetests, though, regardless of the number of processes being used.  For example, if I manually invoke just test_base.py and test_interpolatedimage.py, then something like 60% of lines under src/ are reported as covered.  If I use nosetests to invoke these two files though, then only 4% are reported covered.  Using pytest to invoke returns results similar to manually invoking, so the difference is thankfully not just due to `if name == '__main__'` blocks.  So I think I'd like to continue working on this branch a bit to add a scons flag which would set pytest as the test runner when calling `scons tests`.

rmjarvis:
You can switch to py.test for all `scons tests` runs if you want.  It is nicer for python 3 in particular.

jmeyers314:
>You can switch to py.test for all scons tests runs if you want.

Done.

@rmjarvis, can you check that `fink install pytest-27` is a real thing?  I made that up in Install.md.

I left in place the warnings about `np.testing.assert_raises` requiring nose, since as far as I can tell that's still actually true.

I also left .travis.yml alone.

rmjarvis:
> can you check that fink install pytest-27 is a real thing? 

Close.  It's pytest-py27

jmeyers314:
>>can you check that fink install pytest-27 is a real thing?

> Close. It's pytest-py27

Ah, of course.  How else would we know that it's python?

Turns out that's what I had entered in Sconstruct already, just messed it up here on github.

jmeyers314:
@rmjarvis , any other comments on this PR?

SConstruct
```diff
@@ -2151,14 +2148,14 @@ if not GetOption('help'):
         subdirs += ['examples']
 
     if 'tests' in COMMAND_LINE_TARGETS:
-        if env['NOSETESTS'] == '':
-            nosetests = which('nosetests')
-            if nosetests is None:
-                env['NOSETESTS'] = None
+        if env['PYTEST'] == '':
+            pytest = which('pytest')
+            if pytest is None:
```
rmjarvis: I think it would be worth adding a check for `py.test` at this point.  Apparently, the `pytest` executable name is newish -- version 3.0 -- so if it doesn't exist, it would be nice to check for `py.test` automatically rather than fail.

I didn't even know that `pytest` was valid, since they are keeping the `py.test` executable around still, just deprecated:
> [...] so from pytest 3.0 we will supported and recommend use of pytest as the main command instead of py.test. It's possible that in future we will deprecate py.test and potentially even remove it.

SConstruct
```diff
@@ -2141,14 +2148,14 @@ if not GetOption('help'):
...
+                env['PYTEST'] = None
             else:
-                env['NOSETESTS'] = nosetests
-        if env['NOSETESTS']:
-            GetNosetestsVersion(env)
+                env['PYTEST'] = pytest
+            if env['PYTEST']:
+                GetPytestVersion(env)
```
rmjarvis: Looks like this if block is wrongly indented. It should be at the same level of indentation as the above `if env['PYTEST'] == '':`.

jmeyers314:
Thanks @rmjarvis .  I think this is ready now.  I'll plan to leave it until Friday and if I hear nothing more I'll merge it in.

PR #894, opened 2017-07-21
More python3 fixes

timj:
This is the second batch of fixes required to build GalSim with a Python3 SCons.

I have successfully built GalSim using `scons-3.0.0.alpha.20170614.tar.gz` with Python3 but there is one issue I encountered with building of boost-python because those include files have bad Unicode in them for an author string. I've reported this upstream (and I have a local fix).

### DIFF
--- a/src/SConscript
+++ b/src/SConscript
@@ -12,7 +12,8 @@ try:
 except:
     # If SCons is using an old python (2.4 is relatively common still), then the above may fail.
     # In this case, we need to be a little trickier.
-    execfile(os.path.join(GetBuildPath('#galsim'),'_version.py'))
+    vfilename = os.path.join(GetBuildPath('#galsim'),'_version.py')
+    exec(compile(open(vfilename, "rb").read(), vfilename, 'exec'))
     version = __version__
     version_info = __version_info__
 
--- a/SConstruct
+++ b/SConstruct
@@ -183,7 +183,7 @@ def ErrorExit(*args, **kwargs):
 
     import shutil
 
-    out = open("gs_error.txt","wb")
+    out = open("gs_error.txt","w")
 
     # Start with the error message to output both to the screen and to gs_error.txt:
     print()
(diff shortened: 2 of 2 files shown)


timj:
Everything did build and this patch only touches the scons files so I'm not quite sure why the tests failed.

rmjarvis:
It looks like the culprit is astropy 2.0.  The default pip version for astropy switched from 1.3 to 2.0, and I'm guessing there must be something in there that is breaking some of the tests.

@jmeyers314 It's also happening with [your PR](https://github.com/GalSim-developers/GalSim/pull/892).  Could you take a look?  The failures seem to be mostly with tests that you wrote, so you are probably best equipped to evaluate whether (and how) the test needs to change or if there is something in the backend code that needs to be fixed to accommodate their change.  (My off-the-cuff guess is it's something with the astropy.units module.)

@timj In the meanwhile, if you want to change the pip install line in .travis.yaml to use astropy==1.3, that would probably make this PR pass the tests.

jmeyers314:
Looking at this now...  Looks like at least part of the problem is that Planck's constant has changed! (maybe related to the upcoming redefinition of the kilogram?)...  Working on a fix...

rmjarvis:
Yep.  Looks like they switched from CODATA 2010 values to CODATA 2014.
image

jmeyers314:
I think this is fixed now on commit [d4c602b](https://github.com/GalSim-developers/GalSim/pull/892/commits/d4c602b7fcda639960b17c50fd086aabab7ade3a) if you want to cherry-pick that @timj.  I also made a few other changes on PR #892 for things that were failing on my laptop, but not on Travis.

cwwalter:
@rmjarvis I'm curious: where did you find that plot? I found this:

http://www.codata.org/uploads/RMP.88.035009.pdf

But that has a somewhat different version.

rmjarvis:
http://www.nature.com/nphys/journal/v12/n1/full/nphys3612.html

cwwalter:
Thanks!

src/SConscript
```diff
@@ -12,7 +12,8 @@ try:
 except:
     # If SCons is using an old python (2.4 is relatively common still), then the above may fail.
```
timj: One thing I forgot to say was that the above code *always* fails so I'm not entirely sure what the purpose of the try block is. The comment is obviously out of date.

rmjarvis: Good point.  The `nada()` call looks like maybe a way to force the except block for testing.  `git blame` says it was me that added it, so probably I just forgot to remove it after testing that the except block was working properly.  Oops.  Did now.  (Of course, the "2.4 is relatively common still" comment is indeed out of date.  None of my systems have that anymore.)

Issue #898, opened 2017-07-30
INSTALL.md comments about numpy


rmandelb:
I'm finding that in the GalSim master branch, we use numpy.moveaxis, which first appeared in v1.11.  The INSTALL file says we regularly test on 1.11, but other versions may work.  I think we should explicitly say that numpy versions >=1.11 may work, but that versions below that definitely will not.  Please let me know if there are any objections to that change, otherwise I will make it on master.

rmjarvis:
Actually, I just discovered that the numpy at nersc doesn't have moveaxis, so I think I'm in favor of fixing our code to not require this feature.  At least for now...

rmandelb:
Closing #898.

PR #900, opened 2017-08-02
#898 Two fixes to make GalSim run at Nersc

rmjarvis:
1. Don't use `np.moveaxis` if it doesn't exist.  I think the workaround (using two `swapaxes` calls) may be equally efficient, but I'm not sure, so I left in the `moveaxis` version to use when possible and only do the workaround if necessary.

2. I was getting a failure in the test of the `retry_io` feature.  It worked fine when running `python test_config_output.py` or `nosetests`, but not `scons tests` which runs nosetests in multiprocessing mode.  I still don't completely understand why it was not working properly, but I switched the test to setup the rng more like how we normally do things in config, using `galsim.config.GetRNG` rather than using a local variable.  That made it work, so I don't think there is any bug in the backend code.  Just the test seems to have been a bit fragile.

### DIFF
--- a/galsim/real.py
+++ b/galsim/real.py
@@ -1294,8 +1294,13 @@ class ChromaticRealGalaxy(ChromaticSum):
             NSED, Nim, nk, nk)
 
         # Reorder these so they correspond to (NSED, nky, nkx) and (NSED, NSED, nky, nkx) shapes.
-        coef = np.moveaxis(coef, (0,1,2), (1,2,0))
-        Sigma = np.moveaxis(Sigma, (0,1,2,3), (2,3,0,1))
+        try:
+            coef = np.moveaxis(coef, (0,1,2), (1,2,0))
+            Sigma = np.moveaxis(Sigma, (0,1,2,3), (2,3,0,1))
+        except AttributeError:  # pragma: no cover (numpy < 1.11)
+            # I think this prob less efficient, so only use the workaround for old numpy versions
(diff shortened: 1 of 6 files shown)


galsim/real.py
```diff
@@ -1294,8 +1294,13 @@ def _initialize(self, imgs, bands, xis, PSFs,
             NSED, Nim, nk, nk)
 
         # Reorder these so they correspond to (NSED, nky, nkx) and (NSED, NSED, nky, nkx) shapes.
-        coef = np.moveaxis(coef, (0,1,2), (1,2,0))
-        Sigma = np.moveaxis(Sigma, (0,1,2,3), (2,3,0,1))
+        try:
```
jmeyers314: Just use transpose?
```
coef = np.transpose(coef, (1,2,0))
Sigma = np.transpose(Sigma, (2, 3, 0, 1))
```

rmjarvis: Thanks.  That's better.  I didn't know about the version of transpose that takes a list of ints.

Although, the convention is different, so the first one is actually (2,1,0), not (1,2,0).  And of course the [example](https://docs.scipy.org/doc/numpy/reference/generated/numpy.transpose.html) that numpy gives in the documentation is ambiguous about which way the permutation happens.  :-P

rmjarvis:
Plan to merge Wednesday unless there are any objections.  This will get Josh's astropy fixes onto master (which I cherry-picked here).

PR #902, opened 2017-08-16
#809d Remove boost special functions

rmjarvis:
This is another foray into removing boost from GalSim.  This one focuses on the boost special functions that we had been using in a few places in the C++ layer.

It turns out that there is some public domain code that does what we needed at netlib called [SLATEC]( http://www.netlib.org/slatec/fnlib/).  Here is their "license":
> The SLATEC Library is intended to have no restriction on its distribution;
    therefore, new routines must be in the public domain.  This is generally
    not a problem since most authors are proud of their work and would like
    their routines to be used widely.

So that's the good news.  The bad news is that it's in Fortran, and some of it was probably the worst spaghetti code I'd ever seen.  (cf. [dbesj](http://www.netlib.org/slatec/src/dbesj.f) in particular)  But I did convert all of it into more modern looking C++, getting rid of all the goto's.  The implementation is still pretty obscure, so the checks are mostly that I didn't break any of our unit tests along with some comparisons to scipy results.

I also did a bit of minor code reorg to make a galsim::math namespace in C++ to hold all of these along with a few things we already had like this (sinc, Si, BesselJ0Root).


src/math/BesselK.cpp
```diff
@@ -0,0 +1,1037 @@
...
+        const int nulim[2] = { 35,70 };
+
+        const double elim = -std::log(std::numeric_limits<double>::min() * 1.e3);
+        const double xlim = std::numeric_limits<double>::min() * 1.e3;
+
+        assert(fnu >= 0.);
+        assert(x > 0.);
+
+        if (x < xlim)
+            throw std::runtime_error("DBESK OVERFLOW, FNU OR N TOO LARGE OR X TOO SMALL");
```
rmandelb: I'm a little confused by this error, because it only checks that x is too small, but then says that it could also be that fnu or n is too large?

src/math/BesselY.cpp
```diff
@@ -0,0 +1,1432 @@
...
+
+        const double xlim = std::numeric_limits<double>::min() * 1.e3;
+        const double elim = -std::log(xlim);
+        const int nulim[2] = { 70, 100 };
+
+        assert(fnu >= 0.);
+        assert(x > 0.);
+
+        if (x < xlim)
+            throw std::runtime_error("DBESY OVERFLOW, FNU OR N TOO LARGE OR X TOO SMALL");
```
rmandelb: Same question about this runtime error as above (text vs. test mismatch).

src/math/Bessel.cpp
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    double cyl_bessel_y(double nu, double x)
+    {
+        // Negative arguments yield complex values, so don't do that.
+        // And Y_nu(0) = inf.
+        if (x <= 0)
+            throw std::runtime_error("cyl_bessel_y x must be > 0");
+
+        // Identity: Y_-nu(x) = cos(pi nu) Y_nu(x) + sin(pi nu) J_nu(x)
```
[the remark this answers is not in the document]
rmjarvis: Actually, J is correct here.  Similar to J_-nu(x) = cos(pi nu) J_nu(x)  - sin(pi nu) Y_nu(x).  It's derivable from eqn 9 here: http://mathworld.wolfram.com/BesselFunctionoftheSecondKind.html

src/math/BesselI.cpp
```diff
@@ -0,0 +1,956 @@
...
+        //    750101  DATE WRITTEN
+        //    890531  Changed all specific intrinsics to generic.  (WRB)
+        //    890911  Removed unnecessary intrinsics.  (WRB)
+        //    890911  REVISION DATE from Version 3.2
+        //    891214  Prologue converted to Version 4.0 format.  (BAB)
+        //    900315  CALLs to XERROR changed to CALLs to XERMSG.  (THJ)
+        //    900326  Removed duplicate information from DESCRIPTION section.
+        //            (WRB)
+        //    920501  Reformatted the REFERENCES section.  (WRB)
+        //    170203  Converted to C++, and modified to only cover n==1 option. (MJ)
```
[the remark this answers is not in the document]
rmjarvis: lol.  Probably not.  But, I'm not pushing my update back up to netlib either...

src/math/BesselK.cpp
rmjarvis: They had one error message that can happen from a few cases in the code.  I didn't bother to change the error message.  But yeah, in this case, it's just the x too small part.  Below there are other ones that can hit if nu is very large.

src/math/BesselY.cpp
rmjarvis: Ditto.  There are 3 other cases that emit this error message, which can be because nu is large.

rmjarvis:
@jmeyers314 Did you want to look at this?  I'm assuming no one else does.  If not, I'll plan to merge on Monday.

jmeyers314:
>@jmeyers314 Did you want to look at this? I'm assuming no one else does. If not, I'll plan to merge on Monday.

I'll take a look tomorrow.

jmeyers314:
Looks good to me.  Just out of curiosity, why the netlib conversion instead of just copying over the boost code straight into the GalSim repo (like the random number stuff)?  Too many other boost dependencies?

rmjarvis:
> Too many other boost dependencies?

To be honest, it didn't occur to me.  But yeah, their code is a sticky web of dependencies.  It was a huge pain to cut all the tendrils from the boost.random stuff, so I'm not sure if it would have been easier to do that again for the math stuff or do the netlib conversion that I did.

PR #904, opened 2017-09-02
Deprecations for 2.0

rmjarvis:
In working on preparing for version 2.0, I've changed a few parts of the API for various reasons.  This is technically allowed in a major version upgrade, but of course, we'd like to avoid it as much as possible.    In order to lessen the pain of end users making the switch, it will be helpful to have as many of these as possible show up as deprecation warnings in a 1.x branch prior to 2.0, so here they are.

The first few are to match the API of Angle and CelestialCoord in the LSSTDESC repo, which we'll be using in lieu of the GalSim versions of these classes.  Most of the API is identical (since it was based on our classes), but I did make a few what-I-think-are-improvements when I ported it over to its own repo.

* `Angle.rad` is now a property rather than a method.
* `HMS_Angle` and `DMS_Angle` functions are now class methods `Angle.from_hms` and `Angle.from_dms`
* `_ecliptic_to_equatorial` is now class method `CelestialCoord.from_ecliptic`
* There is also a `CelestialCoord.from_galactic` that we don't use, but I back-ported to here for completeness.

The only other change when switching to use the LSSTDESC.Coord classes is the way project and deproject work.  Those have different conventions for the inputs and outputs (with better units checking), such that I don't think we can make this work with a proper deprecation.  So that usage will just have to be called out as a change in the CHANGELOG for 2.0.

Others are mostly to make the API a bit more Pythonic or to remove implementation details from the public API.

* `ShapeletSize` function is now `Shapelet.size` class method.
* `FitShapelet` is now `Shapelet.fit` class method.
* `Interpolant` base class constructor, which used to act as a factory function is now `Interpolant.from_name`.
* `GSObject` base class constructor is now invalid.
* `GSObject.SBProfile` attribute is now `_sbp` and is officially an implementation detail, not part of the public API.
* Similarly, `image` attribute of `Image` instance is now `_image` and officially an implementation detail.  (It was already not documented, so that should be pretty safe.)
* `PhotonArray.addTo(image)` takes a regular Image instance, not a C++-layer image class.  I'm pretty sure that had been the only public API use of the C++-layer image type.
* `PhotonArray.get*` functions are deprecated in lieu of the simpler properties.  Also `PhotonArray.setPhoton` is deprecated as also unnecessary.  (These all remain as C++ functions.)
* We had lots of other `get*` methods that were equivalent to a property.  E.g. `obj.getFlux() == obj.flux`.  This was mostly a result of C++ programmers writing GalSim as they were learning Python and inappropriately transporting their (my) C++ ways over to the Python layer.  It seems worth switching to the more Pythonic style and have "one obvious way" to get this information (i.e. the property/attribute).  Most of the documentation was already using the property syntax, so in most cases, I doubt this will cause much hardship for people.

Let me know if you have any objections to any of these.  Or if there are other aspects of the current API that you think ought to be tweaked slightly in advance of v2.0.


rmandelb:
Given other deadlines looming, the earliest I could do this is a week for now.  Apologies for the delay.

rmjarvis:
No problem.  I think I'm done adding to it now.  The main thing is really to read through the listing in the PR description and make sure you are ok with all of these deprecations.  Going through the code itself is probably not super exciting (as with most of my recent PRs).

Also, for perspective on the larger goal of #809, I think after the current #809e, there will be just one more code PR to review.  Then I'll probably produce two branches for people to try out.  One using cffi and one using pybind11.  I think I'll have all the wrapping requirements simple enough by then that I can do both without too much trouble.  (Although I reserve the right to give up on one of them if need be.)  Then we can see what system-dependent snags we run into for each of them and make a decision about which way we want to go.

TL;DR I think we can probably have a GalSim 2.0 ready to go in ~2 months or so using setup.py rather than boost::python.

rmandelb:
Great.  If you were trying to incentivize me to review these PRs, the "~two months to GalSim 2.0" did the job.  I hope to be back to a more normal reviewing schedule once the holidays are over.

rmjarvis:
Good.  That was my intent.  :)

Do this one first.  Once this is merged in (and then merged over to noboost), it will reduce the number of lines to review in #809e considerably.

jmeyers314:
Hey @rmjarvis ,  this all compiles for me in py2 and py3.  I think the list of deprecations you mentioned above all make sense.  I'm looking through the modules now that I've developed to see if there are any API changes I'd like to make there.

jmeyers314:
I came up with a couple of ideas for consistency and implementation hiding:

- Demote `ChromaticObject.obj` to an implementation detail `ChromaticObject._obj` since not all subclasses implement this.
- Consistently pick either `obj_list` (used in compound.py)  or `objlist` (used in chromatic.py) for compound objects.
- Demote `OpticalScreen.coef_array` -> `OpticalScreen._coef_array`.

Other than that, this all looks good to me.

rmandelb:
@rmjarvis - taking your suggestion, I read through the list of deprecations in the first comment in this issue.  These all sound OK to me.  I was also learning python as we were starting to put GalSim together and now find some of the older code a bit un-pythonic, so I'm in favor of updates to fix that as we move towards v2.0.

Tests pass on both of my systems.

Please let me know if you need a proper review rather than just this quick look.

rmjarvis:
Thanks both.  @jmeyers314 I added the items you suggested.  

Probably worth at least one of you doing a quick code review, or at least signing off on it by giving an approved review if you're happy enough to trust me that the code is ok.

jmeyers314:
I can take a look through the code on Wednesday if there's no hurry.  

One final thought: I think you could potentially go further with the getters -> properties.  I'd vote for `stepK`, `maxK`, `nyquistScale`, `hasHardEdges`, `isAxisymmetric`, `isAnalyticX`, `isAnalyticK`, `centroid`, and `maxSB` to all be made properties too.  I suppose the problem is that it's not possible to deprecate the method while making a property of the same name?  (At least, not without mucking around with `.__getattr__`?)

rmjarvis:
Sure.  It is possible, and I know how (cf. Angle.rad on this branch), so I'll go ahead and make those changes too.  I was thinking they were all semi-implementation details that users don't need much, so I wasn't worried about it.  But in fact, in the upcoming '#809f', they are all implemented with leading underscore attributes anyway, so it would be in that same vein to make the top level access also be properties.

galsim/chromatic.py
```diff
@@ -211,17 +218,18 @@ def _fiducial_profile(self, bandpass):
 
     def __eq__(self, other):
         return (isinstance(other, galsim.ChromaticObject) and
-                self.obj == other.obj)
+                hasattr(other, '_obj') and  # not all ChromaticObjects have an _obj attribute.
```
jmeyers314: This feels a little weird to me.  Two `ChromaticObject`s can only compare equal if they both have `_obj` attrs (and they're equal).  As it turns out, I think only `ChromaticObject` proper and `ChromaticFourierSqrt` can trigger this comparison, since the other subclasses override with their own `__eq__`.  Since both `ChromaticObject` and `ChromaticFourierSqrt` both have `_obj` defined, this works (actually, I suppose this means the`hasattr` check isn't actually needed).  But if we ever defined another subclass without a `._obj` and without it's own `__eq__`, this might inappropriately fail.  Not sure if this really needs attention or not, but I thought I'd flag it.

rmjarvis: I think it is required for e.g. `ChromaticObject() == ChromaticConvolution()`.

rmjarvis: Actually, I guess Python's decision tree about how to call `__eq__` is [more complicated than I thought](https://stackoverflow.com/questions/3588776/how-is-eq-handled-in-python-and-in-what-order), so I guess it wasn't actually necessary.  Still it seems like an appropriate thing to include here.

rmjarvis:
Thanks Josh.  I fixed the examples to not use the deprecated syntax.  Thanks for spotting that.

Issue #905, opened 2017-09-08
Opaque raise of bad redshift in sed.withMagnitude()


matroxel:
Mike asked to submit example of opaque traceback of impact of weird redshifts on sed.withMagnitude(). Example below, along with raise and traceback. Probably sufficient to add check on valid redshifts in sed.atRedshift().

Example:
sedpath = os.path.join(galsim.meta_data.share_dir, 'SEDs', 'CWW_Sbc_ext.sed')
galaxy_sed = galsim.SED(sedpath, wave_type='Ang', flux_type='flambda')
galaxy_sed = galaxy_sed.atRedshift(-99)
galaxy_sed = galaxy_sed.withMagnitude(22,wfirst.getBandpasses()['F184']) * galsim.wfirst.collecting_area * galsim.wfirst.exptime

Raise:
ValueError: zero-size array to reduction operation minimum which has no identity

Traceback:
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/users/PCON0003/cond0083/lib/python2.7/site-packages/galsim/sed.py", line 626, in withMagnitude
    current_magnitude = self.calculateMagnitude(bandpass)
  File "/users/PCON0003/cond0083/lib/python2.7/site-packages/galsim/sed.py", line 700, in calculateMagnitude
    flux = self.calculateFlux(bandpass)
  File "/users/PCON0003/cond0083/lib/python2.7/site-packages/galsim/sed.py", line 678, in calculateFlux
    return np.trapz(bandpass(x) * self(x), x)
  File "/users/PCON0003/cond0083/lib/python2.7/site-packages/galsim/sed.py", line 385, in __call__
    return self._call_fast(wave)
  File "/users/PCON0003/cond0083/lib/python2.7/site-packages/galsim/sed.py", line 321, in _call_fast
    self._check_bounds(wave)
  File "/users/PCON0003/cond0083/lib/python2.7/site-packages/galsim/sed.py", line 274, in _check_bounds
    wmin = np.min(wave)
  File "/users/PCON0003/cond0083/.conda/envs/local/lib/python2.7/site-packages/numpy/core/fromnumeric.py", line 2397, in amin
    out=out, **kwargs)
  File "/users/PCON0003/cond0083/.conda/envs/local/lib/python2.7/site-packages/numpy/core/_methods.py", line 29, in _amin
    return umr_minimum(a, axis, None, out, keepdims)
ValueError: zero-size array to reduction operation minimum which has no identity

[...]

PR #907, opened 2017-09-09
#905 Check for valid redshifts

jmeyers314:
Adds some sanity checks and unit tests for valid redshift inputs to `SED.atRedshift`.

### DIFF
--- a/galsim/sed.py
+++ b/galsim/sed.py
@@ -228,9 +228,6 @@ class SED(object):
     def dimensionless(self):  # for convenience
         return not self.spectral
 
-    def _wavelength_intersection(self, other):
-        return max([self.blue_limit, other.blue_limit]), min([self.red_limit, other.red_limit])
-
     def _rest_nm_to_photons(self, wave):
         from astropy import units
         _photons = units.astrophys.photon/(units.s * units.cm**2 * units.nm)
@@ -644,7 +641,7 @@ class SED(object):
             depr('Using calculateFlux(bandpass=None) to compute a bolometric flux', 1.5, '',
                  "If you need this functionality, you can use a pseudo-bolometric Bandpass created "
                  "with:  bp = Bandpass('1', blue_limit=sed.blue_limit, red_limit=sed.red_limit)")
-            bp = galsim.Bandpass('1', self.blue_limit, self.red_limit)
+            bp = galsim.Bandpass('1', 'nm', self.blue_limit, self.red_limit)
             return self.calculateFlux(bp)
         else: # do flux through bandpass
             if len(bandpass.wave_list) > 0 or len(self.wave_list) > 0:
--- a/galsim/sed.py
+++ b/galsim/sed.py
@@ -634,6 +634,8 @@ class SED(object):
 
         @returns the redshifted SED.
         """
+        if redshift <= -1:
+            raise ValueError("Invalid redshift {0}".format(redshift))
         wave_factor = (1.0 + redshift) / (1.0 + self.redshift)
         wave_list = self.wave_list * wave_factor
         blue_limit = self.blue_limit
--- a/galsim/sed.py
+++ b/galsim/sed.py
@@ -640,7 +640,12 @@ class SED(object):
         if self.dimensionless:
             raise TypeError("Cannot calculate flux of dimensionless SED.")
         if bandpass is None: # do bolometric flux
-            return galsim.integ.int1d(self, 0.0, float('inf'))
+            from galsim.deprecated import depr
+            depr('Using calculateFlux(bandpass=None) to compute a bolometric flux', 1.5, '',
+                 "If you need this functionality, you can use a pseudo-bolometric Bandpass created "
+                 "with:  bp = Bandpass('1', blue_limit=sed.blue_limit, red_limit=sed.red_limit)")
+            bp = galsim.Bandpass('1', self.blue_limit, self.red_limit)
+            return self.calculateFlux(bp)
         else: # do flux through bandpass
             if len(bandpass.wave_list) > 0 or len(self.wave_list) > 0:
                 slop = 1e-6 # nm
--- a/galsim/sed.py
+++ b/galsim/sed.py
@@ -640,7 +640,8 @@ class SED(object):
             from galsim.deprecated import depr
             depr('Using calculateFlux(bandpass=None) to compute a bolometric flux', 1.5, '',
                  "If you need this functionality, you can use a pseudo-bolometric Bandpass created "
-                 "with:  bp = Bandpass('1', blue_limit=sed.blue_limit, red_limit=sed.red_limit)")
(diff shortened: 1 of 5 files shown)


galsim/utilities.py
```diff
@@ -1145,6 +1145,8 @@ def combine_wave_list(*args):
             red_limit = min(red_limit, obj.red_limit)
         wave_list = np.union1d(wave_list, obj.wave_list)
     wave_list = wave_list[(wave_list >= blue_limit) & (wave_list <= red_limit)]
+    if blue_limit > red_limit:
```
rmjarvis: Should this maybe check if wave_list is empty?  Or can that only happen if blue_limit > red_limit?  

Also, even only having a single element is likely a problem, so might be worth giving an error for that too.

jmeyers314: Actually, I did try `if len(wave_list) == 0` before settling on `if blue_limit > red_limit`.  The former led to a dozen or so errors in the unit tests though.  Empty wave_lists are valid if you have an SED or Bandpass with an analytic wavelength dependence instead of one based off a LookupTable, so I don't think that's the right thing to check for.

I agree a single element would be suspicious.  I think the right way to handle that may be to change `>` -> `>=` above, rather than looking at wave_list directly.

rmjarvis:
Here's a quick script that might be added to your unit tests somewhere:
```
sed = galsim.SED(galsim.LookupTable([1,2,3,4,5], [1.1,1.9,1.4,1.8,2]),
                 wave_type='nm', flux_type='fphotons')
bp = galsim.Bandpass(galsim.LookupTable([4,5],[1,1], interpolant='linear'),
                     wave_type='nm')

for z in [0, 1, 2, 3, 3.5, 3.99, 4, 4.01]:
    print('z = ',z,' flux = ',sed.atRedshift(z).calculateFlux(bp))
```
The question is really what should happen for the last two redshifts.  Currently this is the output:
```
z =  0  flux =  1.9
z =  1  flux =  1.78470982143
z =  2  flux =  1.63869047619
z =  3  flux =  1.24875837054
z =  3.5  flux =  0.58363707623
z =  3.99  flux =  0.0110121850677
z =  4  flux =  0.0
  File "junk.py", line 10, in <module>
    print('z = ',z,' flux = ',sed.atRedshift(z).calculateFlux(bp))
  File "/Users/Mike/lib/python2.7/site-packages/galsim/sed.py", line 679, in calculateFlux
    x, _, _ = galsim.utilities.combine_wave_list(self, bandpass)
  File "/Users/Mike/lib/python2.7/site-packages/galsim/utilities.py", line 1149, in combine_wave_list
    raise RuntimeError("Empty wave_list intersection.")
RuntimeError: Empty wave_list intersection.
```
I actually think that is probably fine for z=4 where the intersection has 1 element.  The flux comes out 0, which is correct.  

For the last one though, should we allow this without error, and give a flux of 0?  Or do you think the error is the right runtime behavior?  (And if so, maybe something from calculateFlux would be better, since this is still a little obscure for a user to figure out.)

jmeyers314:
>For the last one though, should we allow this without error, and give a flux of 0? Or do you think the error is the right runtime behavior?

We raise an exception when requesting the flux density of an SED outside of its defined range, so I think the right behavior is is to raise an exception for calculateFlux when that function would have asked about an out-of-bounds wavelength.  I agree that it would be helpful for that message to come from calculateFlux itself though.

(Note that the opposite behavior is true for Bandpass though, where requesting the throughput for an out-of-bounds wavelength does return 0.0.  My intuition for both of these choices was that users probably don't mean that an object's SED is really zero outside the range they specify, it's just unspecified.  But they probably do mean that a Bandpass throughput is zero outside of the specified range.)

rmjarvis:
> users probably don't mean that an object's SED is really zero outside the range they specify,

I think I agree with that.  But in that case, I guess there should be an error thrown if any part of the bandpass is outside the sed defined range.  So z=3.5 should raise an exception too in my example above.  (Also 3.01.)

jmeyers314:
>I guess there should be an error thrown if any part of the bandpass is outside the sed defined range.

Yep.  I agree with that.  Will fix...

jmeyers314:
This is making me think we should also remove the bolometric flux calculation, which should nominally require the SED to be defined all the way from 0 -> infinity (so only possible using an analytic function SED, not a LookupTable.)  It's pretty easy to roll a new bandpass with throughput 1 and blue/red limits equal to that of the SED if anyone actually does want/use this.

rmjarvis:
> we should also remove the bolometric flux calculation

I don't have any preference.  Probably not many people use it.  But if we, at this point, it would just be a deprecation, so we can see if anyone complains...

jmeyers314:
Okay, I think I fixed the invalid evaluation of SEDs inside `calculateFlux` bug, and also deprecated the bolometric flux.

I also simplified the default values of `SED.blue_limit` and `SED.red_limit` to be 0 and infinity instead of `None`.  This eliminated a handful of `if blue_limit is not None` kinds of checks that are no longer necessary.

galsim/sed.py
```diff
@@ -230,21 +229,7 @@ def dimensionless(self):  # for convenience
...
-
-        red_limit = self.red_limit
-        if other.red_limit is not None:
-            if red_limit is None:
-                red_limit = other.red_limit
-            else:
-                red_limit = min([red_limit, other.red_limit])
-
-        return blue_limit, red_limit
+        return max([self.blue_limit, other.blue_limit]), min([self.red_limit, other.red_limit])
```
rmjarvis: Apparently, this function is not actually used by anything, so could just remove it.

galsim/sed.py
```diff
@@ -651,29 +632,26 @@ def calculateFlux(self, bandpass):
...
-            if len(self.wave_list) > 0:
-                return(np.trapz(self(self.wave_list), self.wave_list))
-            else:
-                return galsim.integ.int1d(self, blue_limit, red_limit)
+            from galsim.deprecated import depr
+            depr('Using calculateFlux(bandpass=None) to compute a bolometric flux', 1.5, '',
+                 "If you need this functionality, you can use a pseudo-bolometric Bandpass created "
+                 "with:  bp = Bandpass('1', blue_limit=sed.blue_limit, red_limit=sed.red_limit)")
+            bp = galsim.Bandpass('1', self.blue_limit, self.red_limit)
+            return self.calculateFlux(bp)
```
rmjarvis: This needs a test in test_deprecated.  Normally the existing test of this functionality would be moved over to there with the check_dep call.  But there wasn't one, so I guess you'll need to make something for this functionality.  Probably could make a simple sed that has an analytic bolometric flux and make sure it comes out right.

jmeyers314:
Thanks @rmjarvis.  I think I fixed the missing test coverages.  Let me know if you see anything else missing.  Otherwise, I'll plan to merge tomorrow.

galsim/sed.py
rmjarvis: Looks like the depr message should get the 'nm' edit too...

PR #909, opened 2017-09-18
- add the real second moments into the result data structure

ocordes:
Hi, for some further analysis we need to have the measured 2nd order moments directly in the results structure instead of the calculated values for e1 and e2.  I simply add the m_xx, m_yy and m_xy values to the CppShape structures. Hopefully you can merge these changes in the main code! 

Cheers,
        Oliver

### DIFF
--- a/galsim/hsm.py
+++ b/galsim/hsm.py
@@ -152,6 +152,9 @@ class ShapeData(object):
             self.moments_amp = data.moments_amp
             self.moments_centroid = data.moments_centroid
             self.moments_rho4 = data.moments_rho4
+            self.moments_m_xx = data.moments_m_xx
+            self.moments_m_yy = data.moments_m_yy
+            self.moments_m_xy = data.moments_m_xy
             self.moments_n_iter = data.moments_n_iter
             self.correction_status = data.correction_status
             self.corrected_e1 = data.corrected_e1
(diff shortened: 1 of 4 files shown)


rmandelb:
Note that the second moments structure has e1, e2, and sigma = det(M)^(1/4) stored.  Those uniquely determine the xx, yy, and xy moments already.  It's not clear that it makes sense to carry around 3 additional numbers when the information is already there, but perhaps we could provide a utility that takes the input ShapeData structure and calculates Mxx, Myy, Mxy from the information it contains?

ocordes:
Yes, but you're loosing precision for values which are already there ... for a simple analysis this maybe a good solution just to use e1,e2 and sigma but we need the accurate moments for a Deimos code implementation!

rmandelb:
Sorry, I don't fully understand this point.  You are saying that with e1, e2, and sigma stored as floats, the numerical error in using the equations to get Mxx, Myy, and Mxy leads to errors that actually matter, when compared to storing Mxx, Myy, and Mxy directly?  I've never had that problem myself, it all seems to work at 10^-6 accuracy.

rmjarvis:
I agree with Rachel.  It doesn't make much sense to duplicate information like this.  

If your problem is that the e1,e2,sigma values are not high enough precision for your use case (which seems unlikely, but I suppose possible), then I'd rather switch the underlying storage to double in C++ rather than float.

If you just want it to be easier to access the raw moments, then I'd rather add properties for mxx, mxy, myy that calculate these from the e1,e2,sigma values.

PR #910, opened 2017-09-20
Permit zero-flux drawPhot

jmeyers314:
I think drawing a zero-flux object with photon shooting should just return a 0 photon image.  This comes up occasionally when photon-shooting chromatic objects, where `sed(w)*bandpass(w)` can sometimes be zero at one of the sampled wavelengths.  (Photon-shooting chromatic objects doesn't quite work yet, but I've started thinking about it...)

This PR is just to avoid the divide-by-zero error currently encountered on master when photon-shooting a zero-flux object.

### DIFF
--- a/tests/test_draw.py
+++ b/tests/test_draw.py
@@ -1012,7 +1012,7 @@ def test_shoot():
 
     # Test that shooting with 0.0 flux makes a zero-photons image.
     image4 = (obj*0).drawImage(method='phot')
-    np.testing.assert_equal(image4.array, image4.array*0)
+    np.testing.assert_equal(image4.array, 0)
 
 
 @timer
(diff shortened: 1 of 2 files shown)


tests/test_draw.py
```diff
@@ -1010,6 +1010,10 @@ def test_shoot():
     # It's not exactly the same, since the rngs are realized in a different order.
     np.testing.assert_allclose(image3.array, image1.array, rtol=0.25)
 
+    # Test that shooting with 0.0 flux makes a zero-photons image.
+    image4 = (obj*0).drawImage(method='phot')
+    np.testing.assert_equal(image4.array, image4.array*0)
```
rmjarvis: This is fine, but slightly wasteful.  You could just compare it to a literal `0`.

jmeyers314: Oh, interesting.  I didn't realize it would do the broadcast.  Thanks.

PR #911, opened 2017-09-21
#809e Switch from boost::shared_ptr to std::shared_ptr

rmjarvis:
Here is the next installment in the effort to rid GalSim of the boost dependency.

This PR nominally is about switching from `boost::shared_ptr` to either `std::shared_ptr` or `std::tr1::shared_ptr` depending on which is available.  At least one of these is available back to GCC 4.1, so I don't think there is any worry about people not having this option.

The other big result from this effort is to get rid of the `NumpyHelper.h` file.  This has a lot of complicated wrapping details, especially when we wanted to return a numpy array allocated in C++, that were hard to transfer over to pybind11 (never mind cffi or other wrapping modes).  Now everything that used to return a numpy array instead has the numpy array allocated in python and then passed in to C++ to be filled in.  This vastly simplifies the wrapping layer code with respect to numpy arrays.

Along the way, i made a few other changes, partly to make that switch easier (in the pysrc code) and partly just as general code cleanup.  Some of these also ended up in PR #904 as deprecations.

* Switched from `SBProfile` to `_sbp` as the name of the SBProfile attribute.  Mostly so the C++-layer SBProfile object is not part of the public API.  This means we don't need to have the SBProfile classes be picklable, which simplifies a lot of the wrapping requirements.  All the python-layer classes are still fully picklable, but their `_sbp` attributes are not necessarily so.
* Changed the Image class `image` attribute to `_image` to make it an implementation detail rather than something that users should ever use.
* Rewrote Bounds and Position in python with `_b` and `_p` properties to return the version that is needed for any calls to the C++ layer.
* Similarly, rewrote `GSParams`, `ShapeParams`, and `HSMParams` in python with attributes that are used for C++ calls.  This means that the python layer is responsible for the memory handling, which means that shared_ptrs are not required for passing these things into C++ function calls.
* Rewrote all the `Interpolant` classes in python with `_i` as the (lazy) property to get the version needed for C++ calls, which again now don't need the shared_ptr in the wrapping.  This was the impetus for switching to the `Interpolant.from_name` syntax that I introduced in #904.  When this is written in python, it's not especially easy to preserve the old version that has that functionality in the constructor.  Which pointed out how non-pythonic that syntax was. 
* Rewrote `PhotonArray` in python, so the python layer handles all the memory management.  Now the `shoot` functions that used to return a PhotonArray take an empty PhotonArray as an argument and fill it.  This means that we don't need complicated numpy array wrapping in pysrc.
* Removed the wrapping of `LVector`, since it was an implementation detail already anyway.  Now `Shapelet` just uses a numpy array directly in python for the b vector. This means `SBShapelet` doesn't need the numpy stuff in pysrc anymore either.
* Switched `FitShapelet` and `ShapeletSize` to class methods `Shapelet.fit` and `Shapelet.size`, which seem more pythonic than having these extra free functions in the `galsim` namespace.
* Rewrote `LookupTable` in python, again to avoid the numpy array handling in pysrc.  `LookupTable` and `LookupTable2D` are now responsible for the memory management of everything accessible from python (the C++ layer might allocate memory for the spline interpolation, but that array is just an implementation detail in C++).  So again, the numpy stuff is no longer needed in pysrc.
* Simplified the return value of LookupTable calls.  We had been aggressively trying to match the output type to the input.  But I think in python, it is better to just return a consistent type (numpy array here) and let duck typing allow for the multiple types of input.  This is a (mild) API change that I mention in the CHANGELOG.
* Simplified some of the code in `NFWHalo` and `PowerSpectrum`, including particularly the return values.  Again, I think keeping the return values simpler is more pythonic where duck typing should suffice for any use case when people might have expected lists or tuples rather than numpy arrays.

Implementation notes:
* I had to port over a few more boost functions into our copy of boost/random that used to be included when we included `<boost/shared_ptr>`, so you'll see those as added files.  As usual, they are just things from the 1.48 version of the boost distro.
* I switched the C++ `Table` class from being a template to using double explicitly.  We only ever used the <double,double> template anyway, so it seemed a bit gratuitous to have it be a template.  It also (IMO) makes the code easier to read.
* Also, there is now a subclass `TableBuilder` which can be used to handle the memory management in C++ when needed.  Things like the SBSersic Hankel function calculation that uses a lookup table can use that to handle the memory and still use the same backend `Table` class for the interpolation.
* Aside from pickle tests, the one place in the test suite where we used the SBProfile attribute explicitly was a call to `prof.SBProfile.xValue(galsim.PositionD(r,0))` in a tight loop to avoid the slowish `parse_pos_args()` call.  This use case is now achieved with `prof._xValue(galsim.PositionD(r,0))`.


jmeyers314:
Hi @rmjarvis,

I'm getting a pair of test failures on this one (noboost branch tests fine).  Both look fairly innocuous; probably relaxing the test tolerances a little would be okay.

[tests.log](https://github.com/GalSim-developers/GalSim/files/1327656/tests.log)

rmjarvis:
Thanks Josh.  I found a system that reproduced your error, and I was able to fix it.  One was a new test, which I was able to make more accurate and pass as is.  

The other was the PhaseScreenPSFs not using enough photons for the shooting test to pass reliably.  I changed the implementation of it slightly, which changed the order that it used the random numbers.  Not sure why it passes on some machines but not others, but probably something about the numerical details of the compiler.  With a non-deterministic seed, it would pass ~half the time in both directions, so I convinced myself it was just a spurious pass previously.  

Anyway, I was already using 10x more photons for InterpolatedImage profiles, and the PhaseScreenPSF (at least sometimes) uses an InterpolatedImage behind the scenes, so I just have it also use 10x more photons now, and it passes.

pysrc/Table.cpp
```diff
@@ -18,276 +18,120 @@
...
-            table2d.gradient(x, y, dfdx, dfdy);
-            return bp::make_tuple(dfdx, dfdy);
+            return new Table2D(x, y, vals, Nx, Ny, i);
         }
 
-        static void GradientMany(const Table2D<double,double>& table2d,
-                                 const bp::object& x, const bp::object& y,
-                                 const bp::object& dfdx, const bp::object& dfdy)
+        static void interpMany(const Table2D& table2d,
+                               size_t ix, size_t iy, size_t ivals, int N)
```
jmeyers314: Where do these size_t's come from?  Just the type that gets assigned to blah.ctypes.data?  Feels weird, but I guess it works....

galsim/base.py
```diff
@@ -147,12 +147,6 @@ def __str__(self):
 
 _galsim.SBGaussian.__getinitargs__ = lambda self: (
         self.getSigma(), self.getFlux(), self.getGSParams())
-# SBProfile defines __getstate__ and __setstate__.  We don't actually want to use those here.
```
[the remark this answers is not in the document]
rmjarvis: Mostly it's that we don't have anything that returns a bare SBProfile to python.  So we always know what actual type it is, and we don't need the generic prescription for pickling an SBProfile object.  The specific ones (e.g. `_galsim.SBGaussian.__getinitargs__`) are sufficient.

Aside: In the next #809 PR, I'll go another step and remove even these, since I won't depend on the various SBProfile types being picklable to make the python type picklable.  In fact, it will even be option whether a specific GSObject subclass has any `_sbp` attribute at all or whether the implementation is defined in pure python.

galsim/interpolatedimage.py
```diff
@@ -579,11 +574,75 @@ def __setstate__(self, d):
...
+                       use_true_center=True, offset=None, gsparams=None,
+                       force_stepk=0., force_maxk=0.):
+    """Approximately equivalent to InterpolatedImage, but with fewer options and no sanity checks.
+
+    Some notable reductions in functionality relative to InterpolatedImage:
+
+    1. There are no padding options. The image must be provided with all padding already applied.
+    2. The stepk and maxk values will not be calculated.  If you want to use values for these other
+       than the default, you may provide them as _force_stepk and _force_maxk.  Otherwise
+       stepk ~= 2pi / image_size and max_k ~= 2pi / pixel_scale.
```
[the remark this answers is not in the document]
rmjarvis: Yes.  Good catch.

include/galsim/SBAiryImpl.h
```diff
@@ -96,7 +95,7 @@ namespace galsim {
     class AiryInfoObs : public AiryInfo
     {
     public:
-        AiryInfoObs(double obscuration, const GSParamsPtr& _gsparams);
+        AiryInfoObs(double obscuration, GSParamsPtr _gsparams);
```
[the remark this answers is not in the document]
rmjarvis: No good reason.  I think it's a remnant of some code changes I was trying out where I got rid of GSParamsPtr entirely and switched to copying the GSParams object into the C++ layer, but then I decided I wanted to keep it this way, so I put them back in and didn't use the same syntax.

So I can switch it back.  The const at least is probably worth keeping as a contract that this won't change the gsparam values.  The reference isn't really that important when passing a smart pointer though, since one point of the smart pointer is to make it safe to pass by value.

pysrc/Table.cpp
rmjarvis: These are basically `void*` pointers, which isn't something python understands directly, so the trick is to call them `size_t`, and then `reinterpret_cast` them back into the kind of pointer you actually have.  This is what I use in TreeCorr all the time to pass arrays to C++ with `ctypes.data` (which is a `double*`).  I've also already used it in GalSim for the WCS speedup functions `ApplyPV`, `ApplyCD`, etc.

jmeyers314: Okay, but why `size_t` then?  If it's arbitrary why not just call them `double*` from the start?  (I guess it's not actually arbitrary?)

rmjarvis: Right.  It's not arbitrary.  `ctypes.data` in python really is returning an integer, since python can't deal with any kind of pointer directly.  So the dance is that numpy's C-layer data object is really a `double*`.  Then in python you can access it as an integer and pass it to your C-layer functions.  Then you recast it back to a `double*`.  I use `size_t`, because that's the C++ name for the integer type that is guaranteed to hold the same number of bytes as a pointer, which makes it safe to do this.

For more information, [here](https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.ndarray.ctypes.html) is the numpy description of the ctypes.data attribute which says it is "A pointer to the memory area of the array as a Python integer."

jmeyers314: Okay, sounds good.  Thanks for the link.

rmjarvis:
Thanks Josh.  I made the edits you suggested.

@rmandelb I'm not ready with the next one of these yet, so there's no rush to merge this if you think you'll want to take a look.  But if not, we can probably go ahead and merge this now.

rmandelb:
Realistically I think it'll be hard for me to look in the next week, so you can go ahead if you want.

PR #912, opened 2017-09-25
Refactored image integrators to use less RAM.

jmeyers314:
I modified the code that integrates `ChromaticObject`s over wavelength so that it accumulates monochromatic images into a result image as it iterates over wavelengths instead of storing all of the monochromatic images in memory before adding them together.  For my test script below, memory usage decreased from ~1.7GB to ~0.1GB.

```python
import galsim
sed = galsim.SED("CWW_E_ext.sed", wave_type="A", flux_type="flambda")
bandpass = galsim.Bandpass("LSST_r.dat", wave_type="nm")
obj = galsim.Gaussian(fwhm=0.1)+galsim.Gaussian(fwhm=10.0)
obj = galsim.ChromaticObject(obj).dilate(lambda w: (w/500)**1.05)*sed
image = obj.drawImage(bandpass)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,10 @@ API Changes
   If you want a view, you should use the more intuitive image.view().  (#873)
 - Changed behaviour of the `preload` option in RealGalaxyCatalog and
   COSMOSCatalog to preload data in memory, not just the fits HDUs (#884)
+- Changed the type of acceptable integration rules for
+  galsim.integ.ImageIntegrator subclasses.  This should be transparent if you
+  have been using strings to specify integrators for chromatic object
+  drawImage.  (#887).
 
 
 Dependency Changes
--- a/galsim/integ.py
+++ b/galsim/integ.py
@@ -102,6 +102,42 @@ def trapz(func, min, max, points=10000):
     return np.trapz(func(points),points)
 
 
+def midptRule(f, xs):
+    """Midpoint rule for integration.
+
+    @param f   Function to integrate.
+    @param xs  Locations at which to evaluate f.
+
+    @returns  Midpoint rule approximation to the integral.
+    """
+    if len(xs) < 2:
+        raise ValueError("Not enough points for midptRule integration")
+    x, xp = xs[:2]
+    result = f(x)*(xp-x)
+    for x, xp, xpp in zip(xs[0:-2], xs[1:-1], xs[2:]):
+        result += 0.5*f(xp)*(xpp-x)
+    result += f(xpp)*(xpp-xp)
+    return result
+
+
+def trapzRule(f, xs):
+    """Trapezoidal rule for integration.
+
+    @param f   Function to integrate.
+    @param xs  Locations at which to evaluate f.
+
+    @returns  Trapezoidal rule approximation to the integral.
+    """
+    if len(xs) < 2:
+        raise ValueError("Not enough points for trapzRule integration")
+    x, xp = xs[:2]
+    result = 0.5*f(x)*(xp-x)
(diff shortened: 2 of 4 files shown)


tests/test_chromatic.py
```diff
@@ -164,7 +164,7 @@ def test_draw_add_commutativity():
     chromatic_kimage = galsim.ImageCD(stamp_size, stamp_size, scale=pixel_scale)
     # use chromatic parent class to draw without ChromaticConvolution acceleration...
     t4 = time.time()
-    integrator = galsim.integ.ContinuousIntegrator(galsim.integ.midpt, N=N, use_endpoints=False)
+    integrator = galsim.integ.ContinuousIntegrator(galsim.integ.midptRule, N=N, use_endpoints=False)
```
rmjarvis: I think I'd rather this be merely a deprecation at this point, rather than a direct API change.  I don't think there is a need to break backwards compatibility here.

It seems like this could be done in the constructors for `SampleIntegrator` and `ContinuousIntegrator`. 
 You could add an explicit check for `midpt`, `trapz` or `np.trapz` (which was also previously documented as allowed), and switch the `rule` over to the corresponding new function with a deprecation warning.

CHANGELOG.md
```diff
@@ -45,6 +45,10 @@ API Changes
   If you want a view, you should use the more intuitive image.view().  (#873)
 - Changed behaviour of the `preload` option in RealGalaxyCatalog and
   COSMOSCatalog to preload data in memory, not just the fits HDUs (#884)
+- Changed the type of acceptable integration rules for
```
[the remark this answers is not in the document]
rmjarvis: Suggest this comment move to the deprecations section (if you accept my suggestion to make it such) and add the RAM reduction to New Features.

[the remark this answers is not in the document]
jmeyers314: Yep.  I reworked the CHANGELOG to address both comments.

galsim/integ.py
```diff
@@ -146,17 +183,27 @@ class SampleIntegrator(ImageIntegrator):
...
+                            galsim.integ.trapzRule  --  Use the trapezoidal integration rule
     """
     def __init__(self, rule):
+        if rule == midpt:
+            from galsim.deprecated import depr
+            depr('galsim.integ.midpt', 1.5, 'galsim.integ.midptRule')
+            rule = midptRule
+        elif rule == np.trapz:
+            from galsim.deprecated import depr
+            depr('galsim.integ.trapz', 1.5, 'galsim.integ.trapzRule')
```
rmjarvis: I think this should read `depr('np.trapz', ...`.  Also, should include a check for `if rule == galsim.inted.trapz` with this message.

rmjarvis:
One more thing.  There are some tests in `test_integ.py` that check the `galsim.integ.midpt` and `galsim.integ.trapz` funtions.  These should probably be switched over to the new functions too, sine the old ones aren't actually used for anything.  (You can move those tests over to test_deprecated.py probably.)

galsim/integ.py
jmeyers314: >I think this should read depr('np.trapz', .... 

Yes, good catch.  

>Also, should include a check for if rule == galsim.inted.trapz with this message.

`galsim.integ.trapz` was never an option for an `ImageIntegrator`.  It has a different syntax than `np.trapz` or `galsim.integ.midpt`.  I think Alex added it as a convenience function when he added the sampling wavelengths from `SED`s code.

rmjarvis: I see.  Yeah, he uses it in the tests for the SED sampling.  OK.

PR #915, opened 2017-10-12
Use char * rather than std::string in C++ interfaces

esheldon:
This addresses an incompatibility found when python/boost/numpy etc. are compiled with gcc `7.2.0` and galsim is compiled with `6.2.0`.  This is an issue for interfaces between the C++ and python generated by boost-python

With these changes the tests all pass.

I only modified the interfaces for which a test failed, I did not exhaustively change all interefaces.  There could be some not covered by tests which would still exhibit this error.

### DIFF
--- a/include/galsim/Noise.h
+++ b/include/galsim/Noise.h
@@ -50,7 +50,7 @@ namespace galsim {
          * @param[in] rng   The BaseDeviate to use for the random number generation.
          */
         BaseNoise(boost::shared_ptr<BaseDeviate> rng) : _rng(rng)
-        { if (!_rng) _rng.reset(new BaseDeviate(0)); }
+        { if (!_rng) _rng.reset(new BaseDeviate( (long)0)); }
 
         /**
          * @brief Copy constructor shares the underlying rng.
--- a/include/galsim/Random.h
+++ b/include/galsim/Random.h
@@ -132,8 +132,14 @@ namespace galsim {
         /**
          * @brief Construct a new BaseDeviate from a serialization string
          */
-        BaseDeviate(const std::string& str) : _rng(new rng_type())
+        BaseDeviate(const char * str_c) : _rng(new rng_type())
         {
+            std::string str;
+            if (str_c==NULL) {
+                str="";
+            } else {
+                str=str_c;
+            }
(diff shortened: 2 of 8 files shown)


esheldon:
actually I just verified that `run_examples.py` works, not `scons tests`

esheldon:
The remaining failures in `scons tests`  I don't understand.  They have to do with missing entries in configurations, and it isn't clear to me how this is related to any C++ issues.

e.g.
```
        # However, an input field that does have nobj will return something for nobjects.                                                                                                                                                                                                              
        # This catalog has 3 rows, so equivalent to nobjects = 3                                                                                                                                                                                                                                       
        config['input'] = { 'catalog' : { 'dir' : 'config_input', 'file_name' : 'catalog.txt' } }                                                                                                                                                                                                      
>       del config['input_objs']                                                                                                                                                                                                                                                                       
E       KeyError: 'input_objs'                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                       
test_config_image.py:773: KeyError  
```

Maybe @rmjarvis will understand these

esheldon:
Note however, that just running nosetests works fine, and travis is happy

So maybe the above are  just a bug in `scons tests`?

esheldon:
I used to have write access, but I guess I no longer do. Can someone please merge?

rmjarvis:
We usually give a little more time than that in case anyone else wants to take a look.  But I'll merge this tomorrow unless anyone objects that they want more time.

dannygoldstein:
I also got those 'input_objs' errors when running scons tests. Below is a gist showing the output of the command:

https://gist.github.com/dannygoldstein/e18866ebb9c39a2739f7b9f16440e2f5

I found that all of these errors were fixed I installed nose, which it turned out was not present on the machine I was running on.

rmjarvis:
Thanks Danny.  I knew nose was required for assert_raises (although Josh removed that dependency in PR #932), but I didn't realize that there must be some other subtle dependency on it for some of these tests.  Glad to hear installing nose fixed these errors.

For version 2.0 (cf. PR #950), I added nose as a test dependency, so it should get installed automatically for test runs (but not be required just for installing galsim).

PR #916, opened 2017-10-24
Newsensor

craiglagegit:
I've made the following changes:
(1) Updated the share/sensors/lsst_* files with pixel distortion information from the latest simulations.  This now includes both ITL sensors and preliminary sims of E2V sensors.  The new pixel distortion info has fixed the discrepancy between the Poisson_CCD simulator and GalSim.  The agreement can be seen at devel/brighter-fatter/GalSim_24Oct17.pdf.

(2) Added tree rings to the pixel distortion calculations.  The tests can be seen in devel/treerings/Spot_Shift.pdf, and in tests/test_sensor.py.  The user can specify the distortion pattern in one of three ways, all three of which are tested in test_sensor.py:
     (A) By specifying the center and period of a default cosine pattern.
     (B) By specifying the center and an arbitrary function f(r).
     (C) By specifying the center and a lookup table of r and f(r) values.

(3) Updated the Si absorption length calculation, which was using an approximate function, to use a lookup table derived from astropy.  This should be more accurate and faster.  The lookup table data is in share/sensors/absorption, and there is a plot at share/sensors/absorption/Abs_Length.png that shows a comparison of the new and old absorption lengths.

A couple of comments:
(1) When I added the absorption length lookup table to be passed to C++, I apparently discovered a limit of 15 parameters that can be passed to the C++ functions.  I fixed this by combining the treeRingCenterx and treeRingCentery parameters into a single PositionD instance, which probably makes more sense anyway.

(2) The speed of the treering calculation is interesting.  If the treeRingAmplitude value is small, say 0.01 pixels, then the treerings have no impact on the speed.  However, when the treeRingAmplitude parameter is large it slows down substantially, and is about 3X slower when treeRingAmplitude=1.0 pixels.  I think this is showing that the speed-up Mike did to look first in the undistorted pixel is working, because with these large distortions most of the electrons don't land in the undistorted pixel.  This shouldn't make any difference in real usage, because the tree ring distortions are typically quite small, but I thought it was interesting.

Craig Lage


craiglagegit:
Mike,


Looks good to me.  No need to keep the old PixelSize option from my standpoint.  Help me understand what happens next.  It looks like you've made the changes in the newsensor branch.  Will they be merged into a mainstream branch eventually, or has that already happened?  Which mainstream branch will they be merged into: master, releases/1.4, or something else?  I ask because I want to pull the changes down and do a few simple tests, and I wanted to pull from the right place.


Thanks,

Craig

________________________________

rmjarvis:
> It looks like you've made the changes in the newsensor branch.  Will they be merged into a mainstream branch eventually, or has that already happened?

Yes, the newsensor branch has the changes I made, so you can download from there if you want to test it.  This will be merged into master and become part of the upcoming v1.5 release.  Let's leave this open until Friday in case anyone else wants to take a look at the changes, but if no one objects, I'll try to remember to merge it into master Friday morning before I leave for Australia.

PR #918, opened 2017-11-02
#675

rmandelb:
This PR is for updates in the WFIRST module as listed in https://github.com/GalSim-developers/GalSim/issues/675#issuecomment-335673276 (which describes what was done, and what was deferred for later).


galsim/wfirst/wfirst_wcs.py
```diff
@@ -39,31 +39,40 @@
...
 # testing.
 sca_rot = np.zeros_like(sca_xc_mm)
 
-# Rotation of focal plane axes relative to payload axes: positive CCW rotation of the f2 axis
-# relative to -Y_pl, and of f1 relative to +X_pl.
-theta_fpa = 32.5*galsim.degrees
+# Rotation of focal plane axes relative to payload axes: this is expressed as a CCW rotation of the
+# f2 axis relative to -Y_pl, and of f1 relative to +X_pl.  In cycle 5, this was around +30 degrees,
+# but in cycle 6 there is a 90 degrees CCW rotation with respect to that.
+theta_fpa = -60.0*galsim.degrees
```
rmjarvis: Is it exactly -60 now?  Your comment implies that it should be 32.5 - 90 = -57.5.  But maybe the update also included 32.5 -> 30.

tests/test_wfirst.py
```diff
@@ -37,107 +37,50 @@
 
 @timer
 def test_wfirst_wcs():
-    """Test the WFIRST WCS routines against those from software provided by WFIRST project office.
+    """Test the WFIRST WCS routines against the one produced by code from Chris Hirata.
```
rmjarvis: I wonder if it would be worth keeping the old test in here, but disabling it with a comment that we need to get a new updated WCS from the WFIRST project office to compare against.  (e.g. rename it `dont_test_wfirst_wcs`)  Then give your current test here  a different name like `test_wfirst_find_sca` where we only compare against Chris's SCA choices.  Because there are quite a few tests that got removed by the change here, which would be nice to be able to bring back eventually.

galsim/wfirst/__init__.py
```diff
@@ -231,7 +232,7 @@
                         [0.001270391, 0.016129619, 0.001200137] ])
 ipc_kernel /= np.sum(ipc_kernel)
 ipc_kernel = galsim.Image(ipc_kernel)
-persistence_coefficients = np.array([0.2246,0.0225,0.0085,0.0043,0.0025,0.0016,0.0011,0.0008])/100.
+persistence_coefficients = np.array([0.045707683,0.014959818,0.009115737,0.00656769,0.005135571,0.004217028,0.003577534,0.003106601])/100.
```
rmjarvis: We don't have any tests of most of these numbers.  Not exactly sure what those tests should be, but maybe worth adding some dummy tests in the test suite to remind ourselves of what we're missing.  E.g.
```
def test_wfirst_persistence():
    """Test the WFIRST persistence coefficients against official WFIRST project data
    """
    # Values from WFIRST project office.
    ref_coeffs =[0.045707683, 0.014959818, 0.009115737, 0.00656769, 0.005135571, 0.004217028, 0.003577534, 0.003106601] # percent
    ref_coeffs = np.array(ref_coeffs) / 100.
    np.testing.assert_array_equal(galsim.wfirst.persistence_coefficients, ref_coeffs)
    # TODO: Some kind of better check here?  Or maybe reference against an official WFIRST document/white paper, etc.?
```
Maybe this is gratuitous, but I'm always a little worried when code changes and no test needed to change along with it.

galsim/wfirst/wfirst_wcs.py
rmandelb: Yes, it's exactly -60 (confirmed with Chris and others).  Comment should have a twiddle; it's around 90, not exactly.  Will push this change in next round of commits.

tests/test_wfirst.py
rmandelb: Yes, that's a good idea.  To be honest it isn't clear when/if we'll get something to compare against from the project office (I've talked with Jeff about it but he is very busy).  Still,  doing what you suggest will serve as a reminder of what kinds of tests we may want to re-institute in some other way later on.  Will do in a sec.

galsim/wfirst/__init__.py
rmandelb: OK... that's pretty gratuitous, and I was going to save beefing up the unit tests for the open issue on that topic, but I guess it's better to get this in before we tag v1.5.  Done.

rmandelb:
OK, this is kind of weird: I have the new `dont_test_wfirst_wcs` routine in `test_wfirst.py`, and it's not listed as a test we're supposed to run, but travis is trying to run it anyway and then complaining when it doesn't work.  @rmjarvis - do you know how to tell travis not to run that test?

rmjarvis:
> do you know how to tell travis not to run that test?

I thought it only ran tests that start with "test", but maybe it runs anything with test in the name.  So try renaming it something else.  Like "skip_wfirst_wcs" or something.

rmandelb:
That seems to have done the job.  Thanks.

rmjarvis:
I'm fine with merging this if you want.  I don't think anyone else will probably want to review it, nor do I think it is necessary.

Then we should start the process of releasing version 1.5.

rmandelb:
Merging.

Issue #917, opened 2017-11-01
Allow HSM adaptive moments to use circular weight function


rearmstr:
The HSM adaptive moments code that is exposed uses an elliptical weight function when computing moments.  There is also existing code in the HSM source file to use a circular weight function.  This issue is to modify the call to the adaptive moments code to allow for the use of a circular weight function.

rearmstr:
@rmandelb, I think the code is working now, but now I'm not sure what is the best radius to return.  

The ```find_mom_2``` function returns the width of the best fit Gaussian and a set of moments.  My first thought was to use the returned width, but doing some measurements on simple sheared Gaussians, this number seems to give the same results as using the elliptical weights.  Maybe I am doing something wrong here?

rmandelb:
Are you comparing the `moments_sigma` outputs for the two cases?  If so, then I think what you're finding is what we expect.  The `moments_sigma` output is based on the determinant so it should be relatively insensitive to shear.  The issue we were having with the EstimateShear resolution factor is because it is based on the trace (which is what we want for noisy images because it's linear in the moments, but it has this problematic shape dependence).  

So the idea (I thought) was that you'd have the new code return the `moments_sigma` and `observed_shape` from the circularly-weighted moments, and we'd use that to get a trace estimate that is less sensitive to the shear.  But the determinant which is in `moments_sigma` is already insensitive to shape for a Gaussian model.

rmandelb:
BTW I tried using your code, but was getting exceptions.  Is the latest version pushed?

rearmstr:
> Are you comparing the moments_sigma outputs for the two cases? If so, then I think what you're finding is what we expect. The moments_sigma output is based on the determinant so it should be relatively insensitive to shear. The issue we were having with the EstimateShear resolution factor is because it is based on the trace (which is what we want for noisy images because it's linear in the moments, but it has this problematic shape dependence).

Yes, that is what I was comparing.  I see what you mean.  How would you recommend returning the observed shape?  There is a moments matrix that is calculated, but I'm not sure what is computed exactly.

> BTW I tried using your code, but was getting exceptions. Is the latest version pushed?
Strange.  Yes, pushed version should be the latest.

rmandelb:
> There is a moments matrix that is calculated, but I'm not sure what is computed exactly.

[...]

PR #922, opened 2017-11-16
#917

rearmstr:
I've added the option to compute the HSM adaptive moments using a circular weight function.  The code to do this was already in PSFCorr.cpp, but it was not accessible to the outside.  I've added a simple argument to FindAdaptiveMom to specify if you want to use a round weight function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Deprecated Features
 
 New Features
 ------------
+- Use circular weight function in HSM adaptive moments code. (#917)
 
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ Deprecated Features
 
 New Features
 ------------
-- Use circular weight function in HSM adaptive moments code. (#917)
+- Add option to use circular weight function in HSM adaptive moments code. (#917)
 
 
 
--- a/src/hsm/PSFCorr.cpp
+++ b/src/hsm/PSFCorr.cpp
@@ -1874,5 +1874,6 @@ namespace hsm {
         const BaseImage<int>& object_image, const BaseImage<int> &object_mask_image,
         double guess_sig, double precision, galsim::Position<double> guess_centroid,
         bool round_moments, boost::shared_ptr<HSMParams> hsmparams);
+
 }
 }
--- a/src/hsm/PSFCorr.cpp
+++ b/src/hsm/PSFCorr.cpp
@@ -276,8 +276,8 @@ namespace hsm {
                   dbg<<"Repackaging find_mom_2 results"<<std::endl;
                   results.moments_amp =  3.544907701811*sig*moments(0,0);
                   results.moments_sigma = sig;
-                  results.observed_e1 = (moments(2,0)-moments(0,2)) / moments(0,0);
-                  results.observed_e2 =(1./std::sqrt(2.))*(moments(1,1)) / moments(0,0);
+                  results.observed_e1 = 0.5*(moments(2,0)-moments(0,2)) / moments(0,0);
+                  results.observed_e2 = (1./std::sqrt(2.))*(moments(1,1)) / moments(0,0);
                   results.moments_status = 0;
              }
         }
--- a/src/hsm/PSFCorr.cpp
+++ b/src/hsm/PSFCorr.cpp
@@ -270,7 +270,7 @@ namespace hsm {
                   dbg<<"About to get moments using find_mom_2"<<std::endl;
                   tmv::Matrix<double> moments(3,3);
                   double sig = guess_sig;
-                  find_mom_2(masked_object_image_cview, moments, 3,
+                  find_mom_2(masked_object_image_cview, moments, 2,
                              results.moments_centroid.x, results.moments_centroid.y, sig,
                              hsmparams->convergence_threshold, results.moments_n_iter, hsmparams);
                   dbg<<"Repackaging find_mom_2 results"<<std::endl;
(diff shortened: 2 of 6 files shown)


tests/test_hsm.py
```diff
@@ -139,6 +139,24 @@ def test_moments_basic():
...
+                # test for moments with a circular weight function
+                result_round = gal_image.FindAdaptiveMom(round_moments=True)
+                q = np.exp(-2.*math.atanh(total_shear))
+                theta = 0.5*math.atan2(g2, g1)
+                e_round = (1 - q**2)/(1 + q**2 + 2*q)
+                e1_round = e_round*math.cos(2*theta)
+                e2_round = e_round*math.sin(2*theta)
+
+                np.testing.assert_almost_equal(np.fabs(result_round.moments_sigma-sig/pixel_scale), 0.0,
+                                               err_msg = "- incorrect round dsigma", decimal = decimal)
```
rmandelb: I think `dsigma` should just be `sigma` here.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ Deprecated Features
 
 New Features
 ------------
+- Use circular weight function in HSM adaptive moments code. (#917)
```
rmandelb: I recommend making clear this is optional - something like "Add option to use circular..."

src/hsm/PSFCorr.cpp
```diff
@@ -247,18 +253,33 @@ namespace hsm {
...
+                  results.moments_status = 0;
+             } else {
+                  dbg<<"About to get moments using find_mom_2"<<std::endl;
+                  tmv::Matrix<double> moments(3,3);
+                  double sig = guess_sig;
+                  find_mom_2(masked_object_image_cview, moments, 2,
+                             results.moments_centroid.x, results.moments_centroid.y, sig,
+                             hsmparams->convergence_threshold, results.moments_n_iter, hsmparams);
+                  dbg<<"Repackaging find_mom_2 results"<<std::endl;
+                  results.moments_amp =  3.544907701811*sig*moments(0,0);
```
rmandelb: Maybe refer to where these magic numbers / formulae come from?  (Point to later in the code in this file.)

tests/test_hsm.py
```diff
@@ -139,6 +139,24 @@ def test_moments_basic():
                 np.testing.assert_almost_equal(result.observed_shape.e2,
                                                distortion_2, err_msg = "- incorrect e2",
                                                decimal = decimal_shape)
+
+                # test for moments with a circular weight function
+                result_round = gal_image.FindAdaptiveMom(round_moments=True)
+                q = np.exp(-2.*math.atanh(total_shear))
+                theta = 0.5*math.atan2(g2, g1)
+                e_round = (1 - q**2)/(1 + q**2 + 2*q)
```
rmandelb: Should say where this formula comes from: it's what we expect when integrating an elliptical Gaussian light profile with a round Gaussian weight function that has the same sigma to get the moments (can confirm analytically).

tests/test_hsm.py
rearmstr: I'm happy to make the change, but all the other hsm tests use ```dismga``` when testing size.

src/hsm/PSFCorr.cpp
rearmstr: Where do they come from?  There is no documentation in the later part of the code where they are used.

rmandelb: Hmmm, I thought I found their origin in Hirata & Seljak (2003), based on some math in Bernstein & Jarvis (2002), but the actual equations in those papers seem to point to a slightly different answer.  But then I realized it's because some of the equations use different basis functions for the moments (cartesian vs. polar) than others - see comments starting around line 1357, which describe manipulations of `find_mom_1` outputs, for what made me realize this.  Bottom line is I think I can put it all together into an informative comment if I work carefully on connecting but I don't have time to do it today.  Are you OK with a slight delay?

rearmstr: I'm not too concerned about where they come from, but it would be good to have in the code at some point.  The delay is fine if you want to this soon.

tests/test_hsm.py
rmandelb: Huh.  I wonder why.

src/hsm/PSFCorr.cpp
rmandelb: Well, I have very good intentions but not necessarily the best follow-through at this time of the semester (teaching obligations etc.).  Since it's a matter of comments, let me suggest you not delay this PR for me.  If it's otherwise ready to merge, then we can do so and open an issue to remind me to improve the documentation of these numbers later.

rearmstr:
The documentation changes have been addressed and pushed.  Are there any other comments?  

It would be nice to merge this in the next couple of days.

rmandelb:
Thanks Bob!  I just approved the changes and it looks like Mike just did too, so we are good to go.  Maybe let's give it until tonight to let anybody else who wants to comment, and plan to merge first thing in the morning?

Does this need to go into a tagged version in order to be useful to you?

rearmstr:
Yeah, I think it needs to be a tagged version.

rmandelb:
@rmjarvis - even though it's technically a (quite minor) new feature, would you be OK with us making a version 1.5.1 that includes this so that it's in a tagged version?

rmjarvis:
Yes.  I can do that this week even.  Unless you want to wait for the WFIRST zeropoint to be included as well.

rmjarvis:
And maybe a fix for the LSST DESC folks if I can manage to figure out something that works there.  (I do hate conda, but I can probably take a stab at that Thursday.)

rmandelb:
I think we're close enough on the WFIRST stuff that that could be merged later this week, so let's wait for that as well.

In the meantime, I will merge this to master (and we can cherry-pick over to releases/1.5 later).

Issue #924, opened 2017-11-20
Build failures with anaconda python 3.6.3


timj:
I'm having trouble building GalSim (1.4 or 1.5) with Anaconda python 3.6.3 on macOS 10.13 in an LSST stack.

The conftests are linked with:

```
g++ -o .sconf_temp/conftest_109_mod/check_python.so -bundle .sconf_temp/conftest_109.os -L/Users/timj/work/lsstsw3/miniconda/lib/python3.6/config-3.6m-darwin -L/Users/timj/work/lsstsw3/miniconda/lib -Llib -L/Users/timj/work/lsstsw3/stack/DarwinX86/boost/1.60.lsst1+2/lib -L/Users/timj/work/lsstsw3/stack/DarwinX86/tmv/0.73.lsst1+2/lib -L/Users/timj/work/lsstsw3/stack/DarwinX86/fftw/3.3.4.lsst2/lib -L/Users/timj/work/lsstsw3/stack/DarwinX86/tmv/0.73.lsst1+2/lib -lpython3.6m -ltmv_symband -lfftw3 -ltmv
```
but when you try to import a module built this way you get this error:
```
Fatal Python error: PyThreadState_Get: no current thread
```
The problem is that `/Users/timj/work/lsstsw3/miniconda/lib/python3.6/config-3.6m-darwin` includes a `libpython3.6m.a` library and you get two python interpreters: one from the main python and one attached to the shared module. If you change the library search path to be `/Users/timj/work/lsstsw3/miniconda/lib/` then linking is much faster but it also fails to import. This is because Anaconda 3.6.3 doesn't link against an external python shared library itself.

```
$ otool -L `which python`
/Users/timj/work/lsstsw3/miniconda/bin/python:
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 853.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
```

Everything works fine if I do not link against `libpython` at all and add `-undefined dynamic_lookup` to the link. This is precisely what `distutils` does when it builds a library. If I look at the `distutils` code it seems to use `sysconfig.get_config_var("LDSHARED")` to determine how to build a shared library module and it does not explicitly link against `libpython`.

```
$ python3
Python 3.6.3 |Anaconda, Inc.| (default, Nov  8 2017, 18:10:31) 
[...]

PR #927, opened 2017-12-06
#924 Fix build problem for Anaconda Python 3.6

rmjarvis:
I reproduced the build problem @timj reported in #924 and fixed it on my laptop.  Tim, please see if this fixes the problem for you.  If so, we can get this into a release version 1.5.1.


rmjarvis:
Great!  Thanks, Tim.

Issue #928, opened 2017-12-08
cutout_row/cutout_col for MEDS output type when using offsets.


NiallMac:
I believe the cutout_row and cutout_col columns in the OBJECT_DATA extension of a MEDS file should give the (zero-offset) position of the object in pixel coordinates (@esheldon can confirm).  Currently this is set to the origin of the stamp wcs in WriteMeds in des_meds.py (called 'row0' and 'col0' in des_meds.py, see e.g. line 263). 

I think this behavior will become problematic as soon as you have more than one exposure of an object with different offsets. I think the solution is to have MEDSBuilder access the stamp offsets (I think this could be done with an extra output type, but maybe there is an easier way?), and make these an attribute of MultiExposureObject. Then WriteMEDS would also be able to access this info. I can go ahead and try to implement this, but keen to hear if @rmjarvis @esheldon or others have thoughts on this before I do so.

rmjarvis:
This seems like a good place to use test-driven design.  Let's start by writing a unit test that uses a few different kinds of WCS types (both Celestial and not at least) and asserting what we think des_meds should be doing with the offsets in each case in terms of cutout_row and cutout_col.

Then we can try to work out what code would be required to make that work correctly.

esheldon:
You are correct, see docs here: https://github.com/esheldon/meds/wiki/MEDS-Format

The "origin of the stamp" is to be stored in `orig_row, orig_col`

NiallMac:
Thanks both. In that case the current wcs treatment is probably fine. I think the key point is that the MEDSBuilder needs to know about the offsets, as these determine cutout_row and cutout_col, whereas currently it only has accesses to the images (after the objects have been drawn with offsets). I'll take a look at the current test_des.py, I think a test of this behaviour should be straightforward to add.

NiallMac:
@esheldon does 'zero-offset' mean the origin for cutout_row/col is at the center of pixel [0,0] or the bottom-left of pixel [0,0]?

esheldon:
SExtractor uses 1.0 = center of the first pixel.

NiallMac:
I'm not sure that answers my question - do you mean that cutout_row/col use the same convention as SExtractor?

esheldon:
Yes, all calculations for the MEDS creation work in the sextractor coordinates, except for the subtraction of 1 to put things into zero offset.

NiallMac:
[...]

PR #930, opened 2017-12-23
#928 Fix various small bugs in config processing

rmjarvis:
This PR includes a number of small bug fixes that @NiallMac and I ran into when trying to do some recent DES simulations.

1. The des_meds maker wasn't setting the position of the object within the stamp correctly.  It was just giving the center of the stamp, not including any sub-pixel offset.  This is now fixed.
2. The FocalPlane type (in the galsim_extra repo) uses multiple rng cadences to get some values to repeat for each chip in an exposure.  This required the seed dict for one of them to know the number of objects in the exposure.  But if the number of objects itself was a random variate, this led to a chicken and egg problem where the requisite rng wasn't set before it was needed.  The solution is to set the first rng in the list before parsing later ones.
3. FocalPlane needs to do some special setup before `SetupConfigRNG` is called.  There used to be a hacky workaround to make this kind of work.  But it was buggy, and the easiest fix seemed to be to just add an OutputBuilder function called `setup` that gets called at the start of building a file.  The default behavior for this function is just to call `SetupConfigRNG`, but it provides a non-hacky place for FocalPlane to do its extra setup stuff, which now seems to work correctly.
4. There was a very subtle bug in how the Eval code was working.  When checking which `eval_varaibles` to load up, it used to just check `if key in str` to see which keys were somewhere in the eval string.  But this found shorter words than were intended.  E.g. `focal_r` when `focal_rmax` was in the string.  This led to `focal_r` being evaluated as well as `focal_rmax`.  Normally this would be fine, but if this happened at image-processing scope rather than object-processing scope, the wrong rng could be used, which would not be re-calculated for the first object in an image, since the code would think it was already current.  TL;DR we now check for the whole word in the string using regular expressions.
5. There was a minor bug in SetupDefaultIndex where a non-default index_key wasn't being correctly propagated into the new index field.
6. Normally `world_center` is just calculated as the world position of the image center.  This is used whenever we need to do a tangent-plane projection of the world coordinates.  But FocalPlane uses the exposure center instead.  It had been using secret knowledge of GalSim's config code to effect this, but here I made this a real parameter in the `image` field that users can set, which seems like a better hook to provide.
7. The `eval_base_variables` list is now imported into the galsim.config namespace to make it more convenient for the user to add to.


rmjarvis:
@NiallMac I think with this branch for GalSim and the focal_plane branch in galsim_extra, everything seems to be working.  Please check it out when you get a chance.

galsim/config/output.py
```diff
@@ -418,6 +415,20 @@ class OutputBuilder(object):
     # A class attribute that sub-classes may override.
     default_ext = '.fits'
 
+    def setup(self, config, base, file_num, logger):
+        """Do any necessary setup at the start of processing a file)
```
rmandelb: random parenthesis

galsim/config/process.py
```diff
@@ -1019,21 +1031,22 @@ def GetIndex(config, base, is_sequence=False):
...
     """Get the appropriate current rng according to whatever the current index_key is.
 
     If a logger is provided, then it will emit a warning if there is no current rng setup.
 
     @param config           The configuration dict for the current item being worked on.
     @param base             The base configuration dict.
     @param logger           If given, a logger object to log progress. [default: None]
-    @param tag              If given, an appropriate name for the current item to use ing the
+    @param tag              If given, an appropriate name for the current item to use in the
                             warning message. [default: None]
```
rmandelb: default: '' (not None)

galsim/config/process.py
```diff
@@ -1046,14 +1059,33 @@ def GetRNG(config, base, logger=None, tag=None):
...
 
+def CleanConfig(config): # pragma: no cover
+    """Return a "clean" config dict without any leading-underscore values
+
+    GalSim config dicts store a lot of ancillary information internally to help improve
+    efficiency.  However, some of these are actually pointers to other places in the dict, so
+    printing a config dict, or even what should be a small portion of one, can have infinite loops.
+
+    This helper function is useful when debugging config processing to strip out all of these
+    leading-underscore values, so that printing the dict is reasonable.
```
rmandelb: Nice.  I will definitely appreciate this when debugging config-related issues.

tests/test_des.py
```diff
@@ -267,6 +267,13 @@ def test_meds_config():
     g1 = -0.17
     g2 = 0.23
 
+    # generate offsets that depend on the on the object num so can be easily reproduced
```
rmandelb: repeats "on the on the"

rmjarvis:
Thanks Rachel.  Typos fixed.

NiallMac:
All looks good to me, thanks both! 

I'll have a think about some more FocalPlane unit tests that test stuff at the image level as well as the truth catalogs...

PR #932, opened 2017-12-30
Scons3

jmeyers314:
I started a fresh python3.6 installation a few days ago and noticed a few problems while installing GalSim.  The first, which I think is related to using scons 3, is that the tests/SConscript file needed a few `.decode('utf-8')`s to be python3 (and still python2) compatible.

The second thing I noticed is that even after fixing the scons3 issue, a number of unit tests were failing.  It turns out that these were all cases where we had a `try/except ImportError` wrapper around an `assert_raises` test.  I hadn't installed nose yet, and so these were all hitting the except part of the try/except.  The problem was that a few of the try blocks had side effects after the exception got raised that we actually depended on below the try block.  When those parts of the try block got skipped, the test then failed.

To solve this second issue, I pulled the `assert_raises` part of nose directly into galsim_test_helpers.py, and then went through the unit tests to replace `np.testing.assert_raises` with `galsim_test_helpers.assert_raises`.  

I also took the liberty of using the `assert_raises` context manager syntax where I thought it read better.  I.e.:

```python
assert_raises(Exception, function, *args)
```

gets rewritten as

```python
with assert_raises(Exception):
    function(*args)
```

which I think is usually nicer since the function call syntax is the same as if you were calling it outside the testing context.  The exception to this is when several `assert_raises` are called in sequence (e.g., with slightly different arguments), and rewriting them all as context managers would muddy the nice sequence of nearly identical rows, so I continued to use the old syntax in those cases.

In the summary, scons3 now works, and I think nose is now completely dropped as even an optional dependency, and the try/except blocks previously guarding against a lack of nose are now executed without exception (handling).


rmandelb:
On my Mac and the Linux cluster (both still on python 2.7) I am getting lots of
```
test_bandpass.py:22: in <module>
    from galsim_test_helpers import *
galsim_test_helpers.py:608: in <module>
    assert_warns = getattr(_t, 'assertWarns')
E   AttributeError: 'Dummy' object has no attribute 'assertWarns'
```
when I do `scons tests` (all over the place).

jmeyers314:
Yeah, I just found out that `assertWarns` wasn't added to the `unittest` standard library package until python 3.2.  So I'll have to figure something else out.

rmjarvis:
In case you haven't gotten too far on this yet, I've been playing around with it while watching football.  I think this works:
```
if sys.version_info > (3,2):
    assert_warns = getattr(_t, 'assertWarns')
else:
    from contextlib import contextmanager
    import warnings
    @contextmanager
    def assert_warns(wtype, *args, **kwargs):
        if len(args) == 0:
            # When used as a context manager
            with warnings.catch_warnings(record=True) as w:
                warnings.simplefilter("always")
                yield w
            assert len(w) >= 1, \
                    "Calling %s did not raise a warning with %s, %s"%(f, args, kwargs)
            assert issubclass(w[0].category, wtype), \
                    "%s with %s, %s raised the wrong warning type"%(f, args, kwargs)
        else:
            # When used as a regular function
            func = args[0]
            args = args[1:]
            with assert_warns(wtype):
                res = func(*args, **kwargs)
            yield res
```
I have it locally, so I can commit it.  But I don't want to step on your toes if you already have something yourself.

rmjarvis:
Assuming you're not working now, so I went ahead and committed.

rmandelb:
That fixed it for me on both systems.

PR #940, opened 2018-01-22
VonKarman profile

jmeyers314:
This PR adds the VonKarman profile, which is a generalization of a Kolmogorov profile that adds an outer scale to the modeled atmospheric turbulence power spectrum.

This code emerged as part of the effort to implement a "second kick" profile that would help with photon shooting an atmospheric phase screen PSF.  Rather than trying to cherry-pick specific commits from that development branch, I found it easier to just copy the appropriate subset of files into a new branch and make a single commit.  So I'm afraid reviewers won't get to see my normal set of false-starts and other fumblings on this PR, (though they're still available in the `secondkick` branch).

The main surprise I encountered with the VonKarman profile was the realization that at its center, there's formally a delta-function component.  By default, the code doesn't actually implement this, since for most parameter combinations its fractional amplitude is ~10^-100 or smaller.  If the user really does want to acknowledge the delta-function though, it's possible via a keyword.

### DIFF
--- a/galsim/vonkarman.py
+++ b/galsim/vonkarman.py
@@ -55,9 +55,11 @@ class VonKarman(GSObject):
         exp(-0.5*0.172*(r0/L0)^(-5/3))
 
     In almost all cases of interest this evaluates to something tiny, often on the order of 10^-100
-    or smaller.  By default, GalSim will ignore this delta function entirely.  If for some reason
-    you want to keep the delta function, though, then you can pass the do_delta=True argument to the
-    VonKarman initializer.
+    or smaller.  By default, GalSim will ignore this delta function entirely since it usually
+    doesn't make any difference, but can complicate some calculations like drawing using
+    method='real_space' or by formally requiring huge Fourier transforms for drawing using
+    method='fft'.  If for some reason you want to keep the delta function, though, then you can pass
+    the do_delta=True argument to the VonKarman initializer.
 
     @param lam               Wavelength in nanometers
     @param r0                Fried parameter in meters.
--- a/galsim/vonkarman.py
+++ b/galsim/vonkarman.py
@@ -33,12 +33,13 @@ class VonKarman(GSObject):
     profile is that the von Karman profile includes a parameter for the outer scale of atmospheric
     turbulence, which is a physical scale beyond which fluctuations in the refractive index stop
     growing, typically between 10 and 100 meters.  Quantitatively, the von Karman phase fluctuation
-    power spectrum is proportional to
+    power spectrum at spatial frequency f is proportional to
 
-        (f^2 + L0^-2)^(-11/6)
+        r0^(-5/3) (f^2 + L0^-2)^(-11/6)
 
-    where f is a spatial frequency and L0 is the outer scale in meters.  The Kolmogorov power
-    spectrum proportional to f^(-11/3) is recovered as L0 -> infinity,
+    where r0 is the Fried parameter which sets the overall turbulence amplitude and L0 is the outer
+    scale in meters.  The Kolmogorov power spectrum proportional to r0^(-5/3) f^(-11/3) is recovered
+    as L0 -> infinity.
 
     For more information, we recommend the following references:
 
--- a/galsim/vonkarman.py
+++ b/galsim/vonkarman.py
@@ -144,7 +144,7 @@ class VonKarman(GSObject):
     def half_light_radius(self):
         return self._sbvk.getHalfLightRadius()
 
-    def structureFunction(self, rho):
+    def _structure_function(self, rho):  # pragma: nocover
         return self._sbvk.structureFunction(rho)
 
     def __eq__(self, other):
@@ -168,6 +168,8 @@ class VonKarman(GSObject):
             out += ", scale_unit=%r"%self.scale_unit
         if self.do_delta:
             out += ", do_delta=True"
+        if self._suppress_warning:
+            out += ", suppress_warning=True"
         out += ", gsparams=%r"%self.gsparams
         out += ")"
         return out
--- a/galsim/vonkarman.py
+++ b/galsim/vonkarman.py
@@ -61,7 +61,7 @@ class VonKarman(GSObject):
 
     @param lam               Wavelength in nanometers
     @param r0                Fried parameter in meters.
-    @param L0                Outer scale in meters.  [default: np.inf]
+    @param L0                Outer scale in meters.  [default: 25.0]
     @param flux              The flux (in photons/cm^2/s) of the profile. [default: 1]
     @param scale_unit        Units assumed when drawing this profile or evaluating xValue, kValue,
                              etc.  Should be a galsim.AngleUnit or a string that can be used to
@@ -77,7 +77,7 @@ class VonKarman(GSObject):
     @param gsparams          An optional GSParams argument.  See the docstring for GSParams for
                              details. [default: None]
     """
-    def __init__(self, lam, r0, L0=np.inf, flux=1, scale_unit=galsim.arcsec,
+    def __init__(self, lam, r0, L0=25.0, flux=1, scale_unit=galsim.arcsec,
                  do_delta=False, suppress_warning=False, gsparams=None):
         # We lose stability if L0 gets too large.  This should be close enough to infinity for
         # all practical purposes though.
--- a/galsim/vonkarman.py
+++ b/galsim/vonkarman.py
@@ -16,9 +16,7 @@
 #    and/or other materials provided with the distribution.
 #
 """@file base.py
-This file implements the von Karman atmospheric PSF profile.  A version which has the underlying
-turbulence power spectrum truncated below a given scale is also available as a correction when using
-geometric shooting through an atmospheric PhaseScreenPSF.
+This file implements the von Karman atmospheric PSF profile.
 """
 
 import numpy as np
--- a/galsim/vonkarman.py
+++ b/galsim/vonkarman.py
@@ -142,8 +142,8 @@ class VonKarman(GSObject):
     def half_light_radius(self):
         return self._sbvk.getHalfLightRadius()
 
-    def _structure_function(self, rho):  # pragma: nocover
-        return self._sbvk.structureFunction(rho)
+    def _structure_function(self, rho):
+        return self._sbvk.structureFunction(rho)  # pragma: nocover
 
     def __eq__(self, other):
         return (isinstance(other, galsim.VonKarman) and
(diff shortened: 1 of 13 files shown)


src/SBVonKarman.cpp
```diff
@@ -0,0 +1,395 @@
...
+         }
+    private:
+        const double _mkt;
+        const VonKarmanInfo& _vki;
+    };
+
+    VonKarmanInfo::VonKarmanInfo(double lam, double r0, double L0, bool doDelta,
+                                 const GSParamsPtr& gsparams) :
+        _lam(lam), _r0(r0), _L0(L0), _r0L0m53(pow(r0/L0, -5./3)), _gsparams(gsparams),
+        _deltaAmplitude(exp(-0.5*magic4*_r0L0m53)),
```
rmjarvis: This is the only place magic4 shows up.  Probably better to just define it right above this function, so the math is visible to anyone coming back to it later.

src/SBVonKarman.cpp
```diff
@@ -0,0 +1,395 @@
...
+    }
+
+    double VonKarmanInfo::structureFunction(double rho) const {
+    // rho in meters
+        double rhoL0 = rho/_L0;
+        if (rhoL0 < 1e-6) {
+            return -magic3*fast_pow(2*M_PI*rho/_r0, 5./3);
+        } else {
+            double x = 2*M_PI*rhoL0;
+            return magic1*_r0L0m53*(magic2-fast_pow(x, 5./6)*boost::math::cyl_bessel_k(5./6, x));
```
rmjarvis: Similarly, this is the only place magic1,2,3 are used.  These could be static const variables inside this function (which the compiler will compute at compile time, so no runtime cost), or global const variables defined just before this function.  Either way, it would be helpful to future developers looking at this code to see the definitions on the same screen as where they are used.

tests/test_vonkarman.py
```diff
@@ -0,0 +1,209 @@
...
+    t1 = time.time()
+    print("Time to create/draw first time: {:6.3f}s".format(t1-t0))
+    for i in range(10):
+        vk.drawImage(nx=16, ny=16, scale=0.2)
+    t2 = time.time()
+    print("Time to draw 10 more: {:6.3f}s".format(t2-t1))
+    for i in range(100):
+        vk.drawImage(nx=16, ny=16, scale=0.2, method='phot', n_photons=50000)
+    t3 = time.time()
+    print("Time to photon-shoot 100 more with 50000 photons each: {:6.3f}s".format(t3-t2))
```
rmjarvis: Worth putting a summary of what you discovered with this benchmark test somewhere in here?

tests/test_vonkarman.py
```diff
@@ -0,0 +1,209 @@
...
+    # Try to see if we can catch the warning first
+    with assert_warns(UserWarning):
+        vk = galsim.VonKarman(**kwargs)
+
+    kwargs['suppress_warning'] = True
+    vk = galsim.VonKarman(**kwargs)
+    do_pickle(vk)
+
+    # This profile has more than 15% of its flux in the delta-function component.
+    np.testing.assert_array_less(0.15, vk.delta_amplitude/vk.flux)
```
rmjarvis: This would probably read better to just use the assert statement, rather than np.testing.assert_array_less.  (Not least because it's not an array on the rhs.)
```
assert vk.delta_amplitude > 0.15 * vk.flux
```

tests/test_vonkarman.py
```diff
@@ -0,0 +1,209 @@
...
+
+@timer
+def test_vk(slow=False):
+    """Test the generation of VonKarman profiles
+    """
+    if __name__ == '__main__' and slow:
+        lams = [300.0, 500.0, 1100.0]
+        r0_500s = [0.05, 0.15, 0.3]
+        L0s = [1e10, 25.0, 10.0]
+        do_deltas = [False, True]
```
rmjarvis: How slow is this?  We have some pretty slow tests already in the name==main branches.  Even if it's a few minutes, I think we should probably remove the `slow` option and just always do this when run from the command line. 

One benefit of this is that `run_all_tests` will hit these slow tests, which is something I tend to run before releases to make sure even the weird stuff hasn't broken at some point along the way during development.

jmeyers314: Thanks for asking; I think I actually must not have run all of these since adding the `do_shoot` and `do_kvalue` tests, since these failed on my laptop.  As far as I can tell, the problem boils down to having an extreme maxk/stepk ratio for a vonKarman that includes a delta function component.  (The maxk in particular become infinite).  Since the tests pass (after some minor tweaks to `gsparams.maximum_fft_size` and `do_shoot`) when `do_delta=False` though, I'm satisfied things are okay, and opted to just skip these when `do_delta=True` (the c++ part of the profile is the same in either case).

These are pretty slow, about 13 minutes.  Removing the `args.slow` is your call; I'm happy either way.

tests/test_vonkarman.py
jmeyers314: Sure.  I added comments inline with the print statements showing the speeds on my laptop:
~0.7s for the first vk, ~0.07s for the next 10, and ~0.9s for drawing 100x50,000 photons.

src/SBVonKarman.cpp
```diff
@@ -0,0 +1,395 @@
...
+            }
+        }
+        dbg<<"Finished building radial function.\n";
+        dbg<<"R = "<<R<<" arcsec\n";
+        dbg<<"HLR = "<<_hlr<<" arcsec\n";
+        R = std::max(R, _gsparams->stepk_minimum_hlr*_hlr);
+        _stepk = M_PI / R;
+        dbg<<"stepk = "<<_stepk<<" arcsec^-1\n";
+        dbg<<"sum = "<<sum<<"   (should be ~= 0.997)\n";
+        if (sum < 0.997)
```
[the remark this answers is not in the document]
jmeyers314: I think this was mostly for testing when I was having a hard time picking values of `dr` and `dlogr` that would end up satisfying the `1-folding_threshold` constraint with a reasonable amount of compute effort.  I went ahead and changed the test to use `folding_threshold` explicitly, and added a unit test for constructing with some non-standard values.

tests/test_vonkarman.py
rmjarvis: I picked up your gauntlet and worked on speeding up the code some.  

Now running test_vonkarman via nosetests takes 6.6 seconds for the whole thing (down from 91 seconds).  

And running directly with slow=True takes 330 seconds (down from 1900 sec for me).  I think that's fast enough now to include by default for command-line runs, so I switched it.

Most of the rest of the slow parts are calling kValue (via drawKImage).  I feel like there's probably something we could do with the large k asymptotic form for BesselK, but I didn't try to work out the algebra with that to see if there is some speed up possible there.  Maybe not, since there is still an exp inside the exp, which doesn't sound like it would simplify nicely, although it's possible.

jmeyers314: Thanks Mike, this is really nice!  Do you recall what the biggest improvement was?

rmjarvis:
For real space, definitely using _radial for the xValue.  Also, the wider dlogr, which I think also speeds up the photon shooting runs.

galsim/vonkarman.py
```diff
@@ -0,0 +1,183 @@
+# Copyright (c) 2012-2017 by the GalSim developers team on GitHub
```
rmandelb: Are we going to do a 2017->2018 switch at some point?  (Preferably before 2019, but doesn't have to be in this PR.  Open new issue reminding Mike to use his magic script?)

galsim/vonkarman.py
```diff
@@ -0,0 +1,183 @@
...
+        Martinez et al.  2010  A&A  vol. 516
+        Conan  2008  JOSA  vol. 25
+
+    Notes
+    -----
+
+    If one blindly follows the math for converting the von Karman power spectrum into a PSF, one
+    finds that the PSF contains a delta-function at the origin with fractional flux of
+
+        exp(-0.5*0.172*(r0/L0)^(-5/3))
```
rmandelb: We haven't defined r0 yet.  As you did below the first equation in the docstring, should define any new quantities (in this case only r0).

galsim/vonkarman.py
```diff
@@ -0,0 +1,183 @@
...
+
+    In almost all cases of interest this evaluates to something tiny, often on the order of 10^-100
+    or smaller.  By default, GalSim will ignore this delta function entirely.  If for some reason
+    you want to keep the delta function, though, then you can pass the do_delta=True argument to the
+    VonKarman initializer.
+
+    @param lam               Wavelength in nanometers
+    @param r0                Fried parameter in meters.
+    @param L0                Outer scale in meters.  [default: np.inf]
```
rmandelb: So our default makes von Karman equivalent to Kolmogorov?  Hmmm.  I guess choosing some random non-infinite value between 10-100 meters seems too arbitrary as a default?

galsim/vonkarman.py
```diff
@@ -0,0 +1,183 @@
...
+    @param flux              The flux (in photons/cm^2/s) of the profile. [default: 1]
+    @param scale_unit        Units assumed when drawing this profile or evaluating xValue, kValue,
+                             etc.  Should be a galsim.AngleUnit or a string that can be used to
+                             construct one (e.g., 'arcsec', 'radians', etc.).
+                             [default: galsim.arcsec]
+    @param do_delta          Include delta-function at origin? (not recommended; see above).
```
rmandelb: Not recommended just because it's usually so small?  Or does it cause some numerical problem?  If the latter, should say this (because so far you've only mentioned the smallness issue).

galsim/vonkarman.py
jmeyers314: #944

galsim/vonkarman.py
jmeyers314: Good point.  I extended the description above to include `r0`.

galsim/vonkarman.py
jmeyers314: Yeah, I'm not sure what the best value is here.  I did go ahead and change this to 25.0 m though, which makes it consistent with the default value in `AtmosphericScreen`.

galsim/vonkarman.py
jmeyers314: There are some numerical problems which led, for example, to restricting some of the unit tests to only operate on profiles without the delta function.  I modified some of the preceding text to say: 

```
    By default, GalSim will ignore this delta function entirely since it usually
    doesn't make any difference, but can complicate some calculations like drawing using
    method='real_space' or by formally requiring huge Fourier transforms for drawing using 
    method='fft'.
```

jmeyers314:
Thanks @rmjarvis and @rmandelb .  Assuming that travis et al. still pass and I don't hear any further feedback, I'll plan on merging this tomorrow.

PR #941, opened 2018-02-08
Euclid fork created with no future and yaml dependencies

mmarseille:
Hi,
I'm Matthieu Marseille, I just entered the Euclid System and Common Tools support team and currently working on GalSim 1.5 integration into Euclid.
I created a fork of GalSim 1.5.1 two days ago and worked in this sense.
The result is
- no more dependency to future
- no dependency to python YAML (we support only JSON from native python 3.6 package)

Building and tests are OK (only those including YAML configuration skipped).

I still have some work to to in this fork, but I crated this pull request so you can validate a potential merge of it in the future.

Cheers, Matthieu

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ tmp*
 examples_bin
 *~
 .cache
+.idea
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -317,3 +317,13 @@ Bug Fix
 -------
 
 - Fixed a build problem for Anaconda on Macs using Python 3.6. (#924)
+
+Fork for Euclid (M. Marseille) from 1.5.1 version
+=================================================
+
+New Feature
+-----------
+
+- Dependency from future removed
+
+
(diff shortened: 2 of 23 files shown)


rmjarvis:
I'm afraid removing yaml support is kind of a non-starter for ever merging back into the main GalSim code base.  It's already only required if you actually use the feature, so you should be able install and use GalSim without yaml.  You just can't use any of the functionality that uses YAML files (of course).

As for future, it's not necessarily a non-starter if you can make a good case for why to prefer six over future.  However, I don't see any advantage to using six, and I do see some (mild) advantages in style to the future way of doing things over six's style.  So this seems like a fairly gratuitous change to me.

So you're welcome to fork and maintain your repo separate from the mainline GalSim repo.  But we're not going to merge it back in with these changes.

mmarseille:
I understand your point of view.
The main point here is : we do not have the choice on Euclid. The python-future and PyYAML packages are not part of our current referential (which is the default Python 3.6 distribution deployed by centos7 repos).
At least, you know that such a fork exists and why.
Cheers.

Issue #942, opened 2018-02-09
Fix memory leak in PhaseScreenPSF


jmeyers314:
When drawing `PhaseScreenPSF`s using geometric photon-shooting, I think I neglected to remove the PSFs from the `PhaseScreenList._pending` list.  This means that even when those PSFs otherwise fall out of scope or are deleted, there's still a reference to them.  I think this can be cleaned up by simply deleting a PSF from the PSF._pending list once it has been photon-shot.

PR #943, opened 2018-02-10
Fix memory leak in PhaseScreenPSF

jmeyers314:
This PR fixes the memory leak described in #942.

I decided my original plan to just delete PSFs from `_pending` after drawing them using either `method='phot'` or `method='fft'` doesn't actually work, since it ought be possible to do both in sequence without losing the speed advantage offered via `_pending`.

My new idea is to instead use the`weakref` module.  This lets a `PhaseScreenList` still refer to any dependent `PhaseScreenPSF`s in the `_pending` list, but won't keep them alive if those weak references are the only references to the PSFs, thus eliminating the memory leak.

### DIFF
--- a/galsim/phase_psf.py
+++ b/galsim/phase_psf.py
@@ -69,6 +69,7 @@ from past.builtins import basestring
 from itertools import chain
 from builtins import range
 from heapq import heappush, heappop
+import weakref
 
 import numpy as np
 import galsim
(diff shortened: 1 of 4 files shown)


galsim/phase_psf.py
```diff
@@ -755,15 +758,17 @@ def _prepareDraw(self):
...
-            # If that PSF's next possible update time doesn't extend past its exptime, then
-            # push it back on the heap.
-            t += psf.time_step
-            if t < psf.t0 + psf.exptime:
-                heappush(self._update_time_heap, (t, i))
-            else:
-                psf._finalize()
+            psfref = self._pending[i]
+            psf = psfref()
+            if psf is not None:
```
rmjarvis: I'm sure this is right, but why is this check required (both here and above)?  What does it mean for a weak_ref to return None?  Any why is it ok to not do anything in that case?

jmeyers314: Thanks Mike.  The `weakref` returns None if the referent no longer exists (was deleted, went out of scope, ...).  I was following the pattern described [here](https://docs.python.org/3/library/weakref.html#weak-reference-objects).

rmjarvis: I guess we must be getting some None's, since codecov thought you hit both branches.  Is that related to this commit: ed660d1?  So if some got deleted, but others are still around, then we want to skip them in this bit of code?

jmeyers314: Yep.  I added that commit b/c of codecov.

We do want to skip any deleted PSFs, since a) there's no supported API for the user to access them at this point (and the user has deleted their own handles to the objects), and b) the referents (`None`s) don't have the required `_step` or `_finalize` methods.

Looking at this piece of code again though, I realized two things: 1) There's no reason to advance the time of the phase screens for a PSF we don't draw.  So I moved that bit of code down into the if statement.  2) I don't really need both `_pending` and `_update_time_heap`, since the latter is just an index into the former.  So a rewrote this bit to use only the `_pending` attribute, which acts like the former `_update_time_heap` except that it stores the weakref to the PSF directly, instead of through an index into another list.

Issue #945, opened 2018-02-21
Running galsim demo11.yaml  gives │NameError: global name 'origpf' is not defined


gijzelaerr:
```
Traceback (most recent call last):
  File "/usr/bin/galsim", line 275, in <module>
    main()
  File "/usr/bin/galsim", line 256, in main
    except_abort=args.except_abort)
  File "/usr/lib/python2.7/dist-packages/galsim/config/process.py", line 780, in Process
    except_abort=except_abort)
  File "/usr/lib/python2.7/dist-packages/galsim/config/output.py", line 58, in BuildFiles
    galsim.config.ProcessInput(config, logger=logger, safe_only=True)
  File "/usr/lib/python2.7/dist-packages/galsim/config/input.py", line 175, in ProcessInput
    input_obj = getattr(config['input_manager'],tag)(**kwargs)
  File "/usr/lib/python2.7/multiprocessing/managers.py", line 667, in temp
    token, exp = self._create(typeid, *args, **kwds)
  File "/usr/lib/python2.7/multiprocessing/managers.py", line 567, in _create
    id, exposed = dispatch(conn, None, 'create', (typeid,)+args, kwds)
  File "/usr/lib/python2.7/multiprocessing/managers.py", line 105, in dispatch
    raise convert_to_error(kind, result)
multiprocessing.managers.RemoteError: 
---------------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python2.7/multiprocessing/managers.py", line 207, in handle_request
    result = func(c, *args, **kwds)
  File "/usr/lib/python2.7/multiprocessing/managers.py", line 386, in create
    obj = callable(*args, **kwds)
  File "/usr/lib/python2.7/dist-packages/galsim/lensing_ps.py", line 198, in __init__
    self._convert_power_function(self.e_power_function,'e_power_function')
  File "/usr/lib/python2.7/dist-packages/galsim/lensing_ps.py", line 675, in _convert_power_function
    "Input provided: {0}\n".format(origpf)+
NameError: global name 'origpf' is not defined
---------------------------------------------------------------------------
```

gijzelaerr:
looks like one of the paths in `demo1.yaml` are wrong in my case, but this error is a bit confusing

rmjarvis:
[...]

Issue #855, opened 2017-01-02
Improve test suite coverage of lensing modules


rmjarvis:
Our three [worst-covered](https://codecov.io/gh/GalSim-developers/GalSim/tree/20142beef4301bfd37f1cd035b7f9c16d1d3c3f4/galsim) (non-wfirst) files are currently (in order) pse.py, lensing_ps.py, and nfw_halo.py.

This issue is to address that by
1. adding more tests to cover functionality that is currently insufficiently tested.
2. removing some functionality that we probably don't need.  I'm thinking especially of the 4 different return values for many functions in lensing_ps.py.  Probably 2 would suffice -- a single value or a numpy array probably covers all the relevant use cases.
3. adding `# pragma: no cover` for some branches that might not actually need coverage.

I think all three files are related enough that they can be done as a joint effort, but anyone who takes this on can feel free to only do one or two of the files and leave the rest for someone else.

rmjarvis:
Done via PR #946

PR #946, opened 2018-02-22
#855 Improve test coverage of lensing module

rmjarvis:
This PR is mostly just additional tests to improve the test coverage of lensing_ps.py, nfw_halo.py and pse.py.  In a couple cases I marked stuff as `pragma: no cover` rather than write the tests for functions that are never really used within GalSim.  And in a couple other cases where we never use options, I went ahead and deprecated them.

Deprecated:
* passing `Image` arguments to `kappaKaiserSquires` function
* `interpolant` argument for PowerSpectrum methods `getShear`, `getConvergence`, `getMagnification`, and `getLensing`.  

Explicitly uncovered:
* `calculateXi` (There is some coverage, but not nearly for all the possible parameter options, and I didn't want to bother with it.  This was relevant during Great3, but I'm not sure how useful it is anymore.  I wouldn't mind deprecating this function.)
* `subsampleGrid`  (Never used by anything.  Again, I wouldn't mind deprecating it.)

API change:
* The return values from `getShear`, etc. for both `NFWHalo` and `PowerSpectrum` used to try to match the input type being a tuple, list, or numpy array.  Now if it gets any kind of list-like input, it just returns a numpy array.  Python duck typing should suffice for essentially all uses here, so I think this will be transparent to all users, but technically it is an API change.

The other item that is now covered, but might nonetheless merit deprecation is the `weight_*` options for `PowerSpectrumEstimator`.  We never use them (or `theory_func`), and when I wrote unit tests for them, the accuracy is significantly worse than not weighting.  If someone can make a case for why they are useful, we should keep them, but if not, I'm inclined to deprecate them.  @joezuntz may have something to say on this, since he was the original author of pse.py.

BTW, I actually started this over a year ago, then never got back to it, and ended up forgetting I had even started it.  But since @gijzelaerr ran into a bug in one of our uncovered error messages, I figured it was worth doing, so I finished this up.  This PR would resolve both #855 and #945.

### DIFF
--- a/galsim/lensing_ps.py
+++ b/galsim/lensing_ps.py
@@ -182,7 +182,7 @@ class PowerSpectrum(object):
         if units == galsim.arcsec:
             self.scale = 1
         else:
-            self.scale = 1. * units / galsim.arcsec
+            self.scale = units / galsim.arcsec
 
     def __repr__(self):
         s = 'galsim.PowerSpectrum(e_power_function=%r'%self.e_power_function
--- a/galsim/lensing_ps.py
+++ b/galsim/lensing_ps.py
@@ -33,43 +33,17 @@ def theoryToObserved(gamma1, gamma2, kappa):
 
     @param gamma1       The first shear component, which must be the NON-reduced shear.  This and
                         all other inputs should be supplied either as individual floating point
-                        numbers, tuples, lists, or NumPy arrays.
+                        numbers or NumPy arrays.
     @param gamma2       The second (x) shear component, which must be the NON-reduced shear.
     @param kappa        The convergence.
 
     @returns the reduced shear and magnification as a tuple `(g1, g2, mu)` where each item has the
              same form as the input gamma1, gamma2, and kappa.
     """
-    # check nature of inputs to make sure they are appropriate
-    if type(gamma1) != type(gamma2):
-        raise ValueError("Input shear components must be of the same type!")
-    if type(kappa) != type(gamma1):
-        raise ValueError("Input shear and convergence must be of the same type!")
-    gamma1_tmp = np.array(gamma1)
-    gamma2_tmp = np.array(gamma2)
-    kappa_tmp = np.array(kappa)
-    if gamma1_tmp.shape != gamma2_tmp.shape:
-        raise ValueError("Shear arrays passed to theoryToObserved() do not have the same shape!")
-    if kappa_tmp.shape != gamma1_tmp.shape:
-        raise ValueError(
-           "Convergence and shear arrays passed to theoryToObserved() do not have the same shape!")
-
-    # Now convert to reduced shear and magnification
-    g1 = gamma1_tmp/(1.-kappa_tmp)
-    g2 = gamma2_tmp/(1.-kappa_tmp)
-    mu = 1./((1.-kappa_tmp)**2 - (gamma1_tmp**2 + gamma2_tmp**2))
-
-    # Put back into same format as inputs
-    if isinstance(gamma1, float):
-        return float(g1), float(g2), float(mu)
-    elif isinstance(gamma1, list):
-        return list(g1), list(g2), list(mu)
-    elif isinstance(gamma1, tuple):
-        return tuple(g1), tuple(g2), tuple(mu)
-    elif isinstance(gamma1, np.ndarray):
-        return g1, g2, mu
-    else:
-        raise ValueError("Unknown input type for shears, convergences: %s",type(gamma1))
+    g1 = gamma1/(1.-kappa)
+    g2 = gamma2/(1.-kappa)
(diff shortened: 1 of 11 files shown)


galsim/lensing_ps.py
```diff
@@ -33,43 +33,17 @@ def theoryToObserved(gamma1, gamma2, kappa):
...
-        return float(g1), float(g2), float(mu)
-    elif isinstance(gamma1, list):
-        return list(g1), list(g2), list(mu)
-    elif isinstance(gamma1, tuple):
-        return tuple(g1), tuple(g2), tuple(mu)
-    elif isinstance(gamma1, np.ndarray):
-        return g1, g2, mu
-    else:
-        raise ValueError("Unknown input type for shears, convergences: %s",type(gamma1))
+    g1 = gamma1/(1.-kappa)
```
jmeyers314: Maybe leave open the possibility of passing in a python list (similar to LookupTable e.g.)?  I.e.
```python
gamma1 = np.array(gamma1, dtype=float)
```
Simplifying the output is definitely a good idea.

galsim/nfw_halo.py
```diff
@@ -195,91 +184,66 @@ def __omega(self, a):
...
-        if out is None:
-            out = np.zeros_like(x)
+        out = np.zeros_like(x, dtype=float)
 
         # 3 cases: x > 1, x < 1, and |x-1| < 0.001
-        mask = (x < 0.999)
-        if mask.any():
-            a = ((1.-x[mask])/(x[mask]+1.))**0.5
-            out[mask] = 0.5*np.log((1.+a)/(1.-a))/(1-x[mask]**2)**0.5
+        mask = np.where(x < 0.999)[0]
```
jmeyers314: Why `np.where` (Here and below)?  I think the original is easier to read.  (Don't have to remember why there's a `[0]`).  Getting rid of the `mask.any()` is a definite improvement though.

tests/test_lensing.py
```diff
@@ -617,6 +707,18 @@ def test_shear_units():
...
+    ps = galsim.PowerSpectrum('k', 'k**2', False, 'arcsec')
+    assert ps == galsim.PowerSpectrum(e_power_function='k', b_power_function='k**2')
+    assert ps == galsim.PowerSpectrum(e_power_function='k', b_power_function='k**2',
+                                      delta2=False, units=galsim.arcsec)
+    for ps2 in [ galsim.PowerSpectrum('k**2', 'k**2', False, 'arcsec'),
+                 galsim.PowerSpectrum('k', 'k', False, 'arcsec'),
+                 galsim.PowerSpectrum('k', 'k**2', True, 'arcsec'),
+                 galsim.PowerSpectrum('k', 'k**2', False, 'arcmin'),
+               ]:
+        assert ps != ps2
```
jmeyers314: `all_obj_diff` (here and for `PowerSpectrumRealizer`)?

galsim/nfw_halo.py
rmjarvis: The reason for this is efficiency.  Indexing with an array of bools tends to be slower than indexing with an array of integers saying where the bools were True.  Especially if there are far fewer True than False, which I suppose will be the case for typically 2 out of these 3 cases.

rmandelb:
Hi @rmjarvis - I would indeed like to take a look.  Will do so this weekend.  Thanks!

rmandelb:
Before I look at the code, a few comments:

- I feel like passing different interpolants to `getLensing` (etc.) is a useful thing to be able to do, at least to be able to check that you aren't that sensitive to the choice of interpolant (because if you are, that usually means you've set up the grid in a way that is suboptimal for the problem of choice).

- I agree about deprecating `subsampleGrid` (which you haven't done yet but I thought I would comment since you mentioned it as an option).  That was mostly for testing when we were putting the basic functionality in place.

- I personally do use `calculateXi` still.  It's useful to be able to predict how much suppression of shear power you'll have due to your finite grid size.  We could deprecate some arguments while keeping the basic functionality?

rmandelb:
I looked at the code and didn't really have anything to add - the new tests and other changes look good.  I edited my above comment about planned changes, so please read on GitHub rather than by e-mail.

rmjarvis:
Thanks Rachel.  I deprecated subsampleGrid.  

And since you say you use calculateXi, I went ahead and added coverage for it.  Technically, I don't test all the possible input parameters, but I refactored the code slightly, so it now shares the implementation for handling delta2, scale, and bandlimit with buildGrid.  So I think that's probably sufficient coverage.

Also, there were indeed bugs with how calculateXi handled B-modes, which the tests uncovered, and which I fixed.

rmandelb:
Thanks Mike.  Indeed our previous calculateXi tests were done only with E-mode power spectra.

PR #951, opened 2018-03-03
Factor out Zernike code into submodule.

jmeyers314:
This PR is to factor out the Zernike polynomial code that had been embedded inside the `OpticalScreen` class into a separate submodule.  I think this is an improvement both because it increases the clarity of `OpticalScreen` by hiding implementation details, and because it allows other code to more easily use the Zernike polynomial code.  (We don't currently have any of the latter inside galsim, but I do use this already outside of GalSim for modeling quantities that vary across the focal plane, and I could envision something like this being used for a double Zernike PSF model in galsim in the future).

I also added a couple of useful functions to handle transforming Zernike coefficients under coordinate system rotations, and conveniently constructing Zernike basis vectors (which is useful for fitting a Zernike series to data, for example).

Note, I had thought about completely factoring this code out of GalSim and into a separate package (similar to `LSSTDESC.Coord`, say), but decided it would be easier to just make it into a GalSim submodule.  If anyone thinks it'd be useful to be able to manipulate Zernikes without depending on GalSim, however, please say so.


galsim/zernike.py
```diff
@@ -0,0 +1,398 @@
...
+    with outer radius not equal to 1.  Taken together, these generalizations yield the
+    orthonormality condition:
+
+        \int_{annulus} Z_i Z_j dA = \pi (R_outer^2 - R_inner^2) \delta_{i, j}
+
+    where 0 < R_inner < R_outer indicate the inner and outer radii of the annulus over which the
+    polynomials are orthonormal.
+
+    The indexing convention for i and j is that from Noll, J. Opt. Soc. Am. 66, 207-2011(1976),
+    though offset by 1 because python is 0-indexed while the Noll indices start at 1.  For example,
```
rmjarvis: This seems potentially confusing and error-prone.  People think of Z4 as defocus, so it seems like it would be easy for people to forget to subtract 1 when using this class.  (Or when setting up their array to pass to Zernike.) 

Can we just hide the 1->0 offset internally and make the public API use the real Noll indices?  (Like we did for the `aberrations` parameter of OpticalPSF.)

galsim/zernike.py
```diff
@@ -0,0 +1,398 @@
...
+             1      |       2      |      2 x
+             2      |       3      |      2 y
+             3      |       4      | sqrt(3) (2 (x^2 + y^2) - 1)
+
+
+    @param coef   Zernike series coefficients.  Note that coef[i] corresponds to Z_(i+1) using the
+                  Noll index convention.  (I.e., coef[0] is 'piston', coef[3] is 'defocus', ...)
+    @param eps    Linear fractional obscuration for defining annular Zernike polynomials.
+                  [default: 0.0]
+    @param diam   Diameter over which polynomials are orthonormal.  [Default: 2.0 (for unit circle)]
```
rmjarvis: From the above docstring, I expected the parameters to be `R_inner` and `R_outer`, rather than `eps` and `diam`.  `eps` in particular was not intuitive to me without reading the param description.  

I'd suggest switching the parameters to those, which I thought were pretty clear in your description.

galsim/zernike.py
```diff
@@ -0,0 +1,398 @@
...
+def _nCr(n, r):
+    if 0 <= r <= n:
+        ntok = 1
+        rtok = 1
+        for t in range(1, min(r, n - r) + 1):
+            ntok *= n
+            rtok *= t
+            n -= 1
+        return ntok // rtok
+    else:
```
rmjarvis: This branch isn't covered.  Seems like an easy test to add.  Also, maybe consider moving this function to utilities an remove the underscore?

galsim/zernike.py
```diff
@@ -0,0 +1,398 @@
...
+    @param n            Zernike radial coefficient
+    @param m            Zernike azimuthal coefficient
+    @param obscuration  Linear obscuration fraction.
+    @param shape        Output array shape
+
+    @returns    2D array of coefficients in |r|^2 and r, where r = u + 1j * v, and u, v are unit
+                disk coordinates.
+    """
+    if shape is None:
+        shape = ((n//2)+1, abs(m)+1)
```
rmjarvis: This is also not covered.  Probably could just remove this.

galsim/zernike.py
```diff
@@ -0,0 +1,398 @@
...
+        ni, mi = _noll_to_zern(i+1)
+        for j in range(max(0, i-1), min(i+2, jmax)):
+            nj, mj = _noll_to_zern(j+1)
+            if ni != nj:
+                continue
+            if abs(mi) != abs(mj):
+                continue
+            if mi == mj:
+                R[i, j] = np.cos(mj * theta)
+            elif mi == -mj:
```
rmjarvis: This can just be else.  The `abs(mi) != abs(mj)` above means that this is the only option left.

galsim/zernike.py
jmeyers314: Done.  I also did some timing testing on this one.  Turns out that just writing:
```python
def nCr(n, r):
    from math import factorial
    if 0 <= r <= n:
        return factorial(n) // (factorial(r) * factorial(n-r))
    else:
        return 0
```
is usually faster than the more complicated thing that was there before, so I replaced it with the above.  I think the exceptions are when n >> r and n >> 1, in which case the complicated math is faster.  I don't think this should hardly ever occur in GalSim though -- certainly not for Zernike coefficients anyway, which is the only place we currently use this.

rmjarvis: That feels wrong to me, but I guess it's ok in python, since they have the unlimited size long type.  20! is the largest value that can be stored in an int (C long).  Above that, python switches to its extended precision long.
```
>>> math.factorial(20)
2432902008176640000
>>> math.factorial(21)
51090942171709440000L
>>> type(math.factorial(21))
<type 'long'>
>>> type(math.factorial(20))
<type 'int'>
```
So in C, n>20 would fail with the naive code.  But I would suspect that this is also the transition to where in Python, the naive code gets slower.  Maybe preserve the old code in a branch `if n > 20`?  Or do you think this is not particularly helpful.  (Premature optimization and all that.)

jmeyers314: This was fun to look into a little further.  Here's a plot showing the log time ratio for the complex (for lack of a better word) algorithm vs the naive algorithm up to n=100.  Reds are where the naive algorithm is faster, blue/greens where the complex algorithm is faster.  

unknown

There's definitely a transition at n=20, but that's not the dominant effect.  My guess is that the python loop in the complex algorithm (O(min(r, n-r)) is eating a lot of time, while the naive algorithm is O(1) in python once you've memoized the factorials (there are surely some C loops in the arbitrary precision arithmetic though).

So while we could potentially switch algorithms using something like `if n > 20 and min(r, n-r) < n//8`, I don't think it's actually worth doing, given how unlikely this is to be a tall pole anywhere.

jmeyers314: ... looks like the region where the naive algorithm wins does get narrower as n increases.  Here's the same plot up to n=1000, but only doing every 25th n and r:

unknown-1

rmjarvis: Succumbing to my premature optimization weakness, I looked into this a bit.  The naive version does seem to get significantly (~30%) slower when n crosses from 20 to 21.  It depends a bit on the value of r whether it becomes slower than the old code or not.

If n >> 20 (like 100 or so), and r is relatively small, then the difference can be quite large in favor of the old code.  I guess that's probably never a use case we'll care about, but a potential concern nonetheless.

Adding an extra if statement does slow things down a bit, especially for the typically small n,r values we use.  But if I tweak the old code a bit to make it more efficient, then I think this is almost always faster than the version with factorial.
```
def nCr(n, r):
    if 0 <= r <= n:
        r = min(r, n-r)
        numer = n
        denom = r
        while r > 1:
            n -= 1
            r -= 1
            numer *= n
            denom *= r
        return numer // denom
    else:
        return 0
```
At least on my laptop, the factorial version was never more than about 10% faster, and this version could be up to a factor of 2 faster (e.g. for r=1 cases) for small n and many times faster for large n (n>30 or so).

rmjarvis: Weird.  Your posts didn't show up until I hit reload.  Anyway, good to see we both have the same weakness.  :)  Curious what your plot looks like with my modified version.

jmeyers314: Lol, yep, I'm already on it.  If the fate of the world ever depends on distracting Josh and Mike, now they know how to do it :).

rmandelb: I am enjoying this conversation from the outside.  :)

jmeyers314: Okay, new plots.  Main caveat to following is this is only on my laptop.

Here's the time for each (n, r) up to n=100.

unknown

So pretty flat for the naive algorithm, and a pretty significant dependence on n and r for the speedy algorithms.

Same thing as ratios:

unknown-1

So the new speedy algorithm is indeed speedier for min(r, n-r) << n/2, but slower nearer the center.

Zooming out:

unknown-2

Naive depends on n, but not so much on r.  Both speedies scale with both r and n.

unknown-3

New and old speedy are about the same in the r, n ~= 100s range.  It's also clear here that, as Mike said, the potential relative speed gains are much larger than the potential speed losses comparing speedy vs naive.

That said, the places where speedy is much speedier are speedy already.  I.e., the fractional improvements are large, but the absolute improvements not so much.  If I just add up all the times for all the (n, r) combinations for each algorithm, I get:

n<100
-------
naive: 10.26
speedy_old: 17.06
speedy_new: 19.16

n < 1000
---------
naive: 38.84
speedy_old: 36.99
speedy_new: 37.01

So assuming a uniform prior over (n, r) with n<1000 (reasonable?), it really doesn't make much difference to the expected value of calculation time.  Over n<100, then naive clearly wins.

rmjarvis: Very interesting.  I wonder if math.factorial got faster in python 3.  My results were using python 2.7, and your plots are red in places where the new code definitely won on my laptop.  

E.g. nCr(10,9) on my laptop is more than a factor of 2 faster with the new algorithm, but it shows up as pink in your t_speedy_new/t_naive plot.  And there were only a few combinations I could find with n < 20 where the naive algorithm was faster.  (Mostly where n ~ 2r as you would expect.)

Anyway, fine to leave it with the factorial version.  Python 3 is the future anyway, so might as well optimize for that to the extent we should bother optimizing this at all.

jmeyers314: >I wonder if math.factorial got faster in python 3.

Yep, this looks like it is the case.  Here's the first plot again, but in py2.7

unknown-4

So a strong dependence on n instead of being mostly flat (for the naive algorithm).

>Python 3 is the future anyway, so might as well optimize for that to the extent we should bother optimizing this at all.

Sounds good to me.

galsim/zernike.py
jmeyers314: This one gave me a little pause, especially now that there's output that needs to be conforming (e.g. zernikeBasis) as well as input.  I think I probably would have preferred the opposite convention if we were starting from scratch, but since there's already a well-established convention from OpticalPSF, and it's good to stay consistent, I went ahead and hid the 1->0 offset as you suggest.

jmeyers314:
Thanks for the review Mike; I think I've addressed all your comments now.  I'll plan to merge this on Friday unless I hear more.

Issue #952, opened 2018-03-06
Excessively long execution time for galsim.Kolmogorov


jchiang87:
I've encountered a set of parameters that cause `galsim.Kolmogorov` to run seemingly indefinitely.  I'm attaching a pickle file with those parameters.  Here's what I see:
```
Python 3.6.2 |Continuum Analytics, Inc.| (default, Jul 20 2017, 13:51:32) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pickle
>>> import galsim
>>> galsim.__version__
'1.5.1'
>>> pars = pickle.load(open('kolmogorov_pars.txt', 'rb'))
>>> pars
{'fwhm': 0.55344217545630736, 'gsparams': galsim.GSParams(128,4096,0.0008316873626901008,5.0,0.001,1e-05,1e-05,1.0,0.0001,1e-06,1e-06,1e-08,1e-05,0.81,32,0.0001)}
>>> galsim.Kolmogorov(**pars)
```
I get the same behavior on cori-haswell at NERSC and in a CentOS 7 docker container.

[kolmogorov_pars.txt](https://github.com/GalSim-developers/GalSim/files/1787008/kolmogorov_pars.txt)

rmjarvis:
Yep.  I can confirm that this gets into an endless loop when building the sampler.  Looking now into what is causing it.

rmjarvis:
Fixed in v1.5.2

jchiang87:
Thanks Mike!  @danielsf Can we get v1.5.2 into the next lsst_sims version please?

danielsf:
@jchiang87 sure thing

PR #953, opened 2018-03-06
#952 Fix error in findExtremum not preserving intended invariants

rmjarvis:
@jchiang87 reported in #952 a combination of input parameters that sent Kolmogorov into an endless loop.  Turns out the problem was in OneDimensionalDeviate.  Specifically the `findExtremum` call ended up having x1 increase without bound until it was inf (and then kept looping unchanged).

The underlying problem seems to be the fatRight case not being properly set up.  It set new values for x1,x2,x3 that didn't preserve the intended invariance that x1 < x2 < x3.  In fact, I'm not even sure how this code used to function properly (which apparently it usually did).

Anyway, I restructured the algorithm a bit and added (when DEBUGLOGGING is enabled) some assert statements to make sure the loop invariants are preserved properly.  It does seem to be working properly now.

TODO on merge:
* [ ] Release version 1.5.2 with this bug fix
* [ ] Cherry pick this commit to master
* [ ] Cherry pick this commit to noboost (new test will be in test_kolmogorov.py there)

### DIFF
--- a/src/OneDimensionalDeviate.cpp
+++ b/src/OneDimensionalDeviate.cpp
@@ -84,23 +84,24 @@ namespace galsim {
         xdbg<<"f("<<x2<<") = "<<f2<<" = "<<function(x2)<<std::endl;
         xdbg<<"f("<<x3<<") = "<<f3<<" = "<<function(x3)<<std::endl;
 
-        // First guess is that minimum is in half of bracket with lowest gradient
-        bool fatLeft = std::abs(df1) < std::abs(df2);
+        // Fat left tells which side is the fatter one.  Keep splitting the fatter side.
+        bool fatLeft = (x2-x1) > (x3-x2);
 
         // Then use golden sections to localize - could use Brent's method for speed.
+        // Based on Numerical Recipes 10.1.
         const double GOLDEN = 2./(1+sqrt(5.));
         while (std::abs(x3-x1) > xTolerance) {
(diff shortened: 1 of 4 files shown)


src/OneDimensionalDeviate.cpp
```diff
@@ -78,56 +79,77 @@ namespace galsim {
...
-                    x1 = xTrial;
-                    f1 = fTrial;
-                    df1 = dfTrial;
-                    fatLeft = false;
                 }
             } else {
+                xdbg<<"fat right\n";
                 // Split right-hand interval (2 / trial / 3)
-                double xTrial = x2 - GOLDEN*(x3-x2);
+                double xTrial = x3 - GOLDEN*(x3-x2);
```
rmjarvis: This was the principal bug in the old code.  I don't even understand how the code used to work with this being x2.  Must have gotten lucky somehow.

src/OneDimensionalDeviate.cpp
```diff
@@ -78,56 +79,77 @@ namespace galsim {
...
-                    fatLeft = false;
-                } else {
-                    // Extremum is in 1 / 2 / trial
-                    x3 = xTrial;
-                    f3 = fTrial;
-                    df2 = dfTrial;
-                    fatLeft = true;
                 }
             }
+            fatLeft = std::abs(df1) < std::abs(df2);
```
rmjarvis: This didn't used to be preserved in the old version.  But I think not preserving this could only cause a slight inefficiency, not cause the code to fail.

Issue #957, opened 2018-03-17
VonKarman profile doesn't like certain r0 values


cpadavis:
@areshernandez found the following bug while looking at the VonKarman profile. Certain values of r0 are bad. So, for example, 0.14606888419121025 and 0.146068885 give the below error, but 0.146068883 and 0.14606889 do not. This is a problem we've run into when using solvers with the VonKarman profile.

```python
RuntimeError                              Traceback (most recent call last)
<ipython-input-30-77c962a179fd> in <module>()
      1 import galsim
      2 #0.14606888419121025
----> 3 galsim.VonKarman(L0=25.,lam=700.,r0=0.146068884)

/nfs/slac/g/ki/ki19/des/aresh/Old_Stuff/Code_Depository/anaconda2/lib/python2.7/site-packages/galsim/vonkarman.pyc in __init__(self, lam, r0, L0, flux, scale_unit, do_delta, suppress_warning, gsparams)
     91         self._scale_unit = scale_unit
     92         scale = scale_unit/galsim.arcsec
---> 93         self._sbvk = _galsim.SBVonKarman(lam, r0, L0, flux, scale, do_delta, gsparams)
     94         self._delta_amplitude = self._sbvk.getDeltaAmplitude()
     95         self._suppress_warning = suppress_warning

RuntimeError: SB Error: Could not find folding_threshold
```

gsparams definitely has folding_threshold in it.

This was tested on the master branch.

rmandelb:
I reproduce this error on my system as well.  (BTW the error, which is not terribly clear, does not result from failure to find `folding_threshold` in gsparams -- but rather that a comparison against its value has failed.  We should make that error message more useful!)

I did not have time to look into this deeply, but when running with debuglogging I found that for the failing cases, there is a weird glitch in the radial profile that screws up the calculation of accumulated flux (and that same glitch does not occur for the cases that work).  So basically, it's very bad behavior in [this line of code](https://github.com/GalSim-developers/GalSim/blob/master/src/SBVonKarman.cpp#L281), which gets added to an accumulated sum in line 288.

@jmeyers314 , thoughts on this?

rmjarvis:
Fix is in PR #958.

rmjarvis:
PR #958 merged.

PR #958, opened 2018-03-18
Add explicit split points for the VK rawXValue integral

rmjarvis:
Fix the bug @areshernandez found for certain r0 values in initializing a VonKarman profile, reported in #957.

The resolution was to add explicit split points for the GKP integration at the zeros of the Bessel function.  This is a trick we've employed for both Moffat and Sersic with similar integrals that involve Bessel functions.  In general, forcing splits at the first few zeros of an oscillating function makes the integral much more stable in my experience.

### DIFF
--- a/src/SBVonKarman.cpp
+++ b/src/SBVonKarman.cpp
@@ -216,9 +216,19 @@ namespace galsim {
     // _radial lookup table is built, use that.
     double VonKarmanInfo::rawXValue(double r) const
     {
+        xdbg<<"rawXValue at r = "<<r<<std::endl;
         // r in arcsec
         VKXIntegrand I(r, *this);
         integ::IntRegion<double> reg(0, integ::MOCK_INF);
+        if (r > 0.) {
+            // Add explicit splits at first several roots of J0.
+            // This tends to make the integral more accurate.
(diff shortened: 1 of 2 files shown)


jmeyers314:
Looks good to me.  Thanks @cpadavis and @areshernandez for the report, and to @rmjarvis for the fix.

rmjarvis:
@jmeyers314 Do you mind turning that comment into an official review?  :)  Thanks.

src/SBVonKarman.cpp
```diff
@@ -303,7 +313,7 @@ namespace galsim {
         sum *= 2.*M_PI * dlogr;
         dbg<<"sum = "<<sum<<"   (should be > 0.995)\n";
         if (sum < 1-_gsparams->folding_threshold)
-            throw SBError("Could not find folding_threshold");
+            throw SBError("Could determine appropriate stepk, given folding_threshold");
```
rmandelb: Missing key word "not"

rmjarvis: Yep.  Thanks.

rmjarvis:
I didn't put this in the CHANGELOG, since VonKarman is new in 1.6, so this is just fixing up a new feature in this release.

Issue #832, opened 2016-11-11
Analytic Zernike gradients


jmeyers314:
During the DESC hack week, I implemented photon shooting for optical PSFs by computing finite difference gradients of optical phase screens.  

It would probably be faster and more accurate to use an analytic expression for the gradients of Zernike polynomials, however.  

[This paper](https://www.osapublishing.org/josaa/abstract.cfm?uri=josaa-31-4-708) shows how to express circular Zernike gradients as Zernike polynomials with different coefficients, though it may also be possible to compute the derivatives directly from OpticalScreen.coef_array, which would then apply to both circular and annular Zernikes.

cpadavis:
just so I have this right, because I think I have a version of doing this that I could add to galsim:

if the wavefront W(rho, phi) = sum_i a_i Z_i(rho, phi), you want:

dW/dx (rho, phi) = sum_i b^x_i Z_i(rho, phi) (and dW/dy, too)

where b_i = gamma^x_{ij} a_j (ie a matrix multiplication).

I also have code that takes a_i Z_i * a_j Z_j = sum_k b_k Z_k (ie zernike multiplication), which I have found is super useful if you need to compute integrals of products of zernikes.

jmeyers314:
Yeah, I think that looks right.  Does your code happen to apply to both circular and annular Zernikes, by chance?

cpadavis:
I have put very little thought into annular zernikes. All I really do with the code is implement noll's instructions for how to get the derivatives of the zernikes

jmeyers314:
I guess we can try it and see!

I still think it may be worth thinking about whether OpticalScreen.coef_array can be used directly (which may be faster than generating new Zernike coefficients(?), and should definitely work for both circular and annular Zernikes).  In that case though, your code would still be useful in a unit test.

rmjarvis:
Merge via PR #962

PR #962, opened 2018-04-12
Analytic Zernike Gradients #832

jmeyers314:
This PR 
1) Adds analytic Zernike gradients (replacing the finite difference gradients I had been using for optical screen geometric photon-shooting)
2) Changes the Zernike implementation to use coefficients in x and y coordinates instead of coefficients in (x + iy) and (x^2 + y^2) coordinates.

The first change increases speed by only requiring two calls to `horner2d` instead of three, and also eliminating subsequent array operations to create finite differences.

The second change increases speed since all of the work performed becomes part of the final result, compared to before where we would perform work using complex numbers and at the last step only keep the real parts.

Overall, I find a speed increase for evaluating Zernikes of about a factor of 2, and for evaluating Zernike gradients of about a factor of 4.

### DIFF
--- a/galsim/zernike.py
+++ b/galsim/zernike.py
@@ -464,7 +464,9 @@ class Zernike(object):
         return _coef_array_xy
 
     @lazy_property
-    def _gradX(self):
+    def gradX(self):
+        """The x-derivative of this Zernike as a new Zernike object.
+        """
         j = len(self.coef)-1
         newCoef = np.hstack([
             [0],
@@ -479,7 +481,9 @@ class Zernike(object):
         return Zernike(newCoef, R_outer=self.R_outer, R_inner=self.R_inner)
 
     @lazy_property
-    def _gradY(self):
+    def gradY(self):
+        """The y-derivative of this Zernike as a new Zernike object.
+        """
         j = len(self.coef)-1
         newCoef = np.hstack([
             [0],
@@ -514,21 +518,7 @@ class Zernike(object):
         return horner2d(x, y, self._coef_array_xy)
 
     def evalCartesianGrad(self, x, y):
-        return self.gradX().evalCartesian(x, y), self.gradY().evalCartesian(x, y)
-
-    def gradX(self):
-        """Compute the x-derivative of this Zernike.
-
-        @returns  d(Zernike)/dx as new Zernike object.
-        """
-        return self._gradX
-
-    def gradY(self):
-        """Compute the y-derivative of this Zernike.
-
-        @returns  d(Zernike)/dy as new Zernike object.
-        """
-        return self._gradY
+        return self.gradX.evalCartesian(x, y), self.gradY.evalCartesian(x, y)
 
     def rotate(self, theta):
         """Return new Zernike polynomial series rotated by angle theta.  For example:
--- a/galsim/zernike.py
+++ b/galsim/zernike.py
@@ -417,6 +417,20 @@ class Zernike(object):
         self.R_outer = float(R_outer)
         self.R_inner = float(R_inner)
 
+    # This exists in support of the deprecated OpticalPSF.coef_array attribute.  It can be deleted
+    # in version 2.0.
+    @lazy_property
+    def _coef_array(self):
+        _coef_array = np.dot(
+            _noll_coef_array(len(self.coef)-1, self.R_inner/self.R_outer),
+            self.coef[1:]
+        )
+
+        if self.R_outer != 1.0:
+            shape = _coef_array.shape
+            _coef_array /= self.R_outer**np.sum(np.mgrid[0:2*shape[0]:2, 0:shape[1]], axis=0)
+
+
     @lazy_property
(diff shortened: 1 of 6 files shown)


jmeyers314:
Actually...  I think I just figured out a better interface for this...  It's not too hard to figure out the coefficients of a new Zernike instance that corresponds to an x or y derivative.  That would enable easy nth derivatives and laplacians, so is probably worth doing.  I'll post again when I either get that working or give up on it; so probably hold off reviewing until then.

jmeyers314:
Okay, I think I got the return-a-new-Zernike working.  Assuming it passes travis, this is review-ready again.

galsim/zernike.py
```diff
@@ -239,17 +433,65 @@ class Zernike(object):
...
+
+    @lazy_property
+    def _gradY(self):
+        j = len(self.coef)-1
+        newCoef = np.hstack([
+            [0],
+            np.dot(
+                _noll_coef_array_grady(j, self.R_inner/self.R_outer),
+                self.coef[1:]
+            )
```
rmjarvis: I see a few of these `np.dot` calls.  I always found it clearer to write `A.dot(B)` rather than `np.dot(A,B)`.  Infix rather than prefix is I guess closer to the infix math notation `A * B`.  If you prefer to leave it this way, it's fine, but I just thought I'd put that out there.

tests/test_zernike.py
```diff
@@ -349,6 +348,63 @@ def test_fit():
...
+    Z11 = galsim.zernike.Zernike([0]*11+[1])
+
+    x = np.linspace(-1, 1, 100)
+    x, y = np.meshgrid(x, x)
+
+    def Z11_grad(x, y):
+        r2 = x**2 + y**2
+        gradx = 12*np.sqrt(5)*x*(2*r2-1)
+        grady = 12*np.sqrt(5)*y*(2*r2-1)
+        return gradx, grady
```
rmjarvis: This and the Z28 test below would benefit from some mathy comments.  E.g. write out the Noll definition of Z11, so we can read off that this is the right value of the derivative.

tests/test_zernike.py
```diff
@@ -349,6 +348,63 @@ def test_fit():
...
+        return ((Z.evalCartesian(x+dh, y)-Z.evalCartesian(x-dh, y))/(2*dh),
+                (Z.evalCartesian(x, y+dh)-Z.evalCartesian(x, y-dh))/(2*dh))
+
+    u = galsim.UniformDeviate(1234)
+
+    for j in range(25):
+        nj = 1+int(u()*55)
+        R_inner = 0.2+0.6*u()
+        R_outer = R_inner + 0.2+0.6*u()
+        Z = galsim.zernike.Zernike([0]+[u() for _ in range(nj)], R_inner=R_inner, R_outer=R_outer)
```
rmjarvis: This could also use some inline commenting.  I think I understand that all this is just to generate 25 random zernike arrays with random inner and outer radii, so it constitutes a fairly comprehensive test of the correctness of the derivative.  But it would be helpful to just explain that a bit in comments.

tests/test_zernike.py
```diff
@@ -149,10 +149,10 @@ def noll_to_zern(j):
         m = (-1)**j * ((n % 2) + 2 * int((j1+((n+1) % 2)) / 2.0))
         return (n, m)
 
-    # Test that the version of _noll_to_zern in phase_screens.py is accurate.
+    # Test that the version of noll_to_zern in phase_screens.py is accurate.
```
rmjarvis: phase_screens.py -> zernike.py

galsim/zernike.py
```diff
@@ -269,12 +509,26 @@ def evalPolar(self, rho, theta):
...
-        return horner2d(rsqr, r, self._coef_array, dtype=complex).real
+    def evalCartesianGrad(self, x, y):
+        return self.gradX().evalCartesian(x, y), self.gradY().evalCartesian(x, y)
+
+    def gradX(self):
+        """Compute the x-derivative of this Zernike.
+
+        @returns  d(Zernike)/dx as new Zernike object.
+        """
+        return self._gradX
```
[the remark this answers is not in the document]
jmeyers314: My thought was to indicate that computing the gradient -- at least the first time, before it's in the cache -- is a non-instantaneous calculation.  But I guess the cache would actually take over in almost any real world situation where the speed mattered (e.g., tight loop).  And it's actually pretty fast anyway.  Lazy_property it is.

galsim/zernike.py
jmeyers314: I guess I normally prefer `np.dot(A, B)` for its symmetry.  That said, in this case, since the dots are really matrix-vector multiplication, the operation isn't symmetric, so I went ahead and switched 'em.

jmeyers314:
Thanks Mike,  I think I hit all your comments.  I'll plan to merge this on Wednesday if I don't hear anything more.

PR #963, opened 2018-04-16
Various updates to speed up drawing sky photons with the Silicon sensor.

rmjarvis:
This update is prompted by @jchiang87 building sky images (especially flats, but even just sky background on normal images) with ImSim for LSST DESC DC2 simulations.

The main thing we realized is that most of the time was spent updating the pixel distortions at the start of the accumulate call, even though brighter-fatter is not really relevant for such low surface brightness levels.

The principal update in this PR is to add a `resume` keyword argument to the `accumulate` function, which indicates that the Silicon object should just continue where it left off at the end of a previous `accumulate` call and resume adding more photons to the same image.  This leads to a very significant speed up for building images of sky photons run through a Silicon layer.

Along the way, there were a few other minor optimization tweaks, plus a few bug fixes.  The bugs I discovered and fixed are as follows:
* In the C++ layer, `_nrecalc` was stored as an int, rather than a float.  This is a problem if `nrecalc` is set to a very large value (e.g. 1.e300 as I recommended Jim use), since it can wrap to negative numbers as an int value.  Fixed this to use double.
* Conversely `_qDist` is an int that was stored as a double.  It didn't cause big problems, but I fixed it nonetheless.
* The tree ring code had a `<` that should have been `<=`, which meant that the last row and column of the image did not have the tree ring distortions applied to it.
* Rarely, a photon falls into a dead zone between two pixels and isn't found to be in either one due to numerical inaccuracies.  There is special handling for that to randomly pick one of them.  However, it also did that handling when the photon falls off the edge of the image and is missed for that reason.  (This is much more important for sky images than images of normal objects.)  I fixed this to just let photons that fall off the outer edge of one of the edge photons disappear as they would have if the nominal pixel was already off the edge.
* The code to check if a photon is trivially in or out of a pixel assumed that the center of the pixel is close to (0.5,0.5).  This is normally true, even with B/F.  However, not with tree rings.  I fixed this to use the actual center of the pixel.

The upshot of this change is that drawing a single amp of size 2000 x 500 with skyCounts = 800 (cf. `dev/lsst/treering_skybg_check.py`) takes about 8 minutes on my laptop, down from something like an hour or so without resume.


rmjarvis:
Just in case anyone is thinking about it already, don't review this yet.  @cwwalter suggested another feature that would make sky images even faster, so I'm going to add that to this branch.

Specifically, we could have the Silicon class fill an image with the net area of each pixel.  Then, rather than rain down photons and decide which pixel each falls in, we can just take the areas of each pixel and use that times the surface brightness as the mean of a Poisson realization.

I think this idea will even work for high surface-brightness flats.  We can add this sky flux back into the Silicon object, get new areas that include B/F correlations, and repeat until we have the full sky level we want for the flat.  Should be loads faster than raining photons.

rmjarvis:
OK.  I think this is ready for review now.  I added a feature to create an image of all the pixel areas according to the Silicon model for a given flux image.  I also added a script for simulating a flat image this way, which is **WAY** faster than doing it with actual photons.  So I think this is probably what imSim is going to do for making flats for DC2.

tests/test_sensor.py
```diff
@@ -384,6 +385,86 @@ def test_silicon_fft():
     print('check |r1-r3| = %f >? %f'%(np.abs(r1-r3), 2.*sigma_r))
     assert r1-r3 > 2.*sigma_r
 
+@timer
+def test_silicon_area():
+    """Test the Silicon class calculate_pixel_areas() function.
```
craiglagegit: I ran a test where I ran a small (sigma = 0.02) spot with 80,000 electrons, which is what I used in the Poisson_CCD code to build the pixel distortion file, and then compared the areas between GalSim and Poisson_CCD.  They agree very well - to better than 0.01%.  I could add this to the unit test, but then I would have to add the area file to share/sensor, so I have this to compare to.  Let me know if you want me to do this.
 Let me know if you want me to do this.

rmjarvis: That sounds good.  Don't add that file to share/sensor though.  Rather, put it somewhere in the tests directory.  E.g. a new subdirectory tests/sensor_validation perhaps.  You'll see there are lots of other similar directories there that serve a similar purpose, validating against various kinds of external sources of truth.

rmjarvis:
That's a good point that we need to be careful with the tree rings when `transpose=True`.  The doc string for the `treering_center` parameter is currently:
```
    @param treering_center  A PositionD object with the center of the tree ring pattern in pixel
                            coordinates, which may be outside the pixel region. [default: None;
                            required if treering_func is provided]
```
which implies that it is in the coordinates of the actual image.  Not a transposed version.  So I think the current behavior is at least acting as documented.

So the question is really, is this the best API to have for this?  Should `treering_center` be in the transposed coordinates or the coordinates of the actual image being simulated?  

@jchiang87 Any thoughts on which way you thinks makes more sense?  Since you've actually used this feature in the wild.  Would it be more confusing to have `treering_center`'s (x,y) values refer to the transposed image (which I think is what is in the reference data)?  Or to have to swap (x,y) -> (y,x) when you pass it to Silicon along with `transpose=True`?

danielsf:
I'm not sure what I'm being asked to comment on.

The lsst_sims code that returns pixel coordinates for astrophysical objects currently returns those coordinates in the Camera coordinate system (in which the x axis is parallel to the serial readout direction/perpendicular to the serial register*).  This is accomplished by the `LSSTCameraWrapper` class defined here

https://github.com/lsst/sims_GalSimInterface/blob/master/python/lsst/sims/GalSimInterface/galSimCameraWrapper.py#L471

Does this answer the question at hand?
* I apologize if I got those backwards; I'm still acclimating myself to what all of those terms mean.  I do know that we are returning pixel coordinates in the Camera system, which is rotated 90 degrees from the DM system.

jchiang87:
The issue is that the brighter-fatter effect cares which direction the parallel and serial transfer directions are in that the deposited charge is redistributed more easily along the parallel transfer direction, and this is conventionally taken to be the y-direction.  However, the CCS has the parallel transfers along in the x-axis, and the sims code returns object positions and orientations in the CCS.  The galsim code treats the BF effect as if the parallel transfers are in the y-direction, so in order to get the BF anisotropy correct we either have to tell galsim to transpose the x and y axes for the BF calculations or have the sims code provide object positions and orientations in a transposed version of the CCS.

danielsf:
sims_GalSimInterface only returns pixel positions in the Camera coordinate system because that is the system that PhoSim eImages are in, and I thought we were trying to build a tool that exactly mimicked PhoSim eImages.  It would be easy (and, in some ways, much less convoluted) to have sims_GalSimInterface/ImSim return pixel positions in the DM coordinate system (with x parallel to the serial register).  I don't have an opinion on this question, but feel like it should get kicked up to the people who plan to deal with both PhoSim and ImSim images, since they might care about consistency between the two.

jchiang87:
Just so we understand the prior context:  Isn't it true that the dithering problem with the DC1 phosim deep data came about because the sims code wasn't using the CCS correctly back then?  And so phosim was actually doing the rotations correctly given the official Project conventions, but because the sims code wasn't, the inconsistency caused problems with those data.  My opinion is that we either get everyone to switch or we isolate the issue here.  I personally think it's completely clear what's going on since the default is to do what is conventional wrt parallel transfers along the y-axis, and the user has to explicitly ask for the transpose in the LSST simulation case.

danielsf:
The DC1 depth problem was due to an error in the handedness convention of rotSkyPos.  That is related, but ultimately independent of whether we use the Camera coordinate system or the DM coordinate system for precise pixel positions.  The DM and Camera coordinate systems both agree on the direction of the z axis.  The DC1 depth problem arose because I had reversed the z axis in the CatSim coordinate transformation code.  Now that this has been fixed, we *could* go to the DM coordinate system without recreating that error.

Put another way: the problem we had in DC1 is that the code which determined which chip a given RA, Dec fell on was wrong.  The question of whether to use the Camera coordinate system or the DM coordinate system impacts the detailed pixel position of a given RA, Dec within the chip.  Both coordinate systems agree on which chip an RA, Dec position falls on.  They just disagree on the name of the exact pixel it falls on.

That being said, it's hard to argue with "if it ain't broke, don't fix it."  You are probably right.

rmjarvis:
I think the conversation here has strayed a bit from the question at hand.  The question is what to do about the `treering_center` keyword when `transpose=True`.

I'm inclined to keep the behavior as is, since I think it matches the doc string, and I think the doc string is sensible.  I should add a bit of text to the doc string pointing out that the treering centers are in the pixel coordinates of the given image, even if `transpose=True`.  I.e. they are not also transposed.  Just the B/F effect.

Then it's the responsibility of the user to give the tree ring center in the coordinates of the image they actually give to Silicon, which seems reasonable.

Furthermore, ImSim will likely be the only user of this feature for the foreseeable future.  For now, the tree ring centers are random anyway, so the transpose is irrelevant.  And even if they ever become relevant (e.g. to match real lab-measured locations), they are implicitly in CCS coordinates in ImSim, which means that the current behavior would be the desired one anyway.

Did I summarize that correctly to everyone's satisfaction?

craiglagegit:
OK, I'll shut up about the X vs Y.  I pushed up a new version of test_sensor that compares the GalSim areas to the areas calculated from Poisson_CCD.  These area files are in tests/sensor_validation.  While I was at it, I also added the 8 vertex E2V files to share/sensors.

Issue #821, opened 2016-10-31
Add ability to draw from arbitrary distributions in RandomWalk profile


esheldon:
Currently we take the limit `nsep -> infinity` in order to draw from a Gaussian rather than actually doing a step-by-step random walk.  This is a very good approximation, even for low number of steps.

Consider extending this to allow drawing from other distributions, the most natural being other surface brightness profiles.

rmjarvis:
It occurs to me that you could probably do this for any photon-shootable profile by shooting npoints photons and use the (x,y) values as the locations of the knots.  This functionality is accessible in python now in branch #799.

So the code would look something like this:
```
ud = UniformDeviate(rng)
photons = knot_profile.SBProfile.shoot(npoints, ud)
self._points = np.array([ photons.getX(i), photons.getY(i) for i in range(npoints) ])
```
Currently, the PhotonArray class doesn't have accessors for the x,y arrays, but we could add that to improve the efficiency.  But that's the gist of the idea anyway.

esheldon:
That's pretty cool.

When #799 gets merged, lets revisit.  It should be easy to implement from there.

rmjarvis:
Done.  #973

PR #973, opened 2018-05-25
#821

esheldon:
Use a profile and photon shooting to generate the point sources in the RandomWalk

This addressed issue #821

### DIFF
--- a/galsim/randwalk.py
+++ b/galsim/randwalk.py
@@ -192,10 +192,16 @@ class RandomWalk(GSObject):
     @property
     def input_half_light_radius(self):
         """
-        The input half-light radius is not necessarily the realized hlr.
+        Get the input half light radius (HLR).
 
-        If a profile is sent, this will be -1 if it was a transformation
-        object
+        Note the input HLR is not necessarily the realized HLR,
+        due to the finite number of points used in the profile.
+
+        If a profile is sent, and that profile is a Transformation object (e.g.
+        it has been sheared, its flux set, etc), then this value will be None.
+
+        You can get the *calculated* half light radius using the calculateHLR
+        method.  That value will be valid in all cases.
         """
         return self._half_light_radius
 
--- a/galsim/randwalk.py
+++ b/galsim/randwalk.py
@@ -129,8 +129,12 @@ class RandomWalk(GSObject):
 
         self._npoints = int(npoints)
 
+        if rng is None:
+            rng = BaseDeviate()
+        self._rng=rng
+
+        self._flux = float(flux)
         if profile is None:
-            self._flux = float(flux)
             if half_light_radius is None:
                 # use runtime error, because this is a fatal error
                 raise RuntimeError("send a half_light_radius "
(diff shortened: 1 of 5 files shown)


esheldon:
In the tests I added a config with the profile specified and it works fine.  But I did not figure out how to specify the profile using strings, I had to generate it in advance.   So I'm not sure how to do this in a pure yaml context.   This is marked as TODO. I could use some help with that.

esheldon:
I'm having difficulty fixing the codecov stuff, partly because I don't know how to check it without pushing a change and waiting half an hour to see what it gives.  Is there a way to check coverage locally on my machine?

rmjarvis:
My usual trick is to add lines like `raise '1'`, `raise '2'`, etc. wherever codecov said a line wasn't covered.  Then I write tests that should hit those lines.  Once I get the error about not being allowed to use strings as exceptions, I remove that line and know that it is covered.  Once all of these `raise 'N'` lines are gone, I push and see if codecov finds anything else uncovered.
But no, I don't know of a really easy way to interpret the coverage output locally, unfortunately.

esheldon:
OK, the patch is now covered but not the project.  But I don't understand how those areas affected are related to this PR.

esheldon:
maybe some of the changes i made ended up removing a code path into that other file which I didn't touch.

by the way, thanks @rmjarvis for the avice it helped a lot

galsim/randwalk.py
```diff
@@ -47,12 +51,21 @@ class RandomWalk(GSObject):
...
+                                    of points.  A single instance will be noisy.
+                                    [default None]
     @param  flux                    Optional total flux in all point sources.
+                                    This value is used for a Gaussian distribution
+                                    if an explicit profile is not sent. 
                                     [default: 1]
+    @param  profile                 Optional profile to use for drawing points.
+                                    If a profile is sent, the half_light_radius
+                                    and flux keywords are ignored.
+                                    [default: Gaussian]
```
rmjarvis: I think default here should be listed as `None`.  The text explains that this means it will automatically make a Gaussian profile in that case.

galsim/randwalk.py
```diff
@@ -100,21 +117,57 @@ class RandomWalk(GSObject):
...
         self._rng=rng
+
+        self._flux = float(flux)
         self._verify()
 
-        self._set_gaussian_rng()
+        if profile is None:
+            if half_light_radius is None:
+                # use runtime error, because this is a fatal error
+                raise RuntimeError("send a half_light_radius "
```
rmjarvis: I think this would normally be a TypeError.  (What you usually get when you don't provide required arguments to a function.)  Or, I've usually been calling this kind of thing a `GalSimIncompatibleValuesError`, which is a subtype of TypeError.

galsim/randwalk.py
```diff
@@ -100,21 +117,57 @@ class RandomWalk(GSObject):
...
+                flux=flux,
+            )
+
+            self._half_light_radius=half_light_radius
+
+            self._set_gaussian_rng()
+        else:
+            if not isinstance(profile, GSObject):
+                raise TypeError("profile must be a GSObject")
+            # the half light radius is not used
```
rmjarvis: Do we maybe want to check here that the input `half_light_radius` and `flux` are None?  (I.e. not specified.)

galsim/randwalk.py
```diff
@@ -133,6 +186,9 @@ def _sbp(self):
     def input_half_light_radius(self):
         """
         The input half-light radius is not necessarily the realized hlr.
+
+        If a profile is sent, this will be -1 if it was a transformation
```
rmjarvis: I think this is incorrect.  Above, you set it to None if the `half_light_radius` doesn't work.

galsim/randwalk.py
esheldon: fixed

galsim/randwalk.py
esheldon: fixed

galsim/randwalk.py
esheldon: done

galsim/randwalk.py
esheldon: fixed

esheldon:
I think this is ready to go.  It says "You’re not authorized to merge this pull request. " I thought I did have authorization.  Anyway, can someone who does go ahead?

Issue #982, opened 2018-08-21
Add bicubic Table2D interpolation


jmeyers314:
I've been playing around with bicubic interpolation for `batoid` surfaces, so thought I'd add them here too.  (Might enable us to use less finely gridded atmospheric screens, for instance).

In 2D, the type of cubic spline most commonly used (a Catmull-Rom spline) is a bit different than what we use in GalSim for the 1D cubic spline (a natural spline).  The natural cubic spline ensures continuous first and second derivatives (requires solving an Ngrid x Ngrid band diagonal matrix equation).  The Catmull-Rom spline, in contrast, only has continuous first derivatives, but is more locally defined, requiring no nPoint x nPoint matrix solutions.  I think this is actually pretty important, since in 2D, the matrix would really be nGrid^2 x nGrid^2 where nGrid is the number of points on one side of the grid, and the matrix would be less bandy too, (though still Toeplitz I think(?)).

One other thought I had while tinkering with the lookup tables is to template on the interpolation enum, rather than using an interpolation-type-targeted class member function pointer as we do currently.  I think this ought to eliminate a pointer dereference in a tight loop, so maybe speed things up a bit.  (Though maybe not, I'll benchmark to find out).  The downside is a bit more bookkeeping since there are now separate class template types for each interpolation type.

jmeyers314:
>I think this ought to eliminate a pointer dereference in a tight loop, so maybe speed things up a bit. 

So I just checked this, and for `interpMany`, it seems the improvement is ~3-8%.  So not huge, but not nothing.

rmjarvis:
While you're at it, any chance you could try implementing Lanczos too?  (cf. #751)  Or if not, maybe at least think a little bit about the way the code is structured there, so the right hooks are in place to eventually include that.

jmeyers314:
>any chance you could try implementing Lanczos too? 

Sure, I'll take a look.


I have a design question: there are a few options for potential restructuring to avoid the dereference in the inner loop:

1.  Just use inheritance.
  I.e., derive (non-virtual) subclasses from Table2DImpl for different interpolants.  Probably lots of code duplication to avoid using virtual functions here though.
2.  Templatize both Table2D and Table2DImpl on the interpolant
  Straightforward, but potentially adds unnecessary `template <Interp>` lines in front of every function definition, and unnecessary Table2D<> symbols into the module.
3. Templatize just Table2DImpl
[...]

PR #986, opened 2018-09-11
#982 LookupTable improvements

jmeyers314:
This PR ...
• Adds bicubic spline interpolation to LookupTable2D
• Lets both LookupTable and LookupTable2D use a galsim.Interpolant object as their interpolant.
• Refactors lensing_ps.py to use LUT2D instead of InterpolatedImage
• Refactors both LUT and LUT2D cpp-layers to use the curiously recurring template pattern, which moves the polymorphism for different interpolants from a runtime decision to a compile-time decision.  There's a small speed advantage from this.

Currently there's no gradient method for either the 1D LUT or for the 2D LUT when the interpolant is a galsim.Interpolant.  We can add these if there's a demand, but I couldn't think of any reason we'd need them for now.

One other item I refactored a bit is the way bounds are set in lensing_ps.py.  With the InterpolatedImage implementation, the bounds were set to the corners of the corner pixels in the image grid.  For the LUT2D implementation, I thought setting the bounds to the actual grid points (corresponding to the centers of the corner pixels in the previous implementation) made more sense, so there's no extrapolation allowed instead of a half-pixel extrapolation allowed.


rmjarvis:
Thanks for the ping.  I'll do that this week.

tests/test_table.py
```diff
@@ -548,6 +643,204 @@ def f(x_, y_):
...
+        newx = np.linspace(0.2, 3.1, 45)
+        newy = np.linspace(0.3, 10.1, 85)
+        newyy, newxx = np.meshgrid(newy, newx)
+
+        np.testing.assert_allclose(tab2d(newxx, newyy), f(newxx, newyy), atol=1e-10, rtol=0)
+        ref_dfdx = dfdx(newxx, newyy)
+        ref_dfdy = dfdy(newxx, newyy)
+        test_dfdx, test_dfdy = tab2d.gradient(newxx, newyy)
+        np.testing.assert_allclose(test_dfdx, ref_dfdx, atol=1e-10, rtol=0)
+        np.testing.assert_allclose(test_dfdy, ref_dfdy, atol=1e-10, rtol=0)
```
rmjarvis: Maybe worth adding a test here of a function that is not an exact cubic, but which is only a function of x or y, that the results from using LookupTable2D matches what we get from the regular LookupTable of just that one dimension (to some precision).  Since the implementations of the spline is different in the two cases, this would be a non-trivial check.

src/Table.cpp
```diff
@@ -359,191 +433,440 @@ namespace galsim {
...
+            // 0 and 1, and der0 and der1 are the derivatives at 0 and 1.
+
+            // I'm assuming the compiler will reduce this all down...
+            double a = 2*(val0-val1) + der0 + der1;
+            double b = 3*(val1-val0) - 2*der0 - der1;
+            double c = der0;
+            double d = val0;
+
+            return d + x*(c + x*(b + x*a));
+        }
```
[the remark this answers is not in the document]
rmjarvis: Also, it would probably be worth an inline comment about the math you are doing here. This calculation isn't "true by inspection". (At least for me.)  This might help make clear the connection to how this relates to the 1D spline calculation, which looks very different.

tests/test_table.py
```diff
@@ -548,6 +643,204 @@ def f(x_, y_):
...
+        return 4*y_
+
+    def f5(x_, y_):
+        return 2*y_*y_ + 3*x_*x_ + 4*x_*y_
+    def df5dx(x_, y_):
+        return 6*x_ + 4*y_
+    def df5dy(x_, y_):
+        return 4*y_ + 4*x_
+    def d2f5dxdy(x_, y_):
+        return np.ones_like(x_)*4
```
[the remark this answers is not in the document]
jmeyers314: Yep.  I think the tests are fine without this fn, so I just nixed it.

src/Table.cpp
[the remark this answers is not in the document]
jmeyers314: I think this is the difference between a "natural" spline (what `setupSpline` does) and a "Catmull-Rom" spline.  The natural spline is smoother, since it enforces the continuous first derivatives.  To get that smoothness though, you have to solve ~N simultaneous (though  sparse) equations.  This isn't so bad in 1D - it's inverting a tri-diagonal matrix.  In 2+D though, you lose the tridiagonality since points in more than one dimension aren't well-ordered.  It's still sparse though, so I'm sure it's still possible, but I didn't try it.  (...and all the bicubic spline algorithms I found online were either the Catmull-Rom type, or the cubic convolution, which makes a third variant).

So it is a different design and will yield different results than what we do in 1D.  I think that's okay though; I don't think I'd advocate changing either one.

FWIW, one potential advantage of the Catmull-Rom spline is that it's local.  Changing the value of a point only changes the interpolation near that point, which isn't true for the natural spline.

I did go and make some additional comments to elucidate the maths here.

galsim/table.py
```diff
@@ -407,66 +444,159 @@ class LookupTable2D(object):
     @param x              Strictly increasing array of `x` positions at which to create table.
     @param y              Strictly increasing array of `y` positions at which to create table.
     @param f              Nx by Ny input array of function values.
-    @param interpolant    Interpolant to use.  One of 'floor', 'ceil', 'nearest', or 'linear'.
+    @param dfdx           Optional first derivative of f wrt x.  Only used if interpolant='spline'.
```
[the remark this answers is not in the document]
jmeyers314: Maybe...  I think we'd have to switch from natural -> Catmull-Rom to enable this though.  Or add the spline-type as a kwarg.

src/Table.cpp
```diff
@@ -137,63 +140,135 @@ namespace galsim {
...
+        using Table::TableImpl::TableImpl;
+
+        double lookup(double a) const override {
+            int i = _args.upperIndex(a);
+            return static_cast<const T*>(this)->interp(a, i);
+        }
+
+        void interpMany(const double* xvec, double* valvec, int N) const override {
+            for (int k=0; k<N; k++) {
+                int i = _args.upperIndex(xvec[k]);
```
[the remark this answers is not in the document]
jmeyers314: Hmmm... Do you have some specific speed-up ideas for this?  I just took a quick go at moving the `equalSpaced` check outside of the loop and keeping the rest basically the same (thinking maybe there'd be a bump from better cache friendliness).  That didn't seem to make a difference though.

src/Table.cpp
```diff
@@ -359,191 +433,440 @@ namespace galsim {
...
+        double lookup(double x, double y) const {
+            int i = _xargs.upperIndex(x);
+            int j = _yargs.upperIndex(y);
+            return static_cast<const T*>(this)->interp(x, y, i, j);
+        }
+
+        void interpMany(const double* xvec, const double* yvec, double* valvec, int N) const {
+            for (int k=0; k<N; k++) {
+                int i = _xargs.upperIndex(xvec[k]);
+                int j = _yargs.upperIndex(yvec[k]);
```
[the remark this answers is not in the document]
jmeyers314: Sure.  I added a `grid=True` kwarg to `__call__` and `__gradient__` that will cause the returned values to be on the outer product of `x` and `y` input arrays instead of returning something congruent to both of them.  This sped up a 100x100 interpolation by a factor of ~1.25, a 1000x1000 by ~2, and a 10000x10000 by ~4.3.  Since the number of calls to upperIndex should be going from N^2 to 2N, it seems like upperIndex must not be all that dominant to the runtime (using bilinear interpolation).

src/Table.cpp
```diff
@@ -359,191 +433,440 @@ namespace galsim {
...
+                         const Interpolant* gsinterp) :
+            T2DCRTP<T2DGSInterpolant>(xargs, yargs, vals, Nx, Ny), _nx(Nx), _gsinterp(*gsinterp) {}
+
+        double interp(double x, double y, int i, int j) const {
+            double dxgrid = _xargs[i] - _xargs[i-1];
+            double dygrid = _yargs[j] - _yargs[j-1];
+            double dx = (x - _xargs[i-1])/dxgrid;
+            double dy = (y - _yargs[j-1])/dygrid;
+
+            // Stealing from XTable::interpolate
```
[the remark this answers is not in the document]
jmeyers314: >  Would it be possible to refactor this such that this calculation is written only once somewhere?

I assume so, but note getting this to work wasn't a copy and paste job.  If I recall correctly, the storage order here and in XTable are switched (row-major, col-major), and the indices are different too (different origins).  The table here has independent x&y dimensions, whereas I think the XTable dims are assumed to be the same (or maybe it was N x N/2).  I can take another look at this, but probably not until this weekend....

tests/test_table.py
jmeyers314: I think for this proposed test we'd actually expect different results (for better or worse) at the moment (see comments above about different types of splines).

jmeyers314:
Hi Mike,

I'm not sure I actually successfully addressed all of your questions, but I think I'm at least at a point where renewed feedback would make sense.  Let me know what you think of the comments/changes I've made.

src/Table.cpp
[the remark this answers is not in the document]
rmjarvis: OK.  Don't worry about it then.  If/when I get around to tackling XTable and KTable, I'll try to remember to have this bit call out to the new code and make sure it's all working compatibly.

galsim/table.py
```diff
@@ -149,19 +138,35 @@ def __init__(self, x, f, interpolant='spline', x_log=False, f_log=False):
         if self.f_log and np.any(self.f <= 0.):
             raise GalSimValueError("Cannot interpolate in log(f) when table contains f<=0.", f)
 
+        # Check equal-spaced arrays
+        if self._interp1d is not None:
+            if self.x_log:
+                logx = np.log(self.x)
+                dlogx = np.diff(np.log(self.x))
```
rmjarvis: First, it looks like logx isn't used.  I'm sure you intended to use it in the dlogx line.

But also, suspect it might be faster to use division rather than log for this:
```
ratio = self.x[1:] / self.x[:-1]
if not np.allclose(ratio, ratio[0]):

File: galsim/table.py [hunk 9]
```diff
@@ -645,76 +666,110 @@ def _call_constant(self, x, y):
...
+                # Start with constant array, then interpolate good positions
+                f = np.empty_like(x, dtype=float)
+                f.fill(self.constant)
+                good = ((x >= self.x[0]) & (x <= self.x[-1]) &
+                        (y >= self.y[0]) & (y <= self.y[-1]))
+                xx = np.ascontiguousarray(x[good].ravel(), dtype=float)
+                yy = np.ascontiguousarray(y[good].ravel(), dtype=float)
+                tmp = np.empty_like(xx, dtype=float)
+                self._tab.interpMany(xx.ctypes.data, yy.ctypes.data, tmp.ctypes.data, len(xx))
+                f[good] = tmp
```
rmjarvis: I'm curious about the asymmetry of good vs bad implementations here.  I don't think I have a particularly good intuition about which is faster for typical use cases.  Was there some motivation for the difference?  Or just that you coded them up at different times and chose arbitrarily and differently.  :)

galsim/table.py
```diff
@@ -645,76 +666,110 @@ def _call_constant(self, x, y):
...
+                return f
+
+    def _call_raise(self, x, y, grid=False):
         if not self._inbounds(x, y):
             raise GalSimBoundsError("Extrapolating beyond input range.",
                                     PositionD(x,y), self._bounds)
```
rmjarvis: I think I missed this before, but won't these not work right when x,y are arrays?  Similarly other places that do `PositionD(x,y)`.

galsim/table.py
jmeyers314: No good reason.  The bad reason though was that I haven't found a satisfying way to assign to a numpy array indexed with a boolean mask in more than one dimension, i.e., something like

```python
f[goodx, goody] = blah
```

FWIW, I'm not really sure we _have_ a typical use case for `grid=True`, though I think it is a nice option to have.

rmjarvis: OK.  I thought the phase_psf stuff was on a grid.  No?

jmeyers314: We might be able to use a grid for a pure optical screen, but since the atmospheric screens depend on time and they're moving around, we generally aren't evaluating those on a grid.

galsim/table.py
jmeyers314: Yep, good find.  I hacked together a solution to find a representative out-of-bounds point and raise an exception using that.

jmeyers314:
Looks like this is ready to go now.  (Thanks for the final cleanup Mike).  I'll merge this afternoon if I don't hear anything more.

Issue #163, opened 2012-05-22
Fixups and enhancements for photon shooting


gbernstein:
A laundry list of things that are needed or desired to improve in the photon-shooting code and interfaces.  I have tried to cull these from all the traffic during the pull request but may have forgotten some, so add your own here.  We'll keep as one issue for now.
- `SBAiry` and `SBInterpolatedImage` should be efficiently copying their `OneDimensionalDeviate` and `ProbabilityTree` structures when they are `duplicate()`d.  Right now those structures are wastefully recomputed on every copy.  Should address this during the big push on #107 to make SBProfiles immutable.
- Detailed numerical comparisons of photon-shooting to DFTs and real-space convolutions, and shear accuracy tests
- Do efficiency testing where there are multiple ways to do the same thing (unit-circle randoms vs sin/cos)
- Add the option for unit-circle trick to `SBExponential` and maybe other places
- Potentially shuffle the photons of `SBAdd` so that we can eliminate the shuffle done on on `SBConvolve` photon shooting.
- Might be the case that use of `BinomialDeviate` in `SBAdd` is very slow, perhaps calling `UniformDeviate` many times, in which case we might as well call it ourselves.
- See whether the Newton-Raphson method of `SBExponential` is really faster than the `OneDimensionalDeviate` routine that you get when you request `SBSersic` with n=1.
- Other speed optimizations
- Add code that will let `drawShoot()` add Poisson variation to the total flux of the drawn object.
- Time DFT vs photon shooting so we can offer advice to users - results will be specific to certain profiles.
- Have the `drawShoot()` code detect requests for large numbers of photons, and break it up into smaller batches of photons whose images are then averaged.  To avoid hogging memory to hold large `PhotonArray`s.
- Have the `drawShoot()` method return the fraction of photons that fell outside the `Image`.

PR #989, opened 2018-11-04
Change OneDimensionalDeviate to use linear approximation rather than constant

rmjarvis:
I've been looking into ways to speed up the atmospheric photon shooting being used for LSST DC2, and it turns out that one of the tall poles for the brighter objects is the OneDimensionalDevaite::shoot function.  In particular when used by the Airy profile (which is used as part of the second kick).

OneDimensionalDeviate had been treating the profile as piecewise constant intervals, but then (usually) using variable flux to reweight the photons in each interval to get the right net profile.
cf. This comment:
```
If `_useRejectionMethod=false`, then no rejection is done.  The photon is kept, but is given
a flux value equal to the FluxDensity at x relative to the mean over the interval.  This is
faster but makes the statistics of the photons harder to interpret.
```

For many profiles, including Airy, the function call required to reweight the photon's flux is fairly slow.  So I've switched the implementation to treating each interval as approximately linear, rather than approximately constant, and getting the photon's position to be correct given this approximation.  Then, there is no need for the flux reweighting, and thus no need for a function call.

I think this all works.  Certainly, all the unit tests pass.  And the statistics of the photons are no longer "hard to interpret".  But hopefully @gbernstein will have a chance to take a look to see if he thinks there is anything I removed from the original design that ought to be kept.

With Airy, this led to a pretty massive speedup of this part of the code.  Overall, the second_kick.shoot function was almost a factor of 2 faster as used by my time_phase_phot.py timing script.  And I checked that the time for exponential (which has a much simpler function call) is not slowed down.  (Nor sped up -- the time was basically the same for the two implementations.)

The next thing to tackle is related to the Table2D gradient calculation, but I want to pull in Josh's changes in #982 before digging into this, since I think some of what he did there will already help.

### DIFF
--- a/include/galsim/OneDimensionalDeviate.h
+++ b/include/galsim/OneDimensionalDeviate.h
@@ -179,7 +179,7 @@ namespace galsim {
          * @param[in] smallFlux Flux below which a sub-interval is not further split.
          * @returns List contiguous Intervals whose union is this one.
          */
-        std::list<Interval> split(double smallFlux);
+        std::list<shared_ptr<Interval> > split(double toler);
 
     private:
 
--- a/include/galsim/OneDimensionalDeviate.h
+++ b/include/galsim/OneDimensionalDeviate.h
@@ -70,8 +70,6 @@ namespace galsim {
      *
      * The `drawWithin()` method will select one photon (and flux) drawn from within this interval
      * or annulus, such that the expected flux distribution matches the FluxDensity function.
-     * Each photon is given a flux value equal to the FluxDensity at x relative to the mean over
-     * the interval.
      *
      * See the `OneDimensionalDeviate` docstrings for more information.
      */
@@ -135,9 +133,7 @@ namespace galsim {
          * @brief Draw one photon position and flux from within this interval
          * @param[in] unitRandom An initial uniform deviate to select photon
          * @param[out] x (or radial) coordinate of the selected photon.
-         * @param[out] flux flux of the selected photon, nominally +-1, but can differ if not
-         *             using rejection.
-         * @param[out] ud UniformDeviate used for rejection sampling, if needed.
+         * @param[out] flux flux of the selected photon = +-1
          */
         void drawWithin(double unitRandom, double& x, double& flux) const;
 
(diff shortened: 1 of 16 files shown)


src/SBExponential.cpp
```diff
@@ -592,14 +592,22 @@ namespace galsim {
...
+            dbg<<"dy, r = \n";
+            dbg<<dy<<"  "<<r<<std::endl;
             while ( std::abs(dy) > Y_TOLERANCE) {
-                r = r + (1.+r)*dy/r;
+                // Newton step: dy/dr = r / (1+r)
+                r += (1.+r)*dy/r;
                 dy = y - r + std::log(1.+r);
+                dbg<<dy<<"  "<<r<<std::endl;
             }
+
```
rmjarvis: These changes are part of a commented out bit of code.  I was re-investigating the idea of using Newton's method directly here rather than using ODD.  (Original investigation [here](https://github.com/GalSim-developers/GalSim/issues/163#issuecomment-6523419))  I did manage to speed this up a fair bit, but it's still slower than the ODD implementation.

include/galsim/OneDimensionalDeviate.h
```diff
@@ -147,8 +139,7 @@ namespace galsim {
          *             using rejection.
          * @param[out] ud UniformDeviate used for rejection sampling, if needed.
          */
-        void drawWithin(double unitRandom, double& x, double& flux,
-                        UniformDeviate ud) const;
+        void drawWithin(double unitRandom, double& x, double& flux) const;
```
jmeyers314: Method documentation looks out of date.

src/OneDimensionalDeviate.cpp
```diff
@@ -175,40 +175,61 @@ namespace galsim {
         // of the flux in this Interval if the function were constant
         // over the interval.
         if (_isRadial) {
-            double rsq = _xLower*_xLower*(1.-fraction) + _xUpper*_xUpper*fraction;
-            return sqrt(rsq);
+            // The model is pdf(r) = f0 r + (f1-f0)/(r1-r0) * (r-r0) r
```
jmeyers314: Above.  "if the function were linear over the interval."

jmeyers314:
Looks good to me Mike.  Just a couple documentation cleanups.

rmjarvis:
If there are no other comments, I'll plan to merge this on Friday.  Thanks, Josh.

Issue #990, opened 2018-11-12
Fix `PhaseScreenPSF.withGSParams`


jmeyers314:
`PhaseScreenPSF.withGSParams` uses `copy.copy`, which uses pickling, which currently draws the PSF in FFT mode before serializing the state of the PSF object.  This adds unnecessary dependencies on the underlying phase screen `wavefront()` method and unnecessary computation in the case where the PSF is only intended to be drawn using photon-shooting.  See (https://github.com/LSSTDESC/imSim/pull/182).

I think part of the solution here will be to avoid using `copy.copy` in `PhaseScreenPSF.withGSParams`.  The goal is that the result of using `withGSParams` ought to be indistinguishable from creating a new `PhaseScreenPSF` directly but with different gsparams.

This probably also means we should rethink pickling for this class to make sure it's commutative with the rewritten `withGSParams`.  This likely means including the underlying `PhaseScreenList` in the pickle (which could mean a large pickle file, but is really the only correct way to proceed).

I'm also considering making `PhaseScreenPSF.img` and `.finalize` hidden attributes, since these are really implementation details for drawing with `method='FFT'`, and potentially exposing `PhaseScreenPSF._screen_list`, since it seems to be formally required and definitely is part of the PSF definition.

rmjarvis:
Closed via #991

PR #991, opened 2018-11-13
#990 fix PhaseScreenPSF.withGSParams

jmeyers314:
I think I'm happy with this now.  I avoided using copy/pickle in `withGSParams()`, and also updated the pickling so that it doesn't assume that `prepareDraw` has been called.  There a bunch more tests in place to check that A) `_wavefront()` isn't needed in order to pickle/copy/change GSParams or draw using photon-shooting and B) make sure that copying, changing GSParams, and drawing are appropriately commutative.

### DIFF
--- a/galsim/phase_psf.py
+++ b/galsim/phase_psf.py
@@ -1194,7 +1194,7 @@ class PhaseScreenPSF(GSObject):
         # Hidden `_bar` kwarg can be used with astropy.console.utils.ProgressBar to print out a
         # progress bar during long calculations.
 
-        self.screen_list = screen_list
+        self._screen_list = screen_list
         self.t0 = float(t0)
         self.lam = float(lam)
         self.exptime = float(exptime)
@@ -1204,7 +1204,7 @@ class PhaseScreenPSF(GSObject):
             if 'diam' not in kwargs:
                 raise GalSimIncompatibleValuesError(
                     "Diameter required if aperture not specified directly.", diam=None, aper=aper)
-            aper = Aperture(lam=lam, screen_list=self.screen_list, gsparams=gsparams, **kwargs)
+            aper = Aperture(lam=lam, screen_list=self._screen_list, gsparams=gsparams, **kwargs)
         elif gsparams is None:
             gsparams = aper.gsparams
         else:
@@ -1238,7 +1238,7 @@ class PhaseScreenPSF(GSObject):
         self._kcrit = kcrit
         # We'll set these more intelligently as needed below
         self._second_kick = second_kick
-        self.screen_list._delayCalculation(self)
+        self._screen_list._delayCalculation(self)
         self._finalized = False
 
     @lazy_property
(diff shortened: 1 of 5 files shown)


tests/test_phase_psf.py
```diff
@@ -985,6 +985,83 @@ def test_gc():
...
+        assert psf2 is not psfGSP
+        rng = galsim.BaseDeviate(57721) # reset rng
+        testimgs.append(psf2.drawImage(rng=rng, **drawKwargs))
+    # Draw the (fresh) withGSP PSF and compare
+    psf = psl.makePSF(exptime=0.02, time_step=0.01, diam=1.1, lam=1000.0)
+    psfGSP = psf.withGSParams(galsim.GSParams(folding_threshold=6e-3))
+    rng = galsim.BaseDeviate(57721)
+    img = psfGSP.drawImage(rng=rng, **drawKwargs)
+    for testimg in testimgs:
+        assert testimg == img
```
rmjarvis: I think we should maybe also do at least some of the above tests for OpticalPSF.  For instance, I think the following should probably fail currently:
```
opt_psf2 = opt_psf.withGSParams(gsp)
assert isinstance(opt_psf2, galsim.OpticalPSF)
```
which I think used to work. (?)  Also, I believe my suggested implementation of `withGSParams` should do this right, but I'm not 100% positive whether it will pass all of these other tests.

jmeyers314: I think this actually still works.  `OpticalPSF` isn't a subclass of `PhaseScreenPSF` (instead it holds one as a lazy property).  So `OpticalPSF.withGSParams` is really `GSObject.withGSParams`, which sets the `._gsparams` attr before the lazy property is evaluated.  I will certainly add this test though.

rmjarvis: Ah.  Ok.  I assumed it was, but you're right.

galsim/phase_psf.py
```diff
@@ -1324,27 +1329,30 @@ def flux(self):
...
-        ret.aper = self.aper.withGSParams(gsparams)
-        return ret
+        gsparams = GSParams.check(gsparams)
+        aper = self.aper.withGSParams(gsparams)
+        return PhaseScreenPSF(
+            self.screen_list, self.lam, self.t0, self.exptime, self.time_step, self.flux,
+            self.theta, self.interpolant, self.scale_unit, self._ii_pad_factor,
+            self._suppress_warning, self._geometric_shooting, aper, self._second_kick, self._kcrit,
+            gsparams, self._force_stepk, self._force_maxk
+        )
```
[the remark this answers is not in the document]
jmeyers314: Sure; streamlined this.  I modified your code a bit to make sure that any lazy properties that depend on gsparams are also recomputed in the new object.

rmjarvis:
I went ahead and merged this now, so I can submit my next PR about speeding up the phase_psf stuff, which I started off the tip of this branch.

villarrealas:
Apologies for the delay on testing - files were missing from the base docker images that needed to be tracked down. We should be on target for getting some test runs finished early next week. I take it I should use the latest GalSim master then?

rmjarvis:
You can still use this branch if you prefer.  Or master is fine too.  Both will have the fix.

PR #993, opened 2018-11-16
Speedups for photon shooting PhasePSF

rmjarvis:
I've been chipping away at the running time for shooting PhasePSF when nphotons is fairly small.  (I've been using 10 photons for my tests.)  There wasn't really one huge thing that sped things up, but all told, I've managed to speed up the script in `devel/lsst/time_phase_phot.py` from 445 seconds down to 250 seconds, which is more than a factor of two for the part of this that is actually involved in shooting photons.

If you're curious about which changes here had more or less impact on the speed, I've been keeping a log of the times on my laptop for each commit:

* 444.8 04a7163f1 Update time_phase_phot.py to test faint objects
* 415.4 e5ae8f70d Cache the illuminated portions of u,v for an aperture
* 331.4 f5d05f6cf Don't build _ii for positive and negative flux if geometric shooting.
* 329.8 de81e0ecd Directly call _call_wrap and _gradient_wrap for phase screens
* 325.3 e76330522 Have _call and _gradient functions always use numpy array inputs
* 319.9 56ecf9c2b Comment out all asserts in main galsim code
* 318.7 352fc75cb Update gradx, grady in place
* 308.0 ec613d3e2 Don't compute more powers of R_outer than necessary
* 297.7 e22de1fa1 Simplify the PhaseScreen __getitem__ implementation
* 293.7 1c1724023 Only use one numpy allocation for temporary array in horner2d
* 284.1 38e9cf008 Add triangle option to horner2d
* 282.2 1db00fc4a Don't do theta calculation if theta == 0
* 280.5 ace99501b Write custom SSE code for wrapping args for Table2D
* 280.0 dcc74cf13 Streamline upperIndexMany slightly
* 277.4 54e55f103 Avoid copying arrays in phase_screen as much as possible
* 274.6 fa1f5a30e Make _pa a lazy property
* 262.2 ca98a39be Implement Horner in C++
* 254.7 7c876312d Implement Horner2D in c++
* 250.4 ea53b4850 Allow wcs toImage and toWorld to take optional offset and flux_ratio
* 235.9 9546ed6f2 Avoid gratuitous constructions of UniformDeviates in python (*)

So the biggest one was avoiding the calculation of the dummy_ii when using geometric shooting via the calls to `positive_flux` and `negative_flux`.  I think that calculation was actually wrong anyway, since when geometric shooting, all the photons are positive.

After that, a lot of it was in speeding up the horner2d code, which is now done completely in C++, but you can follow the evolution of that improvement above if you're curious.  :)

Anyway, this will all be useful for speeding up the DC 2.1i ImSim runs.  Probably another 10-15% speedup overall on that.  Which means millions of CPU hours saved.  Or alternatively maybe one more year of depth.

(*) I think this had a slight effect on the part we care about, but most of this improvement was from a change in `fft_vs_geom_movie.py` to use a consistent BaseDeviate, rather than implicitly passing `rng=None` to `drawImage`.


galsim/utilities.py
```diff
@@ -589,17 +590,35 @@ def horner(x, coef, dtype=None):
...
-    result = np.zeros_like(x, dtype=dtype)
-    if len(coef) == 0: return result
-    result += coef[-1]
-    for c in coef[-2::-1]:
-        result *= x
-        if c != 0: result += c
-    #np.testing.assert_almost_equal(result, np.polynomial.polynomial.polyval(x,coef))
+    result = np.empty_like(x, dtype=dtype)
+    # Make sure everything is an array
+    if result.dtype == float:
```
jmeyers314: Do we ever not want a float?

src/Table.cpp
```diff
@@ -1024,4 +1015,40 @@ namespace galsim {
...
+
+        if (n2) {
+            __m128d xx0 = _mm_set1_pd(x0);
+            __m128d xperiod = _mm_set1_pd(period);
+            __m128d xzero = _mm_set1_pd(0.);
+            __m128d* xx = reinterpret_cast<__m128d*>(x);
+            do {
+                __m128d xoffset = _mm_sub_pd(*xx, xx0);
+                __m128d nperiod = _mm_div_pd(xoffset,xperiod);
+                __m128d floornp = _mm_cvtepi32_pd(_mm_cvttpd_epi32(nperiod));
```
jmeyers314: FWIW, I tried using _mm_floor_pd from SSE4.1 here, but it didn't make things any faster...  :(

src/math/Horner.cpp
```diff
@@ -0,0 +1,51 @@
...
+    void Horner(double* x, const int nx, double* coef, const int nc, double* result)
+    {
+        // Start at highest power
+        double* c = coef + nc-1;
+        // Ignore any trailing zeros
+        while (*c == 0. && c > coef) --c;
+        // Repeatedly multiply by x and add next coefficient
+        for (int i=0; i<nx; ++i) result[i] = *c;
+        while(--c >= coef) {
+            for (int i=0; i<nx; ++i) result[i] = result[i]*x[i] + *c;
```
jmeyers314: Maybe another opportunity for SSE?  I think KNL also has fused multiply and add instructions if we wanted to try that.

src/Table.cpp
```diff
@@ -1024,4 +1015,40 @@ namespace galsim {
         _pimpl->gradientGrid(xvec, yvec, dfdxvec, dfdyvec, Nx, Ny);
     }
 
+    void WrapArrayToPeriod(double* x, int n, double x0, double period)
+    {
+#ifdef __SSE2__
+        for (; n && !IsAligned(x); --n)
+            *x -= period * floor((*x-x0)/period);
```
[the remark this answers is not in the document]
rmjarvis: It is usually aligned, but this is a system-dependent statement, so I think it's worth keeping this here.  Thanks for catching the error.

galsim/utilities.py
rmjarvis: Not in anything we currently do.  But I can imagine wanting to do this calculation for int or complex outputs.

src/Table.cpp
rmjarvis: Yeah, I was kind of disappointed that this didn't have a bigger impact.  I thought even just going to C++ for this would have been a big help.  But somehow the pure numpy thing was already close to as fast as it could get.

src/math/Horner.cpp
rmjarvis: I'm pretty sure this is already in a format that the compiler will be able to turn this into good SSE commands for us, but I didn't try it to confirm, since just moving this calculation to C++ was already a huge improvement.

jmeyers314: Yeah, looks like you're right: https://godbolt.org/z/M1pvoV  I think the inner loop starts at 132.  I'm a bit surprised there's no FMA, but I'm sure this is still quite efficient.

rmjarvis: Update: I was wrong.  Putting in a custom SSE implementation for this did indeed shave off a few more seconds.  c0645a8ae

rmjarvis: That's on my laptop though.  Using mostly the standard python compiler options, which do not include `march`, but do include `-O3` and  `-msse2`.  So maybe it isn't being quite as good in that case as when it's compiling for knl.  I am kind of surprised that my inexpert SSE skillz are better than clang's.

rmjarvis:
Final tally with your hstack removal and the SSE Horner implementation is 229.8 seconds.  :)

src/math/Horner.cpp
jmeyers314: That's really interesting.  I just tried adding `-march=native` and `-mtune=native` on my laptop, and things actually got a little bit slower.

I also played a bit with changing the compiler in the compiler-explorer example.  GCC and icc both like to use the fused multiply and add; clang never does - even on trunk.  Everybody likes to use the 512-bit bit zmm registers though, which does make me wonder if we're leaving something on the table (for KNL, not on our laptops) by coding directly to the 128-bit SSE registers.  Maybe we should do some profiling on KNL directly?

rmjarvis: We might be leaving a little bit on the table, but `_horner2d` is down to only 4.2 seconds out of 96.6 seconds total for `PhaseScreenPSF._shoot`.  So only at most a couple percent further speed up could be possible.  We're probably at the point now where the SED stuff and the LSSTCam WCS calculations are now much taller poles than anything left in GalSim.

jmeyers314: Fair enough.

rmjarvis:
Merging this now to simplify speed tests of ImSim at ANL, so they can just use the current master branch.  However, further comments still welcome if anyone wanted to comment on any of this.

PR #995, opened 2018-12-07
 Make sure FitsWCS works if input header has lowercase keys

rmjarvis:
@beckermr ran into an issue where he had a fits "header" (really just a dict) that had lowercase keys.

The [FITS standard dictates uppercase letters only in header keys](https://fits.gsfc.nasa.gov/standard40/fits_standard40aa-le.pdf) (4.1.2.1), so this is technically user error.  But we should be accommodating, so this PR basically just makes sure to convert all keys to uppercase on creation of a FitsHeader object so the input dict can use either format.

### DIFF
--- a/galsim/fits.py
+++ b/galsim/fits.py
@@ -1118,9 +1118,6 @@ class FitsHeader(object):
             self.header = pyfits.Header()
             if header is not None:
                 if hasattr(header, 'items'):
-                    # For dicts, make sure all the keys are uppercase.
-                    if isinstance(header, dict):
-                        header = {k.upper(): v for k, v in header.items()}
                     # update() should handle anything that acts like a dict.
                     self.update(header)
                 else:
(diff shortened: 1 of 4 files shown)


beckermr:
Matthew “technically committed a user error” Becker is what they call me!

Seriously though, thanks for putting in this PR!

galsim/fits.py
```diff
@@ -1119,6 +1119,7 @@ def __init__(self, header=None, file_name=None, dir=None, hdu_list=None, hdu=Non
             if header is not None:
                 if hasattr(header, 'items'):
                     # update() should handle anything that acts like a dict.
+                    header = {k.upper(): v for k, v in header.items()}
```
beckermr: You may have trouble in python 2 with this line...

rmjarvis: It's fine in 2.7.  It was only invalid in 2.6, which we don't support anymore.

The test failures are something weird about how PyAst is pickling ZPX WCS's.  It apparently works fine in 3.6, but not 2.7 or 3.5.  I'll figure it out.

beckermr: Ahhh good to know.

PR #996, opened 2018-12-10
Make sure PoissonDeviate works with mean=0

rmjarvis:
@jchiang87 and @villarrealas were getting seg faults when trying to generate random Poisson deviates with mean=0.  

This PR fixes this, so that PoissonDeviate(mean=0) always generates 0's.  This seems to be the mathematically consistent thing to do, although I'll note that [Wikipedia](https://en.wikipedia.org/wiki/Poisson_distribution) says that the Poisson deviate is only defined for mean > 0.

I also make sure that it raises an appropriate exception when mean < 0.  Also, while debugging this I noticed that our repr for PoissonDeviate actually looked like a BaseDeviate.  We never really check the repr of deviates, since we don't expect them to round trip.  But they should at least round trip to the correct type if not an equivalent object.  So I fixed that too.

I'd request a fast turnaround on a review, since we'd like to get the mean=0 fix into v2.1.2 quickly for the LSST DC2 run, where this bug turned up.

### DIFF
--- a/tests/test_random.py
+++ b/tests/test_random.py
@@ -791,6 +791,7 @@ def test_poisson_zeromean():
     p = galsim.PoissonDeviate(testseed, mean=0)
     p2 = p.duplicate()
     p3 = galsim.PoissonDeviate(p.serialize(), mean=0)
+    do_pickle(p)
 
     # Test direct draws
     testResult = (p(), p(), p())
--- a/tests/test_random.py
+++ b/tests/test_random.py
@@ -232,6 +232,10 @@ def test_uniform():
     do_pickle(u, lambda x: x.serialize())
     do_pickle(u, lambda x: (x(), x(), x(), x()))
     do_pickle(u)
+    assert 'UniformDeviate' in repr(u)
+    assert 'UniformDeviate' in str(u)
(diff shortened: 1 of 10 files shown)


tests/test_random.py
```diff
@@ -768,6 +784,51 @@ def test_poisson_highmean():
                 err_msg='Realized variance for PoissonNoise did not match getVariance()')
 
 
+@timer
+def test_poisson_zeromean():
```
jmeyers314: Need to add to `if __name__ == '__main__'` block

tests/test_random.py
```diff
@@ -768,6 +784,51 @@ def test_poisson_highmean():
...
+    """
+    p = galsim.PoissonDeviate(testseed, mean=0)
+    p2 = p.duplicate()
+    p3 = galsim.PoissonDeviate(p.serialize(), mean=0)
+
+    # Test direct draws
+    testResult = (p(), p(), p())
+    testResult2 = (p2(), p2(), p2())
+    testResult3 = (p3(), p3(), p3())
+    np.testing.assert_allclose(testResult, 0)
```
jmeyers314: Use assert_equal here?

rmjarvis:
Thanks Josh!

PR #997, opened 2018-12-12
DOC add note about the drawing method to use for DES PSFEx PSFs

beckermr:
I added a note on something that tripped me up. Comments welcome!

@rmjarvis for viz


beckermr:
Yep. I cannot merge this PR so you'll have to do the honors. :)

Thanks!

Issue #984, opened 2018-08-29
also look for pybind in /usr


gijzelaerr:
I have pybind installed (debian package) but galsim can't find it:
```
$  dpkg -L pybind11-dev
/.
/usr
/usr/include
/usr/include/pybind11
/usr/include/pybind11/attr.h
/usr/include/pybind11/cast.h
/usr/include/pybind11/chrono.h
/usr/include/pybind11/common.h
[...]
```

but when i python setup.py galsim i get:
```
Looking for pybind11 header files:
   /home/gijs/.local/include/python2.7  (no)
   /usr/local/include/python2.7  (no)
Could not find pybind11 header files.
They should have been in one of the following two locations:
    /home/gijs/.local/include/python2.7
    /usr/local/include/python2.7
error: Could not find PyBind11
```

rmjarvis:
It looks like your python-installed pybind11 is incompatible with where you actually have the header files.  Those two directories that it looked in are the directories reported by `pybind11.get_include()`.

So perhaps try reinstalling pybind11 with `pip install -U pybind11`.  Or you can install using SCons instead, where there is more flexibility about specifying precisely which directory you want it to look for these files.

gijzelaerr:
hm, that is odd:
```
$  python
Python 2.7.15rc1 (default, Apr 15 2018, 21:51:34)
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import pybind11
>>> pybind11.__path__
['/usr/lib/python2.7/dist-packages/pybind11']
>>> pybind11.get_include()
'/usr/include'
```

gijzelaerr:
looks like the setup.py script is using a pybind11 from a local egg and not the system wide pybind11:

/scratch/packaging/kern/packaging/build/galsim/.eggs/pybind11-2.2.3-py2.7.egg/pybind11

to clarify, i'm trying to update our debian package for galsim so we actually want to use that specific pybind11.

gijzelaerr:
it works if I modify the code to look for `/usr/include`

rmjarvis:
Is there any action to do here?  Or should this be closed?

EiffL:
:-/ just had the same problem, had to manually go in and change the path to pybind11 headers, which for archlinux are located in `/usr/include`. Not sure how to resolve this automatically though...

EiffL:
So, this is a known issue, pybind is not going to necessarily return the path to the headers. 
[...]

PR #999, opened 2018-12-29
Adds additional location to look for PyBind11

EiffL:
This PR aims to address #984 , as documented in that issue, pybind11 on its own is pretty dumb and wont necessarily the correct path to its header files. 
This is a problem on linux systems where pybind is typically installed as a proper package, not using pip. On Arch and Debian/Ubuntu, the default installation path to pybind11 is `/usr/include` so this PR looks in that folder before aborting the build process, it's not full proof, I haven't found a robust way of locating the headers without cmake, but at least, this should cover 80% of current potential failures on linux.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -535,16 +535,20 @@ def add_dirs(builder, output=False):
     import pybind11
     print('PyBind11 is version ',pybind11.__version__)
     print('Looking for pybind11 header files: ')
-    for user in [True, False, None]:
-        if user is None:
+    locations = [pybind11.get_include(user=True),
+                 pybind11.get_include(user=False),
+                 '/usr/include',
+                 None]
+    for try_dir in locations:
+        if try_dir is None:
             # Last time through, raise an error.
             print("Could not find pybind11 header files.")
             print("They should have been in one of the following two locations:")
-            print("   ",pybind11.get_include(True))
-            print("   ",pybind11.get_include(False))
+            for l in locations:
+                if l is not None:
+                    print("   ", l)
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -535,16 +535,20 @@ def add_dirs(builder, output=False):
...
-        if user is None:
+    locations = [pybind11.get_include(user=True),
+                 pybind11.get_include(user=False),
+                 '/usr/include',
+                 None]
+    for try_dir in locations:
+        if try_dir is None:
             # Last time through, raise an error.
             print("Could not find pybind11 header files.")
             print("They should have been in one of the following two locations:")
```
rmjarvis: Probably remove two from this line.

setup.py
```diff
@@ -535,16 +535,20 @@ def add_dirs(builder, output=False):
     import pybind11
     print('PyBind11 is version ',pybind11.__version__)
     print('Looking for pybind11 header files: ')
-    for user in [True, False, None]:
-        if user is None:
+    locations = [pybind11.get_include(user=True),
+                 pybind11.get_include(user=False),
+                 '/usr/include',
```
rmjarvis: Maybe also include "/usr/local/include"?  Seems like that's another place things could normally be installed, so maybe worth adding that too.

EiffL: Yep, that doesn't cost much, done

setup.py
EiffL: Agreed :-) Done

EiffL:
Thanks Mike, should be good to go.

PR #1000, opened 2018-12-31
WIP: Support for drawing light profiles from tensorflow in GalSim

EiffL:
This PR aims to provide support for drawing light profiles using a pre-trained tensorflow module. This is functional but still a *work in progress*, not ready for full review, but I have a few questions.

This will allow GalSim to draw light profiles from any tensorflow module (packaged as tensorflow-hub modules see here: https://www.tensorflow.org/hub) taking as an input some scalar parameters, and outputting a light profile represented as a GalSim InterpolatedImage. The tensorflow module specifies which parameters it needs, as well as the size and resolution of the generated postage stamps.
I'm independently building some deep learning based generative models, but note that this could be used to execute any sort of light profile defined as a tensorflow module in parallel using the GPU (including all the various kinds of parametric profiles, my point is that it doesn't have to come from a neural network).

Here are some open questions:

  - [ ] I have put these functionalities into an optional galsim.tensorflow module (this could potentially grow and provide alternative GPU-based methods for a bunch of galsim functions). Is that appropriate, and if so, is there a procedure for installing the dependencies of optional modules ?

  - [ ] With the yaml driver, every galaxy is drawn independently, so that means the tensorflow model is called as many times as there are galaxies, which is extremely inefficient given that tensorflow typically works in parallel by batches. Is there any way to
draw the galaxies by batches with the Yaml driver ? Or is it against the GalSim spirit ? 

  - [ ]  Another technicality, only one process at a time can use the GPU, so this is incompatible with the multiprocessing spirit of GalSim. In a python script, I guess we can work around that by only doing some tasks with multiprocessing, but this would be something to think about if proper GPU support was to be added to GalSim further down the road. Thoughts on this?


beckermr:
My advice here would be to write a separate package that hooks into the galsim config parser. See the galsim extras the mike and Erin maintain. 

This has a lot of advantages. 

1. Historically the tensorflow api has moved very quickly compared to a lot of things. Decoupling maintenance updates to the ever new TF api from the galsim release cycle seems like a good idea. 

2. It allows you to ship and dev code at your own pace with lower standards as your code matures. Galsim as I am sure you are aware is an extremely mature code base and keeping more experimental features separated seems like a good idea. 

3. Galsim’s own use of hardware accelerators is a bigger issue that I would not want my own code coupled to. In other words, If the proposal is to use tensorflow as an interface to mediate calls to backend hardware accelerators, this in and of itself would need much discussion and planning for all of the reasons you stated. 

The rule of thumb I am using here is an old adage that code should be easy to delete. Making a separate package that hooks into the parser is easy to delete from galsim since we never added anything. OTOH if we add a tensorflow interface, we will be committing to maintain it for a fair bit until we can fully deprecate it should we want to. 

Finally, on the issue of a single call to the graph each time you draw an object, this is clearly not efficient as you state. I don’t see why one could not instantiate a generator that has a cache of drawn objects it pulls from. This cache could be updated in reasonably sized minibatches to enable efficient gpu usage. 

One last thought is that pytorch might be better here but I am not 100% sure. It has a much more python first approach and might let you more explicitly handle gpu memory etc.

beckermr:
Ack another thing. It is possible to write sklearn style estimators that use tensorflow flow and also pickle. See a package on GitHub called muffnn.

rmjarvis:
I think I agree with Matt.  This feels like it would be a better fit as a [galsim_extra](https://github.com/esheldon/galsim_extra) PR, rather than the core GalSim software.  Or your own repo if you prefer.  (The advantage of including in galsim_extra is mostly just so other people can find it more easily.)

EiffL:
Hi Matt and Mike, thanks a lot for the feedback! I think you are right, I was going back and forth between opening this PR or keeping it in a separate package. 

The reason I went ahead with this PR anyway is that these tensorflow modules are completely packaged externally, so there is no TF code to maintain in GalSim apart from the 5-10 lines that open the module file and execute it. So, it's essentially like supporting a new type of input files, leaving it to the user to define what the module file does. 
It also decoupled the deep learning stuff I'm working on with this feature, I wanted to formalize how ml models could be integrated with GalSim. 

I'm closing this PR and putting back this code in my repo for the time being, but will probably reopen one in GalSim extra down the line. 

Also but let me know if you ever want to discuss integrating TF with GalSim. I hear the pythonic goodness of pytorch, but TF is way better for production level stuff, and it is also native C++ so it could be used from lowlevel GalSim.

Issue #919, opened 2017-11-03
More WFIRST Updates


rmjarvis:
Some updates that Rachel didn't get to in branch #675.  I'm making a new issue for these to refocus the effort to this list of remaining tasks:

* Updating Zernikes since cycle 5
* Incorporating their variation across the SCAs
* Any other changes in cycle 7 besides the ones listed above as going into this PR!
* Make WFIRST module components accessible in config
* Dither sequences (re-think connection to Chris's code)
* Simpler persistence specifications
* Brighter-fatter: empirically-based Antilogus-type implementation (CDModel) using coefficients measured by Chris
* Correlated read noise: should store a correlation function in WFIRST module and use it to generate the read noise. [no new coding needed]
* Individual pixel effects: ability to put in a flat field including effects down to the sub-pixel level. Cross-hatching pattern could then be simulated.
* Defects like cosmic rays
* Re-enable WCS tests that were removed because we no longer have a sample WCS from the WFIRST project office.

People interested in doing a subset of these should feel free to spin off separate issues as appropriate.

rmandelb:
I received an update to the cycle 7 WCS from the WFIRST project office, so now I can do a fully self-consistent update to the whole module and re-enable some of the unit tests.

rmandelb:
@rmjarvis - no rush, but I could use some advice about the user interface for some of the updates I'm planning for the WFIRST PSFs.  (And by "I'm planning" I mean "I already got all the inputs set up, so the user interface is the bit that I'm stuck on".)

As you may recall, currently getPSF() returns a dict containing PSFs for each SCA.  These PSFs will be either ChromaticOpticalPSFs or, if an effective wavelength was specified, an OpticalPSF.  Use of `approximate_struts` means it'll use 6 radially-oriented equally-spaced struts rather than the full pupil plane image (ignoring the variation in pupil plane image for long- and short-wavelength bands).  We also only used the Zernike info for the center of each SCA.   Because of the assumptions here -- no variation within an SCA and no variation of pupil plane image with band -- we really only needed to return a dict of PSFs indexed by SCA.

But now I want to enable the following:
* Different pupil plane image for long- and short-wavelength bands (only relevant if `approximate_struts=False`).
* variation of Zernikes across an SCA

[...]

PR #1002, opened 2019-01-08
#919 - WFIRST module updates to cycle 7

rmandelb:
This PR has some very long-awaited updates to the WFIRST module, which Troxel has already been using off on branch #919.  They get the WCS and PSF up to Cycle 7 specifications with some added functionality, and a simplification of the PSF instantiation routine.  Some helper routines that weren't being used in practice got removed.

[This comment](https://github.com/GalSim-developers/GalSim/issues/919#issuecomment-347320054) has a list at the bottom of things that got punted for later.

I was a bit aggressive with API changes without deprecating stuff, but in practice I know that Troxel's code respects the new getPSF() already, and there are not tons of users for this submodule, so I thought it might be OK.


galsim/wfirst/wfirst_psfs.py
```diff
@@ -161,7 +215,10 @@ def getPSF(SCAs=None, approximate_struts=False, n_waves=None, extra_aberrations=
 
             # In this case, we need to pad the edges of the pupil plane image, so we cannot just use
             # the stored file.
-            tmp_pupil_plane_im = galsim.fits.read(galsim.wfirst.pupil_plane_file)
+            if pupil_plane_type == 'long':
+                tmp_pupil_plane_im = galsim.fits.read(galsim.wfirst.pupil_plane_file_longwave)
```
rmjarvis: Not covered.  This is the high_accuracy one, so probably don't want to actually run this in the travis runs.  Maybe can use the same trick to do an assert_raises check that will hit this along the way.

galsim/wfirst/wfirst_psfs.py
```diff
@@ -172,269 +229,114 @@ def getPSF(SCAs=None, approximate_struts=False, n_waves=None, extra_aberrations=
...
-        if logger: logger.debug('   ... SCA %d'%SCA)
-        if wavelength is None:
-            if approximate_struts:
-                PSF = galsim.ChromaticOpticalPSF(
-                    lam=zemax_wavelength,
-                    diam=galsim.wfirst.diameter, aberrations=use_aberrations,
-                    obscuration=galsim.wfirst.obscuration, nstruts=6,
-                    oversampling=oversampling, gsparams=gsparams)
+            if pupil_plane_type == 'long':
+                pupil_plane_im = galsim.wfirst.pupil_plane_file_longwave
```
rmjarvis: Also not covered.  This one is the fastest case, so we should probably actually check it somehow.  Maybe as simple as just checking that the pupil planes generated for Z and F are different.

galsim/wfirst/wfirst_psfs.py
```diff
@@ -172,269 +229,114 @@ def getPSF(SCAs=None, approximate_struts=False, n_waves=None, extra_aberrations=
...
-        PSF_dict[SCA]=PSF
+    # Start reading in the aberrations for that SCA
+    if logger: logger.debug('Beginning to get the PSF aberrations for SCA %d.'%SCA)
+    aberrations, x_pos, y_pos = _read_aberrations(SCA)
+    # Do bilinear interpolation, unless we're exactly at the center (default).
+    use_aberrations = _interp_aberrations_bilinear(aberrations, x_pos, y_pos, SCA_pos)
 
-    return PSF_dict
+    if extra_aberrations is not None:
+        use_aberrations += extra_aberrations
```
rmjarvis: Not covered.  Should add a test that uses this somehow.

galsim/wfirst/wfirst_wcs.py
```diff
@@ -299,6 +249,71 @@ def getWCS(world_pos, PA=None, date=None, SCAs=None, PA_is_FPA=False):
...
+    @param PA_is_FPA If True, then the position angle that was provided was the PA of the focal
+                     plane array, not the observatory. [default: False]
+    @param tol       Tolerance for errors due to distortions, as a galsim.Angle. 
+                     [default: 0.5*galsim.arcsec]
+    @returns a CelestialCoord object indicating the center of the focal plane array.
+    """
+    if not isinstance(SCA, int):
+        raise TypeError("Must pass in an int corresponding to the SCA")
+    if not isinstance(tol, galsim.Angle):
+        raise TypeError("tol must be a galsim.Angle")
```
rmjarvis: These two aren't covered.

matroxel:
I've changed my code to work with this new api already (unless there were further changes in the last few weeks).

rmandelb:
@matroxel - there  were further changes in the past few days, but I specifically looked at your code to make sure it wasn't doing one of the things I was disabling (specifically, getting multiple PSFs simultaneously using getPSF()) - the call to getPSF() appeared to only be requesting a single object, so that change should not affect you.

@rmjarvis - really sorry, I meant to mention (and failed) that I was lazily planning to use the coverage report after making the PR to find the changed bits that need better unit tests.  (Is there a way to do that before the PR?)  I realized it was inadequate but I wanted codecov to tell me what to do rather than trying to figure it out myself.  ;)  Sorry for wasting your time on that bit - now that I know where to look, I'll address that in the coming days, in addition to your other comments.

rmjarvis:
> Sorry for wasting your time on that bit 

No worries.  This was a fine task to do during a telecon that I didn't really need to pay attention to.

>  (Is there a way to do that before the PR?) 

Not easily.  That's usually what I do too.  Once you get that, a neat trick is to add lines like `raise 1`, `raise 2`, etc. where the coverage was missing.  Then add tests until you hit the TypeError for the invalid raise, at which point remove it and make sure the test still works.  Once all the invalid raises are gone, push and see if you missed anything.  It works for everything except `assert_raises(TypeError)` tests.  (For those, I often put in `raise KeyError(3)` instead.)

rmandelb:
@rmjarvis - thanks for the review!  All your comments have been addressed, and we're back to full test coverage.

rmandelb:
Thanks for the approval!  I will plan to merge this on the weekend unless there are objections from anybody.

rmandelb:
@rmjarvis - do you understand why GitHub is offering to "Rebase and merge" rather than just merge?  I was going to go ahead but I didn't fully understand why that was necessary so I thought I'd check (to avoid breaking everything!)

rmjarvis:
Some time ago, I switched it to make this our preferred way to merge.  It makes for a nicer history.  (Mostly linear, rather than a complex web of joins and splits.)  So feel free to click it.  You won't break anything.  :)

If you want to see the difference before and after, do 
```
git log --graph --oneline
```
and page down a ways.  It will be obvious when you cross over to the time before "rebase and merge".

rmandelb:
Great, thanks!

rmandelb:
@matroxel - please start using master branch rather than #919 .

PR #1005, opened 2019-02-18
Add VonKarman r0_500 keyword.

jmeyers314:
Similar to`Kolmogorov`, allow either `r0` or `r0_500` to be specified for `VonKarman`.


jmeyers314:
Cool new draft PR feature from github.  Sadly, doesn't appear that travis-ci is triggered until PR is marked ready for review (at least yet).

rmjarvis:
> Sadly, doesn't appear that travis-ci is triggered until PR is marked ready for review (at least yet).

:(  I was thinking that was going to be the primary benefit of it.  That I could fix up my coverage and then mark when it was ready for review.  Even worse, it looks like it's still not running the Travis tests even after you marked it ready.  Probably need to push a dummy commit to bump it.

jmeyers314:
Actually, looks like travis has still not been triggered after marking PR ready for review...  I'm not sure what's going on, or even how to manually trigger... :(

jmeyers314:
> :( I was thinking that was going to be the primary benefit of it. That I could fix up my coverage and then mark when it was ready for review.

Exactly!  Would have been nice.  Hopefully they're working on it (it's only a 4-day old feature at the moment).  Somewhere there's a joke here about missing coverage analysis for a new feature, but I can't quite make it work...

> Probably need to push a dummy commit to bump it. 

Looks like that did the trick.

galsim/vonkarman.py
```diff
@@ -82,24 +86,35 @@ class VonKarman(GSObject):
     @param gsparams          An optional GSParams argument.  See the docstring for GSParams for
                              details. [default: None]
     """
-    _req_params = { "lam" : float, "r0" : float }
-    _opt_params = { "L0" : float, "flux" : float, "scale_unit" : str, "do_delta" : bool }
-    _single_params = []
+    _req_params = { "lam" : float }
+    _opt_params = { "r0" : float, "r0_500" : float, "L0" : float, "flux" : float,
```
rmjarvis: If r0 and r0_500 are in single_params, then they don't need to be in opt_params.  Not actually sure if it's a problem to have them in both places, but it's at least unnecessary.

Issue #992, opened 2018-11-15
Add wfirst persistence


chienhaolin:
Our previous implementation of the wfirst persistence effect is based on the general applyPersistence, which applies constant weighting coefficients that are independent of illumination level on previous images. The new implementation uses the fermi model to describe the dependence of persistence on x (illumination level) and t (time):  A* (x/x0)^a * (t/1000.)^(-r) / (.exp( -(x-x0)/dx ) +1. ). In addition, persistence effect is put after applyNonlinearity and applyIPC.

rmjarvis:
Closed via #1011.  Thanks Chien-Hao.

PR #1011, opened 2019-03-18
#992

chienhaolin:
This pull request adds the feature of more complex wfirst persistence model.

The edits are mainly in the following three files:

- galsim/utilities.py (add the dynamic array LinkedList)
- galsim/wfirst/__init__.py (add parameters and the fermi function)
- galsim/wfirst/wfirst_detectors.py (implement the persistence model)


chienhaolin:
I will close the pull request for now and update the implementation of galsim.wfirst.addPersistence to better work with wfirst_imsim.

rmjarvis:
For future reference, it is ok to work more on a branch that is already in PR.  Just add a note that you aren't ready for code review yet.  And then another one when you are ready for review.  This might be easier than closing this PR and later making a new one.

chienhaolin:
I've updated the implementation (with LinkedList removed) and just reopened this PR. It is now ready for code review.

rmandelb:
@chienhaolin - the pull request has some failing checks due to insufficient unit test coverage.  Can you please add test coverage for the lines that are not covered (you can see the coverage report linked above) until the checks pass?  If you need help understanding the coverage report, we can talk in person later today.

galsim/wfirst/wfirst_detectors.py
```diff
@@ -83,28 +83,53 @@ def applyIPC(img, edge_treatment='extend', fill_value=None):
...
+        n_exp = min(len(prev_exposures), galsim.wfirst.max_exps)
+        for i in range(n_exp):
+            ### The length of time where slews and pixel resets happen should be specified.
+            ### Now it is simply assumed to be 0 until we get more information.
+            img += galsim.wfirst.fermi_linear(prev_exposures[i].array)/(1.+ 2*i)
 
-    n_exp = min(len(prev_exposures),len(galsim.wfirst.persistence_coefficients))
-    img.applyPersistence(prev_exposures[:n_exp],galsim.wfirst.persistence_coefficients[:n_exp])
+    else:
+        raise ValueError("applyPersistence only accepts 'linear' or 'fermi' methods.")
```
rmjarvis: This should be GalSimValueError.  Also, you need a test to make sure it gets raised appropriately when an invalid `method` is provided.  (`assert_raises`)

galsim/wfirst/wfirst_detectors.py
```diff
@@ -83,28 +83,53 @@ def applyIPC(img, edge_treatment='extend', fill_value=None):
...
+
     if not hasattr(prev_exposures,'__iter__'):
-        raise TypeError("In wfirst.applyPersistence, prev_exposures must be a list of Image "
-                        "instances")
+        raise TypeError("In wfirst.applyPersistence, prev_exposures must be a list of Image instances")
+
+    if method == 'linear':
+
+        n_exp = min(len(prev_exposures),len(galsim.wfirst.persistence_coefficients))
+        img.applyPersistence(prev_exposures[:n_exp],galsim.wfirst.persistence_coefficients[:n_exp])
```
rmjarvis: You lost coverage of this code.  Need tests of both linear and fermi options.

galsim/wfirst/wfirst_detectors.py
```diff
@@ -83,28 +83,53 @@ def applyIPC(img, edge_treatment='extend', fill_value=None):
...
+            ### Now it is simply assumed to be 0 until we get more information.
+            img += galsim.wfirst.fermi_linear(prev_exposures[i].array)/(1.+ 2*i)
 
-    n_exp = min(len(prev_exposures),len(galsim.wfirst.persistence_coefficients))
-    img.applyPersistence(prev_exposures[:n_exp],galsim.wfirst.persistence_coefficients[:n_exp])
+    else:
+        raise ValueError("applyPersistence only accepts 'linear' or 'fermi' methods.")
 
-def allDetectorEffects(img, prev_exposures=(), rng=None, exptime=default_exptime):
+def allDetectorEffects(img, prev_exposures=[], rng=None, exptime=default_exptime):
```
rmjarvis: Revert this change.  You should never have a list be a default parameter for a function in python.  cf [this](https://docs.python-guide.org/writing/gotchas/) or [this](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html) for why this is bad.

galsim/wfirst/wfirst_detectors.py
```diff
@@ -149,20 +173,19 @@ def allDetectorEffects(img, prev_exposures=(), rng=None, exptime=default_exptime
...
-
     # Nonlinearity (use WFIRST routine).
     applyNonlinearity(img)
 
     # IPC (use WFIRST routine).
     applyIPC(img)
 
+    # Persistence (use WFIRST H4RG-lo fermi model)
+    prev_exposures = list(prev_exposures)
+    applyPersistence(img, prev_exposures, method='fermi')
```
rmjarvis: I thought persistence worked like an extra dark current.  Why did this move?  Shouldn't it be applied prior to NL and IPC?

tests/test_wfirst.py
```diff
@@ -541,9 +541,9 @@ def test_wfirst_detectors():
     assert im_2.wcs == im_1.wcs
     assert im_2.dtype == im_1.dtype
     assert im_2.bounds == im_1.bounds
-    np.testing.assert_array_equal(
-        im_2.array, im_1.array,
-        err_msg='Persistence results depend on function used.')
+    #np.testing.assert_array_equal(
+    #    im_2.array, im_1.array,
+    #    err_msg='Persistence results depend on function used.')
```
rmjarvis: Probably should revert this change and use the linear model above so it matches what happens to im_1.

Then add an additional unit tests that actually check that the fermi persistence model is doing what you think it should be doing.  Right now, the only check is that there are no catastrophic errors when you run the function.  Nothing here checks that the math is right.

rmandelb:
@chienhaolin - I wanted to check in about the status about this?

chienhaolin:
@rmandelb @rmjarvis - The pull request is ready for code review again. Could you please go over it and see whether more changes are needed? Thanks!

CHANGELOG.md
```diff
@@ -19,6 +19,11 @@ New Features
   pupil plane image for shorter- and longer-wavelength bands.  (#919)
 - Enabled Zernikes up to 22 (previously 11) in WFIRST PSFs, and added dependence on position
   within the SCA. (#919)
+- Added spline as LookupTable2D interpolant. (#982)
```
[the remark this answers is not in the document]
chienhaolin: I guess this is because my local CHANGELOG was outdated. The lines for #982 are removed.

galsim/wfirst/wfirst_detectors.py
```diff
@@ -83,26 +82,79 @@ def applyIPC(img, edge_treatment='extend', fill_value=None):
...
+
+    A, x0, dx, a, r, half_well = galsim.wfirst.persistence_fermi_parameters
+    ps    = A* (    x    /x0)**a * (t/1000.)**(-r)/(np.exp( -(x-x0)/dx) +1.)
+    ps_hf = A* (half_well/x0)**a * (t/1000.)**(-r)/(np.exp( -(x-x0)/dx) +1.)
+
+    mask1 = x > half_well
+    mask2 = (x > 0.) & (x <= half_well)
+
+    y[mask1] += ps[mask1]
+    y[mask2] += ps_hf[mask2]
```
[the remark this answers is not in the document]
chienhaolin: It is actually a bug. Thanks for spotting it! I've updated the implementation and checked the math of it to make sure the output value is right.

galsim/wfirst/wfirst_detectors.py
```diff
@@ -138,24 +189,22 @@ def allDetectorEffects(img, prev_exposures=(), rng=None, exptime=default_exptime
...
 
-    # Quantize: have an integer number of photons in every pixel after inclusion of sky noise.
-    img.quantize()
-
     # Reciprocity failure (use WFIRST routine, with the supplied exposure time).
     addReciprocityFailure(img, exptime=exptime)
 
+    # Quantize.
+    img.quantize()
+
```
rmjarvis: I think this is wrong.  Didn't we decide that it goes before reciprocity failure, and that if it's there, the poisson noise already makes integers, so we don't need to do anything else?

tests/test_wfirst.py
```diff
@@ -536,15 +536,24 @@ def test_wfirst_detectors():
...
+    galsim.wfirst.applyPersistence(im_2, im_list, method='linear') #check the linear method
     assert im_2.scale == im_1.scale
     assert im_2.wcs == im_1.wcs
     assert im_2.dtype == im_1.dtype
     assert im_2.bounds == im_1.bounds
     np.testing.assert_array_equal(
-        im_2.array, im_1.array,
-        err_msg='Persistence results depend on function used.')
+       im_2.array, im_1.array,
+       err_msg='Persistence results depend on function used.')
```
rmjarvis: indentation error (we use 4 spaces, not 3)

tests/test_wfirst.py
```diff
@@ -536,15 +536,24 @@ def test_wfirst_detectors():
...
+
+    im_unit = galsim.Image(np.ones((2,2)), copy=True)
+    im_f = im_unit*0.0
+    im_f_list = [im_unit*1.E3, im_unit*1.E5]
+    galsim.wfirst.applyPersistence(im_f, im_f_list, method='fermi') #check fermi method
+    #Check the functionality of the fermi method by comparing with pre-calculated values.
+    assert np.allclose( im_f.array, np.ones((2,2))*8.30515, rtol=1.E-06 ), 'Error in Fermi persistence model'
```
rmjarvis: Explain where this 8.30515 value comes from.

tests/test_wfirst.py
```diff
@@ -536,15 +536,24 @@ def test_wfirst_detectors():
...
+
     assert_raises(TypeError, galsim.wfirst.applyPersistence, im_2, im0)
+    assert_raises(galsim.GalSimValueError, galsim.wfirst.applyPersistence, im_2, im_list, method='wrong method')
```
rmjarvis: Also need a test that running `applyPersistence` with no specified `method` is the same as `'fermi'`, not `'linear'`.

galsim/wfirst/wfirst_detectors.py
```diff
@@ -83,26 +82,80 @@ def applyIPC(img, edge_treatment='extend', fill_value=None):
...
+        n_exp = min(len(prev_exposures),len(galsim.wfirst.persistence_coefficients))
+        img.applyPersistence(prev_exposures[:n_exp],galsim.wfirst.persistence_coefficients[:n_exp])
+
+    elif method == 'fermi':
+
+        n_exp = len(prev_exposures)
+        for i in range(n_exp):
+            ### The length of time where slews and pixel resets takes should be specified.
+            ### Now it is simply assumed to be 0 until we get more information.
+            img += fermi_linear(prev_exposures[i].array)/(1.+ 2*i)
```
rmjarvis: While this works, it would be slightly clearer imo to write
```
img.array += fermi_linear(...
```
I was actually surprised that image += array was valid, but it's a byproduct of our code to allow image += float in addition to image += image.

galsim/wfirst/wfirst_detectors.py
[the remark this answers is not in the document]
rmjarvis: You should add a test that would have caught the difference between these two versions.  The test you currently have is just on an input image of all ones.  Not one that passes over the half-well line.  Should make sure the code works properly both above and below half-well.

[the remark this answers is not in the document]
rmjarvis: Or, never mind, it has 1e3 and 1e5, not ones.  But still I think should have a few below and one just above half-well to check linearity below half-well and continuity across half-well.  Then some higher ones to check the fermi part.

galsim/wfirst/wfirst_detectors.py
rmandelb: c.f. discussion in https://github.com/GalSim-developers/GalSim/pull/1022/files

galsim/wfirst/wfirst_detectors.py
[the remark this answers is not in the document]
chienhaolin: I added more points in the test function. Now it has 1e3, 1e4, 4e4, 4.99e4, 5.01e4, 1e5, 1e6.

galsim/wfirst/wfirst_detectors.py
chienhaolin: It's moved to the front of the reciprocity failure.

tests/test_wfirst.py
chienhaolin: The value is pre-calculated using my local function that reproduces the reference curve provided by Chris. 
avg_persistence_602
persistence_model

This value is updated since we changed the illumination level of previous exposures in the test.
Here's the log that shows how the this value is computed:

exposure_id /  mid-time /  illumination  / persistence 
     0        70.125       1e+03       3.04e-01  
     1       210.375       1e+04       1.01e+00  
     2       350.625       4e+04       2.43e+00  
     3       490.875       5e+04       2.16e+00  
     4       631.125       5e+04       1.69e+00  
     5       771.375       1e+05       2.18e+00  
     6       911.625       1e+06       2.97e+00  
total persistence 12.748214217103115

galsim/wfirst/wfirst_detectors.py
chienhaolin: This line is updated as suggested.

In addition, I actually make an approximation here in this line. Since the persistence current is inversely proportional to time, the integrated persistence signal over the exposure time is
image, where t0 is the starting time of a previous exposure since reset and t_exp is the exposure time (140.25 s). The integrated persistence signal is sensitive t0, but the timeline of observation is  currently not available. So I take the persistence current at the mid-time of a exposure as the average persistence current and approximate the integrated persistence signal as 
image; i.e., I'm approximating the 1/t curve locally with a straight line. 
time_integrated
For more distant previous exposures (t0 > 140s), it doesn't make much difference. For the last one previous exposure, t0 (the reset time) is likely to be somewhere between 15-30s. This approximation can only provide a rough estimation for the last exposure. More precise calculation would require the whole timeline of observations.

tests/test_wfirst.py
chienhaolin: A test function is added.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ New Features
   pupil plane image for shorter- and longer-wavelength bands.  (#919)
 - Enabled Zernikes up to 22 (previously 11) in WFIRST PSFs, and added dependence on position
   within the SCA. (#919)
+- Added wfirst fermi persistence model. (#992)
```
rmandelb:
```suggestion
- Added WFIRST fermi persistence model. (#992)
```

galsim/wfirst/wfirst_detectors.py
```diff
@@ -83,26 +82,80 @@ def applyIPC(img, edge_treatment='extend', fill_value=None):
...
+
+    else:
+        raise galsim.GalSimValueError("applyPersistence only accepts 'linear' or 'fermi' methods, got", method)
+
+def fermi_linear(x, t):
+    """
+    The fermi model for persistence: A* (x/x0)**a * (t/1000.)**(-r) / (exp( -(x-x0)/dx ) +1. )
+    For influence level below the half well, the persistence is linear in x.
+
+    @param x        Array of pxiel influence levels in unit of electron counts.
```
rmandelb:
```suggestion
    @param x        Array of pixel influence levels in unit of electron counts.
```

tests/test_wfirst.py
```diff
@@ -536,15 +536,31 @@ def test_wfirst_detectors():
...
+    illuminatin_list = [1.E3, 1.E4, 4.E4, 4.99E4, 5.01E4, 1.E5, 1.0E6]
+    im_f_list = [x*im_unit for x in illuminatin_list]
+
+    galsim.wfirst.applyPersistence(im_f, im_f_list, method='fermi') #check fermi method
+    galsim.wfirst.applyPersistence(im_f1, im_f_list)
+
+    #Check the functionality of the fermi method by comparing with pre-calculated values.
+    assert np.allclose( im_f.array, np.ones((2,2))*12.74821, rtol=1.E-06 ), 'Error in Fermi persistence model'
+    np.testing.assert_array_equal(im_f, im_f1,
+        err_msg='The defaulf method of wfirst.applyPersistence is not fermi.')
```
rmandelb:
```suggestion
        err_msg='The default method of wfirst.applyPersistence is not fermi.')
```

tests/test_wfirst.py
```diff
@@ -536,15 +536,31 @@ def test_wfirst_detectors():
...
+    im_unit = galsim.Image(np.ones((2,2)), copy=True)
+    im_f = im_unit*0.0
+    im_f1 = im_unit*0.0
+    illuminatin_list = [1.E3, 1.E4, 4.E4, 4.99E4, 5.01E4, 1.E5, 1.0E6]
+    im_f_list = [x*im_unit for x in illuminatin_list]
+
+    galsim.wfirst.applyPersistence(im_f, im_f_list, method='fermi') #check fermi method
+    galsim.wfirst.applyPersistence(im_f1, im_f_list)
+
+    #Check the functionality of the fermi method by comparing with pre-calculated values.
```
rmandelb: The comment should say what is the origin of the pre-calculated values, because if the routine changes (for example due to updated coefficients) then whoever changes it needs to know how to precalculate values to update the unit test.

rmjarvis: Even better would be to do the calculations here in the test suite rather than externally.  With inline comments about what formula you are using to do so and where that formula comes from.

chienhaolin: I  added the calculation in the test function. No hard coded value is needed.

galsim/wfirst/wfirst_detectors.py
```diff
@@ -125,7 +125,7 @@ def applyPersistence(img, prev_exposures, method='fermi'):
             ### The slew/settle time and the reset time should be specified.
             ### Now we simply assume them as 0 and take the persitence current at the mid-time of
             ### exposures as the average persistence until we get more information about the observation timeline.
-            img.array += fermi_linear(prev_exposures[i].array, 0.5*(1+i)*galsim.wfirst.exptime)*galsim.wfirst.exptime
+            img._array += fermi_linear(prev_exposures[i].array, (0.5+i)*galsim.wfirst.exptime)*galsim.wfirst.exptime
```
rmjarvis: I guess this is to get around that array is a property.  I think better than using the nominally private attribute would be to write
```
im.array[:,:] += fermi...
```

tests/test_wfirst.py
```diff
@@ -536,15 +536,43 @@ def test_wfirst_detectors():
...
         err_msg='Persistence results depend on function used.')
+
+    im_unit = galsim.Image(np.ones((2,2)), copy=True)
+    im_f = im_unit*0.0
+    im_f1 = im_unit*0.0
+    illuminatin_list = [1.E3, 1.E4, 4.E4, 4.99E4, 5.01E4, 1.E5, 1.0E6]
+    im_f_list = [x*im_unit for x in illuminatin_list]
+
+    galsim.wfirst.applyPersistence(im_f, im_f_list, method='fermi') #check fermi method
+    galsim.wfirst.applyPersistence(im_f1, im_f_list)
```
rmjarvis: Suggest moving this line down to just above where you do the check with it to keep the two things separate.

I.e. first check that method='fermi' works correctly by comparing to your explicit calculation.  (Thanks for that btw.  That looks good.)

And then check that omitting the method is equivalent.

galsim/wfirst/wfirst_detectors.py
chienhaolin: Thanks for the suggestion. This is much better than using the hidden attribute.

chienhaolin:
Thank you, @rmandelb and @rmjarvis for all the review and suggestions!

Issue #1006, opened 2019-02-19
Make shared atmospheric screens safer for multiprocessing


jmeyers314:
In [ImSim](https://github.com/LSSTDESC/imSim), the LSST DESC currently uses a *nix copy-on-write + fork() trick to effectively put read-only atmospheric screens into shared memory where multiple processes can access them.  I've done some experimenting over the past few days and determined that this trick breaks when switching the process initialization from fork() to fork()+exec().  Unfortunately, as pointed out in #998, fork()+exec() initialization is required for subprocess thread safety, which includes any future OpenMP optimization inside GalSim, for example.

I think it's possible to solve this by modifying how atmospheric screens are stored in GalSim.  I've found that using the multiprocessing.Array and RawArray abstractions do permit shared memory access under fork()+exec() subprocess initialization, and that it's relatively straightforward to cast either of these to a numpy ndarray through the buffer interface (See [this blog post](https://research.wmz.ninja/articles/2018/03/on-sharing-large-arrays-when-using-pythons-multiprocessing.html) for an example).  

So we could either 1) make AtmosphericScreen always use a multiprocessing Array under the hood, even when there's no multiprocessing being done, or 2) add an alternative AtmosphericScreen constructor that accepts an externally-generated RawArray (or maybe numpy array that wraps a RawArray?, I'd have to experiment...), i.e., enable the user to handle shared memory outside of GalSim.  

I think I prefer option 1) if I can get it to work (not yet clear...), but either way, some GalSim development seems to be required (sans monkey-patching), which is why I'm raising this issue here instead of in ImSim.

As a bonus, I think this work could also enable generating multiple screens concurrently, which would speed up the startup times for jobs that rely on large screens.

rmjarvis:
If there is no runtime overhead, option 1 sounds good.  But if not, we would probably want this to be optional.  E.g. an optional kwarg to the constructor.

jmeyers314:
The following script demonstrates a _mostly_ working example of how I think this could work:

```python
import os
import multiprocessing as mp
import numpy as np
import ctypes
import time
import hashlib
from galsim.utilities import lazy_property

# Global variable to hold shared memory variables.  We'll form key-value pairs here 
# linking Atm objects to their particular shared memory object.
_share = {}


class Atm:
    def __init__(self, i, shape):
        self.i = i
        self.shape = shape

        # Setup shared state variable.  We'll fill it in lazily as needed later
        size = 1
        for s in shape:
            size *= s
[...]

PR #1021, opened 2019-04-01
Use shared memory for AtmosphericScreen

jmeyers314:
The PR addresses #1006.  I have:

• Explicitly implemented shared memory for atmospheric screens using `multiprocessing.RawArray` and `.RawValue` classes.  The main impact going forward is that we would no longer need to rely on esoteric *nix fork() + copy-on-write behavior to effect shared memory in applications like ImSim.  A secondary impact is that when using the "spawn" launch context, subprocesses should now be able to safely use OpenMP (I hope, this is untested).
• Added an `mp_context` kwarg to `AtmosphericScreen` so that the user can control what multiprocessing launch context to use.  The default of `None` falls back to the python stdlib default, which is "fork".
• Added a couple of helper functions to ease use of shared memory via `multiprocessing.Pool` objects.  See `AtmosphericScreen` docstring for details.
• Enabled parallel instantiation of multiple screens in a `PhaseScreenList`, which should reduce some of the startup overhead in, e.g., ImSim by a factor of ~nlayers.  (Though, this is at the cost of temporarily requiring more memory during screen instantiation).

The tricky bits in getting this working were in using locks correctly when the shared memory is written to, and implementing some not-automatic-reference counting of the shared memory so that it's released when appropriate (and critically, released at the end of each applicable unit test.)

### DIFF
--- a/galsim/phase_psf.py
+++ b/galsim/phase_psf.py
@@ -827,14 +827,26 @@ class PhaseScreenList(object):
                 pass
         self._update_attrs()
 
-    def instantiate(self, _bar=None, **kwargs):
+    def instantiate(self, pool=None, _bar=None, **kwargs):
+        """Instantiate the screens in this PhaseScreenList.
+
+        @param pool      A multiprocessing.Pool object to use to instantiate screens in parallel.
+        @param **kwargs  Keyword arguments to forward to screen.instantiate().
+        """
         for layer in self:
+            results = []
             try:
-                layer.instantiate(**kwargs)
+                if pool is not None:
+                    results.append(pool.apply_async(layer.instantiate, kwds=kwargs))
+                else:
+                    layer.instantiate(**kwargs)
             except AttributeError:
                 pass
             if _bar:  # pragma: no cover
                 _bar.update()
+        if pool is not None:
+            for r in results:
+                r.wait()
 
     def _delayCalculation(self, psf):
         """Add psf to delayed calculation list."""
--- a/galsim/phase_psf.py
+++ b/galsim/phase_psf.py
@@ -835,8 +835,8 @@ class PhaseScreenList(object):
         """
         if pool is not None:
             import sys
-            assert sys.version_info >= (3,4),
-                "instantiating PhaseScreenList with pool requires python version >= 3.4"
+            assert_str = "instantiating PhaseScreenList with pool requires python version >= 3.4"
+            assert sys.version_info >= (3,4), assert_str
         for layer in self:
             results = []
             try:
--- a/galsim/phase_psf.py
+++ b/galsim/phase_psf.py
@@ -833,6 +833,10 @@ class PhaseScreenList(object):
         @param pool      A multiprocessing.Pool object to use to instantiate screens in parallel.
         @param **kwargs  Keyword arguments to forward to screen.instantiate().
         """
+        if pool is not None:
+            import sys
+            assert sys.version_info >= (3,4),
+                "instantiating PhaseScreenList with pool requires python version >= 3.4"
         for layer in self:
             results = []
             try:
--- a/galsim/phase_psf.py
+++ b/galsim/phase_psf.py
@@ -65,6 +65,7 @@ Atmosphere
   Convenience function to quickly assemble multiple AtmosphericScreens into a PhaseScreenList.
 """
 
+import sys
 from past.builtins import basestring
 from itertools import chain
 from builtins import range
@@ -834,7 +835,6 @@ class PhaseScreenList(object):
         @param **kwargs  Keyword arguments to forward to screen.instantiate().
         """
         if pool is not None:
-            import sys
             assert_str = "instantiating PhaseScreenList with pool requires python version >= 3.4"
             assert sys.version_info >= (3,4), assert_str
 
--- a/galsim/phase_psf.py
+++ b/galsim/phase_psf.py
@@ -842,7 +842,7 @@ class PhaseScreenList(object):
             for layer in self:
                 try:
                     results.append(pool.apply_async(layer.instantiate, kwds=kwargs))
-                except AttributeError: # why did I need this?
+                except AttributeError:  # OpticalScreen has no instantiate method
                     pass
                 if _bar:  # pragma: no cover
                     _bar.update()
--- a/galsim/phase_psf.py
+++ b/galsim/phase_psf.py
@@ -837,20 +837,25 @@ class PhaseScreenList(object):
             import sys
             assert_str = "instantiating PhaseScreenList with pool requires python version >= 3.4"
             assert sys.version_info >= (3,4), assert_str
-        for layer in self:
+
             results = []
-            try:
-                if pool is not None:
+            for layer in self:
+                try:
                     results.append(pool.apply_async(layer.instantiate, kwds=kwargs))
-                else:
-                    layer.instantiate(**kwargs)
-            except AttributeError:
-                pass
-            if _bar:  # pragma: no cover
-                _bar.update()
-        if pool is not None:
+                except AttributeError: # why did I need this?
+                    pass
+                if _bar:  # pragma: no cover
+                    _bar.update()
             for r in results:
                 r.wait()
+        else:
+            for layer in self:
+                try:
+                    layer.instantiate(**kwargs)
+                except AttributeError:
+                    pass
+                if _bar:
+                    _bar.update()
 
     def _delayCalculation(self, psf):
         """Add psf to delayed calculation list."""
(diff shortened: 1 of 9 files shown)


jmeyers314:
Looks like the new features added here depend on python3.4+, so in particular won't work with python2.7.  I think that's probably fine, actually; not every feature needs to be supported by every "supported" version of python, especially things that are actually impossible, which I suspect this may be.  (I will go back and make sure the *previously* available features continue to work with py2.7, though).

This makes me wonder if we need something akin to the `deprecated.depr` function to mark such features.  Maybe call it `requires()`?

Also, any thoughts on transitioning our python2.7 support to bugs only?  I.e., make it policy that new features are not necessarily expected to work in python2.7 (though if they do, that's still great of course).

rmjarvis:
I think it's fine to start having py3-only features.  I think most people are now pretty well migrated to python 3 at this point.  We should probably still support the basic things not breaking in py2.7, at least until v3.0.  But I think it's fine to have lines like
```
assert sys.version_info >= (3,4), "This feature requires Python version >= 3.4"
```

jmeyers314:
I think this one's ready.  Note the force push on this branch to clean up some earlier git corruption I accidentally made.  If anybody happened to check this branch out in the last day or so, they may need to `git reset --hard origin/#1006` before re-fetching/-pulling.

galsim/phase_psf.py
```diff
@@ -827,14 +827,35 @@ def _reset(self):
...
-            if _bar:  # pragma: no cover
-                _bar.update()
+    def instantiate(self, pool=None, _bar=None, **kwargs):
+        """Instantiate the screens in this PhaseScreenList.
+
+        @param pool      A multiprocessing.Pool object to use to instantiate screens in parallel.
+        @param **kwargs  Keyword arguments to forward to screen.instantiate().
+        """
+        if pool is not None:
+            import sys
```
rmjarvis: This import is pretty lightweight, so it probably belongs at the top of the file (as per PEP8).  I don't mind having specialized imports inside functions or if clauses, but I think it's better form to move the easy/standard-lib ones up to the top.

galsim/phase_psf.py
```diff
@@ -827,14 +827,35 @@ def _reset(self):
...
+            assert_str = "instantiating PhaseScreenList with pool requires python version >= 3.4"
+            assert sys.version_info >= (3,4), assert_str
+
+            results = []
+            for layer in self:
+                try:
+                    results.append(pool.apply_async(layer.instantiate, kwds=kwargs))
```
rmjarvis: I think this function is available in Python 2.7.  https://docs.python.org/2/library/multiprocessing.html#module-multiprocessing.pool

What part of this requires python 3.4?

galsim/phase_screens.py
```diff
@@ -74,6 +99,73 @@ class AtmosphericScreen(object):
...
+    an atmosphere with 6 screens, each extending 819.2 m and with resolution of 10 cm will consume
+    3 GB of memory.  In contexts where both a realistic atmospheric PSF and high throughput via
+    multiprocessing are required, allocating this 3 GB of memory once in a shared memory space
+    accessible to each subprocess (as opposed to once per subprocess) is highly desireable.  We
+    provide a few functions here to enable such usage:
+
+        - The mp_context keyword argument to AtmosphericScreen.
+            This is used to indicate which multiprocessing process launching context will be used.
+            This is important for setting up the shared memory correctly.
+        - The galsim.phase_screens.initWorker() and initWorkerArgs functions.
```
rmjarvis: For parallelism, should write `initWorkerArgs()`.

galsim/phase_screens.py
```diff
@@ -156,14 +258,72 @@ def __init__(self, screen_size, screen_scale=None, altitude=0.0, r0_500=0.2, L0=
...
+
+    def __del__(self):
+        if sys.version_info >= (3,4):
+            if not hasattr(self, '_objDict'):  # for if __init__ raised exception
+                return
+
+            with self._objDict['lock']:
+                self._objDict['refcount'].value -= 1
+            if self._objDict['refcount'].value == 0:
+                del _GSScreenShare[self._shareKey]
```
rmjarvis: I feel like this del bit should also be inside the lock.  Isn't there a race potential for the refcount to be 0, but someone else is about to increment it?

galsim/phase_screens.py
```diff
@@ -284,6 +473,41 @@ def _random_screen(self):
...
+        self._objDict['x0'].value = tab2d.x0
+        self._objDict['y0'].value = tab2d.y0
+        self._objDict['xperiod'].value = tab2d.xperiod
+        self._objDict['yperiod'].value = tab2d.yperiod
+        # Need to reset this here in case we did a boiling update
+        self._tab2d = _LookupTable2D(
+            xshare, yshare, fshare, 'linear', 'wrap', 0.0,
+            x0=self._objDict['x0'].value, y0=self._objDict['y0'].value,
+            xperiod=self._objDict['xperiod'].value, yperiod=self._objDict['yperiod'].value
+        )
```
rmjarvis: Could avoid some code duplication between this and `_tab2d` below by removing this `self._tab2d = ` bit.  (Maybe also adding a `del self._tab2d` in case boiling update.)  I think the lazy_property will make it when needed in the main process as well as any sub-processes.  

Would this be inefficient?  Or any other problem?

galsim/phase_screens.py
```diff
@@ -131,6 +229,10 @@ def __init__(self, screen_size, screen_scale=None, altitude=0.0, r0_500=0.2, L0=
                 "Setting AtmosphericScreen time_step prohibited when alpha == 1.0.  "
                 "Did you mean to set time_step in makePSF or PhaseScreenPSF?",
                 alpha=alpha, time_step=time_step)
+        if (alpha != 1.0 and mp_context is not None):
+            raise GalSimIncompatibleValuesError(
+                "Shared memory use is not supported for time-evolving screens",
+                alpha=alpha, mp_context=mp_context)
```
rmjarvis: I think this should be a `GalSimNotImplementedError`.

tests/test_phase_psf.py
```diff
@@ -1081,6 +1091,121 @@ def test_withGSP():
...
+    else:
+        ctxs = [None]
+
+    for ctx in ctxs:
+        if ctx is None:
+            ctx = mp.get_context(None)
+        atm = galsim.Atmosphere(r0_500=r0_500, L0=L0, speed=spd, direction=dirn, altitude=alts,
+                                rng=rng.duplicate(),
+                                screen_size=screen_size, screen_scale=screen_scale,
+                                mp_context=ctx)
```
rmjarvis: Looks like you're only checking `mp_context` being a `multiprocessing.context.BaseContext`.  I think better to let `None` stay None, since Atmosphere will run `mp.get_context(None)` for you, and let at least one of the others bet the string version.  Say, leave "forkserver" as a string rather than calling `mp.get_context("forkserver")`.

galsim/phase_screens.py
jmeyers314: Right, I think moving the if and del inside the lock saves you from deallocating in process A while there's a copy happening in process B.  Maybe even use additional `with lock` blocks to give a potential copy an extra chance to increment the refcount.  If the del continues to win a del/copy race, the copy will still fail I guess, but without corrupting memory.  It's probably not possible to ensure that in a copy/delete race the copy always wins.  We could move the del to a garbage collect function, but that seems like more work and more user-responsibility than reasonable.

galsim/phase_screens.py
jmeyers314: Sounds good.  I actually tried deleting this line earlier, but put it back b/c of the boiling update.  I hadn't thought of `del self._tab2d`, which seems like a more elegant solution.

galsim/phase_psf.py
jmeyers314: This line doesn't require py3.4 per se, but I think using the downstream layer.instantiate in a pool but without the shared memory stuff will effectively fail.  I.e., it will create the screens in the subprocesses, but not propagate them back to the parent since w/o py3.4 it's not using shared memory.  (At least, I think that's what would happen, I haven't actually tried it).

I'll try out your suggestion to use the py2.7 versions of RawArray, etc though, in which case I agree this assert should be removed.

jmeyers314: This now works on py2.7 too.

galsim/phase_screens.py
jmeyers314: I think this is now reasonably race-condition resistant.

jmeyers314:
I think I've addressed all of your comments @rmjarvis .  Let me know if this looks ready to you.

rmjarvis:
I'm glad my idea for py2.7 worked.  I think the code looks quite a bit cleaner this way.

jmeyers314:
I added one more guard to avoid errors with boiling and shared memory.  I'll plan to merge this this afternoon unless I hear any objections.

Issue #1020, opened 2019-03-24
BUG integer division in power spectrum module


beckermr:
I think this line might be a bug

https://github.com/GalSim-developers/GalSim/blob/releases/2.1/galsim/lensing_ps.py#L494

It disallows grid spacings less than 1 arcsecond.

rmjarvis:
Yup.  Looks like.  I think this was a casualty of our attempt to convert all the integer division to work right in python 3 several years ago.  `kmax_factor` is an integer, but `grid_spacing` is not, so this should be a regular `/`, not `//`.

rmandelb:
Closed by #1022.

Issue #1019, opened 2019-03-21
a collection of issues from the WFIRST hackathon


rmandelb:
Here are a few small-ish things that came up at the WFIRST hackathon that should be easily dealt with:

- Some of the descriptions of functionality in `__init__.py` are out of step with what the routines now do  as of the latest updates.  Should do a consistency check.

- The effective area that is calculated is too large (does not account for struts).  We should use the number given in Cycle 7 documentation.  Could also consider using different effective area for each filter, based on the mask for certain filters.

I'll quickly fix these and make a PR in the coming days.

rmandelb:
Closed by #1022.

Issue #1016, opened 2019-03-19
Don't quantize after reciprocity failure


rmjarvis:
In our standard treatment of WFIRST detector effects, there is a call to `quantize` directly after the call to `applyReciprocityFailure`.  We discussed whether this was correct at the WFIRST hackathon this week, and we concluded that while it's not completely certain, we think it is more likely that this isn't correct.

Basically, putting quantize where it currently is would maybe be appropriate if the reciprocity failure was essentially a variable quantum efficiency, so it represented a chance of a photon to convert to a hole in the sensor.  But if this were the case, then it should be wavelength dependent, which it apparently isn't for WFIRST sensors.

Instead, it is probably a nonlinearity in the conversion of holes to voltage.  In this case, the resulting value should not be quantized yet.  It's a float value that then undergoes further effects until eventually being quantized by the ADC.  The upshot is that we should remove that call to quantize.  (Or maybe move it to right after the sky noise to make sure that we have an integer number of photons in every pixel.)

rmandelb:
Closed by #1022.

PR #1022, opened 2019-04-04
#1019

rmandelb:
This pull request covers a few small-ish issues: #1016 #1019 #1020  (I started hoping to do #1019 and then added a few other small ones; sorry it's a grab bag).


rmandelb:
@rmjarvis - might you be able to take a look at this in the coming few weeks?  I hope it'll be a quick fix for a few issues.

examples/demo13.py
```diff
@@ -352,6 +352,10 @@ def main(argv):
         # First, we include the expected Poisson noise:
         final_image.addNoise(poisson_noise)
 
+        # At this point in the image generation process, an integer number of photons gets
+        # detected, hence we have to round the pixel values to integers:
+        final_image.quantize()
```
rmjarvis: I believe at this point, the image is already quantized unless any of the pre-noise values were > 2^30.  That's when our Poisson implementation switches over to the Gaussian approximation, which won't necessarily produce integers.
So probably we don't need this line.  Or if you want to keep it, maybe add some commentary about how it's not really necessary for non-saturated pixel values.

rmandelb: Yes - I confirmed this directly as well, so have removed this step and added a comment.

rmandelb:
Thanks for approving, @rmjarvis !  I will  plan to merge on Friday unless anybody comments that they want more time to look this over.  

@beckermr - note that a fix for the bug you reported in #1020 is included in this PR.

beckermr:
Thanks @rmandelb !

rmandelb:
Merging #1022.

Issue #1024, opened 2019-05-03
FitsWCS is reading all the data not just the header


esheldon:
When using `FitsWCS` it is reading all the data rather than just the header.  For example running on some DES images I see it is spending all its time reading and decompressing the data. This is inefficient but also could exhaust memory for very large images
```
   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       21    6.964    0.332    8.140    0.388 {built-in method astropy.io.fits.compression.decompress_hdu}
      895    3.058    0.003    3.058    0.003 {method 'read' of '_io.FileIO' objects}
       21    0.483    0.023    0.483    0.023 {built-in method gc.collect}
    14621    0.274    0.000    0.274    0.000 {built-in method posix.lstat}
   ... snip ...
```

rmjarvis:
I think I see where this is probably happening.  We have a [`_check_hdu`](https://github.com/GalSim-developers/GalSim/blob/releases/2.1/galsim/fits.py#L327) function that we run to see if `hdu.verify('fix')` is required to fix various kinds of broken (but salvageable) fits files.  This process involves making sure `hdu.data` doesn't raise an exception.

I think what we want to do here is to add a parameter `header_only` to this, which `FitsHeader` can set to `True` when it calls it, where it only checks that the header is readable, not the data.

rmjarvis:
Done in #1025

Issue #1009, opened 2019-03-12
isinstance(...,int) check in BaseDeviate failing for numpy.int64


matroxel:
I'm updating code to python3 and switching to master branch from #919, on a new machine. I've run into a failed int check that worked in python2/#919/a different machine in BaseDeviate().

The cause is that I'm passing the sum of three values, two of which are python int and one of which is numpy.int64, which types the sum to numpy.int64. isinstance(...,int) then fails on the passed integer.

rmjarvis:
Can you post a snippet that fails for you, which we could turn into a unit test?

matroxel:
>>> galsim.BaseDeviate(np.array(1).astype(np.int64))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/cvmfs/sw.lsst.eu/linux-x86_64/lsst_distrib/w_2019_08/stack/miniconda3-4.5.4-fcd27eb/Linux64/galsim/2.1.4.lsst/lib/python/galsim/random.py", line 92, in __init__
    self.reset(seed)
  File "/cvmfs/sw.lsst.eu/linux-x86_64/lsst_distrib/w_2019_08/stack/miniconda3-4.5.4-fcd27eb/Linux64/galsim/2.1.4.lsst/lib/python/galsim/random.py", line 128, in reset
    raise TypeError("BaseDeviate must be initialized with either an int or another "
TypeError: BaseDeviate must be initialized with either an int or another BaseDeviate

rmjarvis:
Ok.  I guess I'm not sure why you are using this for the seed of a BaseDeviate.  Can't you just cast it to `int` in your code before passing to BaseDeviate?

matroxel:
Yes, I can, and do now. @rmandelb asked me to open the issue.

rmandelb:
Even though troxel did the local fix for that specific piece of code, I
thought it worth at least considering whether those type checks should be
made more robust to this problem.

rmjarvis:
Sure.  The documentation says "integer", which I guess is ambiguous about whether non-`int`, numpy integer family types are allowed.

rmjarvis:
Done in #1025

PR #1025, opened 2019-05-05
#1024 A few bug fixes I'd like to add to v2.1 series (v2.1.5)

rmjarvis:
This PR includes a few bug fixes I think we should include in the 2.1 series as v2.1.5:

1. @aamon pointed out a bug in the WCS functions where things like `local` would fail when given a PositionI input rather than a PositionD.  This is a casualty of a few-years-old change in numpy behavior to disallow subtracting a float from an int with `-=` (and similar ops).  But apparently it's rare enough that no one noticed it being a problem until now.  Anyway, the fix is just to build most of our numpy arrays with `dtype=float`.  I went through and checked them all and fixed the ones that looked potentially problematic, but I only added a unit test for the error that Alex was getting.
2. PyYAML changed its `load` function to raise ugly errors telling people to switch to `safe_load` instead or explicitly use a non-safe `Loader`.  I'm pretty sure we don't ever need a non-safe loader, so I switched to `safe_load`.
3. I fixed the inefficient decompression step that gets triggered by `FitsHeader`, even though it doesn't actually need the data, just the header info.  (cf. #1024, reported by @esheldon)
4. I cherry-picked @rmandelb's fix to #1020, since I thought we might as well include that in v2.1.5 as well.  (Ping @beckermr)\
5. I fixed the seed checks for BaseDeviate to allow non-int integers (cf. #1009, reported by @matroxel).

Note: This is set to merge to releases/2.1.  I will cherry-pick them over to master by hand.

### DIFF
--- a/tests/test_random.py
+++ b/tests/test_random.py
@@ -1710,7 +1710,7 @@ def test_int64():
               np.array(123).astype(np.int64)]
 
     if sys.version_info < (3,):
-        ivalues.append(long(1234))
+        ivalues.append(long(123))
 
     for i in ivalues:
         rng2 = galsim.BaseDeviate(i)
--- a/tests/test_random.py
+++ b/tests/test_random.py
@@ -1683,6 +1683,40 @@ def test_ne():
     assert repr(d1) == repr(d2)
     assert d1 != d2
 
+@timer
+def test_int64():
+    # cf. #1009
+    # Check that various possible integer types work as seeds.
+
+    rng1 = galsim.BaseDeviate(int(123))
+    # cf. https://www.numpy.org/devdocs/user/basics.types.html
+    ivalues =[np.int8(123),  # Note this one requires i < 128
+              np.int16(123),
+              np.int32(123),
+              np.int64(123),
+              np.uint8(123),
+              np.uint16(123),
+              np.uint32(123),
+              np.uint64(123),
+              np.short(123),
+              np.ushort(123),
+              np.intc(123),
+              np.uintc(123),
+              np.intp(123),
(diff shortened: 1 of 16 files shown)


tests/test_random.py
```diff
@@ -1683,6 +1683,40 @@ def test_ne():
...
+              np.uint32(123),
+              np.uint64(123),
+              np.short(123),
+              np.ushort(123),
+              np.intc(123),
+              np.uintc(123),
+              np.intp(123),
+              np.uintp(123),
+              np.int_(123),
+              np.longlong(123),
```
jmeyers314: I suppose `np.long` and `np.ulong` should be added for completeness.

rmjarvis: I think I got all the numpy valid types.  It's true that 
`np.long` is valid (albeit [undocumented](https://docs.scipy.org/doc/numpy/user/basics.types.html)).  But it is equivalent to `int` in Python 3.x and `long` in Python 2.7, so not really an additional type to check.  

However, I don't think `np.ulong` is valid.
```
>>> np.long
<class 'int'>
>>> np.ulong
AttributeError: module 'numpy' has no attribute 'ulong'
```

jmeyers314: Oh interesting.  I think I somehow ended up at the [v1.10 documentation](https://docs.scipy.org/doc/numpy-1.10.1/user/basics.types.html) which implies that long and ulong are a thing.  But I guess this doesn't exist in recent numpy.

rmjarvis:
Done.  v2.1.5 released.

Issue #1008, opened 2019-03-11
Use OpenMP in Silicon::accumulate function.


rmjarvis:
For future LSST DESC ImSim runs, we would like to potentially use OpenMP to parallelize the Silicon class's [`accumulate`](https://github.com/GalSim-developers/GalSim/blob/releases/2.1/src/Silicon.cpp#L462) method, particularly when accumulating a large number of photons.

Specifically, we should parallelize the loop over photons, but we need to be careful to avoid race conditions (e.g. when calling `updatePixelDistortions` and when adding the flux to the `_delta` array and `addedFlux`).

The call to [`updatePixelDistortions`](https://github.com/GalSim-developers/GalSim/blob/releases/2.1/src/Silicon.cpp#L157) probably need to be atomic in this context, but it too might be able to parallelize its calculation over pixels.

When working on this, tests/test_silicon.py will be a good test that you haven't broken anything.  And devel/time_phot.py will be a decent test of the timing.  (We might want to write some additional timing tests that are more targeted to what we will want, but this isn't too bad.)

jamesp-epcc:
I have started adding OpenMP to this method. I have forked the repo and am checking in my progress here: https://github.com/jamesp-epcc/GalSim/blob/releases/2.1/src/Silicon.cpp . So far I have parallelised the main loop over photons. It's not giving quite the same results as the old serial version, but this may just be because of random number sequences being different, and `updatePixelDistortions` being called at subtly different times (which is going to be difficult to avoid completely). It is speeding things up a bit but not as much as I'd hoped. This may improve if I can get `updatePixelDistortions` to run in parallel as well.
I have also started looking at parallelising `updatePixelDistortions` itself but my first attempt didn't work properly so I will have to try something more complex.

jamesp-epcc:
I have parallelised `updatePixelDistortions` and it is now working. It required some operations in `Polygon::distort` to be made atomic. I will now return to working on the `accumulate` method itself.

jamesp-epcc:
I have tested the parallelised code on Cori KNL and Haswell. The graphs below show the scaling of the code when running the "resume and no resume" test from `test_sensor.py`. This is with only `updatePixelDistortions` parallelised. The scaling is reasonably good even with just this function parallelised. The "resume" case shows much less improvement, probably because it is starting from a much faster baseline. The "no resume" case scales well up to about 8 threads on both Haswell and KNL, though there is less benefit beyond this level.

[...]

PR #1027, opened 2019-05-24
Add OpenMP parallelization to Silicon::accumulate

jamesp-epcc:
This is my initial attempt at adding OpenMP parallelization to the photon loop and the pixel distortions loop in Silicon::accumulate. Discussion and performance results can be found in issue #1008. I suspect this code is not ready to be merged yet but it would be good to get feedback on what needs to be changed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 branches:
     only:
         - master
-        - noboost
+        - releases/2.1
 
 language: python
 python:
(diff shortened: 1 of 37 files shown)


jamesp-epcc:
Sorry, it appears some other commits have ended up in this pull request due to me working on a different branch. I'll close this and open another one once I've cleaned that up.

Issue #1026, opened 2019-05-06
fast, vectorized WCS methods


beckermr:
@esheldon and I noticed a lot of overhead in the current outwardly facing APIs for WCS transformations. Apparently, things are vectorized underneath but it is not fully exposed according to @rmjarvis. After some discussion, @rmjarvis suggested adding methods like this

```
ra, dec = wcs.toWorldRad(x, y)  # returns ra, dec in radians
x, y = wcs.toImageRad(ra, dec)  # takes ra, dec in radians
```

The inputs here are numpy arrays or scalars.

I think these would be added to the Celestial WCS classes?

esheldon:
What do you think about a check in the toWorld type methods for two input arrays rather than a scalar `Position`?  Then everything could take a different code path that is fully vectorized, including a final `np.degrees(ra, out=ra)` ?

rmjarvis:
I don't like having units of angles be implicit.  But perhaps
```
x, y = wcs.toImage(ra, dec, coord.degrees)
ra, dec = wcs.toWorld(x, y, coord.degrees)
```
Then you can specify which unit you are using for the ra, dec arrays explicitly, and it can apply it for you (still vectorized).

esheldon:
looks good to me

rmjarvis:
Done.  #1032

PR #1032, opened 2019-06-04
#1026 Add vectorized versions of toWorld and toImage

rmjarvis:
@esheldon and @beckermr were finding their use of the `toWorld` method of the WCS classes to be slow, since they were applying to many positions at one, and the only public API requires doing these calculations one at a time using PositionD instances for the positions.

The backend code for our WCS classes allows numpy arrays for the `toWorld` function using either `_uv` or `_radec` (depending on which kind of WCS it is), but there was no public method that gave documented access to these methods.

So this PR adds vectorized front end versions of `toWorld` that work with numpy arrays.  I also added vectorized versions of the reverse operation `toImage`, which required a bit of development on the backend as well, since those hadn't always worked correctly with numpy arrays.

The new front end API options are:
```
       >>> u, v = wcs.toWorld(x, y)                 # For EuclideanWCS types
       >>> ra, dec = wcs.toWorld(x, y, units=units) # For CelestialWCS types
       >>> x, y = wcs.toImage(u, v)                 # For EuclideanWCS types
       >>> x, y = wcs.toImage(ra, dec, units=units) # For CelestialWCS types
```
The CelestialWCS classes require a `units` parameter to declare what units you are using for the angles `ra`, `dec`.  As usual, this can be either a `coord.AngleUnit` or a string that can be parsed as such (e.g. `'rad'` or `'deg'`).

### DIFF
--- a/galsim/fitswcs.py
+++ b/galsim/fitswcs.py
@@ -209,19 +209,16 @@ class AstropyWCS(CelestialWCS):
 
         # astropy outputs ra, dec in degrees.  Need to convert to radians.
         factor = degrees / radians
-        ra *= factor
-        dec *= factor
 
-        try:
-            # If the inputs were numpy arrays, return the same
-            len(x)
-        except TypeError:
-            # Otherwise, return scalars
-            #assert len(ra) == 1
-            #assert len(dec) == 1
-            ra = ra[0]
-            dec = dec[0]
-        return ra, dec
+        if np.ndim(x) == np.ndim(y) == 0:
+            return ra[0] * factor, dec[0] * factor
+        else:
+            # Sanity checks that the inputs are the same shape.
+            assert np.ndim(x) == np.ndim(y)
+            assert x.shape == y.shape
+            ra *= factor
+            dec *= factor
+            return ra, dec
 
     def _xy(self, ra, dec, color=None):
         factor = radians / degrees
@@ -232,12 +229,14 @@ class AstropyWCS(CelestialWCS):
         with warnings.catch_warnings():
             warnings.simplefilter("ignore")
             x, y = self.wcs.all_world2pix(r1, d1, 1, ra_dec_order=True)
-        try:
-            len(ra)
-        except TypeError:
-            x = x[0]
-            y = y[0]
-        return x, y
+
+        if np.ndim(ra) == np.ndim(dec) == 0:
+            return x[0], y[0]
+        else:
+            # Sanity checks that the inputs are the same shape.
+            assert np.ndim(ra) == np.ndim(dec)
+            assert ra.shape == dec.shape
+            return x, y
 
     def _newOrigin(self, origin):
         ret = self.copy()
@@ -471,26 +470,25 @@ class PyAstWCS(CelestialWCS):
         ra, dec = self.wcsinfo.tran( xy )
         # PyAst returns ra, dec in radians, so we're good.
 
-        try:
-            len(x)
-        except TypeError:
-            # If the inputs weren't numpy arrays, return scalars
-            #assert len(ra) == 1
-            #assert len(dec) == 1
-            ra = ra[0]
-            dec = dec[0]
-        return ra, dec
+        if np.ndim(x) == np.ndim(y) == 0:
+            return ra[0], dec[0]
+        else:
+            # Sanity checks that the inputs are the same shape.
+            assert np.ndim(x) == np.ndim(y)
+            assert x.shape == y.shape
+            return ra, dec
 
     def _xy(self, ra, dec, color=None):
         rd = np.array([np.atleast_1d(ra), np.atleast_1d(dec)], dtype=float)
         x, y = self.wcsinfo.tran( rd, False )
 
-        try:
-            len(ra)
-        except TypeError:
-            x = x[0]
-            y = y[0]
-        return x, y
+        if np.ndim(ra) == np.ndim(dec) == 0:
+            return x[0], y[0]
+        else:
+            # Sanity checks that the inputs are the same shape.
+            assert np.ndim(ra) == np.ndim(dec)
+            assert ra.shape == dec.shape
+            return x, y
 
     def _newOrigin(self, origin):
         ret = self.copy()
@@ -714,15 +712,13 @@ class WcsToolsWCS(CelestialWCS): # pragma: no cover
         # wcstools reports ra, dec in degrees, so convert to radians
         factor = degrees / radians
 
-        try:
-            len(x)
-            # If the inputs were numpy arrays, return the same
-            return np.array(ra)*factor, np.array(dec)*factor
-        except TypeError:
-            # Otherwise return scalars
-            #assert len(ra) == 1
-            #assert len(dec) == 1
+        if np.ndim(x) == np.ndim(y) == 0:
             return ra[0]*factor, dec[0]*factor
+        else:
+            # Sanity checks that the inputs are the same shape.
+            assert np.ndim(x) == np.ndim(y)
+            assert x.shape == y.shape
+            return np.array(ra)*factor, np.array(dec)*factor
 
     def _xy(self, ra, dec, color=None):
         import subprocess
@@ -773,15 +769,13 @@ class WcsToolsWCS(CelestialWCS): # pragma: no cover
                 x.append(float(vals[4]))
                 y.append(float(vals[5]))
 
-        try:
-            len(ra)
-            # If the inputs were numpy arrays, return the same
-            return np.array(x), np.array(y)
-        except TypeError:
-            # Otherwise return scalars
-            #assert len(ra) == 1
-            #assert len(dec) == 1
+        if np.ndim(ra) == np.ndim(dec) == 0:
             return x[0], y[0]
+        else:
+            # Sanity checks that the inputs are the same shape.
+            assert np.ndim(ra) == np.ndim(dec)
+            assert ra.shape == dec.shape
+            return np.array(x), np.array(y)
 
     def _newOrigin(self, origin):
         ret = self.copy()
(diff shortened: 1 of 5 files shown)


esheldon:
thanks @rmjarvis

beckermr:
Yes thank you!

galsim/wcs.py
```diff
@@ -323,7 +323,19 @@ def toImage(self, arg, **kwargs):
...
-        2. The second converts a surface brightness profile (a GSObject) from world
+           Equivalent to wcs.posToImage(world_pos).
+
+        2. The second is nearly the same, but takes either u and v values or ra and dec values
+           (depending on the kind of wcs being used) directly and returns x and y values.
+           For this version, the inputs maybe be numpy arrays, in which case the returned values
+           are also numpy arrays.
+
+               >>> x, y = wcs.toImage(u, v)                 # For EuclideanWCS types
+               >>> x, y = wcs.toImage(ra, dec, units=units) # For CelestialWCS types
```
esheldon: It would be nice to make these explicit keywords to avoid the order of arguments confusion.  How can the user know whether it should be `u, v` or `v, u`?
```
x, y = wcs.toImage(u=u, v=v)
x, y = wcs.toImage(ra=ra, dec=dec, units=units)
```
Of course the return values also have this ambiguity `x, y` or `y, x`?

galsim/wcs.py
```diff
@@ -1038,6 +1088,29 @@ def x0(self): return self.origin.x
     @property
     def y0(self): return self.origin.y
 
+    def radecToImage(self, ra, dec, units, color=None):
```
esheldon: these versions avoid the order-of-arguments issue because the order is in the method name

esheldon:
I made some comments about conventions but don't feel you need to take them very seriously

galsim/wcs.py
esheldon: at least part of it, `radecToxy` would be even more explicit

galsim/wcs.py
```diff
@@ -2188,35 +2317,60 @@ def world_origin(self): return self._world_origin
 
     def _u(self, x, y, color=None):
         if self._uses_color:
-            return self._ufunc(x,y,color)
+            try:
+                return self._ufunc(x,y,color)
+            except TypeError:
```
beckermr: Is this the only error that can be thrown here, or should we be catching something more general?

galsim/wcs.py
```diff
@@ -2188,35 +2317,60 @@ def world_origin(self): return self._world_origin
...
+                # If that didn't work, we have to do it manually for each position. :(  (SLOW!)
+                return np.array([self._ufunc(x1,y1,color) for [x1,y1] in zip(x,y)])
         else:
-            return self._ufunc(x,y)
+            try:
+                return self._ufunc(x,y)
+            except TypeError:
```
beckermr: Ditto here.

galsim/wcs.py
```diff
@@ -2371,7 +2525,16 @@ def radec_func(self): return self._radec_func
...
 
     def _radec(self, x, y, color=None):
-        return self._radec_func(x,y)
+        try:
+            return self._radec_func(x,y)
+        except Exception as e:
+            try:
+                world = [ self._radec(x1,y1) for (x1,y1) in zip(x,y) ]
+            except Exception:  # pragma: no cover
+                raise e  # Raise the original one if this fails, since it's probably more relevant.
```
beckermr: This try except block seems like a better fit for the ones I flagged above (and a few below that).

galsim/wcs.py
rmjarvis: Sure, I'm fine with changing these names to include the order of the return values.

galsim/wcs.py
rmjarvis: > How can the user know whether it should be u, v or v, u?

Read the documentation?  :) 
I think I'm pretty reticent to add a big infrastructure to parse both args and kwargs for this to make sure all possible ways of specifying things work correctly.  I did point to which more explicitly named functions (with explicit named arguments) are equivalent.  Do you think this is sufficient for those who want their code to be more self-documenting than the overloaded `toImage` and `toWorld`?

esheldon: I don't feel strongly about it, go with what you have.

galsim/fitswcs.py
```diff
@@ -1296,14 +1338,39 @@ def _radec(self, x, y, color=None):
...
+            len(u)
+        except TypeError:
+            with convert_cpp_errors():
+                return _galsim.InvertPV(u, v, self.pv.ctypes.data)
+        else:
+            invpvu = np.empty(len(u))
+            invpvv = np.empty(len(u))
+            for i in range(len(u)):
+                with convert_cpp_errors():
+                    uu, vv = _galsim.InvertPV(u[i], v[i], self.pv.ctypes.data)
```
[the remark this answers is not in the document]
rmjarvis: The `InvertPV` and `InvertAB` functions are already doing a non-linear solution to a polynomial equation, so I'd be very surprised if leaving the outer loop in python is a slow down.  If someone can demonstrate otherwise, we can adjust to let these functions take numpy arrays, but it didn't seem worth worrying about to me.

galsim/wcs.py
```diff
@@ -233,12 +253,47 @@ def posToWorld(self, image_pos, color=None, **kwargs):
...
+                raise TypeError("units is required for CelestialWCS types")
+            elif isinstance(units, str):
+                units = AngleUnit.from_name(units)
+            elif not isinstance(units, AngleUnit):
+                raise GalSimValueError("units must be either an AngleUnit or a string", units,
+                                       AngleUnit.valid_names)
+            return self._xyToWorld(x, y, units, color)
+        else:
+            if units is not None:
+                raise TypeError("units is not valid for EuclideanWCS types")
```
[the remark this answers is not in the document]
rmjarvis: "Arbitrary" WCS is almost not really a thing.  Euclidean and Celestial WCS's are really pretty different beasts.  
I suspect it's more likely for this to help people identify when they thought they should be getting ra, dec back (with units) and accidentally passed a local WCS where that doesn't make sense, than that this would cause hardships of them wanting to use the same function call for both euclidean and celestial wcs types.

galsim/wcs.py
rmjarvis: It's the only kind of error I could think of for people writing code that works for floats but not numpy arrays.  (It's the same except statement that I had elsewhere before this PR.  I just moved it here in some code restructuring.)

galsim/fitswcs.py
```diff
@@ -224,13 +224,19 @@ def _radec(self, x, y, color=None):
...
+        d1 = np.atleast_1d(dec) * factor
+
         with warnings.catch_warnings():
             warnings.simplefilter("ignore")
-            xy = self.wcs.all_world2pix(rd, 1, ra_dec_order=True)[0]
-        x, y = xy
+            x, y = self.wcs.all_world2pix(r1, d1, 1, ra_dec_order=True)
+        try:
+            len(ra)
+        except TypeError:
```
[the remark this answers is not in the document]
rmjarvis: > Also, does the code allow ra, dec that can be broadcasted together? In that case this might fail in weird ways.

Not sure what you mean by this.  Can you give me an example of what you have in mind?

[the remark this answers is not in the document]
beckermr: Right now I think this might give weird results, but IDK if the API allows these inputs

```
self._radec(10, np.array([34, 56, 78]), units='degrees')
```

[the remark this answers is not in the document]
rmjarvis: I see.  Currently, this sometimes goes through and sometimes gives errors.  (I just wrote a unit test to check.)  And some of the time it goes through it gives the right answer, but sometimes not.  
Do you think this is something that a user is likely to try?  If so, should I work to make it always return the right thing, or check and raise an exception?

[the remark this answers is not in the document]
beckermr: I think this depends on how seriously we take the numpy API. Generally, the numpy API supports any combination of operations that broadcast correctly. 

I'd say we either support broadcasting, or explicitly check that either both arrays are the same size or both are scalars. A silent failure here seems to be pretty catastrophic in that it sometimes gives the wrong answers.

In terms of whether or not this is common, I honestly have no idea. I think some extra error checking is the best way to go here mostly because that seems like the least amount of effort and I am happy to make sure the inputs conform for my own code.

galsim/wcs.py
rmjarvis: I went ahead and changed this to catching `Exception`, since it wouldn't be too surprising if I'm just insufficiently imaginative to think of a different exception that could be raised.  I switched these to follow the pattern in `RaDecFunction`.

galsim/fitswcs.py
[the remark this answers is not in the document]
rmjarvis: Hm..   This seems like a bit more handholding than I generally want to do for the user.  There's a balance between checking for likely mistakes and giving helpful errors vs slowing down the code by checking for every possible thing someone could do wrong.  (if statements in python aren't free.)
My compromise now is to include a couple of `assert` statements that the inputs are the same shape at this point when returning numpy arrays.  These tend to be faster than if statements in python and still serve the purpose of telling the user than their code is wrong without bending over backwards to give a helpful error message.  (They can also be turned off with -O if the run is really performance critical, but nobody usually does this.)

rmjarvis:
Thanks Erin and Matt.  I think I've addressed all your suggestions.

Let's merge this Friday unless anyone requests more time to review it.

beckermr:
LGTM!

beckermr:
@rmjarvis I think we are good to go here!

rmjarvis:
Thanks for the reminder.  Merging now.

PR #1034, opened 2019-06-24
removed errant mail message

demitri:


demitri:
Not specifically; just randomly came across it.

PR #1036, opened 2019-06-24
Fix the flux when shooting some profiles to not be off by ~1.e-3

rmjarvis:
@cwwalter ran across a bug in photon shooting where we were systematically shooting a little bit less total flux than intended for some profiles.  The problem occurred with profiles that use OneDimensionalDeviate for their back-end shooting calculations.  Indeed there were actually two problems:
1. Two profiles, Kolmogorov and VonKarman, used the wrong tolerance value for when to stop the radial profile.  We advertise that errors in the shooting flux should be less than gsparams.shoot_accuracy, which defaults to 1.e-5.  These were erroneously using gsparams.folding_threshold/5 for no obvious reason.  The latter value is only 1.e-3, so this led to the shooting flux values being wrong by 1.e-3 in relative value rather than 1.e-5, which would probably have been tolerable.
2. All profiles that use OneDimensionalDeviate normalized the fluxes using the empirical integral of the radial function from 0 to maxr, using whatever maxr was calculated to be acceptable.  This is nominally ok, but it means that the resulting fluxes are all systematically small by a factor of 1-gsparams.shoot_accuracy.  I changed the OneDimensionalDeviate constructor to take in the true analytic integral of the radial function and use that for the normalization.  This means that rather than all photons coming back with flux = 0.99999, they come back with flux=1, and the total flux comes out correct.  (At least when all photons land on the sensor.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,4 +25,7 @@ Bug Fixes
 ---------
 
 - Fixed error in `wcs.makeSkyImage` when crossing ra=0 line for some WCS classes. (#1030)
+- Fixed slight error in the realized flux of some profiles when using photon shooting.
+  The bug was most apparent for Kolmogorov and VonKarman, where the realized flux
+  could be too small by about 1.e-3. (#1036)
 - Fixed error in Sersic class when n is very, very close to 0.5. (#1041)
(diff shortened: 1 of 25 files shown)


rmjarvis:
Thanks Josh.  Merged.

Issue #1038, opened 2019-07-11
Expose Interpolant xval and uval functions to Python


rmjarvis:
The `piff.PixelGrid` class currently reimplements the Lanczos xval functionality from GalSim to calculate the interpolated model at arbitrary grid locations.  

I'd like to have it allow any of the various galsim Interpolants, but rather than reimplement the xval functions for each in the `piff` repo, I think it would make more sense to expose the existing code we have here for all Interpolant classes.

While I'm at it, exposing uval seems like an obvious thing as well, even though I don't have a specific use case for it.

rmjarvis:
Done.  #1040

PR #1040, opened 2019-07-16
#1038 Updates to Interpolant classes

rmjarvis:
The main thing here is to expose some of the Interpolant functions that had been confined to the C++ layer.  In particular the ability to calculate xval and kval for an array of input values.  This is Issue #1038.  We want this functionality (the xval part) for use in Piff.

Along the way, I significantly beefed up the unit testing of the Interpolant classes.  No significant bugs found, which was good.  However, I did notice that our use of the `tol` parameter isn't really consistent with the rest of GalSim.  It was basically being used the way we normally use `kvalue_accuracy`.  And the default of `1.e-4` doesn't match the default `kvalue_accuracy=1.e-5`.  Furthermore, quite a few of the classes don't even use `tol` for anything now that they have working analytic functions for everything.

So I think it makes sense to deprecate the `tol` parameter and encourage users to use `kvalue_accuracy` for this purpose instead.  So that's here too.

### DIFF
--- a/galsim/interpolant.py
+++ b/galsim/interpolant.py
@@ -65,11 +65,11 @@ class Interpolant(object):
         @param tol          [deprecated]
         @param gsparams     An optional GSParams instance [default: None]
         """
-        gsparams = GSParams.check(gsparams)
         if tol is not None:
             from galsim.deprecated import depr
             depr('tol', 2.2, 'gsparams=GSParams(kvalue_accuracy=tol)')
-            gsparams._kvalue_accuracy = tol
+            gsparams = GSParams(kvalue_accuracy=tol)
+        gsparams = GSParams.check(gsparams)
 
         # Do these in rough order of likelihood (most to least)
         if name.lower() == 'quintic':
--- a/galsim/interpolant.py
+++ b/galsim/interpolant.py
@@ -41,7 +41,7 @@ class Interpolant(object):
             "Use one of the subclasses instead, or use the `from_name` factory function.")
 
     @staticmethod
-    def from_name(name, tol=1.e-4, gsparams=None):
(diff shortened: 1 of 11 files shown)


galsim/interpolant.py
```diff
@@ -100,6 +104,14 @@ def from_name(name, tol=1.e-4, gsparams=None):
     def gsparams(self):
         return self._gsparams
 
+    @property
+    def positive_flux(self):
```
jmeyers314: These should probably get a docstring.  Maybe just a single sentence description.

galsim/interpolant.py
```diff
@@ -126,6 +138,53 @@ def __ne__(self, other):
     def __hash__(self):
         return hash(repr(self))
 
+    def xval(self, x):
+        """Calculate the value of the interpolant kernel at one or more x values
+
+        Parameters
```
jmeyers314: This looks like numpy doc style, which is neither current doxygen nor future google style.  Just sayin...

galsim/interpolant.py
```diff
@@ -137,32 +196,42 @@ class Delta(Interpolant):
...
     def __repr__(self):
-        return "galsim.Delta(%r, %r)"%(self._tol, self._gsparams)
+        return "galsim.Delta(gsparams=%r)"%(self._gsparams)
 
     def __str__(self):
-        return "galsim.Delta(%s)"%(self._tol)
+        return "galsim.Delta()"
+
+    @property
+    def tol(self):
```
jmeyers314: Move to parent class `Interpolant`?

rmjarvis:
Thanks Josh.  I make all your suggested edits.

Issue #1041, opened 2019-07-16
Sersic drawImage fails for some values of n just under 0.5


taranu:
```
galsim.Sersic(n=0.499999999999999, half_light_radius=1).drawImage()
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-15-2f9d139261c0> in <module>()
----> 1 galsim.Sersic(n=0.499999999999999, half_light_radius=1).drawImage()

~/.local/lib/python3.6/site-packages/galsim/gsobject.py in drawImage(self, image, nx, ny, bounds, scale, wcs, dtype, method, area, exptime, gain, add_to_image, use_true_center, offset, n_photons, rng, max_extra_noise, poisson_flux, sensor, surface_ops, n_subsample, maxN, save_photons, setup_only)
   1632 
   1633         # Make sure image is setup correctly
-> 1634         image = prof._setup_image(image, nx, ny, bounds, add_to_image, dtype)
   1635         image.wcs = wcs
   1636 

~/.local/lib/python3.6/site-packages/galsim/gsobject.py in _setup_image(self, image, nx, ny, bounds, add_to_image, dtype, odd)
   1068                 image = Image(nx, ny, dtype=dtype)
   1069             else:
-> 1070                 N = self.getGoodImageSize(1.0)
   1071                 if odd: N += 1
   1072                 image = Image(N, N, dtype=dtype)

~/.local/lib/python3.6/site-packages/galsim/gsobject.py in getGoodImageSize(self, pixel_scale)
   1768         """
   1769         # Start with a good size from stepk and the pixel scale
-> 1770         Nd = 2. * math.pi / (pixel_scale * self.stepk)
   1771 
   1772         # Make it an integer

~/.local/lib/python3.6/site-packages/galsim/gsobject.py in stepk(self)
    302     def stepk(self):
    303         "The sampling in k space necessary to avoid folding of image in x space."
--> 304         return self._stepk
    305 
    306     @property

~/.local/lib/python3.6/site-packages/galsim/convolve.py in _stepk(self)
    295         # Assume R_final^2 = Sum(R_i^2)
    296         # So 1/stepk^2 = 1/Sum(1/stepk_i^2)
--> 297         inv_stepksq_list = [obj.stepk**(-2) for obj in self.obj_list]
    298         return np.sum(inv_stepksq_list)**(-0.5)
    299 

[...]

PR #1042, opened 2019-07-17
#1041 Fix brackeyUpper failure for Sersic galaxies with n very very close to 0.5.

rmjarvis:
When n is very very close to 0.5, the Sersic class's calculation to determine a good stepk could fail to converge.  My solution here is the just make sure the initial bracketing is at least 0.01.  With this starting point, the bracketing takes only a few iterations, and the Brent solution takes 1 step.  So I think this is a good, efficient solution for this edge case.


rmjarvis:
Thanks Josh.  There were a couple other lines like the one you changed.  I changed them all to `<=`.  cf.  cb20ae146

Issue #1037, opened 2019-06-28
Moving the Great3 server to Edinburgh


joezuntz:
Hi all,

To help deal with things like #1023 I'd like to move the Great3 server, which hosts the COSMOS data among other things, up to Edinburgh where I can debug things properly.

I guess this would mean changing some paths in download code somewhere, and website text change.  Anything else?

Cheers,
Joe

rmjarvis:
See also #1033.  @EiffL was looking into switching the host to Zenodo.  It seems like that might be a better solution than relying on either university's servers.

rmjarvis:
Superseded by #1033 / #1043

Issue #1033, opened 2019-06-09
Host RealGalaxy Data on Zenodo?


EiffL:
Obviously not a super high priority issue, but this would do two things:
  - Have a nice DOI to cite whenever we use these datasets.
  - Provide a centralized repo from which to download GalSim datasets (because right now, I'm attempting to download the COSMOS_25.2 sample for the 3rd time from great3.jb.man.ac.uk and it keeps failing at about 25% of the download for some reason).

Let me know what you think, I'd be happy to implement the changes in `download_cosmos.py` and move a copy of the data over to Zenodo if that sounds like a good idea.

rmjarvis:
I haven't investigated different hosting options, but it sounds like you have.  I do think it would be a good idea to move to a more official hosting platform if it doesn't cost us anything.  So if you think Zenodo is a good one, that would be great if you could start the process of moving it there.

EiffL:
Great, yes, I think Zenodo is a great solution for this kind of data repository.

I have prepared an upload of the COSMOS dataset but I have a few questions before I can press the `publish` button:
  - I need a list of authors (and optionally contributors). I'm guessing @rmandelb and a few other people as authors ?
  - I need a license for the data, as a default Zenodo suggests `Creative Commons Attribution 4.0 International`
  - A list of grants that supported this work, there I'm guessing I can copy the grants in the acknowledgements of the GREAT3 paper, but let me know.
  - A date of first publication... I have no idea when that data was first released
   
I also went ahead and created a "GalSim" community, which acts as a curated list of datasets compatible with GalSim (it's empty for now): https://zenodo.org/communities/galsim 
There is only one issue with this is that there can only be one curator account. Curators accept or rejects uploads submitted by other people under the given community. I can play this role, but it's kind of weird because I'm not even a galsim dev ^^' Otherwise, someone else needs to create that community under their own personal accounts, or it's possible to create this community under a shared account, like `GalSim-developers`, maybe there is already an email account for this kind purpose ?

EiffL:
I'm working with students who have had difficulties with the download of the galsim  cosmos data... should we move forward with hosting it on zenodo?
Essentially all I need are the answers to the questions in the previous post, mainly a list of authors for the COSMOS_25.2 sample, and associated grant acknowledgements. @rmandelb ?

rmjarvis:
[...]

PR #1043, opened 2019-07-25
Updates download URL for  COSMOS dataset to Zenodo

EiffL:
This PR simply updates the download link for the COSMOS dataset to use the Zenodo hosted copy of the data. This aims to address  #1033 and #1037 in part.

Note that I'm not digging deaper into the  download mechanism here, we may ultimately want to use something a little bit more robust that can handle download interruptions (this is the reason why the current download from Great3 server was failing, because the server could only download 1GB at a time),

### DIFF
--- a/galsim/download_cosmos.py
+++ b/galsim/download_cosmos.py
@@ -394,7 +394,7 @@ def main():
         target_dir = share_dir
         link = False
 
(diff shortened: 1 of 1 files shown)


EiffL:
Thanks Rachel, I made the changes on the wiki. I'm going to go ahead and merge this branch then.

PR #1045, opened 2019-08-05
Convert docs to Sphinx

rmjarvis:
I think I'm basically done converting the docs to Sphinx.

Thanks to @jmeyers314 @tmcclintock @empEvil @matthewkirby for their help at LSST DESC hack day on this.

The documentation is hosted at http://galsim-developers.github.io/GalSim/.  So probably the best way to review this PR would be to just look through that for anything that doesn't look right.

For the most part, there is not too much new documentation.  The old doxygen format was converted to ReST.  And I ported over some of the wiki pages into rst files in the docs directory.  I also grabbed a few things from the Quick Reference Guide, most notably the discussion of units.

So I think this completely supersedes the old doxygen documentation, and the QRG, and some of the wiki pages.  I think we can probably move the rest of the wiki pages into this format as well, but I didn't do all of them yet.  E.g. the Installation FAQ is still on the wiki.

The one thing I couldn't figure out is how to get the Sphinx `coverage` module working to tell if there are any docs that I didn't include.  I manually checked things, and I think I have everything that is imported into the `galsim` namespace, as well as the public things in the sub-modules like `galsim.hsm`, `galsim.config`, `galsim.des`, `galsim.utilities` etc.  But it's certainly possible I missed something.  So it would be nice to figure out how to get the coverage thing working.

One caveat to those who have worked on this already.  I rebased the branch onto the current master, so if you want to make further edits at this point, you should do:
```
git fetch
git rebase --hard origin/sphinx
```
before editing.  Don't just `git pull`.


jmeyers314:
Do you know why `__weakref__` seems to appear in the docs for most classes?

rmjarvis:
Yes.  It's because I told it to document :special_members:.  Which I originally thought would just pick out the ones with doc strings, but it seems to be including all special members including things like `__weak_ref__` even if they don't have doc strings.

I've since changed this to not include all the special methods, but just list the appropriate ones to document by hand.  Still local on my machine.

... and now pushed.  :)  

So the weakref entries (along with others like `__eq__`, `__init__`, etc.) should all be gone now except for the ones that are actually useful.  (e.g. some `__call__` methods, `__add__`, `__mul__`, a few others.)

rmjarvis:
Any objection to merging this?  I'd like to start pushing towards a 2.2.0 release, so I'll have a few more PRs coming up.  It would be easier if I had the new doc formats in place for them.

Even if these docs here aren't perfect, I hope people will agree that the basic structure of them is better than before, and we can fix any errors down the line as people notice them.

Also, comments about whether it is ok to get rid of the quick reference guide and some of the wiki pages that are now ported into the rst docs would be welcome.

rmandelb:
I agree about the new docs being an improvement, so it's fine with me if you want to merge this.  In recognition of this fact, I'm about to put an "approve" review, but I will say I only looked through the online docs (didn't review the changes directly in this PR).

If the content in some of our wiki pages / quick ref guide is on sphinx, then I agree it's better to remove those wiki pages / quick ref guide so we don't have to be careful about maintaining both places in future.

README.rst
```diff
@@ -25,75 +25,74 @@ development.  For details of algorithms and code validation, please see
     http://adsabs.harvard.edu/abs/2015A%26C....10..121R
 
 
-Installation
-------------
+Baisc Installation
```
rmandelb:
```suggestion
Basic Installation
```

CHANGELOG.rst
```diff
@@ -0,0 +1,47 @@
...
+  were using it for any purpose, you should now use `galsim.BaseCorrelatedNoise`. (#160)
+- Deprecated the ``tol`` parameter of the various Interpolant classes.  Users should use the
+  ``kvalue_accuracy`` parameter of ``gsparams`` instead. (#1038)
+
+API Changes
+-----------
+
+- Removed functionality to store and re-load WFIRST PSFs, and to obtain multiple WFIRST
+  PSFs simultaneously. (#919)
+- The RealGalaxy COSMOS catalog is now hosted at `Zenodo <https://zenodo.org/record/3242143>`_,
```
rmandelb: Is this really an API change?  I thought this should be pretty transparent to any user (shouldn't the COSMOS downloading functionality just work seamlessly without people changing their code?), unlike most API changes.

CHANGELOG.rst
```diff
@@ -0,0 +1,47 @@
...
+  which seems to have solved some flakiness in the previous hosting at the University of
+  Manchester. (#1033)
+
+Documentaion Updates
```
rmandelb:
```suggestion
Documentation Updates
```

rmjarvis:
Thanks Rachel.  I made your suggested changes.  Also went ahead and removed the doc directory with the old QRG and doxygen docs.  They still live in git of course, so if we have a need to reinstate them, they're available.  But I'm pretty sure all the information there is superseded by the new docs.

Issue #977, opened 2018-07-06
RandomWalk much slower when sheared


esheldon:
The following code runs 10 times faster when no shear is applied.
```python
import numpy as np
import galsim
import time

import argparse

parser=argparse.ArgumentParser()
parser.add_argument('ntrial',type=int)
parser.add_argument('--shear',action='store_true')


class Sim(object):
    def __init__(self, shear=False):
        self.shear=shear
        self.dim=100

        self.flux = 100.0
        self.hlr = 4.0
        self.n_knots=100
        self.psf=galsim.Gaussian(fwhm=0.9)

    def make_im(self):
        obj=self.get_model()
        if self.shear:
            g1,g2=self.get_shape()
            obj = obj.shear(g1=g1,g2=g2)

        obj=galsim.Convolve(obj, self.psf)
        return obj.drawImage(
            scale=0.263,
        ).array

    def get_shape(self):
        while True:
            g1,g2 = np.random.normal(scale=0.2,size=2)
            g=np.sqrt(g1**2 + g2**2)
            if g < 0.99:
                break
        return g1,g2

    def get_model(self):
        profile=galsim.Exponential(
            flux=self.flux,
            half_light_radius=self.hlr,
        )

        obj=galsim.RandomWalk(
            self.n_knots,
            profile=profile,
        )
        return obj

args=parser.parse_args()
sim=Sim(shear=args.shear)

tm=time.time()
for i in range(args.ntrial):
    im=sim.make_im()

print("time:",time.time()-tm)
```

The profiler indicates that it is spending a lot of time in transform.py  Is this something we could optimize?  This is pure peculation, but these are point sources (a delta function is used for each knot) so we should be able to do something smart to avoid this overhead it seems.

```
time: 14.920971870422363
         1162298 function calls (1135947 primitive calls) in 18.439 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      100   14.072    0.141   14.099    0.141 transform.py:500(_drawKImage)
100342/100340    0.171    0.000    0.261    0.000 {built-in method builtins.isinstance}
      100    0.156    0.002    0.196    0.002 randwalk.py:197(_sbp)
      100    0.094    0.001    0.094    0.001 {built-in method galsim._galsim.irfft}
 1388/259    0.088    0.000    0.327    0.001 sre_parse.py:470(_parse)
[...]

PR #1051, opened 2019-08-23
#977 Speed up RandomWalk when sheared

rmjarvis:
@esheldon noticed that RandomWalk galaxies are much slower to draw when they are sheared than when they are not.  This is because the code to apply transformations to the FFT image are not particularly fast. 

It turns out it is much faster to apply the transformation first to the input profile and then use that to shoot the locations of the knots.  This gets the knots in the right place before trying to draw them, so even when using FFT, it is still pretty fast.  There is a script in devel/time_knots.py that showed that the sheared knots used to take about 8 times longer than not shearing.  Now it is about 1.5 timex as long, so a big improvement.

Along the way, we realized that the old RandomWalk class was not serializing properly.  It had not been properly keeping track of the rng that was used to generate the points, so the post-serialization object would generate points in new locations.  I fixed that here by duplicating the input rng and always generating points from that if the points need to be regenerated.

Finally, I am proposing here a name change of the class from RandomWalk to RandomKnots, for the following reasons:

* The name RandomWalk was really an implementation detail about the algorithm used to generate the locations of the points.  It was never really a good description of what the profile looked like when rendered.  
* Most people probably think first of a random walk with constant step lengths, not a Gaussian distribution of steps, which is the version that this algorithm was really implementing.  The result looks nothing like a constant-step random walk.  So I think this was probably confusing.
* We already (even before this PR) were not using the Gaussian random walk algorithm to generate the points.  We instead use the shoot mechanism to distribute knots according to any input profile, Gaussian or otherwise.
* When the given profile is not a Gaussian, the knot placement is not consistent with a random walk, so even more that just that we are not using that algorithm, we cannot use that algorithm for non-Gaussian profiles.
* Finally, I think the name RandomWalk was not particularly intuitive for the ways that people typically use this class. Normally it is used to model a bunch of knots of star formation.  I think the new name should evoke the physical model that users are trying to model.  Thus, I propose RandomKnots.

In this PR, I've deprecated the old name, and converted the code to use the new name.  So nothing will break in code that uses RandomWalk (for now, until v3.0).  But I can easily revert it if people prefer to keep the name RandomWalk.  Or we can switch to something else if a case is made for some other name instead.


tests/test_knots.py
```diff
@@ -334,10 +352,68 @@ def test_randwalk_hlr():
             mess="hlr for npoints: %d outside of expected range" % npoints
             assert abs(mn-hlr) < nstd*std_check, mess
 
+@timer
+def test_randwalk_transform():
```
jmeyers314: s/test_randwalk/test_knots

PR #1053, opened 2019-08-27
Add center option to drawImage command

rmjarvis:
Sorry, I lied.  One more PR for version 2.2.  This one is quick though.  It's a feature I've wanted to add for a while now, but I never bothered to make an issue for.

I added `center` as an optional parameter to the `drawImage` command.  It's kind of just a convenience option, since it doesn't add any real capability that isn't already possible with the `offset` parameter.  But getting a galaxy drawn at the right position with `offset`, especially when letting GalSim automatically size the image for you, is kind of a pain.  An example was in demo9, and it was pretty confusing.

Now a whole bunch of code in demo9.py has gone away, and the draw command changed to
```
final.drawImage(wcs=local_wcs, center=image_pos)
```

This makes me happy.  :)  I didn't propagate this change to anywhere else, but it wouldn't surprise me if there were other places where some code could look significantly simpler using `center` rather than `offset`.

### DIFF
--- a/examples/demo9.py
+++ b/examples/demo9.py
@@ -391,7 +391,7 @@ def main(argv):
             # Draw the stamp image
             # This will construct an appropriately sized postage-stamp image with the galaxy
             # drawn near the center of the image.  The bounds of the postage stamp will be set
-            # such that the given center is close the stamp center.  And the galaxy will be drawn
+            # such that the given center is close to the stamp center.  And the galaxy will be drawn
             # centered at that sub-pixel location on the image.
             # We also need to provide the local wcs at the current position.
             local_wcs = wcs.local(image_pos)
--- a/examples/demo9.py
+++ b/examples/demo9.py
@@ -297,24 +297,6 @@ def main(argv):
             # stamp on the full image.
             image_pos = wcs.toImage(pos)
 
-            # For even-sized postage stamps, the nominal center (available as stamp.center)
-            # cannot be at the true center (available as stamp.true_center) of the postage stamp,
-            # since the nominal center values have to be integers.  Thus, the nominal center is
-            # 1/2 pixel up and to the right of the true center.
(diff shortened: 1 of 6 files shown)


examples/demo9.py
```diff
@@ -407,15 +389,13 @@ def yfunc1(u,v):
             final = galsim.Convolve([gal, psf])
 
             # Draw the stamp image
-            # To draw the image at a position other than the center of the image, you can
-            # use the offset parameter, which applies an offset in pixels relative to the
-            # center of the image.
+            # This will construct an appropriately sized postage-stamp image with the galaxy
+            # drawn near the center of the image.  The bounds of the postage stamp will be set
+            # such that the given center is close the stamp center.  And the galaxy will be drawn
```
rmandelb:
```suggestion
            # such that the given center is close to the stamp center.  And the galaxy will be drawn
```

galsim/gsobject.py
```diff
@@ -1604,14 +1638,15 @@ def drawImage(self, image=None, nx=None, ny=None, bounds=None, scale=None, wcs=N
         # Figure out what wcs we are going to use.
         wcs = self._determine_wcs(scale, wcs, image)
 
-        # Make sure offset is a PositionD
+        # Make sure offset and center are PositionD if entered as tuples.
```
rmandelb: I think these convenience functions also work if offset/center are entered as lists or a numpy array?  Maybe this documentation could be something like "Make sure offset and center are PositionD, converting as necessary from other formats.  If `center` was passed in as None, it will remain that way after `_parse_center` is run"?

rmjarvis:
Thanks both.  Let's plan to release 2.2.0 on Friday.  So if you have any systems you want to check to make sure we didn't break installation or anything, please check them this week.

PR #1054, opened 2019-09-19
Fix horner and horner2d for complex inputs

rmjarvis:
I just discovered that our horner and horner2d functions had a bug when the inputs are complex.

Basically, it was recasting them to float for no good reason.  I guess maybe thinking to recast int or float32 arrays to float, but I don't think that's actually necessary, so better to not bother about that.  Anyway, fixed here.

### DIFF
--- a/galsim/utilities.py
+++ b/galsim/utilities.py
@@ -659,8 +659,10 @@ def horner(x, coef, dtype=None):
     return result
 
 def _horner(x, coef, result):
-    """Equivalent to `horner`, but ``x``, ``coeff``, and ``result`` must be contiguous arrays
-    with dtype == float.
+    """Equivalent to `horner`, but ``x``, ``coef``, and ``result`` must be contiguous arrays.
+
+    In particular, ``result`` must be already allocated as an array in which to put the answer.
+    This is the thing that is returned from the regular `horner`.
 
     Parameters:
         x:      A numpy array of values at which to evaluate the polynomial.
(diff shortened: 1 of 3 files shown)


galsim/utilities.py
```diff
@@ -705,13 +705,12 @@ def horner2d(x, y, coefs, dtype=None, triangle=False):
...
-        coefs = np.array(coefs, copy=False, dtype=float)
+        x = np.array(x, copy=False)
+        y = np.array(y, copy=False)
+        coefs = np.array(coefs, copy=False)
 
-    if len(x) != len(y):
-        raise GalSimIncompatibleValuesError("x and y are not the same size",
-                                            x=x, y=y)
+    if x.shape != y.shape:
+        raise GalSimIncompatibleValuesError("x and y are not the same size", x=x, y=y)
```
jmeyers314: s/size/shape

galsim/utilities.py
```diff
@@ -705,13 +705,12 @@ def horner2d(x, y, coefs, dtype=None, triangle=False):
         y = np.ascontiguousarray(y, dtype=float)
         coefs = np.ascontiguousarray(coefs, dtype=float)
     else:
-        x = np.array(x, copy=False, dtype=float)
-        y = np.array(y, copy=False, dtype=float)
-        coefs = np.array(coefs, copy=False, dtype=float)
+        x = np.array(x, copy=False)
```
jmeyers314: I think this change (and the 1d version too) makes the docstrings of `_horner`, `_horner2d` lies.  The dtype can now be complex.  Actually, looks like the docstrings were previous lies too.

jmeyers314: Also, s/coeff/coef for 1d and s/coeff/coefs for 2d.

rmjarvis:
Thanks Josh.  Fixed.

PR #1057, opened 2019-11-01
AtmosphericScreen pickling

jmeyers314:
See https://github.com/LSSTDESC/imSim/issues/234

### DIFF
--- a/galsim/phase_screens.py
+++ b/galsim/phase_screens.py
@@ -53,79 +53,6 @@ _calcOptStepK = LRU_Cache(__calcOptStepK)
 _GSScreenShare = {}
 
 
-def writeScreenShare(filename):
-    """Write `AtmosphericScreen` shared memory to a file.  See discussion in `AtmosphericScreen`
-    docstring for more information.
-
-    Parameters:
-        filename: Filename to which to write.
-    """
-    import pickle
-    screenShare = {}
-    for k, v in _GSScreenShare.items():
-        oneScreen = {}
-        oneScreen['f'] = np.frombuffer(v['f'], dtype=np.float64)
-        oneScreen['x'] = np.frombuffer(v['x'], dtype=np.float64)
-        oneScreen['y'] = np.frombuffer(v['y'], dtype=np.float64)
-        oneScreen['alpha'] = v['alpha'].value
-        oneScreen['kmin'] = v['kmin'].value
-        oneScreen['kmax'] = v['kmax'].value
-        oneScreen['x0'] = v['x0'].value
-        oneScreen['y0'] = v['y0'].value
-        oneScreen['xperiod'] = v['xperiod'].value
-        oneScreen['yperiod'] = v['yperiod'].value
-        oneScreen['instantiated'] = v['instantiated'].value
-        oneScreen['ctx_name'] = v['ctx_name'].value
-        screenShare[k] = oneScreen
-    with open(filename, 'wb') as f:
-        pickle.dump(screenShare, f)
-
-
-def readScreenShare(filename):
-    """Read `AtmosphericScreen` shared memory from file.  See discussion in `AtmosphericScreen`
-    docstring for more information.
-
-    Parameters:
-        filename: Filename from which to read.
-    """
-    import pickle
-    with open(filename, 'rb') as f:
-        screenShare = pickle.load(f)
-        for k, v in screenShare.items():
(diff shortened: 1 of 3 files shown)


tests/test_phase_psf.py
```diff
@@ -24,6 +24,11 @@
 import unittest
 from galsim_test_helpers import *
 
+try:
+    import cPickle as pickle
+except ImportError:
+    import pickle
```
rmjarvis: Note: I noticed that this pattern wasn't being used in ImSim.  Jim just did `import pickle`.  My understanding is that cPickle is much faster when it is available, so might be worth adding this when you're doing the imsim update.

tests/test_phase_psf.py
```diff
@@ -1265,8 +1270,83 @@ def test_shared_memory():
             initargs=galsim.phase_screens.initWorkerArgs()
         )
 
+@timer
+def test_pickle():
+    # This is response to ImSim issue #234
+    # https://github.com/LSSTDESC/imSim/issues/234
+
+    import multiprocessing
```
rmjarvis: This line is unnecessary now.  (I removed all the bits that had used it, and now you import it as mp below.)

tests/test_phase_psf.py
```diff
@@ -1265,6 +1270,90 @@ def test_shared_memory():
...
+
+    #screen_size = 819.2
+    screen_size = 32
+    screen_scale = 0.1
+    kmax = 1
+
+    if sys.version_info >= (3,4):
+        import multiprocessing as mp
+        if __name__ == "__main__":
+            ctxs = [None, mp.get_context("fork"), mp.get_context("spawn"), "forkserver"]
```
rmjarvis: All four only take 0.4 seconds to run this test on my laptop, so probably don't need to bother with this distinction.  Can just always do all of them in py3.

tests/test_phase_psf.py
jmeyers314: Done on https://github.com/LSSTDESC/imSim/pull/235

jmeyers314:
I think this is ready.  I'll wait for someone to review https://github.com/LSSTDESC/imSim/pull/235 before merging this though.

galsim/phase_screens.py
```diff
@@ -53,6 +53,79 @@ def __calcOptStepK(lam, diam, obscuration, gsparams):
...
+        oneScreen['alpha'] = v['alpha'].value
+        oneScreen['kmin'] = v['kmin'].value
+        oneScreen['kmax'] = v['kmax'].value
+        oneScreen['x0'] = v['x0'].value
+        oneScreen['y0'] = v['y0'].value
+        oneScreen['xperiod'] = v['xperiod'].value
+        oneScreen['yperiod'] = v['yperiod'].value
+        oneScreen['instantiated'] = v['instantiated'].value
+        oneScreen['ctx_name'] = v['ctx_name'].value
+        screenShare[k] = oneScreen
```
rmjarvis: @jchiang87's request would be easy to implement by making the above part of this function be in a method called perhaps `getScreenShare`.  Then this function would be just
```
with open(filename, 'wb') as f:
        pickle.dump(self.getScreenShare(), f)
```
Likewise for the read function, most of it could be in `setScreenShare` or some similar name.

jmeyers314: Jim's proposal [here](https://github.com/LSSTDESC/imSim/pull/235#discussion_r341879513) looks good.  One modification I'd like to propose: instead of using screen `get_picklable()` and `set_data()`, I think I can work things out so that the screens get included in the psf pickle automagically if the pickle is created with in a specific context manager (which temporarily sets a global GalSim pickling variable).  Then when reading back in, the `AtmosphericScreen.__setstate__` method would also load the screen if it happens to find it in the pickle.  E.g.,

```python
pickle.dump(psf1, file1)  # Only saves psf1, no shared screens
with galsim.pickleShared():
    pickle.dump(psf2, file2) # Saves both psf2 and shared screens, but only shared screens associated with psf2

psf1b = pickle.load(file1)  # Works if psf1 shared screens are already loaded
psf2b = pickle.load(file2)  # Works whether or not psf2 shared screens are already loaded
```

rmjarvis: That's cute.  :)  Seems like a nice API.

rmjarvis:
LGTM!

PR #1058, opened 2019-11-02
Add more splits to VK integral for large r values

rmjarvis:
@aaronroodman ran across another r0 value that gave the integration in VonKarman trouble.  (cf. #957, where this cropped up before)  r0 = 0.16879518207956518 failed, but 0.16879518 is fine.

This one failed to converge properly for a large r value, r~=49.  And out there, the oscillations are converging particularly slowly.  So the solution was to add more explicit split points to the integration.  

However, it doesn't make much sense I think to do this for all values of r.  I played around with the numbers a bit, and it seems like things die off reasonably well after around r/2 zero crossings.  So here, that means 25 split points (including 0), but at smaller r values, it would be fewer, so we don't need to waste time doing integrals that are close to 0.

This seems to work for the new r0 failure value, and still works for the old one.  We should keep a lookout for any additional special r0 values that give problems here, but hopefully this will be more robust than the previous implementation that always used 10 split points.


jmeyers314:
Changes look good to me, but I wonder if we can do better.  I let my computer randomly try triplets of `(r0_500, lam, L0)` over the weekend to find more problematic combinations. (Problematic after the applying this PR).  Ended up with 318 (out of 500000).  Values in attached file.
[vk_fail.txt](https://github.com/GalSim-developers/GalSim/files/3805254/vk_fail.txt)

rmjarvis:
Good call.  Every single time, it is because a term comes back negative, so now I check for that and rerun with higher accuracy settings.  Most of them are resolved by adding more split points, but at least one needed relerr to be reduced instead.  So now I do both.

PR #1060, opened 2019-11-09
Add _force_stepk to VonKarman

jmeyers314:
Adds undocumented `_force_stepk` kwarg to `VonKarman`.  If present, then calculation of real-space profile is delayed until needed by photon shooting, `getHalfLightRadius`, or `xValue`.  

I decided that in the case where the table build is triggered, it was simplest to repeat exactly the same calculation as if `_force_stepk` was not specified.  The only difference is that `_stepk` is not updated from its previously forced value.

Most functionality is therefore the same between the forced and unforced version of `VonKarman`, with the notable exceptions of speed for the create-through-drawFFT cycle, and `getGoodImageSize`.

In the Piff atmospheric PSF fitting context, using `_force_stepk` leads to a 10x speed improvement.

### DIFF
--- a/galsim/vonkarman.py
+++ b/galsim/vonkarman.py
@@ -98,7 +98,7 @@ class VonKarman(GSObject):
     _is_analytic_k = True
 
     def __init__(self, lam, r0=None, r0_500=None, L0=25.0, flux=1, scale_unit=arcsec,
-                 do_delta=False, suppress_warning=False, gsparams=None):
+                 do_delta=False, suppress_warning=False, gsparams=None, _force_stepk=0.0):
         # We lose stability if L0 gets too large.  This should be close enough to infinity for
         # all practical purposes though.
         if L0 > 1e10:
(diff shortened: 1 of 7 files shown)


src/SBVonKarman.cpp
```diff
@@ -125,14 +126,15 @@ namespace galsim {
...
+                                 const GSParamsPtr& gsparams, double force_stepk) :
         _lam(lam), _L0(L0),
         _L0_invcuberoot(fast_pow(_L0, -1./3)), _L053(fast_pow(L0, 5./3)),
+        _stepk(0.0), _maxk(0.0),
         _delta(exp(-0.5*magic1*_L053)),
         _deltaScale(1./(1.-_delta)),
         _lam_arcsec(_lam * ARCSEC2RAD / (2.*M_PI)),
         _doDelta(doDelta), _gsparams(gsparams),
-        _radial(Table::spline)
+        _radial(Table::spline), _sampler(nullptr)
```
rmjarvis: `nullptr` is a C++11 thing.  While probably almost all users will have that now, we technically don't require C++11 with SCons installations (using Boost rather than pybind11).
So this should be just `_sampler()` instead.  Or indeed just leaving it off is actually fine.

src/SBVonKarman.cpp
```diff
@@ -160,7 +162,10 @@ namespace galsim {
         dbg<<"_delta = "<<_delta<<'\n';
 
         // build the radial function, and along the way, set _stepk, _hlr.
-        _buildRadialFunc();
+        if (force_stepk == 0.0)
+            _buildRadialFunc();
```
rmjarvis: Is there really any advantage to this?  
Why not just `_stepk = force_stepk` always and let it be built whenever it is needed?

galsim/vonkarman.py
```diff
@@ -98,7 +98,7 @@ class VonKarman(GSObject):
     _is_analytic_k = True
 
     def __init__(self, lam, r0=None, r0_500=None, L0=25.0, flux=1, scale_unit=arcsec,
-                 do_delta=False, suppress_warning=False, gsparams=None):
+                 do_delta=False, suppress_warning=False, gsparams=None, _force_stepk=0.0):
```
rmjarvis: I think we should make this a regular documented parameter, rather than a nominally private one.  After all, you are proposing it for use by an external package (Piff).  Not to enable some internal-to-GalSim feature.

jmeyers314:
Thanks @rmjarvis , I accepted all your suggestions.

Issue #1061, opened 2019-11-09
OpticalPSF repr omits gsparams


jmeyers314:
should add that...

PR #1063, opened 2019-11-16
Clean up OpticalPSF repr

jmeyers314:
Addresses #1061 .  Adds and tests a few more variants of OpticalPSF and specifically how it `obj == eval(repr(obj))` round trips.

### DIFF
--- a/tests/test_optics.py
+++ b/tests/test_optics.py
@@ -104,6 +104,9 @@ def test_OpticalPSF_flux():
     interpolant_test = galsim.OpticalPSF(lam_over_diam=4., interpolant='linear')
     do_pickle(interpolant_test)
 
+    scale_unit_test = galsim.OpticalPSF(lam_over_diam=4., scale_unit=galsim.arcmin)
+    do_pickle(scale_unit_test)
+
 
 @timer
 def test_OpticalPSF_vs_Airy():
--- a/tests/test_optics.py
+++ b/tests/test_optics.py
@@ -101,6 +101,9 @@ def test_OpticalPSF_flux():
     check_basic(optics_test, "OpticalPSF")
     assert optics_test._screens.r0_500_effective is None
 
+    interpolant_test = galsim.OpticalPSF(lam_over_diam=4., interpolant='linear')
+    do_pickle(interpolant_test)
+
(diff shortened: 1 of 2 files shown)


tests/test_optics.py
```diff
@@ -458,7 +479,7 @@ def test_OpticalPSF_pupil_plane():
         do_pickle(test_psf)
 
     # Make a smaller pupil plane image to test the pickling of this, even without slow tests.
-    with assert_warns(galsim.GalSimWarning):
+    with assert_warns(galsim.GalSimWarning) if not do_slow_tests else ExitStack():
```
rmjarvis: I don't understand the point of this.
If do_slow_tests is true, then we don't assert that there is a warning?  Instead we just use ExitStack, which isn't actually doing anything, I guess.  But either way, the next lines get called, so what's supposed to be the difference between when we are willing to do the slow tests versus not?

tests/test_optics.py
```diff
@@ -99,7 +99,16 @@ def test_OpticalPSF_flux():
     do_pickle(optics_test._psf)
     do_pickle(optics_test._psf, lambda x: x.drawImage(nx=20, ny=20, scale=1.7, method='no_pixel'))
     check_basic(optics_test, "OpticalPSF")
-    assert optics_test._screens.r0_500_effective is None
+    assert optics_test._psf._screen_list.r0_500_effective is None
```
rmjarvis: You removed all the places where the `screen_list` property was used in the main code.  That's the drop in coverage.  You could remedy that by switching this to use the non-underscore version.
Or if you want, you could add a real unit test that checks that `psfscreen.psf.screen_list` is what you expect for some initialization where you know what it should be.

tests/test_optics.py
jmeyers314: When do_slow_tests is true, the pupil image is sufficiently high-res to not trigger the warning.  I suppose though that it may be better to bin by a larger factor in that case rather than elide the assert_warns...  I'll look into that.

jmeyers314: Changing the binning factor did the trick, so I removed the ExitStack stuff.

PR #1068, opened 2019-12-01
Add Refraction surface op

jmeyers314:
Adds Refraction surface operator.  Pretty straightforward.  A couple design choices include:
• Assumes surface normal is along z-axis.  I think this is sufficient for GalSim, we can do something more complicated in ImSim via batoid in the future if desired.
• In case of total internal reflection, NaNs the dxdz and dydz values, and zeros the flux values, but does not otherwise raise any errors or warnings.

### DIFF
--- a/galsim/photon_array.py
+++ b/galsim/photon_array.py
@@ -673,7 +673,34 @@ class Refraction(object):
             index_ratio = self.index_ratio(photon_array.wavelength)
         else:
             index_ratio = self.index_ratio
-        normsqr = 1 + photon_array.dxdz**2 + photon_array.dydz**2
+        # Here's the math avoiding any actual trig function calls:
+        #
+        #        x1 = dr/dz
+        #        ------+
+        #         \    |
+        #          \   |
+        #           \  |  dz/dz = 1
+        #            \ |
+        #             \|         n1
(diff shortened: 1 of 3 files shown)


galsim/photon_array.py
```diff
@@ -647,3 +647,36 @@ def applyTo(self, photon_array, local_wcs):
...
+        """Refract photons
+        """
+        if hasattr(self.index_ratio, '__call__'):
+            index_ratio = self.index_ratio(photon_array.wavelength)
+        else:
+            index_ratio = self.index_ratio
+        normsqr = 1 + photon_array.dxdz**2 + photon_array.dydz**2
+        with np.errstate(invalid='ignore'):
+            # NaN below <=> total internal reflection
+            factor = np.sqrt(1 - normsqr*(1-index_ratio**2))
```
rmjarvis: I think this line could use a few lines of inline comment.  I did verify it independently, so I get it now.  But it was far from obvious to me on initial reading.  Took some trig manipulations for me to get here, but you might have a more direct derivation in mind.

rmjarvis: I love the ASCII art you added for this!  :)

PR #1070, opened 2019-12-12
clarify provenance of numbers in WFIRST PSF module

rmandelb:
This is a very short docstring-only PR that originated from a discussion with @matroxel and @rmjarvis about the WFIRST PSFs.  Hopefully this update is more clear.  The PR is to go into master, but I think it could perhaps also be cherry-picked onto releases/2.2?


rmandelb:
Thanks, Mike!  I will go ahead and merge now.

Should I cherry-pick onto releases/2.2 as well?

rmjarvis:
Sure.  If you want to try remaking the docs on 2.2 with this, you can.  Or I can do it.

rmandelb:
I've not actually made the new docs before, so would probably want to wait until winter break to try (would have more time then).

Issue #1069, opened 2019-12-02
Add `focus_depth` kwarg to `SiliconSensor`


jmeyers314:
In the redder bands of LSST, and presumably other cameras too, the best focus occurs when the geometric focus of the optics is located in the bulk of the silicon rather than at the surface.

I think simulating this can be achieved via a `focus_depth` kwarg to `SiliconSensor` and some small changes around https://github.com/GalSim-developers/GalSim/blob/releases/2.2/src/Silicon.cpp#L602.  I guess we'd need to decide if the keyword should refer to the depth within the silicon (so post-refraction distance), or how far the CCD surface needs to be displaced from the optical focus (pre-refraction distance).

jmeyers314:
Actually, now that I think about it, this one can probably also be done as a surface op.  Just need to "rewind" the photons backwards in time a bit before passing them on to the current `SiliconSensor`.

rmjarvis:
And if it is a surface op, you don't need to decide which definition it is.  The user can control that by whether they put this operator before or after the refraction.  Which is actually pretty intuitive I think.

cwwalter:
> Actually, now that I think about it, this one can probably also be done as a surface op. Just need to "rewind" the photons backwards in time a bit before passing them on to the current `SiliconSensor`.

Hi Josh, I can't quite follow what you mean with the 'rewinding' language here.

rmjarvis:
The photons are coming in at an angle, which we parameterize by dx/dz and dy/dz slopes.  Changing the point of convergence is equivalent to moving their x,y back by:
`dx = -focus_depth * dxdz` and `dy = -focus_depth * dydz`.  
Then you can just use the current conversion distance routine in the Silicon code, and some of them will convert above the focus point and others below the focus point.

cwwalter:
Ah, I see.  You will want to call the refraction Surface Op before this then right?  

So, even with batiod will we need to pre store/calculate the focus set point per-filter to do this?  It seems like we should be able to calculate/record where the focus point was in the batoid step and then pass that to this operation. And, if you actually included the sensor in the optics path, you wouldn't need to do the extra refraction either.

rmjarvis:
> You will want to call the refraction Surface Op before this then right?

It depends on how you want to define the focal_depth parameter:
* If you want to define it as the distance the surface of the CCD needs to move to account for the non-surface conversion, then it should come before the refraction op.  
[...]

PR #1072, opened 2019-12-22
#1069 FocusDepth surface operator

jmeyers314:
This was mostly straightforward.  

The tricky bit, that TBH I'm not sure I've fully addressed, was figuring out the units of the photon_array `.x` and `.y` attributes, specifically in the context of surface operation `applyTo` methods.  I think these are pixels, which means that the `depth` keyword here also makes the most sense in units of pixels.  (One could use the localwcs to express the depth in arcsec I suppose, but that feels even more confusing than pixels).  It's also not clear to me what one should do if the pixels aren't actually square, but hopefully that's sufficiently unusual as to not be a problem.

### DIFF
--- a/galsim/photon_array.py
+++ b/galsim/photon_array.py
@@ -49,8 +49,14 @@ class PhotonArray(object):
         x:          The incidence x position at the top of the detector
         y:          The incidence y position at the top of the detector
         flux:       The flux of the photons
-        dxdz:       The tangent of the inclination angles in the x direction
-        dydz:       The tangent of the inclination angles in the y direction
+        dxdz:       The tangent of the inclination angles in the x direction.  Note that we define
+                    the +z direction as towards towards the dielectric medium of the detector and
+                    -z as towards vacuum; consequently, a photon with increasing x in time has
+                    positive dxdz.
+        dydz:       The tangent of the inclination angles in the y direction.  Note that we define
+                    the +z direction as towards towards the dielectric medium of the detector and
+                    -z as towards vacuum; consequently, a photon with increasing y in time has
+                    positive dydz.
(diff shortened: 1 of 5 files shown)


rmjarvis:
I can say right away that x and y in PhotonArray are definitely in units of pixels.  They refer to the x and y position in the final Image object.

galsim/photon_array.py
```diff
@@ -707,3 +707,21 @@ def applyTo(self, photon_array, local_wcs=None):
...
         photon_array.flux = np.where(np.isnan(factor), 0.0, photon_array.flux)
+
+
+class FocusDepth(object):
+    """Surface-layer operator that focuses/defocuses photons by changing the height of the focal
+    surface with respect to the conical beam.
+
+    Parameters:
+        depth:   The z-distance by which to displace the focal surface, in units of pixels.  A
+                 positive (negative) number here indicates an extra- (intra-) focal sensor height.
```
rmjarvis: This could use a translation into non-optics jargon I think.  Like "In other words, using depth>0 will cause the focus to be above the sensor.  Using depth < 0 will cause the focus to be in (or below) the surface of the sensor."  I *think* that's right anyway.

rmjarvis:
One other thing occurred to me.  It might be worth a unit test that checks that z or Y-band LSST images of stars are smaller when using an appropriate depth with the Silicon sensor.

jmeyers314:
Thanks Mike,
I added your suggested test and tried to clarify the definitions of `depth` and also `dxdz`, `dydz`, especially the signs involved.  Hopefully these make more sense now.  Could you take another look?  Feel free to merge yourself if you're satisfied.

rmjarvis:
lgtm

PR #1077, opened 2020-04-30
Adding new sensor files and correcting typo in sensor.py

craiglagegit:
Added new sesnor files and corrected/updated documentation in sensor.py.  No code changes.


galsim/sensor.py
```diff
@@ -84,21 +84,24 @@ class SiliconSensor(Sensor):
     into account the repulsion of previously accumulated electrons (known as the brighter-fatter
     effect).
 
-    There are currently three sensors shipped with GalSim, which you can specify as the ``name``
-    parameter mentioned below.
+    There are currently four up-to-date sensors shipped with GalSim, which you can specify as the ``name``
+    parameter mentioned below. The _50_ indicates 50V back-bias.
 
-        lsst_itl_8
+        lsst_itl_50_8
```
rmjarvis, line 90: Probably should change the default value to this one?  Rather than the now-out-of-date `name='lsst_itl_8'`.

craiglagegit: Good catch. Done!

PR #1078, opened 2020-05-08
Add obj.makePhot method

rmjarvis:
This PR adds a way to draw photons without actually accumulating them onto an image.

It is functionally equivalent to `_, photons = obj.drawPhot(im, ...)` but without requiring the image.

@cwwalter

### DIFF
--- a/galsim/gsobject.py
+++ b/galsim/gsobject.py
@@ -2067,6 +2067,101 @@ class GSObject(object):
         return iN, g
 
 
+    def makePhot(self, n_photons=0, rng=None, max_extra_noise=0., poisson_flux=None,
+                 sensor=None, surface_ops=(), maxN=None, orig_center=PositionI(0,0),
+                 local_wcs=None):
+        """
+        Make photons for a profile.
+
+        This is equivalent to drawPhot, except that the photons are not placed onto
+        an image.  Instead, it just returns the PhotonArray.
+
+        .. note::
+
+            The (x,y) positions returned are in the same units as the distance units
+            of the GSObject being rendered.  If you want (x,y) in pixel coordinates, you
+            should call this function for the profile in image coordinates::
+
+                >>> photons = image.wcs.toImage(obj).makePhot()
+
+            Or if you just want a simple pixel scale conversion from sky coordinates to image
+            coordinates, you can instead do
+
+                >>> photons = obj.makePhot()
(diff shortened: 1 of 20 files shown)


rmjarvis:
>  Just one question: would it be possible for drawPhot to use makePhot? Too tricky with the resume I'm guessing?

That was originally my plan, but with all the maxN and resume stuff, if seemed a little too much to untangle cleanly.  This seemed simpler, and the unit tests make sure they are equivalent for their overlapping use cases.

galsim/gsobject.py
```diff
@@ -2067,6 +2067,101 @@ def _calculate_nphotons(self, n_photons, poisson_flux, max_extra_noise, rng):
         return iN, g
 
 
+    def makePhot(self, n_photons=0, rng=None, max_extra_noise=0., poisson_flux=None,
+                 sensor=None, surface_ops=(), maxN=None, orig_center=PositionI(0,0),
```
cwwalter: Sensor isn't in the parameter list.

galsim/gsobject.py
```diff
@@ -2067,6 +2067,101 @@ def _calculate_nphotons(self, n_photons, poisson_flux, max_extra_noise, rng):
...
+                            applied in order before accumulating the photons on the sensor.
+                            [default: ()]
+            orig_center:    The position of the image center in the original image coordinates.
+                            [default: (0,0)]
+            local_wcs:      The local wcs in the original image. [default: None]
+
+        Returns:
+            - a `PhotonArray` with the data about the photons.
+        """
+        from .sensor import Sensor
```
cwwalter, line 2128: Is this necessary in this function?

galsim/gsobject.py
rmjarvis: Thanks.  Also maxN.  Fixed.

galsim/gsobject.py
rmjarvis: Thanks.  No not needed here.

rmjarvis:
> Also, just for my education/curiosity: Why all the changes to " rng=rng.duplicate()" in the tests? Was it not giving a reproducible sequence before?

These tests check that drawPhot and makePhot produce the same exact photons.  So the first time, for drawPhot, I call rng.duplicate() to fork the RNG sequence.  Then when I do it again with makePhot, the rng runs through the exact same sequence again, so should produce the same photons.

PR #1082, opened 2020-05-15
Remove future as a dependency

rmjarvis:
This is a bit of a "spring cleaning" PR.

I've come to realize that our python usage has very few bits that require the future module, so I'm inclined to remove it as a dependency.  Here are the places where we are currently using `future` and my proposed (via this PR) changes:

1. `basestring`: I found a cute way to define basestring that works in both 2 and 3: `basestring = ("".__class__, u"".__class__, b"".__class__)`.  This captures `str`, `unicode` and `bytes` properly on both.  I put this in utilities.py and import from there when we want it.
2. `exec_`: Turns out the python 3 version of `exec` works just fine in python 2.7.  I think the `future.exec_` workaround must have been for python 2.6, which we don't support anymore.  So I just switched exec_ -> exec, and it works fine.
3. `iteritems`, `itervalues`, `iterkeys`: These have only ever been an efficiency difference between python 2 and 3, and a minuscule one for all of our uses.  So I just switched them all to the regular `items`, `values`, `keys`.
4. `range`, `zip`: Similarly, the difference was just (slight) efficiency.  Just use the regular ones in both now.
5. `object`: I'm pretty sure this was already not needed.  I think `object` in 2.7 is already the python 3 version.
6. `int`: Our usage didn't care about the differences in functionality here, so I just use the native `int` in both and it's fine.
7. `input`: In python 2, this was called `raw_input`.  Easy to handle with a try/except NameError.  Just used in galsim_download_cosmos.py.
8. `super`: Python 2 doesn't have the no-argument version.  It's a little bit of extra typing, but I converted our uses of that to the more verbose one that works in both 2 and 3.  This was only being used in errors.py.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,8 +2,10 @@ Changes from v2.2 to v2.3
 =========================
 
 
-Deprecated Features
--------------------
+Dependency Changes
+------------------
+
+- Removed future as a dependency. (#1082)
 
 
 API Changes
--- a/SConstruct
+++ b/SConstruct
@@ -1782,19 +1782,6 @@ def CheckPyFITS(config):
 
     return 1
 
-def CheckFuture(config):
-    config.Message('Checking for future... ')
-
-    result, output = TryScript(config,"import future",python)
-    if not result:
-        ErrorExit("Unable to import future using the python executable:\n" + python)
-    config.Result(1)
-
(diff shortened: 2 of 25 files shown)


rmjarvis:
Anyone care about this?  There no real code changes here, so if no one objects, I'll merge Friday.

PR #1084, opened 2020-06-03
Completely gratuitous use of hyper exact constant from Fried, 1966 paper

rmjarvis:
I've always been unreasonably bothered by a magic number we've had in the Kolmogorov implementation.  It comes from Racine, 1996, who uses 6.8839 in his defining equation for F(k) in Fourier space.  Anyway, when writing the Piff paper, I discovered that Fried, 1966 actually has an analytic form of this number, which is 2 (24 Gamma(6/5)/5)^5/6.

Now, I'm very sure that no one cares about the precision of this number to 6+ decimal places, since it is multiplied by r0, which is never known to more than 2 decimal places, if that.  But I feel better having the correct analytic description for this magic number, rather than the decimal version without further explanation.

And since I was putting in the explanation, I went ahead and calculated it to full precision, rather than use the version with only 5 significant digits.  This required some updates of various tests at the 6th+ decimal place.


galsim/kolmogorov.py
```diff
@@ -141,7 +141,14 @@ class Kolmogorov(GSObject):
     # We convert this into T(k) = exp(-(k/k0)^5/3) for efficiency,
     # which implies 1/2 6.8839 (lambda/r0 / 2Pi)^5/3 = (1/k0)^5/3
     # k0 * lambda/r0 = 2Pi * (6.8839 / 2)^-3/5 = 2.992934
-    _k0_factor = 2.992934
+    #
+    # Update: It turns out that 6.8839/2 is actually (24 Gamma(6/5) / 5)^(5/6)
```
jmeyers314, line 145: Or 0.5*SBVonKarman::vkStructureFunction::magic4 :)

rmjarvis: Interesting.  Makes sense that this would show up there too.  I added a comment to this effect in SBVonKarman where you define magic4.

rmjarvis:
> Might also see if it's possible to tighten the test tolerances in test_vonkarman:test_vk_eq_kolm() now.

A little bit.  One of them could get a factor of 2 tighter.  The other one didn't get any closer.

Issue #1073, opened 2020-01-09
rename `withOrigin` to `shiftOrigin` for WCS classes?


beckermr:
I am working with a `TanWCS` and am seeing the following. 

This bit of code illustrates what is going on

```python
In [3]:     origin1 = galsim.PositionD(x=10, y=20) 
   ...:     origin2 = galsim.PositionD(x=20, y=40) 
   ...:     shift = origin2 - origin1 
   ...:     scale = 0.25 
   ...:     world_origin = galsim.CelestialCoord( 
   ...:         ra=10 * galsim.degrees, dec=12 * galsim.degrees) 
   ...:     wcs = galsim.TanWCS( 
   ...:         affine=galsim.AffineTransform( 
   ...:             scale, 0, 0, scale, 
   ...:             origin=origin1, 
   ...:             world_origin=galsim.PositionD(0, 0), 
   ...:         ), 
   ...:         world_origin=world_origin, 
   ...:         units=galsim.arcsec, 
   ...:     ) 
   ...:     print("origin1|wcs.toImage(world_origin):", origin1, wcs.toImage(world_origin)) 
   ...:     print("origin2|wcs.withOrigin(origin2).toImage(world_origin):", origin2, wcs.withOrigin(origin2).toImage(world_origin)) 
   ...:     print("shift|wcs.withOrigin(origin2).toImage(world_origin):", shift, wcs.withOrigin(shift).toImage(world_origin)) 
   ...:                                                                                                                                                                                                 
origin1|wcs.toImage(world_origin): galsim.PositionD(10.0,20.0) galsim.PositionD(10.0,20.0000000000229)
origin2|wcs.withOrigin(origin2).toImage(world_origin): galsim.PositionD(20.0,40.0) galsim.PositionD(30.0,60.0000000000229)
shift|wcs.withOrigin(origin2).toImage(world_origin): galsim.PositionD(10.0,20.0) galsim.PositionD(20.0,40.0000000000229)
```

Not that if I call `withOrigin` with where I want `world_origin` to point, It adds to the old origin. If instead I feed it a shift, it does what (at least) I expect.

cc @rmjarvis

beckermr:
Further @rmjarvis pointed out that in the code for images, this is used by feeding it a delta/shift

https://github.com/GalSim-developers/GalSim/blob/releases/2.2/galsim/image.py#L1103

rmjarvis:
[...]

PR #1085, opened 2020-06-11
Change withOrigin -> shiftOrigin for non-local WCS types

rmjarvis:
This is a quick hit to address issue #1073.

For non-local WCS types, the name `withOrigin` is deprecated in favor of `shiftOrigin`.  The functionality hasn't changed, but the name withOrigin is only really appropriate for LocalWCS types.  When the WCS already has a non-zero origin, then the action takes in really to shift the origin, not set a new value.

I kept the name withOrigin for local WCS types where I think it is more intuitive and accurately describes the action taken.  Both names are available for LocalWCS types and they are equivalent.

I also updated the docs a bit to hopefully make the action that is happening here somewhat clearer.

### DIFF
--- a/galsim/wcs.py
+++ b/galsim/wcs.py
@@ -546,18 +546,19 @@ class BaseWCS(object):
                 world_pos = self.toWorld(image_pos, color=color)
             return jac.withOrigin(image_pos, world_pos, color=color)
 
-    def withOrigin(self, origin, world_origin=None, color=None):
-        """Recenter the current WCS function at a new origin location, returning the new WCS.
+    def shiftOrigin(self, origin, world_origin=None, color=None):
+        """Shift the origin of the current WCS function, returning the new WCS.
 
-        This function creates a new WCS instance (always a non-local WCS) that treats
-        the origin position the same way the current WCS treats (x,y) = (0,0).
+        This function creates a new WCS instance (always a non-local WCS) that shifts the
+        origin by the given amount.  In other words, it treats the image position ``origin``
+        the same way the current WCS treats (x,y) = (0,0).
 
         If the current WCS is a local WCS, this essentially declares where on the image
         you want the origin of the world coordinate system to be.  i.e. where is (u,v) = (0,0).
         So, for example, to set a WCS that has a constant pixel size with the world coordinates
         centered at the center of an image, you could write::
 
-            >>> wcs = galsim.PixelScale(scale).withOrigin(im.center)
+            >>> wcs = galsim.PixelScale(scale).shiftOrigin(im.center)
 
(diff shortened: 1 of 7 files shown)


galsim/wcs.py
```diff
@@ -546,18 +546,19 @@ def affine(self, image_pos=None, world_pos=None, color=None):
...
-    def withOrigin(self, origin, world_origin=None, color=None):
-        """Recenter the current WCS function at a new origin location, returning the new WCS.
+    def shiftOrigin(self, origin, world_origin=None, color=None):
+        """Shift the origin of the current WCS function, returning the new WCS.
 
-        This function creates a new WCS instance (always a non-local WCS) that treats
-        the origin position the same way the current WCS treats (x,y) = (0,0).
+        This function creates a new WCS instance (always a non-local WCS) that shifts the
+        origin by the given amount.  In other words, it treats the image position ``origin``
+        the same way the current WCS treats (x,y) = (0,0).
```
beckermr, line 554: The last statement in this paragraph confuses me. I think it is technically correct, but the language throws me. IDK if there is anything to do about this though and I don't have a good suggestion.

galsim/wcs.py
```diff
@@ -568,32 +569,38 @@ def withOrigin(self, origin, world_origin=None, color=None):
...
-            >>> wcs3 = wcs.withOrigin(new_origin, new_world_origin)
+            >>> wcs3 = wcs.shiftOrigin(new_origin, new_world_origin)
             >>> world_pos3 = wcs3.toWorld(new_origin)
             >>> # world_pos3 should be equal to new_world_origin
 
         Parameters:
-            origin:         The image coordinate position to use as the origin.
-            world_origin:   The world coordinate position to use as the origin.  Only valid if
+            origin:         The image coordinate position to use for what is currently treated
+                            as (0,0).
```
beckermr, line 586: Ditto here for language but again I don't have a better suggestion.

galsim/wcs.py
rmjarvis: I guess the point I was trying to make is that the "origin" of the current wcs doesn't have to be (0,0).  It is if it's local, but not necessarily otherwise.  So whatever real position on the image is currently called (0,0), the new wcs will call that position `origin`, the argument being passed to this function.  I'm not sure what would be a better sentence to explain that, but if you can think of something that would make that clearer, I'm game to change it.

beckermr: Yeah I’ve got nothing. I’d say merge the pr and I can make another later to change things if I think I can say it more clearly.

rmjarvis: OK.  :)

Issue #1076, opened 2020-03-08
RuntimeError: implement_array_function method already has a docstring


EiffL:
I am encountering errors when trying to install GalSim on a fresh Python:3.7.5 NumPy:1.18.1 environment. From looking around on the internet it looks that it might be a known NumPy issue that arises when the numpy module gets reloaded from some reason. See: https://github.com/numpy/numpy/issues/14384 

Here is  what I see when installing GalSim:
```  
Installed /tmp/pip-install-g78c5tor/galsim/.eggs/eigency-1.77-py3.7-linux-x86_64.egg
    running egg_info
    creating /tmp/pip-install-g78c5tor/galsim/pip-egg-info/GalSim.egg-info
    writing /tmp/pip-install-g78c5tor/galsim/pip-egg-info/GalSim.egg-info/PKG-INFO
    writing dependency_links to /tmp/pip-install-g78c5tor/galsim/pip-egg-info/GalSim.egg-info/dependency_links.txt
    writing entry points to /tmp/pip-install-g78c5tor/galsim/pip-egg-info/GalSim.egg-info/entry_points.txt
    writing requirements to /tmp/pip-install-g78c5tor/galsim/pip-egg-info/GalSim.egg-info/requires.txt
    writing top-level names to /tmp/pip-install-g78c5tor/galsim/pip-egg-info/GalSim.egg-info/top_level.txt
    writing manifest file '/tmp/pip-install-g78c5tor/galsim/pip-egg-info/GalSim.egg-info/SOURCES.txt'
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-g78c5tor/galsim/setup.py", line 1053, in <module>
        zip_safe=False,
      File "/gpfslocalsup/pub/anaconda-py3/2019.10/lib/python3.7/site-packages/setuptools/__init__.py", line 145, in setup
        return distutils.core.setup(**attrs)
      File "/gpfslocalsup/pub/anaconda-py3/2019.10/lib/python3.7/distutils/core.py", line 148, in setup
        dist.run_commands()
      File "/gpfslocalsup/pub/anaconda-py3/2019.10/lib/python3.7/distutils/dist.py", line 966, in run_commands
        self.run_command(cmd)
      File "/gpfslocalsup/pub/anaconda-py3/2019.10/lib/python3.7/distutils/dist.py", line 985, in run_command
        cmd_obj.run()
[...]

Issue #1035, opened 2019-06-24
Sandbox issue with installation on macOS 10.14+


demitri:
Recent macOS releases are restricting access to certain system directories. This prevents GalSim from being installed from source with `python setup.py build` or via `pip install galsim`. (For the record, `conda install -c conda-forge galsim` does work.)

Though poking around, this may not strictly be a macOS issue - as a Mac programmer I have to deal with the Mac Sandbox, which this may or may not be.

This is the output:

```
% python setup.py build
Using setuptools version 41.0.1
Python version =  3.7.3 (default, Mar 27 2019, 16:54:48) 
[Clang 4.0.1 (tags/RELEASE_401/final)]
Adding eigency to build_dep
GalSim version is 2.1.5
Compiling eigency/conversions.pyx because it changed.
Compiling eigency/core.pyx because it changed.
[1/2] Cythonizing eigency/conversions.pyx
/usr/local/anaconda/lib/python3.7/site-packages/Cython/Compiler/Main.py:367: FutureWarning: Cython directive 'language_level' not set, using 2 for now (Py2). This will change in a later release! File: /private/var/folders/yp/lf1xdvp165n5xb7hcrwzpngc0000gn/T/easy_install-9i_ctei3/eigency-1.77/eigency/conversions.pxd
  tree = Parsing.p_module(s, pxd, full_module_name)
[2/2] Cythonizing eigency/core.pyx
/usr/local/anaconda/lib/python3.7/site-packages/Cython/Compiler/Main.py:367: FutureWarning: Cython directive 'language_level' not set, using 2 for now (Py2). This will change in a later release! File: /private/var/folders/yp/lf1xdvp165n5xb7hcrwzpngc0000gn/T/easy_install-9i_ctei3/eigency-1.77/eigency/core.pxd
  tree = Parsing.p_module(s, pxd, full_module_name)
warning: no files found matching 'eigency/*.pyd'
warning: no previously-included files matching 'CMakeLists.txt' found under directory 'eigency/eigen_3.2.8/Eigen'
Traceback (most recent call last):
  File "/usr/local/anaconda/lib/python3.7/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/usr/local/anaconda/lib/python3.7/distutils/dist.py", line 966, in run_commands
    self.run_command(cmd)
  File "/usr/local/anaconda/lib/python3.7/distutils/dist.py", line 985, in run_command
    cmd_obj.run()
[...]

PR #1086, opened 2020-06-14
Rather than eigency, download eigen from internet if it's not already installed.

rmjarvis:
Since version 2.0 let us do `pip install galsim`, installation problems have been much reduced.  The only thing left that seems to be somewhat unreliable about the installation is Eigen.  Specifically my attempt to make it work automatically via eigency.

This doesn't always work, and lately has been hitting a bug in numpy, which they don't seem to have any plans to fix.  cf. #1076.  (See also #1035, and indeed all of the pip installation issues on the [FAQ](https://github.com/GalSim-developers/GalSim/wiki/Installation-FAQ-(pip,-setup.py)#2-installing-with-pip).)  There have been other bug reports people have sent separately, and I usually recommend `conda install eigen`, which works great, but requires conda, so doesn't work for everyone.

Anyway, this PR tries to fix this by getting rid of the eigency option and instead just download eigen from their website.  It's super quick and doesn't require any compiling, since Eigen is header-only.  So I think this will be much more reliable for the cases where people have had troubles with eigency.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -321,8 +321,8 @@ def find_eigen_dir(output=False):
 
     if output:
         print("Could not find Eigen in any of the standard locations.")
-        print("Will now try to download it from the internet.  This requires an internet")
-        print("connect, so will fail if you are currently offline.")
+        print("Will now try to download it from bitbucket.org. This requires an internet")
+        print("connection, so it will fail if you are currently offline.")
         print("If Eigen is installed in a non-standard location, and you want to use that")
         print("instead, you should make sure the right directory is either in your")
         print("C_INCLUDE_PATH or specified in an EIGEN_DIR environment variable.")
@@ -336,7 +336,7 @@ def find_eigen_dir(output=False):
         url = 'https://bitbucket.org/eigen/eigen/get/3.3.4.tar.bz2'
         if output:
             print("Downloading eigen from ",url)
-        u = urlopen('https://bitbucket.org/eigen/eigen/get/3.3.4.tar.bz2')
+        u = urlopen(url)
         fname = 'eigen.tar.bz2'
         block_sz = 32 * 1024
         with open(fname, 'wb') as f:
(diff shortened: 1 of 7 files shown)


rmjarvis:
@EiffL If you have time, could you see if this resolves the trouble you were having with eigency?  Or in general, if you could do a code review, that would be much appreciated.  (Should be quick.)

setup.py
```diff
@@ -309,10 +317,49 @@ def find_eigen_dir(output=False):
...
+        print("Will now try to download it from bitbucket.org. This requires an internet")
+        print("connection, so it will fail if you are currently offline.")
+        print("If Eigen is installed in a non-standard location, and you want to use that")
+        print("instead, you should make sure the right directory is either in your")
+        print("C_INCLUDE_PATH or specified in an EIGEN_DIR environment variable.")
+
+    try:
+        dir = 'downloaded_eigen'
+        if os.path.isdir(dir):
+            # If this exists, it was tried above and failed.  Something must be wrong with it.
```
jmeyers314, line 332: I think a user message like "Previous attempt at downloading Eigen found.  Deleting and starting again." would be appropriate here.

setup.py
```diff
@@ -309,10 +317,49 @@ def find_eigen_dir(output=False):
...
+        if output:
+            print("Downloading eigen from ",url)
+        u = urlopen(url)
+        fname = 'eigen.tar.bz2'
+        block_sz = 32 * 1024
+        with open(fname, 'wb') as f:
+            buffer = u.read(block_sz)
+            while buffer:
+                f.write(buffer)
+                buffer = u.read(block_sz)
```
jmeyers314, line 345: Why the buffering?  More stable?

rmjarvis: I just copied this from galsim_download_cosmos, where the buffering is probably helpful for such a large file.  (And there, I'm sure I just took something from some StackOverflow answer.)  It's probably not needed here though, since I think this tarball is pretty small.  I don't think it hurts though.

setup.py
rmjarvis: Yes.  Good idea.  Done.

PR #1087, opened 2020-06-16
Add ability to draw chromatic objects in config processing.

rmjarvis:
This PR addresses Issue #510 to implement drawing chromatic objects via the config layer.  Mostly, I targeted implementing demo12 as a YAML config file, which certainly gets most of the interesting chromatic functionality.

This required adding SED, Bandpass, and ChormaticAtmosphere to the config processing.  Only minor updates related to the random number sequence were required in demo12.py to get them to match up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -121,10 +121,10 @@ before_install:
     - if [[ $TRAVIS_OS_NAME == "linux" ]]; then sudo -H apt-get -qq update; sudo -H apt-get install -y python-dev libfftw3-dev libeigen3-dev; fi
     # Shouldn't usually need this, but occasionally might.
     #- if [[ $TRAVIS_OS_NAME == "osx" ]]; then brew update; fi
-    - if [[ $TRAVIS_OS_NAME == "osx" ]]; then brew install fftw eigen wget; fi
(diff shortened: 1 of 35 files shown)


rmjarvis:
Thanks Josh.  You're right, that was an oversight.  I added them here.  

The only thing I punted on was the SEDs option to ChromaticRealGalaxy.  I'm not sure how important that is, since none of the regular ChromaticRealGalaxy tests use that directly.  Only through makeFromImages.  So I figured it's probably ok to wait until someone wants that functionality available from the config layer.

I also added a speed optimization to ChromaticOpticalPSF, which is to save the stepk and maxk from the first time it gets evaluated at any wavelength and use that, scaled by 1/wave, for all future calls.  See e5bba65.  I think that's ok, but could use a sign off from you that you don't think that will mess something up.  (Note: an earlier version that didn't do the wavelength scaling failed one of the other ChromaticOpticalPSF tests, but this version seems to be ok as far as I can tell.)

rmjarvis:
Hi Josh, not sure if you wanted to look again at the three additional classes I enabled in response to your comment.  If not, that's fine, but I was leaving this in case you wanted to take a look.

jmeyers314:
LGTM.

PR #1089, opened 2020-06-20
Improve the speed and quality of the Roman PSFs

rmjarvis:
This PR fixes a number of problems with the Roman PSFs in GalSim noticed by @matroxel in his recent simulations of the Roman telescope.

1. Most importantly, the approximate_struts option doesn't really look at all like the 12-spike PSF that we expect. In other words, just putting in 6 radial struts is a bad approximation to the pupil plane mask.
2. Very bright stars are noticeably boxy, with evident reflections in the diffraction spikes.
3. There are also apparent FFT ringing artifacts when approximate_struts=False, along the x and y axes of the star.
4. Setting high_accuracy=True just takes forever and doesn't improve anything.
5. Setting folding_threshold lower helps with the boxiness and folding, but also takes a vv long time.

The fixes implemented here are as follows:

1. Don't do approximate struts.  Instead for an approximate version, take the real pupil plane image and down sample it (using the Image.bin method) to get a smaller image that is as correct as possible at that resolution.
2. Don't increase oversampling or pad_factor for high_accuracy.  Instead lower folding_threshold to get rid of the folding as per the normal meaning of that parameter.
3. Deprecate both high_accuracy and approximate_struts.  To replace the former, suggest that users lower folding_threshold in the normal way.  To replace the latter, let users set the pupil plane binning with the new pupil_bin parameter.  (Default is 4.)
4. Added a workaround for a weird anomaly in the Roman pupil plane images.  There is a square around the main pupil image with amplitude ~0.03.  This eventually gets turned into 1 when cast as a boolean, so it leads to significant ringing in the rendered PSF image.  I deal with it by simply taking everything <0.5 => 0.0, which seems to work.
5. Make it easier to set individual GSParams by writing e.g. `obj.withGSParams(folding_threshold=2.e-3)` rather than `obj.withGSParams(galsim.GSParams(folding_threshold=2.e-3))`, which is almost equivalent. However, the latter syntax isn't always safe, since it loses any other non-default GSParams that `obj` might already have had set.  The new syntax will preserve any non-default values that aren't explicitly overridden in the kwargs.
6. Change the default `ii_pad_factor` for PhaseScreenPSFs from 4 to 1.5.  I realized when working on this that there was no point to setting it to 4.  The whole reason for InterpolatedImage's pad_factor is to accurately render sheared interpolated images.  PSFs are vv rarely sheared and not much if at all.  So the analysis that led us to recommend pad_factor=4 for (galaxies using) InterpolatedImage doesn't apply.  I think 1.0 would be fine, but I chose 1.5 in case there was some need for a little bit of padding.
7. Cache the apertures that get created.  Since the approximate_struts replacement now does a little more work, reading in a non-trivial file, fixing the 0.03 artifact, and then rebinning, there's a bit more overhead than there used to be.  So now I save the resulting Aperture in a cache indexed by the relevant parameters.  In the normal case where the user makes different PSFs with different aberrations, but otherwise the same bandpass, wavelength, gsparams, then it will be able to use the cache.


rmjarvis:
Some comparisons between old and new.  These are a magnitude 10 star drawn with a typical Roman noise level.  The script that made these is in devel/test_roman_psfs.py.

Old (master branch):
old_roman_psf

New (this branch):
new_roman_psf

The four choices of folding_threshold and pupil_bin are what the corresponding high_accuracy and approximate_struts get converted into along with a deprecation warning.  The lower left is the most accurate, and that looks quite good I think.  In all cases, the new version is (usually much!) faster and looks better.

rmjarvis:
Note that the central pixel of this star has a value of more than 1.e8.  And the flux near the box edges, even in the upper right case is <1000, so the relative accuracy is always better than 1.e-5, which is usually the level that we shoot for with these kinds of things.  So the variation among these options is mostly a cosmetic issue for extremely bright things.  The usual faint galaxies we mostly care about will be just find with even the 8x8 binning and normal folding_threshold.

rmjarvis:
@rmandelb I'm still waiting on responses from project re filter throughputs.  I'm still not positive which ones are the really official ones.  I had thought to update those on this PR too, but if you have time to take a look at what's here, we can go ahead and finish this PR up and I'll make another one for the other cycle 8 updates that need to happen, once I figure out what the right values are.

rmjarvis:
Also, one update since I wrote the overview above: I updated the pupil plane images we had here with the new Cycle 8 ones from the project.  They now have separate images for each SCA, so I use that.  (We already had SCA as a parameter for the getPSF function, so I didn't need any API change to allow this.)  They also don't have the weird 0.03 artifact, so I was able to remove that bit of code to fix that up.

rmandelb:
@rmjarvis - I got eaten up by other things this week, and have a full-afternoon Zoom commitment Thurs+Fri.  Realistically I can't do a deep dive into this until the weekend.  Just wanted to let you know in case that affects your thinking about what to include in this versus punting.  Sorry -

rmjarvis:
That's fine.  I think this is a good set of work to review for now.  There'll be more Roman prs this summer with other bits, but we should get this one merged I think.

CHANGELOG.rst
```diff
@@ -56,6 +67,21 @@ New Features
...
+  the resolution of the pupil plane mask, which involves a trade-off between
+  speed and accuracy of the PSF rendering. (#1089)
+
+
+Performance Improvements
+------------------------
+
+- Improved the rendering of Roman PSFs to always show 12 diffraction spikes
+  (rather than 6 in the now-deprecated approximate_struts mode), remove an
+  FFT artifact in the exact pupil plane mode, and significanly speed up all
```
rmandelb, line 83:
```suggestion
  FFT artifact in the exact pupil plane mode, and significantly speed up all
```

examples/demo13.py
```diff
@@ -132,21 +132,18 @@ def main(argv):
     # (diameter, bandpasses, obscuration, etc.).  Note that we arbitrarily choose a single SCA
     # (Sensor Chip Assembly) rather than all of them, for faster calculations, and use a simple
     # representation of the struts for faster calculations.  To do a more exact calculation of the
-    # chromaticity and pupil plane configuration, remove the `approximate_struts` and the `n_waves`
-    # keyword from the call to getPSF():
+    # chromaticity and pupil plane configuration, remove the `n_waves` keyword from the call to
```
rmandelb, line 135:
```suggestion
    # chromaticity, remove the `n_waves` keyword from the call to
```

rmandelb: (at least I think this is true?)

CHANGELOG.rst
```diff
@@ -27,6 +27,17 @@ API Changes
   shiftOrigin is equivalent to withOrigin, which still exists as a valid
   name for this action. (#1085)
 - Deprecated galsim.wfirst module.  Now called galsim.roman. (#1088)
+- Changed the default ii_pad_factor for PhaseScreenPSF and OpticalPSF to 1.5.
+  The old default of 4.0 (matching the InterpolatedImage default) is not
+  necessary for objects that will not be significantly sheared, which
+  PSF objects typically aren't.  For almost all use cases, this will be
+  simply a performance improvement, but if you need the higher ii_pad_factor
+  for some reason, you will now need to manually specify it. (#1089)
```
rmandelb, line 35: Could be I missed it, but I didn't see in the tests that you had any where you had to manually set a higher value.  Does that mean that all of our tests passed with this lower default?  Interesting...

galsim/phase_psf.py
```diff
@@ -1905,7 +1905,7 @@ def __repr__(self):
             s += ", _force_stepk=%r" % self._force_stepk
         if self._force_maxk != 0.:
             s += ", _force_maxk=%r" % self._force_maxk
-        if self._ii_pad_factor != 4.:
+        if self._ii_pad_factor != 1.5:
```
rmandelb, line 1908: I was wondering if we should have these defaults controlled by a named variable (e.g., `default_ii_pad_factor`, `default_oversampling`, etc.) so that we don't have to have lines of code that depend explicitly on their values?

galsim/roman/roman_psfs.py
```diff
@@ -49,22 +56,29 @@ def getPSF(SCA, bandpass,
...
-    is stored in ``galsim.roman.longwave_bands`` and ``galsim.roman.shortwave_bands``).
-
-    To avoid using the full pupil plane configuration, use the optional keyword
-    ``approximate_struts``.  In this case, the pupil plane will have the correct obscuration and
-    number of struts, but the struts will be purely radial and evenly spaced instead of the true
-    configuration.  The simplicity of this arrangement leads to a much faster calculation, and
-    somewhat simplifies the configuration of the diffraction spikes.  Also note that currently the
-    orientation of the struts is fixed, rather than rotating depending on the orientation of the
-    focal plane.  Rotation of the PSF can easily be affected by the user via::
+    The mask images for the Roman pupil plane are available at from the Roman Reference Information
```
rmandelb, line 59:
```suggestion
    The mask images for the Roman pupil plane are available at the Roman Reference Information
```

galsim/roman/roman_psfs.py
```diff
@@ -49,22 +56,29 @@ def getPSF(SCA, bandpass,
...
+    page: https://roman.gsfc.nasa.gov/science/Roman_Reference_Information.html.
+    There are separate files for each SCA, since the view of the spider pattern varies somwhat
```
rmandelb, line 61:
```suggestion
    There are separate files for each SCA, since the view of the spider pattern varies somewhat
```

galsim/roman/roman_psfs.py
```diff
@@ -49,22 +56,29 @@ def getPSF(SCA, bandpass,
...
+    across the field of view of the wide field camera. Furthermore, the effect of the obscuration
+    is somewhat different at longer wavelengths, so F184 has a different set fo files than the
```
rmandelb, line 63:
```suggestion
    is somewhat different at longer wavelengths, so F184 has a different set of files than the
```

galsim/roman/roman_psfs.py
```diff
@@ -49,22 +56,29 @@ def getPSF(SCA, bandpass,
...
+    other filters.  cf. the ``galsm.roman.longwave_bands`` and ``galsim.roman.shortwave_bands``
+    attributes, which define which bands use which pupil plane images.
```
rmandelb, line 65: Maybe add a comment that GalSim automatically takes care of this internally, so users don't think they are being given this info because there is some implied action for them?

galsim/roman/roman_psfs.py
```diff
@@ -49,22 +56,29 @@ def getPSF(SCA, bandpass,
...
+
+    To avoid using the full pupil plane configuration, use the optional keyword ``pupil_bin``.
```
rmandelb, line 67: Wording at first implies you have to set `pupil_bin` to avoid using the full pupil plane configuration; only later is it clear that the default in fact avoids using it, so you have to set `pupil_bin` if you *do* want to use the full pupil plane configuration.  I think this could be cleaned up.

galsim/roman/roman_psfs.py
```diff
@@ -49,22 +56,29 @@ def getPSF(SCA, bandpass,
...
+    The full pupil-plane images are 4096 x 4096, which is more detail than is typically needed for
+    most applications. The default binning is 4x4, which results in an image that is 1024 x 1024.
+    This provides enough detail for most purposes and is much faster to render than using the full
+    pupil plane image.  Using pupil_bin=8 (resulting in a 512 x 512 image) still provides fairly
+    reasonable results and is even faster to render, but it is not recommended to use higher
+    binning than that, as the diffraction spikes will be noticeably degraded.
+
+    Also note that currently the orientation of the struts is fixed, rather than rotating depending
+    on the orientation of the focal plane.  Rotation of the PSF can easily be affected by the user
+    via::
```
rmandelb, line 77:
```suggestion
    via:
```

galsim/roman/roman_psfs.py
```diff
@@ -166,94 +166,136 @@ def getPSF(SCA, bandpass,
...
+
+    aper = Aperture(lam=wave, diam=diameter,
+                    obscuration=obscuration,
+                    pupil_plane_im=pupil_plane_im,
+                    gsparams=gsparams)
+    return aper
+
+# Usually a given run will only need one or a few different apertures for repeated getPSF calls.
+# So cache those apertures here to avoid having to remake them.
+_make_aperture = LRU_Cache(__make_aperture)
```
rmandelb, line 274: Is it worth mentioning this caching in the docstring for getPSF() so that users are aware that the potentially significant setup costs may not apply for each getPSF() call?

tests/test_roman.py
```diff
@@ -613,13 +612,13 @@ def test_roman_psfs():
     # - achromatic PSFs without loading the pupil plane image.
```
rmandelb, line 612: Is this still true?  (We have some caveats about things we do/don't test, but since you fixed the defaults to be more accurate and faster, I think we do test some of what's mentioned here as not being tested?)

CHANGELOG.rst
rmjarvis: Yes.  I'm pretty sure if you're not shearing an interpolated image, then the pad factor doesn't need to be set very high.  Certainly, dropping to 1.5 for all the phase_psf cases didn't cause any problems, probably because we never shear these objects.  When it's used for a PSF, it would be weird to shear it.  (The shape usually comes from aberrations, not an applied shear.) 

The pad factor is important though if you're going to shear the object, like we often do with RealGalaxy for instance, so the default of 4 is preserved in that case, as well as the generic InterpolatedImage class, where we don't know what it will be used for.

galsim/phase_psf.py
rmjarvis: Yes.  Good idea.  I'm usually usually in favor of avoiding magic numbers in code.  Better to give them a name at least.  So thanks for calling out this case where we didn't.

galsim/roman/roman_psfs.py
rmjarvis: This apparent typo is actually Sphinx markup.  The double colon, line skip, and indent is the start of a code block.

rmjarvis:
Thanks Rachel!  I reworked parts of the doc string to clear things up.  And I took your other suggestions as well.  Thanks for the review.

CHANGELOG.rst
rmandelb: Makes sense, thanks.

galsim/roman/roman_psfs.py
rmandelb: Ah ha, thanks for the reminder!

PR #1091, opened 2020-07-15
Test an equivalent to a simple version of fpack/funpack using galsim.fits functions.

rmjarvis:
I was having trouble getting fpack to work on a system, and I realized that GalSim already did what I needed using readMulti and writeMulti.

Except that it didn't quite work correctly.  I discovered that writeMulti didn't properly write the header information when there were non-trivial headers on the images, so the round trip lost the header values I needed from the original files.

So this PR fixes that, and also adds a note about how to effect at least a very simple version of fpack and funpack with galsim, and tests these operations that they preserve at least most of the header values.  (Some get changed by astropy's verify operations, so I skip those.)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -92,3 +92,5 @@ Bug Fixes
   of (r0, L0, lam). (#1058)
 - Fixed minor bug in repr of OpticalPSF class. (#1061)
 - Fixed bug in RandomKnots when multiplied by an SED. (#1064)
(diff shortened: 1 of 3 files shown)


rmjarvis:
This isn't a particularly sexy PR, so I'm fine if no one wants to review it.  But in the interest of not letting it sit here forever, I'll plan to merge this by end of week unless someone objects.

PR #1092, opened 2020-07-19
Add FittedSIPWCS

jmeyers314:
Adds function that fits a TAN-SIP WCS to a list of reference image and celestial coordinates.

For the reference WCS's in tests/fits_files, this gets down to milliarcsec precision with SIP order 7.  The one exception is the ZPN WCS, which is only good to ~arcsec or so.  I'm not sure what's different about this one.

Not visible in this PR, but ~microarcsec precision appears possible for each Rubin Obs CCD using just a 3rd order SIP, so long as tangent plane point is near the center of the CCD.


tests/test_wcs.py
```diff
@@ -2440,6 +2440,96 @@ def test_fitswcs():
...
+            galsim.Angle.from_dms(ref_list[0][1])
+        )
+        fittedWCS = galsim.FittedSIPWCS(
+            x[:50], y[:50], ra[:50], dec[:50],
+            order=7, center=center
+        )
+        ra_test, dec_test = fittedWCS.xyToradec(x[:50], y[:50], units='rad')
+        check_sphere(
+            ra[:50], dec[:50], ra_test, dec_test,
+            atol=60 if tag == 'ZPN' else 10
```
rmjarvis, line 2522: Possibly interesting extension -- you could have the TAN part of this class be an input variable.  It just sets the kind of projection.  ZPN is a different one, and I guess this file is over a large enough area that the difference in the projections isn't easily captured by a 3rd order polynomial at the place where SIP does it.
But if you let TAN be replace by ZEA, I think it would be much better.

galsim/fitswcs.py
```diff
@@ -1761,3 +1763,92 @@ def FitsWCS(file_name=None, dir=None, hdu=None, header=None, compression='auto',
...
+            for j in range(order+1):
+                if (i+j < 2) or (i+j > order):
+                    a.append(0.0)
+                    b.append(0.0)
+                else:
+                    a.append(params[6+k])
+                    b.append(params[6+k+1])
+                    k += 2
+        a = np.array(a).reshape((order+1, order+1))
+        b = np.array(b).reshape((order+1, order+1))
```
rmjarvis, line 1829: I usually like to allocate the array at the start and fill it in if possible.  But I'm not sure if this actually ends up clearer in this case.  (It seems to be irrelevant for running time, and it's not actually many fewer lines of code.)  Anyway, for your consideration, I believe the following is equivalent:
```
        a = np.zeros((order+1,order+1))
        b = np.zeros((order+1,order+1))
        k = 0
        for i in range(order+1):
            # skip i+j < 2 and i+j > order
            j1 = 0 if i > 1 else 2-i
            j2 = order + 1 - i
            a[i][j1:j2] = params[6+k : 6+k+2*(j2-j1) : 2]
            b[i][j1:j2] = params[7+k : 7+k+2*(j2-j1) : 2]
            k += 2*(j2-j1)
        ab = np.array([a, b])
```
Feel free to ignore this though and leave the code as is.

galsim/fitswcs.py
```diff
@@ -1761,3 +1763,92 @@ def FitsWCS(file_name=None, dir=None, hdu=None, header=None, compression='auto',
...
+    from .celestial import CelestialCoord
+    if center is None:
+        # Use deprojected 3D mean of ra/dec unit sphere points as center
+        wx = np.mean(np.cos(dec)*np.cos(ra))
+        wy = np.mean(np.cos(dec)*np.sin(ra))
+        wz = np.mean(np.sin(dec))
+        center = CelestialCoord.from_xyz(wx, wy, wz)
+
+    if order < 1 or order > 9:
+        raise GalSimValueError("Illegal SIP order", order)
```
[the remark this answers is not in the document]
jmeyers314: Yeah, on page 2 [here](https://fits.gsfc.nasa.gov/registry/sip/SIP_distortion_v1_0.pdf) there's a statement: "A_ORDER and B_ORDER can take on integer values ranging from 2 to 9".  But now that you mention it, I can't see any technical reason these couldn't be larger.  I guess maybe we should allow order > 9, but issue a warning if writing these SIP coefficients to a header since other software may not be able to read them?

And `order=1` just means don't do the SIP stuff.  I should double check that the header writer isn't inserting 'TAN_SIP' or 'A_ORDER', 'B_ORDER' in this case though.

tests/test_wcs.py
jmeyers314: Interesting.  I'm not sure I've ever seen `-SIP` attached to anything other than `TAN` (do we have such cases in the current wcs unit tests?), but the SIP distortion document doesn't seem to require `TAN`, AFAICT, so we probably shouldn't either.  Will investigate.

rmjarvis: Ha.  I've never seen any projection other than TAN in any capacity in the wild.  It's always TAN, TPV, TNX, or TAN-SIP, all of which use a gnomonic projection.  These have the property that great circles project to straight lines.  
I've always thought lambert (aka ZEA) would make more sense for astronomy (preserves area), but that doesn't seem to be the one that caught on.  So this is more of academic interest probably than something that will prove useful in practice.

galsim/fitswcs.py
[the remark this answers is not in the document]
rmjarvis: I don't see a good reason for a limit < 100.  At 100, the FITS header keys would start to be more than 8 characters, which means they will require HEIRARCH cards, so bleh.  But <= 9 seems a bit arbitrary and unnecessary.

galsim/fitswcs.py
```diff
@@ -1761,3 +1763,92 @@ def FitsWCS(file_name=None, dir=None, hdu=None, header=None, compression='auto',
...
+        ab = np.array([a, b])
+
+        _data = [
+            'TAN',
+            crpix,
+            cd,
+            center,
+            None, # pv, unused
+            ab,
+            None  # abp, unused
```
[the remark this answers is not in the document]
jmeyers314: Looks like the current `InvertAB` code called downstream of `radecToxy` always uses an iterative approach assuming `ab` is correct and `abp` is just a good initial guess.  This means I can't use this framework to construct a loss function for `abp`, since it doesn't really return a residual.  My guess is just rewriting `radecToxy` in pure python using the candidate `abp` will probably be sufficient here.  Any thoughts though?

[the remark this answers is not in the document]
rmjarvis: Maybe add a bool option to the InvertAB function about whether to do the iterative step?

tests/test_wcs.py
```diff
@@ -2440,6 +2440,96 @@ def test_fitswcs():
...
+        check_sphere(
+            ra[:50], dec[:50], ra_test, dec_test,
+            atol=1 if tag == 'ZPN' else 1e-3
+        )
+        # Now check on some "reserve" points
+        ra_test, dec_test = fittedWCS.xyToradec(x[50:], y[50:], units='rad')
+        check_sphere(
+            ra[50:], dec[50:], ra_test, dec_test,
+            atol=30 if tag == 'ZPN' else 1e-3
+        )
```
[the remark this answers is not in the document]
jmeyers314: Added abp values.  The code without these appears somewhat unstable anyway (frequent "Unable to solve for image_pos (max iter reached)" encountered).  Even with the abp vals I occasionally got errors with a largish SIP order: ~7 or so.  The reverse direction is still ~50-100 times slower.  I suspect [this python for loop](https://github.com/GalSim-developers/GalSim/blob/releases/2.2/galsim/fitswcs.py#L1370) vs vectorized and maybe even MKL-d instructions for the forward direction.

galsim/fitswcs.py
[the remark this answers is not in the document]
jmeyers314: I went ahead and removed this restriction.  To be fair though, I don't think I'd recommend setting `order` this high.  I believe every fit I looked at actually landed on a better solution when order=7 than when order=9.  So either there are numerical difficulties with high-order non-centered, non-Chebyshev/Legendre polynomials (definitely the case in the inverse case, where it's possible to end up with situations like dx~1000 pixels raised to the 9th power say) or maybe there are just more local minima to get trapped in.  I guess we could gradually increase the order when fitting to at least obtain monotonically improving results, but I haven't bothered here since the smallish field-of-view examples all do well with order=3, including LSST sims.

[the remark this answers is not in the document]
rmjarvis: Sounds good.  9 in the SIP spec might just have been this kind of practical limitation rather than something fundamental.  As you say, it's probably a bad idea to go that high anyway.

tests/test_wcs.py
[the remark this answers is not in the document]
rmjarvis: I don't see an obvious reason why I didn't write the InvertAB function to allow it to do all the x,y values at once.  At the worst, the initial application of the ABP matrices could be done faster doing them all at once, and then have a for loop in the C++ layer to do the refinement separately for each position.  Maybe I'll take this PR as an opportunity to do that, unless you want to give it a go.  50-100x is a lot.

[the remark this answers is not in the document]
jmeyers314: I'm on it.  Unless you've already started :)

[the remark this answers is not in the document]
rmjarvis: Nope.  You go for it.  :)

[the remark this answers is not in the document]
jmeyers314: I managed to vectorize the inversion just using numpy (and horner2d).  horner2d also proved faster than the previous forward-direction SIP/PV code that was there.  So now forward direction is ~10% faster and reverse is only ~3-5x slower instead of 50-100x.  

The numpy inversion is a bit suboptimal in that if just one coordinate requires an extra iteration, then the Newton-Raphson update is repeated for every coordinate.  We could probably chunk the input vector up to mitigate that if desired (or move things back down to c++), but I didn't bother.

galsim/fitswcs.py
jmeyers314: Yeah, I like allocate-then-fill generally too.  In this case, though, I do think the current code is clearer, and as you say, not a time-sink.

tests/test_wcs.py
jmeyers314: Okay, I think this is running now too.  Unfortunately, it doesn't seem to help with the fit of the ZPN case.

jmeyers314:
@rmjarvis , I think this is ready now.

galsim/fitswcs.py
```diff
@@ -1343,41 +1340,65 @@ def _radec(self, x, y, color=None):
...
+            # J^-1 . du
+            det = pvudu*pvvdv - pvudv*pvvdu
+            duu = -(du*pvvdv - dv*pvudv)/det
+            dvv = -(-du*pvvdu + dv*pvudu)/det
+
+            uu += duu
+            vv += dvv
+            # Do at least 2 iterations before spending the time to check for
+            # convergence.
+            if iter > 2 and np.max(np.abs(np.array([duu, dvv]))) < 1e-12:
```
rmjarvis, line 1380: I think this does at least 3 iterations?  iter = 0, 1, 2..check.

tests/test_wcs.py
```diff
@@ -2467,21 +2467,19 @@ def test_fittedsipwcs():
...
-        'TAN': (5.0, 0.01),
-        'TSC': (5.0, 0.01),
-        'STG': (5.0, 0.01),
+        'HPX': (5.0, 0.02),
+        'TAN': (5.0, 0.02),
+        'TSC': (5.0, 0.02),
+        'STG': (5.0, 0.02),
         'ZEA': (10.0, 0.1),
         'ARC': (5.0, 0.02),
         'ZPN': (2500.0, 2.5),
```
rmjarvis, line 2479: Maybe worth commenting here (or below where you delve deeper) that this particular image spans a crazy large portion of the sky.  You can open it in ds9 and see.  
The declination goes from about -18 in the upper right corner to -84 in the lower left, which is more than 60 degrees.  RA ranges from 17:50 to 21:44 hours, so more than 45 degrees of RA, which is actually less than 45 degrees across the sky, but still pretty huge.  It's not necessarily surprising that other projections, even with a SIP step will not work so well.
Making this even more difficult, the projection center of the lambert projection for this particular image is off the image.  They project it at the south pole, which is at CRPIX (-183,22).  The image itself is size (192,192), so it's almost a full image size off the left edge of the image, and down near the bottom.  So when you try to use a ZEA projection here, centered at the center of the image, it's not actually a good match to the original, so it doesn't work.  It's not obvious why it would be a little worse than TAN, but since the center of the projection is not close to the original, I don't think we should expect the ZEA to necessarily be better either.
Anyway, all this is just to say that I think having a vv large tolerance for this image is fine and not indicative of a problem anywhere.  It's unfortunate that this test image is so crazy.  It would have been cute to showcase the ZEA-SIP working here.  But unless you want to add an option to give a manual projection center (which I don't think we care about), it looks like we shouldn't expect anything to do a good job on this one.

rmjarvis: BTW, here is a screen shot of that in ds9:
image

jmeyers314: Looks like it's -39 to -78 to me (staying within the 192 pix bounding box) (?), but yeah, still huge.

I dug a little further and found section 5.1.7 of [Calabretta and Greisen II](https://www.aanda.org/articles/aa/pdf/2002/45/aah3860.pdf).  Looks like ZPN is a generalization of ARC, not ZEA.  Also, this particular header has 8 non-zero polynomial correction terms, which I initially figured meant 3rd order or so.  But the CGII paper says the corrections are only in the radial direction, so this is really a 7th order correction.

So I tried `FittedSIPWCS(..., wcs_type='ARC', order=7, center=south_pole)` (forcing a center was already an option :) ), but that didn't do any better :(.  2D order=7 isn't really the same anyway, since there are 14 extra degrees of freedom, and the 2D polynomial can't do odd powers of r.

galsim/fitswcs.py
jmeyers314: Haha, yes.  Silly me.  The error is in the comment, not the code.  Will fix.

jmeyers314: Actually, just checked my timing script and `iter > 1` is a bit quicker, so I'll go with that.

PR #1094, opened 2020-07-31
Update Roman PSF orientation

rmjarvis:
@troxel pointed out a subtlety with the Roman PSFs that we had been doing wrong in demo13 and which he also did wrong in his WFIRST sims.  Namely, the PSF that is returned by `galsim.roman.getPSF` is oriented with respect to the image coordinates, not world coordinates.  

While it is in arcsec (not pixels, so kind of in world coordinates), the pupil plane rotates with the camera, so observations with different PA values don't have the diffraction spikes rotate correctly as the telescope rotates.  This led to coadd images in Troxel's sims having very pronounced spikes, whereas we actually expect them to be washed out by the varying PA angles.

This PR fixes that by adding an optional `wcs` parameter to the `getPSF` function.  To get the PSF in world coordinates (as is standard everywhere else in GalSim), you need to provide the wcs of the image onto which you will be drawing it.  Otherwise, it will return the PSF oriented along the image directions as before.

### DIFF
--- a/galsim/roman/roman_psfs.py
+++ b/galsim/roman/roman_psfs.py
@@ -40,7 +40,8 @@ pupil_plane_file_longwave = '_full_mask.fits.gz'
 pupil_plane_file_shortwave = '_rim_mask.fits.gz'
 
 def getPSF(SCA, bandpass,
-           SCA_pos=None, pupil_bin=4, n_waves=None, extra_aberrations=None,
+           SCA_pos=None, pupil_bin=4, wcs=None,
+           n_waves=None, extra_aberrations=None,
            wavelength=None, gsparams=None,
            logger=None, high_accuracy=None, approximate_struts=None):
     """Get a single PSF for Roman ST observations.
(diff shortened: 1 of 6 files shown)


rmjarvis:
Note: I confirmed with the project that the pupil plane orientation is along the SCA directions, not the observatory directions, which makes this simpler than it might have been.  cf. https://github.com/spacetelescope/webbpsf/issues/374

galsim/roman/roman_psfs.py
```diff
@@ -86,13 +87,13 @@ def getPSF(SCA, bandpass,
...
-    on the orientation of the focal plane.  Rotation of the PSF can easily be affected by the user
-    via::
-
-       psf = galsim.roman.getPSF(...).rotate(angle)
-
-    which will rotate the entire PSF (including the diffraction spikes and all other features).
+    The PSF that is returned by default will be oriented with respect to the SCA coordinates,
+    not world coordinates as is typical in GalSim.  The pupil plane has a fixed orientation
+    with respect to the focal plane, so the PSF rotates with the telescope.  To obtain a
+    PSF in world coordinates, which can be convolved with galaxies (who are normally described
```
rmandelb, line 93:
```suggestion
    PSF in world coordinates, which can be convolved with galaxies (that are normally described
```

tests/test_roman.py
```diff
@@ -812,6 +813,36 @@ def test_roman_basic_numbers():
...
+    wcs_dict = galsim.roman.getWCS(PA=PA, world_pos=world_pos, date=date)
+    wcs = wcs_dict[5]
+
+    # Get the PSF in real world coordinates with this wcs
+    psf = galsim.roman.getPSF(SCA=5, bandpass='F184', wavelength=1950., pupil_bin=8,
+                              SCA_pos=image_pos, wcs=wcs)
+    # Draw on an image with this wcs.
+    im_wcs = psf.drawImage(bounds=im_scale.bounds, wcs=wcs, center=image_pos)
+
+    np.testing.assert_allclose(im_wcs.array, im_scale.array)
```
rmandelb, line 844: I am actually quite confused... why does this work?  My reading is the first case `im_scale` is in image coords, and the second case `im_wcs` is in world coordinates for one particular date and time and PA, so why would they match?

rmjarvis: Because it was also drawn onto an image that had the same WCS.  So the first time, we built it directly in image coordinates and drew it onto an image with just a pixel_scale WCS.  

The second time, we tell it to project onto sky coordinates according to some WCS.  But we draw it onto an image with that same WCS.  So we should get the same thing in the end as the first case, but this time it round tripped through sky coordinates.

The point is that the PSF really does always look the same on the SCA, regardless of where the sky is.  So if we want to have a PSF in sky coordinates (as we normally do), then we need to tell it the WCS we will eventually draw onto so that it can project to the sky correctly, and then drawImage will project it back.

rmandelb: Thank you!

PR #1095, opened 2020-08-04
Remove the _nobjects_only mechanism from input objects

rmjarvis:
**Background**:

One of the features of input objects in the config layer is that if they have a well-defined number of objects in them (e.g. an input catalog for instance), then the code can automatically use that to set the number of objects to render in the simulation.  Namely, doing all of them in turn.

Furthermore, sometimes the config layer needs to know the number of objects that will be rendered before it needs the rest of the information in the input object.  Say when planning out the number of objects that it will need for each image, so it can plan out the right random number seeds to send to each of several jobs.  If the input object takes a fair amount of time to load in (or potentially uses a lot of memory once it's fully loaded), it would be nice to only load in enough to get the number of objects for the cases where that's all we need at the moment.

**Current Code**:

The mechanism that the current config code uses to deal with that case efficiently is to require that each input object that can return a number of objects must have an optional construction parameter, `_nobjects_only`, which when set to True will only do what is necessary to figure out nobjects, and not bother loading in the rest of the data.

This mechanism works ok if the particular input data has a quick way to figure out the number of objects without having to load in everything.  The RealGalaxyCatalog was the paradigm I used when designing this.  The parametric catalog records the total number of objects.  So there is no need to load the real galaxy data if all you need is the number of objects.

**The Problem**:

The instance catalog we read in in imsim doesn't work like this at all.  It takes a *very* long time to read in, even if all you're doing is counting which objects fall in the image in question.  Furthermore, the nobjects special construction is actually called twice before the real load, so it ends up reading in the instance catalog essentially in full 3 times before getting to actually do anything with it.

Also, the current mechanism of requiring a special (nominally private) construction argument is kind of bad form anyway.  It was a hack that was probably ill-advised from the start.  Most simulation runs will eventually need to load in the full input data anyway, and I suspect it is rare that people run versions where lots of different input objects need to compute their number of objects so it can plan out a long complicated sim run.  I think most people use command-line arguments to set things appropriately for a single image or small set of images using a single input catalog.  So this optimization was probably premature.

**The Solution**:

The better solution to this is to let the input class itself delay loading in the rest of the data if it's useful to do so.  I refactored the Catalog and RealGalaxyCatalog classes to use lazy properties to only load the data once they are needed.  The constructor just does enough to be able to return the number of objects, so if that's all that ever is needed, it doesn't waste time.  But if we only have the one input object, then there is only ever one call to the constructor.  Not 3.  So for the instance catalog situation, it doesn't waste time there.

### DIFF
--- a/galsim/fitswcs.py
+++ b/galsim/fitswcs.py
@@ -1336,61 +1336,39 @@ class GSFitsWCS(CelestialWCS):
             assert x.shape == y.shape
             return ra, dec
 
-    def _invert_pv(self, u, v, pv=None, uu=None, vv=None):
-        # Do the inversion via Newton-Raphson iteration.
-        # We're solving u, v = horner2d(u', v', pv) for u' and v' given u and v.
-        # uu and vv are optional initial guesses, if absent, use u and v.
-        if pv is None:
-            pv = self.pv
-        # Assemble horner2d coefs for derivatives
-        _, nx, ny = pv.shape
-        uducoef = (np.arange(nx)[:,None]*pv[0])[1:,:-1]
-        udvcoef = (np.arange(ny)*pv[0])[:-1,1:]
-        vducoef = (np.arange(nx)[:,None]*pv[1])[1:,:-1]
-        vdvcoef = (np.arange(ny)*pv[1])[:-1,1:]
-        if uu is None:
-            uu = u.copy()
-        if vv is None:
-            vv = v.copy()
+    def _invert_ab(self, u, v, ab, x=None, y=None, abp=None):
+        # This is used both for inverting (u,v) = PV (u',v')
+        # and for inverting (x,y) = AB (x',y')
+        # Here (and in C++) the notation is (u,v) = AB(x,y), even though both (u,v) and (x,y)
+        # in this context are either in CCD coordinates (normally called x,y) or tangent plane
+        # coordinates (normally called u,v).
+        # The provided x and y are optional initial guesses, if absent, use u and v.
+        # abp is an optional set of coefficients to make a good guess for x,y
+        if x is None:
+            x = u.copy()
+        if y is None:
+            y = v.copy()
 
-        # Usually converges in ~3 iterations.
-        for iter in range(10):  # pragma: no branch
-            # Want Jac^-1 . du
-            # du
-            du = horner2d(uu, vv, pv[0], triangle=True) - u
-            dv = horner2d(uu, vv, pv[1], triangle=True) - v
-            # J
-            pvudu = horner2d(uu, vv, uducoef, triangle=True)
-            pvudv = horner2d(uu, vv, udvcoef, triangle=True)
-            pvvdu = horner2d(uu, vv, vducoef, triangle=True)
-            pvvdv = horner2d(uu, vv, vdvcoef, triangle=True)
-            # J^-1 . du
-            det = pvudu*pvvdv - pvudv*pvvdu
-            duu = -(du*pvvdv - dv*pvudv)/det
-            dvv = -(-du*pvvdu + dv*pvudu)/det
+        uu = np.ascontiguousarray(u)  # Don't overwrite the give u,v, since we need it at the end
+        vv = np.ascontiguousarray(v)  # to check it we were provided scalars or arrays.
+        x = np.ascontiguousarray(x)
+        y = np.ascontiguousarray(y)
 
-            uu += duu
-            vv += dvv
-            # Do at least 3 iterations before spending the time to check for
-            # convergence.
-            if iter >= 2 and np.max(np.abs(np.array([duu, dvv]))) < 1e-12:
-                break
-        else:  # pragma: no cover
-            raise GalSimError("Unable to solve for image_pos.")
-        return uu, vv
+        with convert_cpp_errors():
+            nab = ab.shape[1]
+            nabp = abp.shape[1] if abp is not None else 0
+            _galsim.InvertAB(len(x), nab, uu.ctypes.data, vv.ctypes.data, ab.ctypes.data,
+                             x.ctypes.data, y.ctypes.data, self._doiter,
+                             nabp, abp.ctypes.data if abp is not None else 0)
 
-    def _invert_ab(self, x, y):
-        if self.abp is not None:
-            xx = horner2d(x, y, self.abp[0])
-            yy = horner2d(x, y, self.abp[1])
+        # Return the right type for u,v
+        try:
+            len(u)
+        except TypeError:
+            return x[0], y[0]
         else:
-            xx = x.copy()
-            yy = y.copy()
-        if not self._doiter:
-            return xx, yy
+            return x, y
 
-        x, y = self._invert_pv(x, y, pv=self.ab, uu=xx, vv=yy)
-        return x, y
 
     def _xy(self, ra, dec, color=None):
         u, v = self.center.project_rad(ra, dec, projection=self.projection)
@@ -1401,7 +1379,7 @@ class GSFitsWCS(CelestialWCS):
         v *= factor
 
         if self.pv is not None:
-            u, v = self._invert_pv(u, v)
+            u, v = self._invert_ab(u, v, self.pv)
 
         if not hasattr(self, 'cdinv'):
             self.cdinv = np.linalg.inv(self.cd)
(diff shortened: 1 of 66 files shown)


rmjarvis:
OK, I put a few more optimizations here that turned up when working on the imsim config run:

1. The biggest was changing the `SED.atRedshift()` method to use the processed `_spec`, rather than `_orig_spec`.  `_orig_spec` is the original spectrum as passed into the `self` SED.  So in imsim, this is a file name.  This means that every time we did `atRedshift`, we had to re-read the file, which takes like 0.1 sec.  So massively dominant for most of the faint objects.  Switching to `_spec` is much faster, since we cache the finite number of raw SEDs that we redshift to various values.  This broke a unit test related to pickling, but I think not a reason to prefer the old behavior.
2. The RandomKnots generation had been dominated by the `rng.duplicate()` calls, which surprised me, since in my head I considered that a trivial op, so didn't ever consider trying to minimize use of it.  We had been implementing that via the serialization string, which it turns out isn't particularly fast.  Unfortunately, the boost rng code doesn't have something really straightforward to use instead, so the thing I settled on is kind of a hack.  It makes a new rng, and then memcpy's the old state over top of the new one.  It seems to work, and I don't *think* there are any potential problems with it.  But it feels a little dirty.  However, this implementation is about 100x faster than using serialize.  So now it's probably fine to use `rng.duplicate()` semi-gratuitously without thinking about efficiency considerations.
3. I changed some of the details of the LoggerProxy and LoggerWrapper implementations to get rid of some gratuitous communication over the multiprocessing pipes that wasn't really needed.  Surprisingly, the pipe communication for `logging.debug` statements was significant, since it had to send something over the pipe to ask if the logger was enabled for debugging (it wasn't) to then decide not to send the string that would have been emitted.  So now the local copy (the LoggerWrapper) knows how to decide this for itself and only use the pipe if there is actually a string to be emitted by the source logger.

The remaining tall poles are the flux calculation (dominated by `SED.calculateFlux`) and the calculation of (x,y) from (ra,dec) (dominated by `GSFitsWCS._invert_ab`).  I'll look into those next, and if there's something obvious to do to speed them up, I'll do that here too.  So if anyone was planning to code review this, maybe wait a bit.  I may have a couple more additions.

rmjarvis:
More updates:

1. I discovered that `multiprocessing` pipes can't hold more than 32767 messages.  This was leading to deadlocks, since the main process was pushing all the tasks into the worker pipe before starting to watch for and process the results pipe.  But will >100K jobs, all the pipes were full and it couldn't move on.  I already had a mechanism for bundling multiple jobs per task, so the solution was just to do that so the total number of tasks was less than this limit.
2. Another multiprocessing problem was that on nersc, there are 64 or so cores, so I was running with that many processes.  However, each one of these wanted to start 64 threads for doing OpenMP and various numpy blas calls.  For OpenMP, it was easy to switch to single-thread mode when in multiprocessing.  The blas one was harder.  For that you need to manually set environment variables OPENBLAS_NUM_THREADS=1 or MKL_NUM_THREADS=1 for the whole execution (they need to be set before `import numpy` to work, so there's no way to toggle it off and on inside the python program using `os.environ`). Or (this is actually the preferred solution) have `threadpoolctl` installed.  This is a python package designed for just this purpose.  So if it is installed, GalSim will now use it to limit those things to 1 thread each.
3. The `invert_ab` function was taking around 25% of the total time.  It turns out the changes Josh made recently to vectorize it for array inputs made it slower for scalar inputs.  So I went back to implementing this in C++, but using Josh's new, better algorithm using horner2d, rather than the old one we used to have.  Now it's fast for both scalar and array inputs.  About 2x faster than the python horner2d version in Josh's `time_wcs.py` test with array inputs, since it skips some of the temporary array allocations.  And almost 10x faster for scalar inputs (i.e. in the imsim run).
4. I added a few more "leading underscore" functions to avoid some sanity-test cruft: `_lens`, `_PositionD`, and `_PositionI` are now available as faster versions of the regular method/classes without some of the parsing and error checking of the inputs.
5. I refactored `combine_wave_lists` a bit to be faster when the input has a lot of elements outside the final wavelength range, which is the case for the imsim seds.  They extend very far bluer and redder than the bandpass range we're working with, so there was some repeated looping through these values, which end up getting cut out at the end, so it's wasted work.  Now it calculates the limits first and cuts everything to that range immediately before doing the unions.  I think this will be a fairly typical situation, so it should be a broadly helpful optimization.

The next tall pole I think is `SED.calculateFlux(bandpass)`.  This is now taking about 13% of the total time, which feels like a lot.  (It was over 20% before the `combine_wave_lists` speedup.)  I think I have an idea of a better algorithm for this, but it will be a fairly significant bit of new functionality, so I'll do that on a new PR.  I've added enough to this one.  :)

tests/test_wcs.py
```diff
@@ -2488,7 +2488,12 @@ def test_fittedsipwcs():
 
     dir = 'fits_files'
 
-    nref = 100
+    if __name__ != "__main__":
+        # For pytest runs, don't bother with the crazy ones.  We really only care about
+        # testing that the code works correctly.  That can be done with just a couple of these.
+        test_tags = ['TAN', 'TPV', 'SIP', 'TNX']
+
+    nref = 300  # Use more than 256, since that's the block length in C++.
```
rmjarvis, line 2496: Note: I wanted to exercise the n>256 version in the C++ code, so I bumped nref to 300.  This was useful, since it found a bug initially in my looping there.  But once I got things working, some of the new objects failed the old tolerances.  So that's why some of those numbers above changed.  I don't think this is pointing to any bug.  Just that there are more objects being extrapolated a little farther from where the constraints are.

galsim/real.py
```diff
@@ -641,13 +673,23 @@ def getIndexForID(self, id):
...
+            self._preload = False  # Once we've loaded them.  Don't do it again.
+
     def preload(self):
         """Preload the files into memory.
 
         There are memory implications to this, so we don't do this by default.  However, it can be
         a big speedup if memory isn't an issue.
         """
         from ._pyfits import pyfits
+        self.loaded_lock.acquire()
```
jmeyers314, line 692: I see a comment later in this file that the context manager syntax used to not work?  I definitely used in in the [phase_screens stuff](https://github.com/GalSim-developers/GalSim/blob/releases/2.2/galsim/phase_screens.py#L343) though, so maybe that means now it's safe to switch?

include/galsim/Silicon.h
```diff
@@ -81,6 +81,7 @@ namespace galsim
     };
 
     int SetOMPThreads(int num_threads);
+    int GetOMPThreads();
```
jmeyers314, line 84: It feels slightly awkward that these live in Silicon.h.  Would they be better placed in, say, Std.h?

rmjarvis: The declarations can be moved to Std.h, but the instantiations need to live in some .cpp file.  Since Silicon.cpp is the only one using OpenMP at the moment, that seemed the obvious place.  And given that (and that no C++ code ever calls these aside from the pybind11 wrapper), this seems like the right place to put the declarations too.

galsim/fitswcs.py
```diff
@@ -1451,15 +1422,14 @@ def _local(self, image_pos, color=None):
             order = len(self.ab[0])-1
             xpow = x ** np.arange(order+1)
             ypow = y ** np.arange(order+1)
-            p1 += np.dot(np.dot(self.ab, ypow), xpow)
+            p1 = np.dot(np.dot(self.ab, ypow), xpow)
```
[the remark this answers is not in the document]
rmjarvis: I thought so.  I tried it, but it made it slower.  Probably because x,y are always scalars here, so the overhead of turning things into numpy arrays of length 1 in the horner2d code ended up being more than making the short xpow and ypow arrays.

galsim/real.py
rmjarvis: Looks like.  I don't really remember the context where I had problems before.  That comment stems from a commit in 2013 (e1d51a4f53e985).  So could even have been a py2.6 problem.  I switched them all to use the `with lock:` syntax.  Everything seems to work fine on my machine for both 2.7 and 3.7.

galsim/fitswcs.py
```diff
@@ -1332,65 +1333,35 @@ def _radec(self, x, y, color=None):
...
-        if not self._doiter:
-            return x+dx, y+dy
-
-        # Turn an ab matrix into a pv matrix
-        pv = self.ab.copy()
-        pv[1,0,1] += 1
-        pv[0,1,0] += 1
-        x, y = self._invert_pv(x, y, pv=pv, uu=x+dx, vv=y+dy)
-        return x, y
+            return x, y
```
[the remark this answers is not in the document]
rmjarvis: Good catch.  `len(x)` is the wrong thing for nx in that case.  Everything is fine if we use `nx=len(x.ravel())` there.

PR #1096, opened 2020-08-17
ENH shear the position

beckermr:
This PR adds a shear method to the Position objects so that they can be sheared.


rmjarvis:
Retarget this to master.  Not releases/2.2.

rmjarvis:
And probably rebase it from the current master.

beckermr:
Ahhh I did both, but didn't know the workflow here. See https://github.com/GalSim-developers/GalSim/pull/1097

closing!

PR #1097, opened 2020-08-17
ENH shear the position

beckermr:
This PR adds a shear method to the Position objects so that they can be sheared.

### DIFF
--- a/tests/test_shear_position.py
+++ b/tests/test_shear_position.py
@@ -68,5 +68,29 @@ def test_shear_position():
     )
 
 
+@timer
+def test_shear_position_image_integration():
+    wcs = galsim.PixelScale(0.3)
+    obj1 = galsim.Gaussian(sigma=3)
+    obj2 = galsim.Gaussian(sigma=2)
+    pos2 = galsim.PositionD(3, 5)
+    sum = obj1 + obj2.shift(pos2)
+    shear = galsim.Shear(g1=0.1, g2=0.18)
+    im1 = galsim.Image(50, 50, wcs=wcs)
+    sum.shear(shear).drawImage(im1, center=im1.center)
+
+    # Equivalent to shear each object separately and drawing at the sheared position.
+    im2 = galsim.Image(50, 50, wcs=wcs)
+    obj1.shear(shear).drawImage(im2, center=im2.center)
+    obj2.shear(shear).drawImage(
+        im2,
+        add_to_image=True,
+        center=im2.center + wcs.toImage((pos2).shear(shear)),
+    )
+
+    assert np.allclose(im1.array, im2.array, rtol=0, atol=5e-8)
+
+
 if __name__ == "__main__":
     test_shear_position()
+    test_shear_position_image_integration()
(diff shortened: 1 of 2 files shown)


rmjarvis:
Right now, your unit tests is a recapitulation of the implementation, which means it's more like a regression test than a test of correctness.

I'd suggest adding a test that gets more at the point of this featue.  Namely that shearing a sum of two objects does the same thing as shearing the position and drawing them separately:
```
obj1 = galsim.Gaussian(sigma=3)
obj2 = galsim.Gaussian(sigma=2)
pos2 = galsim.PositionD(3,5)
sum = obj1 + obj2.shift(pos2)
shear = galsim.Shear(g1=0.1, g2=0.18)
im1 = galsim.Image(50,50, scale=0.3)
sum.shear(shear).drawImage(im1, center=im1.center)

# Equivalent to shear each object separately and drawing at the sheared position.
im2 = galsim.Image(50,50, scale=0.3)
obj1.shear(shear).drawImage(im2, center=im2.center)
obj2.shear(shear).drawImage(im2, add_to_image=True, center=im2.center + (pos2-im2.center).shear(shear))

np.assert_allclose(im1.array, im2.array)
```
I didn't run this, so there are surely errors in this.  But I think this would confirm that the implementation is at least doing the same this as other ways we use a shear when drawing.  It would also be a useful reference for people (e.g. you) wanting to use this feature.

Even better would be to repeat this on an image with a non-trivial (but uniform) WCS to make sure you know the right way to apply this in the presence of a WCS.  Indeed, it may prompt a further feature addition in the WCS class to handle some of the boilerplate that I think is probably involved there.

beckermr:
Happy to add more beefy tests!

tests/test_shear_position.py
```diff
@@ -0,0 +1,96 @@
...
+    im1 = galsim.Image(50, 50, wcs=wcs)
+    sum.shear(shear).drawImage(im1, center=im1.center)
+
+    # Equivalent to shear each object separately and drawing at the sheared position.
+    im2 = galsim.Image(50, 50, wcs=wcs)
+    obj1.shear(shear).drawImage(im2, center=im2.center)
+    obj2.shear(shear).drawImage(
+        im2,
+        add_to_image=True,
+        center=im2.center + wcs.toImage((pos2).shear(shear)),
```
beckermr, line 88: There is some subtly here in that you have to apply the shear in the tangent plane and not in image coords.

beckermr:
ok @rmjarvis back to you!

tests/test_shear_position.py
```diff
@@ -0,0 +1,131 @@
...
+    obj2.shear(shear).drawImage(
+        im2,
+        add_to_image=True,
+        center=wcs.toImage(pos2.shear(shear)),
+    )
+
+    print("err:", np.max(np.abs(im1.array - im2.array)))
+    assert np.allclose(im1.array, im2.array, rtol=0, atol=2e-7), (
+        np.max(np.abs(im1.array - im2.array))
+    )
```
rmjarvis, line 125: This is fine, but rather than give the extra information in a comma message of the assert, I usually use `np.testing.assert_allclose` which gives a nice message for failures about how many points didn't meat tolerance, and the maximum rel and abs error.

No need to change this though.  Just FYI.

Issue #540, opened 2014-04-29
Photon shooting for Chromatic objects


rmjarvis:
In #535, I updated the drawing routines in chromatic.py to use the new drawImage function, which includes the option to do photon shooting.  I thought this might mean that the chromatic objects would now be rendered correctly with method='phot'.  But thinking more about it, I'm not so sure that the integration stuff is really done correctly when photon shooting.

The right way to do it would be to draw each photon from the SED of the object, scale its flux by the bandpass throughput, and then shoot it onto the image.  At least for separable objects.  It gets more complicated when they aren't separable.

Anyway, this issue is for someone to investigate photon shooting of chromatic objects and make sure it is working correctly (or probably make it work correctly if, as I suspect, it currently isn't).

cwwalter:
A Duke student (@jiwonpark93) is going to start to look at this by building up a image chromatically with photon shooting and comparing with the fft results.

cwwalter:
A quick recap of @jiwonpark93's work so far on implementing chromatic photon shooting in GalSim. By sampling random numbers from a SED, freezing the chromatic object at that frequency and then shooting photons at each frequency she could reproduce simple FFT convolved cases.   In order to get the normalization to work properly she needs to account for the fact that drawImage wants to preserve the total flux of the chromatic object and scale the SED before shooting appropriately.  A Chi2 test shows the expected agreement for given statistics between the two methods on the resulting 2d objects.

She is now working on comparing more complicated compound objects along with chromatic PSFs etc.  Also, after suggestion from @rmandelb she is refactoring the code to better separate the objects and SEDs to more naturally handle compound objects. After the code works in the python layer, the eventual goal is to migrate it into a fast C++ version.  More progress is expected later in the summer.

cwwalter:
I reported on updates to this work by @jiwonpark93 in November, but it was on slack (in the desc-ssim channel) and I think it fell through the cracks a bit because everyone was at the DESC hackweek.  I'm transferring the information to here to get things back on track and to make sure that it is not lost.

------

Ji Won completed a python version of her chromatic photon shooting code as we discussed in this issue above.   It is in:

https://github.com/DukeCosmology/photon-shooting

It is in a slightly different form than we originally discussed (it is in python notebooks) but the skeleton for the code that should be translated into C++ now exists.  Basically this code repeatedly freezes a chromatic object as samples by a SED and then does monochromatic photon shooting from the frozen object thereby building up proper chromatic photon shot distribution.

[...]

PR #1100, opened 2020-09-15
#540 Implement native photon shooting for chromatic profiles

rmjarvis:
The point of this PR is to make photon shooting work better for ChromaticObjects.  Right now, it basically does the same calculation as for fft rendering, where it draws the whole profile at a set of wavelengths and then numerically integrates this to get the net result.  This method is fine for fft drawing, but it's clearly the wrong way to go for photon shooting.

The better approach, which we've known for a long time, is to start with the spectral component (the galaxy) and draw that in the normal way to generate the PhotonArray, assign the right wavelengths to all those photons, and then let each chromatic PSF do its convolution in a wavelength-dependent way.

This turns out to be very similar to what we already do with PhotonOps.  So the first thing this PR does is to allow GSObjects and ChromaticObjects to duck type as PhotonOps.  They can be put into a `photon_ops` list and their convolution will take place at that point in the series of operations.

Making this work for GSObjects was pretty simple, since it's the same procedure as we were already doing in the Convolution shoot function, so I just copied that over.  For ChromaticObjects, we didn't have shoot functions yet, so I had to write those.  ChromaticAiry is basically exact.  ChromaticOpticalPSF is an approximation, but I think not a worse one than we are already doing with geometric_shooting=True.

Then the last step was to have the ChromaticConvolution.drawImage function automatically convert a convolution into a single driver object (the galaxy -- identified as the one with a spectral SED) with all the other (PSF) items moved to a photon_ops list after a WavelengthSampler op.

One slight nuance that turned up is how we do the flux per photon when photon shooting.  Usually it's 1, but for InterpolatedImage, it's not.  We used to handle that with the g value in drawPhot.  But I switched that to have `InterpolatedImage._shoot` apply the scaling so that it works both when it's being run from the drawPhot directly and also when it's applied in a photon_ops list.  There are kind of a lot of lines touched by that change, but it's not really a big functionality change.  It only really affects the new applyTo usage.


tests/test_chromatic.py
```diff
@@ -2547,6 +2706,9 @@ def test_atredshift():
     image1 = final1.drawImage(nx=64, ny=64, scale=0.2, bandpass=bandpass)
     image2 = final2.drawImage(nx=64, ny=64, scale=0.2, bandpass=bandpass)
     np.testing.assert_allclose(image1.array, image2.array, atol=1.e-6)
+    assert gal1.redshift == 0.
```
jmeyers314, line 2709: Not this PR, but appears that this function is `atredshift` but called as `atRedshift` in the ifmain block.

tests/test_interpolatedimage.py
```diff
@@ -1532,9 +1532,18 @@ def test_ii_shoot():
...
-        photons2 = obj.makePhot(poisson_flux=False, rng=rng)
+        photons2 = obj.makePhot(poisson_flux=False, rng=rng.duplicate())
         assert photons2 == photons, "InterpolatedImage makePhot not equivalent to drawPhot"
 
+        # Can treat as a convolution of a delta function and put it in a photon_ops list.
+        delta = galsim.DeltaFunction(flux=flux)
+        psf = galsim.InterpolatedImage(image_in, x_interpolant=interp, scale=3.3, flux=1)
+        photons3 = delta.makePhot(poisson_flux=False, rng=rng.duplicate(), photon_ops=[psf],
+                                  n_photons=len(photons))
+        np.testing.assert_allclose(photons3.x, photons.x)
```
jmeyers314, line 1543: Why is this one (and a few others) `allclose` but not `==` like the rest?

galsim/chromatic.py
```diff
@@ -422,6 +422,9 @@ def drawImage(self, bandpass, image=None, integrator='quadratic', **kwargs):
         Returns:
             the drawn `Image`.
         """
+        from .table import LookupTable
```
jmeyers314, line 425: unused I think.

galsim/chromatic.py
```diff
@@ -435,6 +438,12 @@ def drawImage(self, bandpass, image=None, integrator='quadratic', **kwargs):
         # determine combined self.wave_list and bandpass.wave_list
         wave_list, _, _ = utilities.combine_wave_list(self, bandpass)
 
+        # If there are photon ops, they'll probably need valid wavelengths, so add
+        # WavelengthSampler as the first op in the list.
+        if 'photon_ops' in kwargs:
+            wave_sampler = WavelengthSampler(self.SED, bandpass)
+            kwargs['photon_ops'] = [wave_sampler] + kwargs['photon_ops']
```
jmeyers314, line 445: Does this make it illegal to pass in a `WavelengthSampler` explicitly?  Or does the second one become a no-op?

galsim/chromatic.py
```diff
@@ -2942,16 +3045,26 @@ def evaluateAtWavelength(self, wave):
                     lam=wave, diam=self.diam,
                     aberrations=self.aberrations*wave_factor, scale_unit=self.scale_unit,
                     _force_stepk=self._stepk*wave_factor, _force_maxk=self._maxk*wave_factor,
-                    gsparams=self.gsparams, **self.kwargs)
+                    geometric_shooting=True, gsparams=self.gsparams, **self.kwargs)
```
jmeyers314, line 3048: This is probably fine, as trying to do anything with `geometric_shooting=False` would be really slow.  But I think it's worth mentioning in the docstring that we set this here.  I can imagine someone wanting to see diffraction spikes in their hyper realistic chromatic optics image and being confused when they don't show up.

jmeyers314 (approved):
LGTM.

tests/test_interpolatedimage.py
rmjarvis: I can't remember all the details now, but some of the calculations did things in slightly different orders, so things could be different at the 1.e-15 level.  Others involved different functions doing precisely the same calculations, so they were exactly equal.  I started with all of them as `photonsx == photons`, but this must have been one that was basically the same answer, but not precisely identical.  So I had to pull it apart and check the components separately.

galsim/chromatic.py
rmjarvis: It wouldn't be a no-op.  But it would be redundant.  What do you think?  Should we check for that and not add the WavelengthSampler if there already is one.  Or warn the user?  Or maybe WavelengthSampler itself should issue a warning if wavelengths are already set and you're overwriting them?

galsim/chromatic.py
rmjarvis: We could make it a class parameter maybe?  But I'm not sure how to write the photon shooting bit in that case.  Do you think it would be feasible to write the `applyTo` function if it's not using the geometric approximation?

jmeyers314: I think the algorithm is FFT for every wavelength and photon-shoot from that unfortunately.  So possible but immensely slow.  A potential alternative would be to figure out how to enable photon shooting from `InterpolatedChromaticObject` (which aside, I think we mis-named.  Every other chromatic objects starts with `Chromatic` and I keep guessing the wrong name for this one :( ). But I'm not sure what that algorithm should look like...

galsim/chromatic.py
jmeyers314: Warning-if-wavelengths-are-set sounds good.

galsim/chromatic.py
rmjarvis: I guess maybe slightly better than the current approximation if people want diffraction spikes would be to still only evaluate it at one wavelength and rescale the x,y values proportional to lambda the way I currently am, but have that base evaluation shoot using geometric_shooting=False.
So then we can have this class have a geometric_shooting option just like the regular OpticalPSF and other PhaseScreenPSFs.  The caveat would be that in this case the chromatic shooting is not expected to be as close a match as when the explicit integral is done.  I don't have a good feel for how big a difference there is in this case.  (The math clearly doesn't have a lambda factor just drop out like it did in the geometric_shooting=True case, but maybe it's not so far off.)

jmeyers314: Here's a movie I made a few years ago showing the wavelength dependence of an instantaneous atmospheric PSF: https://www.youtube.com/watch?v=x3dTZk5LuGg  (So same math as here, but different wavefront).  It's definitely more complicated than just a simple scaling.

jmeyers314: Maybe the way to handle this is a combination of what you suggest (scaling by lambda) and an `InterpolatedChromaticObject`.  I.e., draw the FFT images for some small number of control wavelengths, then for a given photon, use its wavelength to pick the nearest FFT image, shoot a photon, and rescale by the difference in wavelengths.  It may even be worth probabilistically selecting between the nearest floor and ceil FFT images to smooth out any kinks between them.

rmjarvis: OK, I think I have a semi-reasonable implementation of the chromatic shooting when geometric_shooting=False.  It just picks the closest profile among the min, mean, and max wavelength in the photon array, and then rescales the positions by the lambda factor.  
It passes my unit test with the extreme wavelengths, which is kind of a cheat, since all the wavelengths are close to either the min or the max, so it doesn't have to interpolate too far, but I also tested it with a Roman PSF and Roman bandpass, and it works well there too (agreeing at the level of 6e-4).  It's too slow to use as a unit test, but it's in the devel dir.
However, it's still not super fast, since it still needs to do 3 FFTs to make those three profiles and shoot from them.  So I'll work on getting shooting to work with InterpolatedChromaticObject, which if I can get to work, I think will be the preferred way to use ChromaticOpticalPSF with photon shooting.  (It's not yet obvious to me how to get the right interpolated x,y positions out of the saved profiles at the fiducial wavelengths.)

rmjarvis:
Hi Josh, I added two more implementations if you want to take a look.  Both the geometric_shooting=False option for ChromaticOpticalPSF (which is actually the default to match the regular OpticalPSF behavior for this parameter) and InterpolatedChromaticObject.  I also added a note in the ChromaticOpticalPSF doc string to point out that most users will probably prefer to use the interpolated approximation rather than shooting directly from the ChromaticOpticalPSF itself.

galsim/chromatic.py
```diff
@@ -1223,6 +1259,50 @@ def evaluateAtWavelength(self, wave):
...
+
+        # Instead of averaging these, we can do the averaging probabilistically by selecting
+        # each photon with a probability equal to the relative weight we would have used
+        # in the average.
+        u = np.empty(len(photons))
+        UniformDeviate(rng).generate(u)
+        use_k = k - (u<f0).astype(int)  # The second term is either 0 or 1.
+
+        # Draw photons from the saved profiles according to when we have selected to use each one.
+        for kk, ww in enumerate(self.waves):
```
jmeyers314, line 1296: Maybe enumerate over `self._objs` instead of `self.waves` (since `ww` is unused).

galsim/chromatic.py
```diff
@@ -1223,6 +1259,50 @@ def evaluateAtWavelength(self, wave):
...
+            use = use_k == kk
```
jmeyers314, line 1297: I think `use = (use_k == kk)` is clearer.  I missed the double equals the first couple times I read this line.

jmeyers314:
> I added two more implementations if you want to take a look.

I added a couple quick comments, but looks good to me.

galsim/chromatic.py
rmjarvis: Good.  Yeah.  I think I originally thought I would need the wavelength, but all that calculation was already handled.

rmjarvis:
Thanks Josh!

Issue #119, opened 2012-04-24
Library naming issues


reikonakajima:
I've successfully compiled and installed GalSim (using SCons).  I've tried running Demo.py, and it gave the following error:

=-=-=-=-=
reiko@tvwna-ip-a-55.princeton.org~/2code/GalSim/examples% Demo.py
Traceback (most recent call last):
  File "Demo.py", line 19, in <module>
    import galsim
  File "/Users/reiko/2code/GalSim/galsim/**init**.py", line 1, in <module>
    from ._galsim import *
ImportError: dlopen(/Users/reiko/2code/GalSim/galsim/_galsim.so, 2): Library not loaded: /usr/local/lib/libgalsim.0.dylib
  Referenced from: /Users/reiko/2code/GalSim/galsim/_galsim.so
  Reason: image not found
=-=-=-=-=

It appears that Demo.py cannot find /usr/local/lib/libgalsim.0.dylib.  When I go to /usr/local/lib, I see "libgalsim.dylib", but not "libgalsim.0.dylib".

PR #1101, opened 2020-10-09
Add support for higher order TPV WCS files.

rmjarvis:
This is a quick hitter in response to [Piff issue #119](https://github.com/rmjarvis/Piff/issues/119) raised by @rgraz.

Romain is using Piff on ZTF (Zwicky Transient Facility) images.  They apparently use the TPV standard, but with a higher order solution.  order=4, rather than the order=3, which was all I had seen so far.  His workaround was to manually load the WCS in astropy and then pass that to the AstropyWCS class.

So first, I fixed the problem he ran into trying to get this to happen automatically.  We had a guard in GalSim, since astropy didn't used to implement TPV properly.  It thought it would load it right, but then produce wrong answers.  They've apparently fixed that problem, so I relaxed the guard to only check for ZPX WCS's (which I confirmed are still not working correctly).

Second, since I now have a real-life example of a TPV solution with order=4, I went ahead and generalized the GSFitsWCS TPV implementation to allow for any order up to 6 (the maximum allowed by the standard).  So now GalSim should be able to properly read and use these WCS solutions using either GSFitsWCS (the default and faster implementation) or AstropyWCS.


tests/test_wcs.py
```diff
@@ -2047,9 +2052,14 @@ def test_astropywcs():
     # These all work, but it is quite slow, so only test one of them for the regular unit tests.
```
jmeyers314, line 2052: "only test one of them" -> "only test a few of them"

And similar comment below.

jmeyers314 (approved):
LGTM.

rmjarvis:
Thanks Josh!

PR #1102, opened 2020-10-13
TableScreen

jmeyers314:


jmeyers314:
This PR adds a new type of PhaseScreen for use with galsim.PhaseScreenPSF.  It wraps a galsim.LookupTable2D to enable arbitrary user provided phases.

Note that I haven't paid super close attention yet to potential 90-degree rotations or flips of phase screen definitions here, or even in the past with AtmosphericScreens or OpticalScreens.  I'd like to address this in a future issue though, after I've had a bit of time to investigate with batoid/Zemax things like is a +ve wavefront defined to be leading or lagging the reference sphere?, and make sure I got things right like is the FFT math consistent with viewing the focal plane from the front or from the back?

tests/test_phase_psf.py
```diff
@@ -1378,6 +1377,73 @@ def test_pickle():
...
+            diam=4.0,
+            aberrations=aberrations,
+            lam_0=750.0
+        )
+
+        u1 = np.linspace(-2, 2, 200)
+        u, v = np.meshgrid(u1, u1)
+        wf = Zscreen.wavefront(u, v)
+
+        table = galsim.LookupTable2D(u1, u1, wf.T, interpolant='spline')
```
jmeyers314, line 1400: Note the `wf.T` here.  LUT2D and `screen.wavefront()` appear to be inconsistent with respect to x/y axes ordering.

rmjarvis: I feel like this is a sign of a bug somewhere. Because AtmosphericScreen even uses a LookupTable2D in places for doing the wavefront and gradient. But those start with a random screen, so I don't think we could have noticed a transposition error if there was one there.

I think we should have the output from wavefront of an OpticalScreen be a valid input into this Table and be consistent. I'll see if I can track down where the root of the disagreement is happening.

rmjarvis: Tests of the Table2D functionality have code like this:
```
    x = np.linspace(0.1, 3.3, 25)
    y = np.linspace(0.2, 10.4, 75)
    yy, xx = np.meshgrid(y, x)  # Note the ordering of both input and output here!
    z = f(xx, yy)
```
whereas phase screens seem to have the standard order
```
    fx, fy = np.meshgrid(fx, fx)
```
So that's probably the source of the disagreement.  I think maybe Table2D is the problem here.  Was there some reason that it needed to be defined with the reverse convention?

jmeyers314: It's possible at the time I had just read [this](https://github.com/GalSim-developers/GalSim/blob/releases/2.2/galsim/image.py#L55), and thought I was being consistent with that.  TBH, I've never fully grokked the numpy/fits/meshgrid/LUT2D/other indexing conventions, I just know that they're not all the same.  Probably time to bite the bullet and pay attention to this.

galsim/phase_screens.py
```diff
@@ -1122,3 +1122,113 @@ def _wavefront_gradient(self, u, v, t, theta):
 class _DummyScreen(OpticalScreen):
     def _wavefront(self, *args):
         raise RuntimeError("Shouldn't reach this")
+
+
+class TableScreen(object):
```
rmjarvis, line 1127: Aside from the .T issue, the only other concern I have with this is the class name.  "Table Screen" doesn't seem particularly self-documenting as to what the class is about.   The fact that it takes a LookupTable2D argument doesn't seem like the right core feature to put into the name.  That's really just an implementation detail about how to define the 2d function.
Other possibilities:
* UserScreen
* CustomScreen
* TwoDScreen
* ...

rmjarvis (changes requested):
All the implementation here looks fine.  Just the .T issue and a request for a name change.

galsim/phase_screens.py
jmeyers314: Good point.  Switched to `UserScreen`.

jmeyers314:
Thanks for the numpy hints.  I think this is good to go now, unless you'd prefer to wait until I've investigated more on https://github.com/GalSim-developers/GalSim/issues/1104.

rmjarvis (approved):

rmjarvis:
Thanks Josh!

Issue #1105, opened 2020-11-07
GSFitsWCS TAN-SIP inversion problems


jmeyers314:
I've been running into some problems with `radecToxy` for TAN-SIP WCS's.  Specifically, occasionally the code to do the inversion gets stuck with a residual > 1e-12 which leads to a max_iter error.  Here's a link to some test data/script: https://www.dropbox.com/sh/oavydz0y8je7bga/AAAKvoe2meptgiUhYeqQv-tla?dl=0.

I think there are a couple of things that could be addressed here.  First, it'd be nice if in an array of many radecToxy transformations, a single failure didn't kill the whole batch.  Maybe there should be an option to just return `NaN` on any failures.  Second, I wonder if 1e-12 is just too small.  It's hard for me to imagine a circumstance where I'd care about errors of even 1e-6 of a pixel.  So I'd vote we either relax this constraint or at least make it user-settable.

rmjarvis:
Fixed on #1106

PR #1106, opened 2020-11-08
#1105 Fix convergence test for InvertAB

rmjarvis:
(#1105) The convergence test for InvertAB had used an absolute step size, rather than a relative one.  If |x| or |y| is of order 10^4, this could spuriously claim that the calculation was not converged, when it actually was converged.

The fix is to use relative steps when |x| or |y| > 1.


tests/test_wcs.py
```diff
@@ -2303,6 +2303,51 @@ def test_gsfitswcs():
     assert_raises(TypeError, galsim.GSFitsWCS, file_name, header='dummy')
     assert_raises(ValueError, galsim.GSFitsWCS, _data=('invalid',0,0,0,0,0,0))
 
+@timer
+def test_inverseab_convergence():
```
jmeyers314, line 2307: Need to add to ifmain block at the bottom.  I noticed a few other tests here lacked a `@timer` decoration too.

I wonder, in fact, if we should just automate the ifmain stuff.  The following snippet seems to work:
``` python
    testfns = [v for k, v in vars().items() if k[:5] == 'test_']
    for testfn in testfns:
        testfn()
```
We could probably do the timer wrapping there too.

Possible downsides include it's slightly harder to enable just a single test this way, since you have to remember its name instead of just commenting out n-1 `test_*` lines.

jmeyers314 (approved):
Generally looks good to me.  I still wonder though about killing the entire batch if just one transformation fails.  If you don't like my return NaN idea, maybe it's at least worth printing some additional information in the runtime error, e.g. which index failed?

tests/test_wcs.py
rmjarvis: Lately, I usually run single tests as e.g. `nosetests -s test_wcs.py:test_inverseab_convergence`, so that kind of automatic thing seems fine with me.  It is a common thing for us to forget.  :)

rmjarvis:
I think this change was really an error in our convergence criterion, so I think it should be very rare now for things to fail.  I think probably only for positions that are way outside the region of applicability of the SIP or TPV coefficients.  But I don't have a problem adding more detail to the error message, since then a user could catch the error and see whether they care about any non-convergence points.

rmjarvis:
I added information to the error message about which indices failed.  It will still fail to continue on with the rest of the radecToxy calculation, but at least the user can detect the problematic positions and redo without them.
It would be significantly more work to have radecToxy finish the rest of the calculation and then keep track that it should raise an exception at the end.  But if you think that's an important use case, we could consider doing that.

jmeyers314:
Thanks!  This all looks good to me now.

PR #1107, opened 2020-12-01
Migrate CI from Travis to GitHub actions

rmjarvis:
Given Travis's new pricing structure, sticking with them is no longer tenable.  This PR switches over our CI backend to use GitHub Actions instead.


rmjarvis:
Update: Everything here is working except for codecov.  As far as I can tell, the report looks valid (`coverage report` looks reasonable), and it's sending the information to codecov, but then codecov doesn't register it properly.  It says "There was an error processing coverage reports."

cf. https://codecov.io/github/GalSim-developers/GalSim/commit/799fa056df3277c2b10859b94a09e90e5b3ac80a

So I'm probably doing something wrong.  I just don't know what.  I have an issue open on the codecov community help, so hopefully someone there can help me resolve it.

cf. https://community.codecov.io/t/github-not-getting-codecov-report-after-switching-from-travis-to-github-actions-for-ci/2332

That issue is technically about Piff, but I've now hit the same problem for TreeCorr, Piff and now GalSim.  Obviously my fault, but I can't figure it out.

rmjarvis:
FYI, I got codecov working using a different uploader (the bash uploader, rather than the `codecov` command), so I think this is ready to go now.

The only code change that I added is to add a `'lens'` option to config, which calls the `GSObject.lens` method rather than separate `shear` and `magnify`.  They should be equivalent, and I don't think there is any significant efficiency difference between them.  But with the separate shear and magnify calls, precisely one object in demo11 ended up with a different transformation matrix at the last decimal place, so the py and yaml versions ended up different (since demo11.py uses the lens method and the config didn't).  This difference had never happened on any other system, but this one has some very slight numerical difference, which made the two versions not identical.

Anyway, I added `lens` as a new option, which can thus match the python call exactly, so now they match.

@jmeyers314 if you can give a quick sign-off on this, we can merge this into your PR and get CI running on that.

jmeyers314 (approved):
LGTM.

Should add the new status badge to the README: 

GalSim CI

rmjarvis:
> Should add the new status badge to the README:

Good point.  Done.

Issue #1108, opened 2020-12-16
Construct PhotonArray from pre-allocated memory


jmeyers314:
I've been thinking a bit more about how we might batch photons together to efficiently raytrace them using `batoid`.  I think it'd be useful to be able to construct a `PhotonArray` from (portions of) previously allocated numpy arrays.  E.g.,

```python

# allocate in shared memory...
N = 10_000
x = np.zeros(N)
y = np.zeros(N)
flux = np.zeros(N)
dxdz = np.zeros(N)
dydz = np.zeros(N)
wavelength = np.zeros(N)

pa_batch = galsim.PhotonArray._fromArrays(x, y, flux, dxdz, dydz, wavelength)
pa1 = galsim.PhotonArray._fromArrays(
    x[:N//2],
    y[:N//2],
    flux[:N//2],
    dxdz[:N//2],
    dydz[:N//2],
    wavelength[:N//2]
)

pa2 = galsim.PhotonArray._fromArrays(
    x[N//2:],
    y[N//2:],
    flux[N//2:],
    dxdz[N//2:],
    dydz[N//2:],
    wavelength[N//2:]
)

# in process 1
obj1._shoot(pa1, rng1)
# simultaneously in process 2
obj2._shoot(pa2, rng2)
# apply photon operations to entire batch without doing any copying...
batoidPhotonOp.applyTo(pa_batch)
# accumulate into sensor...
```

I.e., with `PhotonArray._fromArrays` one could perform the steps in `drawImage`, using multiprocessing to fill in independent objects photons "at the entrance pupil", then use batoid to trace them all simultaneously, and then finish with accumulate.  In fact, I think most of our photonOps, and even the atmospheric kicks, could be done simultaneously, not just batoid.  (Though DCR probably doesn't work, since that requires a local_wcs which will be different for different objects).  I'm not sure how to integrate this into `drawImage()` itself (probably requires some kind of delayed-execution, which would be tricky), but this at least provides a hook to drawing things "manually".

rmjarvis:
This sounds very reasonable.  My one quick API push back would be to not make the function start with an underscore.  There's no reason for this to be a "private" method.  It should be a normal classmethod in good standing for users to use and expect good documentation and some version stability, IMO.  So I think it should just be named `fromArrays`.

jmeyers314:
[...]

PR #1109, opened 2020-12-17
Add PhotonArray.fromArrays constructor

jmeyers314:
Addresses #1108 .
This version never copies an input array, but instead raises an error if a copy would be required.
There is both a sanity-checked "public" ctor `PhotonArray.fromArrays`, and a "private" unchecked ctor `PhotonArray._fromArrays`.


jmeyers314:
I'm sad that GHA cancels all the builds as soon as any fails.

Looks like warnings are only caught by assert_warns in py27 if they are the first occurrence of that warning maybe?  Should we just have assert_warns always succeed in py27?  

(And separately, I think we should start thinking about dropping support for py27...)

rmjarvis:
> I'm sad that GHA cancels all the builds as soon as any fails.

I think the logic is that if something fails, then you'll need to rerun everything anyway, so they're saving cpu cycles.  But I agree, it was nice to be able to see the full set of failures after each push.

> Looks like warnings are only caught by assert_warns in py27 if they are the first occurrence of that warning maybe? Should we just have assert_warns always succeed in py27?

I thought this was always the default behavior.  Once a warning is emitted, I thought python turned off subsequent warnings from that same bit of code.  Otherwise, if you hit a warning in a loop for instance, you can end up with a huge cascade of warnings that are all the same thing.  Our assert_warns context both checks that a warning is emitted and resets the count so that the next time, it will also be emitted.  

The problem is that your new test is emitting that warning, which makes python turn off that warning for the later test that checks for it.

I think the solution is either to fix the test so that it doesn't warn, or document that it is an expected warning by wrapping the line that does so in an `assert_warns` context.

rmjarvis:
> (And separately, I think we should start thinking about dropping support for py27...)

I think I probably agree.  Maybe on our next release, declare that this will be the last one to support py2.7 and get rid of it at that point on main.

jmeyers314:
> I thought this was always the default behavior. 

I guess it could be.  I only saw failures on the py27 branches, and didn't see anything locally.  But maybe there's something about nose tests in py27 that changes the order the tests are run in.

> Our assert_warns context both checks that a warning is emitted and resets the count so that the next time, it will also be emitted.

Ah, I didn't realize it reset the counter.  I wrapped the new code in an `assert_warns`.  Looks like CI is passing now.

rmjarvis (approved):
lgtm!

PR #1110, opened 2021-01-11
Improve the accuracy of Kolmogorov stepk for very low folding_threshold

rmjarvis:
@jchiang87 noticed a bug introduced in commit 56003a938963ba4bef875c2b34fadeb41fb25bf7 (part of PR #1099) where if folding_threshold was too small, R would never be set and stepk was way too large.  

This was a straight bug in that commit.  But in investigating it, I realized that the old implementation of stepk for Kolmogorov was also not very accurate, especially for very low folding_threshold.  The problem is that the accumulation of the integral is just a crude rectangular rule integral, which isn't accurate enough to tell precisely when the integral reaches 0.999999 (for instance).  

So here, I use the high-r asymptotic power law form for the Kolmogorov profile and use that directly to estimate what radius encloses (1-folding_threshold) of the flux, by analytically doing the integral from R to infinity, which gets more accurate for very small folding_threshold, and is plenty accurate for the default value of 5e-3.

The upshot of this is that the "good image size" for folding_threshold=1.e-6, which imSim was using, got a lot larger than the previous version 2.2 values.  I'm convinced this is actually the right answer now, but it means that the imSim function that was doing this probably needs to use a larger folding_threshold.  I suspect 1.e-4 is actually closer to what we should be using.  It looks like this produces fairly reasonable image sizes on this branch.

Note: I applied the same change to VonKarman too, which also seems to be well approximated by a power law, at least for reasonable values of L0.  I didn't investigate too hard whether things changed at extreme values of L0.

### DIFF
--- a/src/SBKolmogorov.cpp
+++ b/src/SBKolmogorov.cpp
@@ -401,16 +401,21 @@ namespace galsim {
         // 2pi F1 r1^n / (n-2) f_t = R^(n-2)
         double r1 = _radial.argMax();
         double F1 = _radial.lookup(r1);
+#ifdef DEBUGLOGGING
         double r2 = r1 * (1-dlogr);
         double F2 = _radial.lookup(r2);
         dbg<<"r1,F1 = "<<r1<<','<<F1<<std::endl;
         dbg<<"r2,F2 = "<<r2<<','<<F2<<std::endl;
         // power law index = dlog(F)/dlog(r)
-        double n = -(std::log(F2)-std::log(F1)) / (std::log(r2)-std::log(r1));
-        // For high folding_threshold, n can come out too low (e.g. < 2), which is bad.
-        // We know the right answer is close to 3.67, so just use that if n < 3.6.
-        if (n < 3.6) n = std::max(n, 3.67);
-        dbg<<"n = "<<n<<std::endl;
+        double n_emp = -(std::log(F2)-std::log(F1)) / (std::log(r2)-std::log(r1));
+        dbg<<"Empirical n = "<<n_emp<<std::endl;
+#endif
+        // Emprically n is very close to 11/3.  This is probably exact, since it's the kind of
+        // fraction that probably comes out of the Komogorov turbulence.  But I (MJ) haven't
+        // figured out how to prove this.
+        // Regardless, let's just always use this for the purpose of estimating stepk,
+        // since any deviations from the exactly correct answer don't matter much.
+        double n = 11./3.;
         double R = fast_pow(2.*M_PI*F1*fast_pow(r1,n)/((n-2)*gsparams->folding_threshold),
                             1./(n-2));
         dbg<<"R = "<<R<<std::endl;
--- a/src/SBKolmogorov.cpp
+++ b/src/SBKolmogorov.cpp
@@ -407,6 +407,9 @@ namespace galsim {
         dbg<<"r2,F2 = "<<r2<<','<<F2<<std::endl;
         // power law index = dlog(F)/dlog(r)
         double n = -(std::log(F2)-std::log(F1)) / (std::log(r2)-std::log(r1));
+        // For high folding_threshold, n can come out too low (e.g. < 2), which is bad.
+        // We know the right answer is close to 3.67, so just use that if n < 3.6.
+        if (n < 3.6) n = std::max(n, 3.67);
         dbg<<"n = "<<n<<std::endl;
         double R = fast_pow(2.*M_PI*F1*fast_pow(r1,n)/((n-2)*gsparams->folding_threshold),
                             1./(n-2));
--- a/src/SBKolmogorov.cpp
+++ b/src/SBKolmogorov.cpp
@@ -367,19 +367,14 @@ namespace galsim {
         // We use a cubic spline for the interpolation, which has an error of O(h^4) max(f'''').
         // As with Sersic (since Kolmogorov is just a Sersic with n=0.6 in reverse), we use
         // max(f'''') ~= 10.  So:
-        // 10 h^4 <= kvalue_accuracy
-        // h = (kvalue_accuracy/10)^0.25
+        // 10 h^4 <= xvalue_accuracy
+        // h = (xvalue_accuracy/10)^0.25
         double dlogr = gsparams->table_spacing * sqrt(sqrt(gsparams->xvalue_accuracy / 10.));
         xdbg<<"dlogr = "<<dlogr<<std::endl;
 
-        // Along the way accumulate the flux integral to determine the radius
-        // that encloses (1-folding_threshold) of the flux.
-        double sum = 0.;
-        double thresh0 = 0.5 / (2.*M_PI*dlogr);
-        double thresh1 = (1.-gsparams->folding_threshold) / (2.*M_PI*dlogr);
-        double thresh2 = gsparams->shoot_accuracy / (2.*M_PI*dlogr);
-        xdbg<<"thresh values = "<<thresh0<<"  "<<thresh1<<"  "<<thresh2<<std::endl;
-        double R = 0., hlr = 0.;
+        // Continue until the missing flux is less than shoot_accuracy.
+        double thresh = gsparams->shoot_accuracy / (2.*M_PI*dlogr);
+        xdbg<<"thresh  = "<<thresh<<std::endl;
         KolmXValue xval_func(*gsparams);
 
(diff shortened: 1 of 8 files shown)


src/SBKolmogorov.cpp
```diff
@@ -367,19 +367,14 @@ namespace galsim {
...
-        // Along the way accumulate the flux integral to determine the radius
-        // that encloses (1-folding_threshold) of the flux.
-        double sum = 0.;
-        double thresh0 = 0.5 / (2.*M_PI*dlogr);
-        double thresh1 = (1.-gsparams->folding_threshold) / (2.*M_PI*dlogr);
-        double thresh2 = gsparams->shoot_accuracy / (2.*M_PI*dlogr);
-        xdbg<<"thresh values = "<<thresh0<<"  "<<thresh1<<"  "<<thresh2<<std::endl;
-        double R = 0., hlr = 0.;
+        // Continue until the missing flux is less than shoot_accuracy.
+        double thresh = gsparams->shoot_accuracy / (2.*M_PI*dlogr);
```
jmeyers314, line 376: Can you double check the factor of `dlogr` here?  Since the integral below is analytic instead of a Riemann sum, I think it may be incorrect.

src/SBKolmogorov.cpp
```diff
@@ -390,33 +385,42 @@ namespace galsim {
...
+        double F1 = _radial.lookup(r1);
+        double r2 = r1 * (1-dlogr);
+        double F2 = _radial.lookup(r2);
+        dbg<<"r1,F1 = "<<r1<<','<<F1<<std::endl;
+        dbg<<"r2,F2 = "<<r2<<','<<F2<<std::endl;
+        // power law index = dlog(F)/dlog(r)
+        double n = -(std::log(F2)-std::log(F1)) / (std::log(r2)-std::log(r1));
+        // For high folding_threshold, n can come out too low (e.g. < 2), which is bad.
+        // We know the right answer is close to 3.67, so just use that if n < 3.6.
+        if (n < 3.6) n = std::max(n, 3.67);
```
jmeyers314, line 412: Feels a little weird to introduce a discontinuity at n=3.6.  Can we just always apply `n = std::max(n, 3.67)`?

src/SBVonKarman.cpp
```diff
@@ -292,23 +290,40 @@ namespace galsim {
...
+        double F1 = _radial.lookup(r1);
+        double r2 = r1 * (1-dlogr);
+        double F2 = _radial.lookup(r2);
+        dbg<<"r1,F1 = "<<r1<<','<<F1<<std::endl;
+        dbg<<"r2,F2 = "<<r2<<','<<F2<<std::endl;
+        // power law index = dlog(F)/dlog(r)
+        double n = -(std::log(F2)-std::log(F1)) / (std::log(r2)-std::log(r1));
+        // For high folding_threshold, n can come out too low (e.g. < 2), which is bad.
+        // We know the right answer when L0=inf is close to 3.67, so just use that if n < 3.
+        if (n < 3.) n = std::max(n, 3.67);
```
jmeyers314, line 316: I guess this one might be trickier...  Is the asymptotic slope for finite L0 < 3.67?  Or could just setting a simple floor (with no discontinuity introduced) work here too?

jmeyers314 (commented):
Generally looks good, but I have a few questions attached.

src/SBKolmogorov.cpp
rmjarvis: Yeah, I wasn't quite sure what the best strategy here is.  Kolmogorov is a specific profile, so there is a right answer.  I originally had just hard-coded `n=3.67`.  But then I thought it might be better to let the code determine it, rather than introduce a special number.  That felt more principled to me, especially since as folding_threshold gets smaller, the estimate should get more accurate, so sticking to 3.67 might be introducing unwarranted inaccuracies in some (extreme) cases. 
But then it failed when folding_threshold was very high, like 0.05 or something, so I added this test, which I agree doesn't feel very principled anymore.
Maybe I'll try to get Maple or Wolfram to figure out a more precise number and just hard code it again.

src/SBVonKarman.cpp
rmjarvis: Here I was less confident about what the right number is.  I didn't actually do any tests of how much the number changes for different L0.  I guess maybe I should do that investigation.  For some reasonable L0, it was empirically also pretty close to 3.67.  So maybe there is a good reason to expect it to be the same as Kolmogorov, at least over reasonable values of L0.

src/SBKolmogorov.cpp
rmjarvis: Thank you!  This was indeed wrong.  The thing that gets compared to thresh now is just the integral divided by 2pi.

src/SBKolmogorov.cpp
rmjarvis: I couldn't coerce either Maple or Wolfram to give me the analytic asymptote.  Maple gave me a closed form solution for the Hankel transform -- a crazy combination of hypergeometric functions; see below for a -> ascii printout of the result.  But any kind of evaluation quickly became inaccurate as r increased, so it couldn't manage anything in the high r limit.

However, emprically, as I increased the accuracy of the calculations, I got answers that seemed to get quite close to 11/3 exactly.  Numbers like 3.66675 for instance.  So I'm putting forth a conjecture that it is exactly 11/3, since that seems like the kind of exponent that might emerge from Kolmogorov turbulence theory.  But I haven't proved it.

Also, the VonKarman gave basically the same answer even for rather small L0 values.  Even when L0=r0, I was still getting n=3.66628.  Once L0 << r0, it started to diverge, but I don't think we care about that limit.  So I just set n=11/3 always in that case too.

```
                          //                     
            1             ||         8          /
------------------------- ||1309 Pi r  hypergeom|
             /3\      /4\ \\                    \
5000000 GAMMA|-| GAMMA|-|                        
             \5/      \5/                        

  [9   16  37  47  26]  [   6  6  7  8  8  9  9]       729     10\ 
  [--, --, --, --, --], [1, -, -, -, -, -, -, -], - --------- r  | 
  [10  15  30  30  15]  [   5  5  5  5  5  5  5]    156250000    / 

       /4\    /1   \            6          /[7   13  31  41  23]  
  GAMMA|-| sin|- Pi| - 7040 Pi r  hypergeom|[--, --, --, --, --], 
       \5/    \5   /                       \[10  15  30  30  15]  

  [4  4     6  7  7  8  8]       729     10\      /3\    /2   \          
  [-, -, 1, -, -, -, -, -], - --------- r  | GAMMA|-| sin|- Pi| + 300000 
  [5  5     5  5  5  5  5]    156250000    /      \5/    \5   /          

   4          /[1  2  5  7  4]  [3  3  4  4  6  6  7  7]  
  r  hypergeom|[-, -, -, -, -], [-, -, -, -, -, -, -, -], 
              \[2  3  6  6  3]  [5  5  5  5  5  5  5  5]  

       729     10\      /3\      /4\    /2   \    /1   \           
  - --------- r  | GAMMA|-| GAMMA|-| sin|- Pi| sin|- Pi| - 1920000 
    156250000    /      \5/      \5/    \5   /    \5   /           

   2          /[3   7   19  29  17]  [2  2  3  3  4     6  6]  
  r  hypergeom|[--, --, --, --, --], [-, -, -, -, -, 1, -, -], 
              \[10  15  30  30  15]  [5  5  5  5  5     5  5]  

                           2                                     
       729     10\      /4\       /3\    /2   \    /1   \        
  - --------- r  | GAMMA|-|  GAMMA|-| sin|- Pi| sin|- Pi| + 9600\
    156250000    /      \5/       \5/    \5   /    \5   /        

               /[1   4   13  23  14]  [1  1  2  2  3  4  4   ]  
  000 hypergeom|[--, --, --, --, --], [-, -, -, -, -, -, -, 1], 
               \[10  15  30  30  15]  [5  5  5  5  5  5  5   ]  

                           2                             \       
       729     10\      /3\       /4\    /2   \    /1   \|    /2 
  - --------- r  | GAMMA|-|  GAMMA|-| sin|- Pi| sin|- Pi|| sin|- 
    156250000    /      \5/       \5/    \5   /    \5   //    \5 

               \
    \    /1   \|
  Pi| sin|- Pi||
    /    \5   //
```

jmeyers314: I bet 11/3 is right too.  Thanks for looking into this.

jmeyers314 (approved):
LGTM once you get CI to pass.

PR #1111, opened 2021-01-15
Bypass fmath on non-intel CPUs.

erykoff:
The fastmath routines require x86 CPUs.  This PR adds in a check and reverts to standard library exponential and logarithms on non-x86 (e.g. Apple Silicon) CPUs.

### DIFF
--- a/include/galsim/fmath/fmath.hpp
+++ b/include/galsim/fmath/fmath.hpp
@@ -33,6 +33,11 @@
 #include <stdlib.h>
 #include <float.h>
 #include <string.h> // for memcpy
+
+#ifdef __x86_64__
+// Optimized fmath required intel cpus
+#define HAS_FMATH
+
 #if defined(_WIN32) && !defined(__GNUC__)
 	#include <intrin.h>
 	#ifndef MIE_ALIGN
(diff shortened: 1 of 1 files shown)


include/galsim/fmath/fmath.hpp
```diff
@@ -849,3 +854,26 @@ __m128 (*const exp_ps)(__m128) = local::C<>::getInstance().exp_ps_;
...
+namespace fmath {
+
+inline float exp(float x)
+{
+  return expf(x);
+}
+
+inline double expd(double x)
+{
+  return exp(x);
```
rmjarvis, line 869: I think I'd prefer `std::exp` here.  Otherwise, I'm not 100% confident that the resolution order won't favor the above float function and lose precision.

include/galsim/fmath/fmath.hpp
```diff
@@ -33,6 +33,11 @@
 #include <stdlib.h>
 #include <float.h>
 #include <string.h> // for memcpy
+
+#ifdef __x86_64__
```
rmjarvis, line 37: I think this is probably too restrictive.  For instance, I think `__i386__` works just fine with fmath.
Can we use `__SSE__` as a proxy for when fmath is expected to work?  That's the fundamental feature it needs, right?

rmjarvis (changes requested):
Thanks Eli.  I have a couple push backs for you.

rmjarvis (approved):

PR #1113, opened 2021-01-19
Add cross-compilation and SCons 4 compatibility to SConscript

erykoff:


rmjarvis (approved):
Seems fine to me.  I don't use SCons myself anymore.  I think it's just the LSST DM folks.  So if it works in that context for you, I presume this is all ok.

PR #1117, opened 2021-02-05
Use the fork context explicitly in some places where spawn fails

rmjarvis:
Duplicate of #1116 but targeted to main (and rebased onto main).  I realized that the CI won't run on #1116, since that branch is set up for Travis, which we've turned off now.  So If this one passes the GH Actions CI, then we can merge both.  Anyone who wants to review can do so on either one, since they are equivalent.

### DIFF
--- a/galsim/config/util.py
+++ b/galsim/config/util.py
@@ -20,6 +20,7 @@ import logging
 import copy
 import sys
 from collections import OrderedDict
+from multiprocessing.managers import BaseManager
 
 from ..utilities import SimpleGenerator
 from ..random import BaseDeviate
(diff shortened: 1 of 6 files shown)


galsim/config/util.py
```diff
@@ -248,6 +249,23 @@ def recursive_copy(d):
...
+    """There are a few places we need a Manager object.  This one uses the 'fork' context,
+    rather than whatever the default is on your system (which may be fork or may be spawn).
+
+    Starting in python 3.8, the spawn context started becoming more popular.  It's supposed to
+    be safer, but it wants to pickle a lot of things that aren't picklable, so it doesn't work.
+    Using this as the base class instead, should work.  And doing it it one place means that we
+    only have one place to change this is there is a different strategy that works better.
+    """
+    def __init__(self):
+        if sys.version_info >= (3,8):
```
jmeyers314, line 262: Can we just use (3,0)?  That'd make it more obvious to simplify in the future when we eventually drop support for py2.

jmeyers314: (and same comment for most of the (3,8) checks...)

jmeyers314 (approved):
LGTM.  One question attached.

galsim/config/util.py
rmjarvis: Yes.  Good point.

PR #1121, opened 2021-05-20
Create config version of demo13.

rmjarvis:
This PR is mostly about getting the Roman detector effects into the config layer.  There wasn't much back-end code changes required.  Indeed the main thing that was required was the implementation of chromatic photon shooting, which I did last fall.  Mostly this involved writing new config modules in the galsim.roman sub-module to automatically handle all the standard detector effects.

Some of the new features showcased in demo13 are:

- RomanPSF object type
- RomanSCA image type
- RomanBandpass bandpass type
- The top-level modules field, which we hadn't used yet in other demos in the main demo set (just the des and great3 extra ones)
- Examples of custom modules.  I used LogNormal and MixedScene, both from galsim_extra.  And I point out that these are both available from that repo.
- Multiple input objects of the same type.
- Multiple rng sequences and the use of rng_num.
- Chromatic photon shooting.

I'm actually a little ambivalent about hosting the Roman config modules in the main galsim repo.  I'm concerned that we will want to update these modules on a different cadence than the regular GalSim version updates.  However, there is enough pedagogical value in this demo, that I think it is probably still worthwhile as a demo.

But I think fairly soon, the Roman HLS SIT may want to host these separately, so we can update them on a quicker cadence as we work on updating some of the numbers.  Especially, we are planning to use different numbers for a number of effects on different SCAs, based on lab measurements of the actual detectors.


rmjarvis:
BTW, here is a screen shot of the suggested ds9 command that you get from running this demo (either .py or .yaml):
image

rmjarvis:
The main reason I decided to add the MixedScene was to get some stars in there, so we could see the spiky Roman PSF.  :)

docs/config_objects.rst
```diff
@@ -134,9 +134,13 @@ Galaxy Types
 
     * ``index`` = *int_value* (default = 'Sequence' from 0 to ``cosmos_catalog.nobjects``-1)  Which item in the catalog to use. Special: If ``index`` is either a 'Sequence' or 'Random' and ``last`` or ``max`` (respectively) is not specified, then it is automatically set to ``real_catalog.nobjects-1``.
     * ``gal_type`` = *str_vale* (required, unless ``real_catalog.use_real`` is False, in which case 'parametric') Which type of galaxy to use.  Options are 'real' or 'parametric'.
+    * ``noise_pad_size`` = *float_val* (default = 5) The size of a padding region around the HST image when ``gal_type='real'``.
```
rmandelb, line 137: I think this could use a little more explanation - what does the value mean?  e.g., does the default 5 mean that the postage stamp is enlarged by 5 pixels in each direction, or a factor of 5, or ...?

docs/config_objects.rst
```diff
@@ -134,9 +134,13 @@ Galaxy Types
...
     * ``sersic_prec`` = *float_value* (default = 0.05) The desired precision on the Sersic index n in parametric galaxies.  GalSim is significantly faster if it gets a smallish number of Sersic values, so it can cache some of the calculations and use them again the next time it gets a galaxy with the same index.  If ``sersic_prec`` is 0.0, then use the exact value of index n from the catalog.  But if it is >0, then round the index to that precision.
+    * ``chromatic`` = *bool_value* (default = False) Whether to build chromatic profiles. (Only valid if ``gal_type='parametric'``.)
+    * ``area`` = *float_value* (default = None, which will use the HST collecting area.)  The effective collecting area in cm**2 of the telescope being simulated.  Used for rescaling the flux values.
+    * ``exptime`` = *float_value* (default = 1)  The exposure time in seconds.  Used for rescaling the flux values.
```
rmandelb, line 143: Somebody might wonder why is the default value "1" and not the length of the COSMOS HST exposures.

examples/demo13.py
```diff
@@ -54,308 +55,302 @@
...
-
-        # Now we're done with the per-galaxy drawing for this image.  The rest will be done for the
-        # entire image at once.
-        logger.info('Postage stamps of all galaxies drawn on a single big image for this filter.')
-        logger.info('Adding the sky level, noise and detector non-idealities.')
+        # GalSim uses the term Bandpass for the class that defines the throughput across a
+        # filter bandpass, partly because "filter" is a reserved word in python.  So we follow
+        # that convention here as well.
+        bandpass = roman_filters[filter_name]
+        y_bandpass = roman_filters['Y106']
```
rmandelb, line 189: Not a big deal, but why do this inside the loop?

examples/demo13.py
```diff
@@ -54,308 +55,302 @@
...
+                obj = galsim.DeltaFunction() * sed
+
+            else:
+                # Bright galaxy
+                logger.debug('Bright galaxy')
+
+                obj = cat2.makeGalaxy(chromatic=True, gal_type='parametric', rng=obj_rng)
+                logger.debug('galaxy index = %s',obj.index)
+
+                # Scale up the area by a factor of 2, and the flux by a factor of 4.
```
rmandelb, line 319: Why?

examples/demo13.py
```diff
@@ -21,8 +21,6 @@
 The thirteenth script in our tutorial about using Galsim in python scripts: examples/demo*.py.
 (This file is designed to be viewed in a window 100 characters wide.)
 
-This script currently doesn't have an equivalent demo*.yaml or demo*.json file.
-
 This script introduces non-idealities arising from NIR detectors, in particular those that will be
```
rmandelb, line 24: I know this isn't something new, but reading this, I think it's missing a single high-level statement that the goal of the demo is to illustrate drawing a sky scene with features expected for images from the Roman Space Telescope.   (Currently it kind of implies that the goal is only to illustrate detector non-idealities, but there is more than that - Roman PSF and WCS etc.)  I suggest adding a 1-sentence preamble along those lines to demo13.py and demo13.yaml.

examples/demo13.yaml
```diff
@@ -0,0 +1,332 @@
...
+    # If a local file has the appropriate Register commands, then importing it will enable
+    # whatever extra types you have defined.  Here we use two user-defined types:
+    #
+    # - A stamp type, MixedScene, defined in the file mixed_scene.py.
+    # - A value type, LogNormal, defined in the file log_normal.py.
+    #
+    # Note: Both of these types are also available in the python repository called galsim_extra.
+    #       cf. https://github.com/esheldon/galsim_extra
+    #       So if that is installed, you could instead put galsim_extra in the modules list
+    #       and both of these (as well as numerous other user-defined types) would be availble.
```
rmandelb, line 91:
```suggestion
    #       and both of these (as well as numerous other user-defined types) would be available.
```

rmandelb (changes requested):
Hi Mike - thanks for getting the various Roman functions into config!  I like the updates you've made to demo13.py as well.  Just a few minor comments for you to consider.

docs/config_objects.rst
rmjarvis: Good point.  Will add.  I had to look it up -- It's the minimum size in arcsec to pad out the noise for the real galaxy stamps.  I guess many of them must be smaller than 5 arcsec, so this at least sets that as a uniform minimum size of the noise region.

docs/config_objects.rst
rmjarvis: I don't know the history of this decision.  Any insight you want to share for me to write here?

examples/demo13.py
rmjarvis: Just for visual impact to get a few more big, bright galaxies that can try to represent brighter galaxies than the bright limit of the 25.2 catalog.

examples/demo13.py
rmjarvis: It's not really an inefficiency, since it's just a dict access.  But I'll pull it out of the loop and add a comment about how we'll be using it.

examples/demo13.py
rmjarvis: Done.  I added:

This script is intended to produce a relatively realistic scene of galaxies and stars as will be observed by the Roman Space Telescope, including the Roman PSF, WCS, and various NIR detector effects.

docs/config_objects.rst
rmandelb: I had to go remind myself, and my answer is not really satisfying.

It seems that somebody in the COSMOS survey decided to normalize their science release images to have units of counts/s instead of counts.  So, to get the number of counts that would've been seen in COSMOS, you effectively have to pretend they had a 1s exposure time.  I don't know why they decided to do that, but we got stuff from them that was normalized this way, so ...

examples/demo13.py
rmandelb: OK, might be worth mentioning.

rmjarvis: I think I did.  The update has:
```
# This is not necessarily physical, but it is intended to add some more big,                
# bright galaxies to the scene to make the final image a bit more interesting.
```
Does that seem ok?

rmandelb: Yes, thanks.

docs/config_objects.rst
rmjarvis: For this, the update has
```
(Note: The native COSMOS galaxy images being used have been rescaled to 1 second exposures, so the default corresponds to a 1 second exposure on HST.)
```
I guess rescaled to 1 second exposures is equivalent to rescaling to counts per second rate.  Not sure if there is any distinction worth making there.

rmandelb: Hmm, I think there are some subtleties we might want to mention for people who are trying to actually match COSMOS-like images.  I checked and I think there is language about this in scene.py and real.py that we could use.  For example, in scene.py it says "The nominal COSMOS zeropoint for single-orbit depth (2000s of usable exposure time, across 4 dithered exposures) is supposedly 25.94.  But the science images that we are using were normalized to count rate, not counts, meaning that an object with mag=25.94 has a count rate of 1 photon/sec, not 1 photon total.  Since we've declared our flux normalization for the outputs to be appropriate for a 1s exposure, we use this zeropoint directly."  Knowing that the science images were normalized to count rates is important because otherwise they might think that matching the quoted zero points requires putting in 2000?  Or maybe there is simply a tweak on the sentence you included that could express this?

rmjarvis: I don't think the user has easy access to the zeropoint directly.  The zeropoint is used to set the flux of the returned object (right after where this comment is in scene.py), so I think it's already correctly applied.  If they want the objects to have the right flux for a different exposure time than 1 second, then I'm pretty sure using that exptime here does the right thing.  I guess I'm probably not grokking what you're worried about the user doing wrong here exactly.

rmandelb: Mmm.  I might be confusing myself, so I'm not sure there is a problem.  Let me try again, and ignore the zero point distraction.  Imagine we have a user who wants to reproduce COSMOS images.  So they look up COSMOS and find that the total exposure time over all images was 2000 s.   Should they be using exptime=1 or 2000 at this point?  And how should the result compare to what is seen in the science images?

rmjarvis: It depends on if they want to reproduce the real COSMOS images that had effective exposure time of 2000s or if they want to reproduce the images in the file we make available.  In the former case, they should use exptime=2000.  In the latter either leave it out or explicitly set exptime=1.

rmjarvis: Or maybe your point is that the "real" COSMOS images (i.e. the processed science images) are the ones that were rescaled to count rate, so after processing, they have an effective exposure time of 1 s.  In that case, using exptime=1 would match the flux level.

rmjarvis: How about this as possibly clearer and getting at this subtlety:
```
(Note: The processed COSMOS ACS/HST science images are given in counts/second, so they have an effective exposure time of 1 second in terms of their flux levels.  The default value of corresponds to a 1 second exposure 
on HST, which will match these processed images.)
```

rmjarvis: I pushed an edit with this language, but please feel free to tweak further if you want.

rmandelb: Thanks for parsing my wandering remarks so effectively.  I understood all this very deeply 6+ years ago but it's been far too long since I thought about it carefully!  I think your updated version is good.

rmandelb (approved):

rmjarvis:
Great.  Thanks Rachel!

Issue #1120, opened 2021-04-27
default wcs when reading from fits


jmeyers314:
I think the documentation of `galsim.fits.read` is incorrect.  Specifically the part [here](https://github.com/GalSim-developers/GalSim/blob/main/galsim/fits.py#L738):

```
    If the FITS header has keywords that start with ``GS_`` keywords, these will be used to
    initialize the bounding box and WCS.  If not, the bounding box will have ``(xmin,ymin)`` at
    ``(1,1)`` and the scale will be set to 1.0.
```

Specifically, if the fits header has standard WCS keywords (CRPIX, CRVAL, CD, ...), even if they don't start with 'GS_', then the wcs will get attached to the returned image.  We should update this.

As a side note, what I'm actually interested in is if there's a simple way from within `galsim.fits.read` to decide if the WCS was absent in the header and the default value used, as opposed the the header did exist and just happens to look like the default.

jmeyers314:
I guess it's a little complicated in that individual WCS components have defaults, not that there's an overall default WCS.  My use case is mostly concerned with whether or not a pixel scale of 1.0 was the default, so I've arrived at the following test:

```python
def _isWCSGalsimDefault(wcs, hdr):
    """Decide if wcs pixel scale is explicitly present in header, or if it's just the galsim default.

    Parameters
    ----------
    wcs : galsim.BaseWCS
        Potentially default WCS.
    hdr : galsim.fits.FitsHeader
        Header as read in by galsim.

    Returns
    -------
    isDefault : bool
        True if default, False if explicitly set in header.
    """
    if wcs != galsim.PixelScale(1.0):
        return False
    if hdr.get('GS_WCS') is not None:
        return False
    if hdr.get('CTYPE1', 'LINEAR') == 'LINEAR':
        return not any(k in hdr for k in ['CD1_1', 'CDELT1'])
    for wcs_type in galsim.fitswcs.fits_wcs_types:
        # If one of these succeeds, then assume result is explicit
        try:
            wcs_type._readHeader(hdr)
            return False
        except Exception:
            pass
    else:
        return not any(k in hdr for k in ['CD1_1', 'CDELT1'])
```

rmjarvis:
Would it be simpler maybe to have GalSim emit a warning when there is no valid WCS in the header, and it uses the default?  Then you could just catch the warning when reading and log that information somewhere.

jmeyers314:
[...]

PR #1122, opened 2021-05-26
Add option to get warning about WCS defaulting to PixelScale (#1120)

rmjarvis:
If a FITS file doesn't have any WCS information in it, we default to a PixelScale.  This is currently done silently.

This PR adds a warning that is emitted by default when reading this via `FitsWCS(file)`, but is suppressed by default when reading via `galsim.fits.read(file)`.

If you want to get the warning, you can now do `galsim.fits.read(file, suppress_warning=False)`, and it will emit the warning.  (So you can potentially catch it and do something particular if this happened.)


jmeyers314 (approved):
LGTM.

rmjarvis:
Thanks Josh!  One to go...  :)

PR #1123, opened 2021-06-03
Fix config world pos to be the real world_pos, not tangent plane projection

rmjarvis:
@beckermr recently ran into a bit of cruft in the config processing that would be worth cleaning up before the next release.  He discovered that config['world_pos'] is not necessarily the real world_pos.  If the wcs is a CelestialWCS, then world_pos is a tangent plane projection.

There isn't any good reason for this behavior, and it's confusing, so the PR changes it to something more sensible.  Specifically, world_pos is whatever the current world_pos is as we normally talk about it.  If world_pos is a CelestialWCS, then we also have a uv_pos, which is the tangent plane project, which enables the use cases that the old behavior used to enable (namely having a flat "world" position for NFW halos or shear power spectra fields).

### DIFF
--- a/galsim/config/stamp.py
+++ b/galsim/config/stamp.py
@@ -718,9 +718,10 @@ class StampBuilder(object):
             image_pos = wcs.toImage(world_pos)
 
         # If the world_pos is a CelestialCoord, then we also call it sky_pos.
-        # If the world_pos is not celestial, then the user may optionally define a sky_pos
-        # value, which gets saved as base['sky_pos'].  This may be useful for things that need
-        # to know where in the sky the pointing is, even if the WCS is not a CelestialWCS.
+        # If the world_pos is not celestial, or the user wants to override it for some reason,
+        # then the user may optionally define a sky_pos value, which gets saved as base['sky_pos'].
+        # This may be useful for things that need to know where in the sky the pointing is,
+        # even if the WCS is not a CelestialWCS.
         if 'sky_pos' in config:
             base['sky_pos'] = ParseValue(config, 'sky_pos', base, CelestialCoord)[0]
         elif isinstance(world_pos, CelestialCoord):
--- a/galsim/config/stamp.py
+++ b/galsim/config/stamp.py
@@ -213,8 +213,8 @@ def SetupConfigStampSize(config, xsize, ysize, image_pos, world_pos, logger=None
 
     Parameters:
         config:         A configuration dict.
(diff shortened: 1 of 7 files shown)


galsim/config/stamp.py
```diff
@@ -717,19 +717,26 @@ def locateStamp(self, config, base, xsize, ysize, image_pos, world_pos, logger):
         elif world_pos is not None and image_pos is None:
             image_pos = wcs.toImage(world_pos)
 
+        # If the world_pos is a CelestialCoord, then we also call it sky_pos.
+        # If the world_pos is not celestial, then the user may optionally define a sky_pos
+        # value, which gets saved as base['sky_pos'].  This may be useful for things that need
+        # to know where in the sky the pointing is, even if the WCS is not a CelestialWCS.
```
beckermr, line 723: The comment here doesn't quite match the logic of the code. Not sure if it has to but worth thinking about what the correct behavior is.

galsim/config/value_eval.py
```diff
@@ -60,7 +60,7 @@ def _type_by_letter(key):
 eval_base_variables = [ 'image_pos', 'world_pos', 'image_center', 'image_origin', 'image_bounds',
                         'image_xsize', 'image_ysize', 'stamp_xsize', 'stamp_ysize', 'pixel_scale',
                         'wcs', 'rng', 'file_num', 'image_num', 'obj_num', 'start_obj_num',
-                        'world_center', 'sky_pos', 'bandpass' ]
+                        'world_center', 'sky_pos', 'uv_pos', 'bandpass' ]
```
beckermr, line 63: Is bandpass already in the docs? If so ignore me. Otherwise I don't think I saw it added in this or though I could have missed it.

beckermr (commented):
I left some comments. I am not familiar with galsims internals and docs so I may be off base. Thanks for working on this!

galsim/config/stamp.py
rmjarvis: Good point.  A user-defined sky_pos takes precedence.  Added "or the user wants to override it for some reason,"

galsim/config/value_eval.py
rmjarvis: Yes.  It's not in the v2.2 docs, since bandpass wasn't allowed then.  But it was already added to the docs on main when I added that functionality.

beckermr (approved):
Lgtm from me though you may want a more experienced galsim developer if you feel you need it. Thanks again!

PR #1124, opened 2021-06-19
Various optics updates

jmeyers314:
This PR has a few loosely related updates related to optics:

- Added function that will describe circular zernikes algebraically (as shown in #1104).
- Delayed some optional calculations when manipulating Zernikes arithmetically.  (This significantly speeds up evaluation the hessian of a Zernike series when you don't actually care what the series coefficients are).
- Added an "fft_sign" kwarg to OpticalPSF and related.  This is primarily aimed at #1104 as a stopgap measure so we can get a new release.  Motivated by @aaronroodman's observation that GalSim's convention doesn't match Zemax's convention when comparing DECam donuts, I made the default `fft_sign='+'`, which is the opposite behavior from before this PR.  I made the required adjustments to unit tests, but I don't know, e.g., if this messes up the Roman optical PSFs, so somebody should check that.
- Cleaned up a few minor items.

Once this is merged, I think we'll be ready for v2.3. :)


rmjarvis (approved):
This all looks good to me, Josh.

The only slight push-back I might have (but not enough of one to even add as a line comment) is that fft_sign being a string feels a little wrong.  There are only 2 possibilities, so bool seems like the right type for this, but with a different name.  Maybe swap_fft_sign?  

But what you have is also fine with me if you prefer this.  It doesn't seem like any of the places where it is used/checked are super time-critical, so the string processing isn't going to be inefficient.  So this is merely a style question.  And I'm happy to defer to you on your preference here.

jmeyers314:
Thanks Mike,
  Yeah, I did consider a bool, but decided that choosing between '+' and '-' was more explicit.  Like you said, probably not a big speed issue so I think I'll just leave it.

Issue #935, opened 2018-01-05
Try to speed up InterpolatedImage calculations


rmjarvis:
A lot of real-world use cases are limited by the calculations in InterpolatedImage.  In particular, the `KTable::interpolate` function.  I think this could be significantly sped up by refactoring the calculation to compute many kvalues at once rather than one at a time to provide more opportunities for vectorizing the calculation.  

Indeed, I think the refactoring can likely get rid of the KTable and XTable classes entirely, just using the Image class instead.

rmjarvis:
Done #1126

PR #1126, opened 2021-06-23
#935 Speed up draw functions for InterpolatedImage

rmjarvis:
This has been an open issue for a while, since I realized that the way the `fillXImage` and `fillKImage` functions in SBInterpolatedImage were not particularly efficient.  They both used caching within the XTable and KTable interp methods, which was effective, but it encouraged looping over the images against the natural stride direction, which meant that they were missing opportunities for vectorization.

The main thing that this PR does is to implement all 4 `fill` functions directly and be smarter about the caching.  The caching that used to be there is now done for whole rows at a time, rather than just a few values, which lets more loops have unit stride, and thus vectorize nicely.  Moreover, for the rectilinear versions (i.e. not sheared or rotated), there is an additional caching option of saving the whole row results for a given y-interpolant value, which wasn't really possible with the old code structure.

The rectilinear versions got a lot faster as a result of this refactoring.  Unfortunately, most use cases do have shear/rotation (from the WCS typically), so the typical speedup is less.  But even the non-rectilinear functions are 10s of percent faster for common use cases.

The use case in particular that I have in mind is a Piff refactoring that I'm working on.  I have an idea about how to implement the composite PSF, which I'm still working through.  But I found that a version of my new approach was dominated by the calls to `_drawReal`.  This PR speeds up that step by a factor of 2.

Another timing data point is that the total run of nosetests on my laptop reduced by about 10% from this PR.  This is pretty significant, presumably because we use InterpolatedImage in a lot of contexts.  The test suite is of course not representative of real-world usage, but these optimizations help in a lot of different contexts.

Finally, since this was almost the last place we still used XTable and KTable, I went ahead and replaced all the other uses with the direct Image FFTs, which were implemented a while back.  The main other place I had to edit was in the HSM code, where I think the new version in terms of image FFTs is significantly more legible.  I removed the FFT.h and FFT.cpp files, since they are no longer used anywhere.

I'd like to include this update in version 2.3 if someone has time to review this.  Thanks!


galsim/interpolatedimage.py
```diff
@@ -966,8 +961,14 @@ def withGSParams(self, gsparams=None, **kwargs):
     def _sbp(self):
         stepk_image = self.stepk / self.kimage.scale  # usually 1, but could be larger
         with convert_cpp_errors():
+            # C++ layer needs Bounds that look like 0, N/2, -N/2, N/2-1
+            # So find the biggest N that works like that.
```
jmeyers314, line 965: Sounds like this leaves a possibility of truncation, is that right?  Should we warn in that case?

src/Image.cpp
```diff
@@ -116,6 +116,21 @@ struct AlignedDeleter {
     void operator()(T* p) const { delete [] ((char**)p)[-1]; }
 };
 
+template <typename T>
+std::shared_ptr<T> allocateAlignedMemory(int n)
+{
+    // This bit is based on the answers here:
```
jmeyers314, line 122: This bit and the comment make me wonder if I could take advantage of aligned memory in, e.g., batoid.  There I just let numpy do all the allocating.  Any idea if numpy defaults to aligning things on 16 byte boundaries?

src/SBInterpolatedImage.cpp
```diff
@@ -864,20 +1516,18 @@ namespace galsim {
...
+        double* data = const_cast<double*>(reinterpret_cast<const double*>(_kimage.getData()));
+        int No2 = _kimage.getBounds().getXMax();
+        int N = 2*No2;
         return ConstImageView<double>(data, shared_ptr<double>(), 1, 2*N,
-                                      Bounds<int>(0,2*N-1,0,N/2));
+                                      Bounds<int>(0,N+1,-No2,No2-1));
     }
 
+    // Not used anymore.
+    // MJ: Maybe time to remove all these SBProfile serialize functions?
```
jmeyers314, line 1530: Fine with me if they're not being used :)

jmeyers314 (approved):
LGTM.  I admit I scanned through SBInterpolatedImage.cpp pretty quickly, but our unit tests are pretty good, so I bet it's fine :)

galsim/interpolatedimage.py
rmjarvis: This is what the old code did implicitly when copying to the KTable in C++.  But honestly, I don't think it's really ever an issue for the kinds of ways we use InterpolatedKImage.  The edges are usually designed to be negligible.  The only bit of potential information loss is if the input x dimension is even, in which case there is one more x value on the negative side than the positive side.  And this throws that column away.  We used to do that implicitly in the C++ code, since we only copied over the +x values to the KTable, so we missed that outermost -x column.

src/Image.cpp
rmjarvis: No, numpy doesn't (necessarily) align on 16 byte boundaries.  Maybe it does for complex128, since then the data type itself is 16 byte, but I haven't tested it.  We have a different hack to force 128 byte allocations in the python-layer Image class.  Look at the _make_empty method of Image in python.

src/SBInterpolatedImage.cpp
rmjarvis: Good.  Done.

PR #1127, opened 2021-07-03
Adds GitHub action to create manylinux and mac os wheels

### LINKED ISSUE #1056: Binary wheel for GalSim? For easy live demos on Colab
GalSim is a dependency in some other projects of mine, and I'm trying to see if I can make it even easier to install. The conda builds certainly work, but I hate conda, I don't want  to have to use a different python than my system python. Also, it becomes slightly painful in some environments like Google Colab. 

I have this demo notebook that uses the conda package to run GalSim on Colab:
https://colab.research.google.com/drive/1il04VsOgcDhQlEXkeq1g9XXvURYyoucr
But it's not completely satisfactory, and rather slow It needs to install a completely new environment.

The other solution I have is to pip install GalSim source package, this reuses maximally the colab environment, but it's super slow to compile from source:
https://colab.research.google.com/drive/14CE91UF_GbIMdjCQkxluJXj9BAS3VYmW

So, what would be great is a binary wheel version of the pip package. Any reason why we don't have one? Before diving into it I thought I would ask

EiffL:
This PR should resolve #1056 , it adds a GitHub action for builiding wheels using the excellent [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/) project. 

When triggered (by default set to trigger upon new release), the action generates manylinux and mac os wheels for python 3.6  to python 3.9. This should cover all use cases, although I don't know what's the deal with the [Apple Silicon](https://cibuildwheel.readthedocs.io/en/stable/faq/#apple-silicon) and I don't really want to know ^^' my understanding is that the current wheels should work under emulation on new macs but could add option to compile natively....

We could add automatic upload to pypi if you want @rmjarvis, it's easy enough, but it's  currently setup for manual upload.
The way it works is the following:
- The action runs and generates a zip artifact file (available for download from the actions tab)
- Maintainer downloads and unzip the file
- Maintainer runs `twine upload *.whl`

And done! I've tested it on the test pypi: https://test.pypi.org/project/GalSim/#files and you can try to install it from any machine like so:
```bash
$ pip install -i https://test.pypi.org/simple/ GalSim
```
To  demonstrate it works on colab: https://colab.research.google.com/drive/10a20mCGadYLiuBVI7_Mwn5DdOUiIh8FF?usp=sharing

Finally, no need for conda anymore \o/


.github/workflows/wheels.yml
```diff
@@ -0,0 +1,38 @@
...
+
+# To publish on every pull request
+# on: [push, pull_request]
+# To publish when a (published) GitHub Release is created, use the following:
+on:
+  push:
+  pull_request:
+  release:
+    types:
+      - published
```
rmjarvis, line 11: Does this run for all tags?  On the GitHub releases page, it lists all our tags.  But I know some people do more formal "releases" with the "Draft a new release" button.  Do I need to do that to trigger this?  Or will just tagging vM.m.r trigger it?

And if the latter, I think we may want to skip some tags.  E.g. those with rc in the name.  So we might want to add a check that the tag name is really tagging a real version number.

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,38 @@
...
+
+jobs:
+  build_wheels:
+    name: Build wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, macos-10.15]
```
rmjarvis, line 19: I think I understand this that the ubuntu one will use docker to install all the various python flavors using "manylinux".  Whatever that means.  :)  

And we don't need to do the multiple python versions in the matrix, since it will automatically spawn all of those?

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+
+      # Used to host cibuildwheel
+      - uses: actions/setup-python@v2
+
+      - name: Install cibuildwheel
+        run: python -m pip install cibuildwheel==2.0.0a4
```
rmjarvis, line 28: Do we really want to pin this to an alpha release?  In fact, do we gain anything from using 2.0, rather than their current stable release (1.11)?

rmjarvis (approved):
Thanks François!  This looks great!  My review is mostly questions, since I don't really know anything about this.  :)  In fact, if you don't know the answers to some of these, that's fine.  Maybe the best way to answer them is to just merge this and start trying it out.

.github/workflows/wheels.yml
rmjarvis: Actually, I'm concerned that this PR itself seems to have triggered the build jobs:
```
Build / Build wheels on ubuntu-20.04 (pull_request) 
Build / Build wheels on macos-10.15 (pull_request) 
```
Was that expected.  We definitely don't want them to trigger on every pull request.  Just on new version releases.

EiffL: ahah, yes, that wasn't expected, but should have been, I didn't think it through. This means running the workflow on push, pr, AND release. We just want the last one. Will fix it.

As to what the "published" release type means, I think this triggers the workflow indeed when you use this interface: https://github.com/GalSim-developers/GalSim/releases/new and click the "publish" button. Here is the list of types of releases: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release Not  clear if/how to trigger just on  tags. I'll give it a try on my fork and let you know.

In any case, you can always trigger the workflow manually from the workflow tab. Given that you don't push new pypi wheels everyday, and that this is not currerntly automatically uploading to pypi, it might be acceptable to trigger manually.

.github/workflows/wheels.yml
EiffL: exactly, the linux matrix entry will just use a series of containers to build the various versions of the wheel. 

The manylinux image is essentially a barebone linux with maximal compatibilty with other linux variants. The script builds the wheel in that environment, and also uses a magical tool to package in the wheel all shared libraries needed  for the execution. So that the final binary is almost completely self contained, only retain some specificity against the python version and platform used to compile it.

.github/workflows/wheels.yml
EiffL: hum yes, I used the latest example from their repo, it looks like the latest stable version is 1.12. They recommend using dependabot for keeping this up to date, but  I have never used it: https://cibuildwheel.readthedocs.io/en/stable/faq/#automatic-updates

So I guess, here using 1.12 would  make a lot of sense. I'll change the version

EiffL:
I've made some changes based on your suggestions, essentially:
- Removed trigger on push and pull requests
- Simplified script using the cibuildwheel action, and switched to last stable release 1.12.0

I think this should now work as expected. Maybe it won't trigger on simple tags... not sure... but you can always trigger it manually from the action menu.

EiffL:
I experimented a bit more, 2 observations:
- Just adding a tag doesn't trigger  the workflow, you need to use the GitHub release API. For instance I added a release on my fork, and it triggered the workflow: https://github.com/EiffL/GalSim/actions/runs/998299940
- I also added the workflow_dispatch trigger event, which allows you to manually trigger the workflow on whatever branch you like: 
image

rmjarvis:
Great.  Let's try it out.  I'll probably eventually automate the twine call too, but probably better to keep that step manual at first.

rmjarvis:
Hm.  I made a [release](https://github.com/GalSim-developers/GalSim/releases/tag/v2.3.0) and published it, but it didn't seem to trigger the action.  :(  Well, I'll do the manual trigger for now.

rmjarvis:
Maybe because I used an existing tag?  Rather than have the release itself make the tag?  Not sure.  I set the target to releases/2.3, and I did it after merging this PR, so it should have been there.

Next time I need to make a release tag, I'll do it via the Release page rather than making the tag by hand first, and see if that triggers it.  No big deal of course if it doesn't work, since you made it so I can trigger it manually, but I'm just confused why it's not working as expected.

EiffL:
Hum.... yeah also confused... let me try to create a release of an existing tag on my fork, see if it triggers

EiffL:
annnnd yes, confirmed. releasing an existing tag doesn't trigger the action :-| oh well...

rmjarvis:
That's fine.  That's a perfectly reasonable workflow to use the Release page to generate both the tag and the wheels.  The last check will be when I upload it to pypi to make sure everything installs properly with pip.  But the action is still running.  Will check that later.

rmjarvis:
Success!  I uploaded them all to pypi, and AFAICT it all seems to be working.  I spot checked a few systems, including both Mac and Linux, using a few different flavors of python, and they all downloaded the whl and installed super fast.  Yay!

Thanks again François!

EiffL:
Fantastic!!! Thanks Mike!

rmandelb:
I'm late to this party, but wanted to say thank you - this looks great!

PR #1128, opened 2021-07-06
Fix shared clib

rmjarvis:
@erykoff has been trying out the new shared library build for use with the LSST DM stack, and found some problems.  Specifically, it had required adding `-lfftw3 -fopenmp` when using the library, which is not ideal.

This PR is to fix the linking steps to work better to not need to explicitly add these link flags whenever using the libgalsim.so library.

Will release 2.3.1 with this fix when Eli gives me the go ahead.


erykoff:
This branch is working for me.

PR #1129, opened 2021-07-12
Remove old stuff

rmjarvis:
This is purely a cleanup PR.  It removes the things we said we'd get rid of after v2.3:

* Python 2.7 support
* TMV support
* Boost Python support
* SCons support

Along the way, I cleaned up various other bits of cruft as I came across them:

* No longer seem to need IgnoreWarnings.h.  (I'll check the GHA builds, but my portability checks didn't find any platforms that required it to avoid compiler warnings now.)
* Removed 1.x-depracated CppEllipse.h, which we should have removed at 2.0
* Removed sizeof_SIFD.cpp, which is no longer useful.
* Moved fds_test.py to devel directory.  Not sure if it's still useful.  But it doesn't need to be in the main galsim distro.
* Updated a few places (mostly in tests) where we had workarounds for deficiencies in python 2.  E.g. contexts that Josh wanted to use, but didn't work right in test_phase_psf.py.
* I re-checked the timings of the gzip/bzip2 options.  It seems that the Python modules improved their speed for writing, so that's no longer faster than the external call, but the read is still slow.  So I made gzip_in_mem and bz2_in_mem the default now for writing.  This in turn means that we don't need the infrastructure for handling multiple write functions that we had there.  (Still need it for reading though.)
* Removed AstronomicalConstants.h, which we don't use, and I'm not actually sure if we ever did.
* Removed our custom implementation of tgamma and lgamma, which we had kept in for non-c++11 compilers.  But we don't support those anymore, so don't need it.
* Removed ups directory.  I think this was only a SCons thing, but @jmeyers314 please correct me if we still need this when using setup.py?


jmeyers314:
The `ups/` directory is an `eups` thing, not a `scons` thing.  I think it will still be useful for me (maybe only me?) for switching between release and dev GalSim versions while inside a stack environment.

rmjarvis:
OK.  I reverted that commit.  I thought when doing a more pip-y install, you wouldn't need these environment variables set anymore.  But I'm fine with leaving it.

cwwalter:
Yes, I think I might need it too. I was worried I might not be able to use it in a pip installed env, but I think I remember Josh figured out how to do it.

jmeyers314:
@cwwalter, my strategy is to install stack as normal (I use `newinstall.sh`).  Then `pip install develop` GalSim.  At this point, for reasons I don't understand, the conda-installed version of GalSim still takes precedence.  But, if you declare the dev version to eups: `eups declare -r $GALSIM_SRC_DIR galsim git`, then you can `setup galsim` (this is where the `ups/` directory is required) to give the dev version precedence, and `unsetup galsim` to go back to the conda-release.

tests/galsim_test_helpers.py
```diff
@@ -434,10 +420,6 @@ def do_pickle(obj1, func = lambda x : x, irreprable=False):
     assert object() != f1
 
     # Test the hash values are equal for two equivalent objects.
-    try:
-        from collections.abc import Hashable
-    except ImportError:
-        from collections import Hashable
     if isinstance(obj1, Hashable):
         #print('hash = ',hash(obj1),hash(obj2))
         assert hash(obj1) == hash(obj2)
```
jmeyers314, line 425: Below on lines 470-528:  This section appears to only barely get run anymore; just random.py features an explicitly coded `__getinitargs__`.  I'm guessing this was more useful in the past when the c++-wrapped classes were old-style in py27 maybe?  (Maybe only in boost_python?)

I believe py3 only features new-style classes, even when inheriting from `object`, so recommend global-search-replace to just make everything look "new-style".

src/Table.cpp
```diff
@@ -666,25 +658,7 @@ namespace galsim {
...
-                rhs(i-1) = 6. * ( (_vals[i+1] - _vals[i]) / (_args[i+1] - _args[i]) -
-                                  (_vals[i] - _vals[i-1]) / (_args[i] - _args[i-1]) );
-            }
-            tmv::Vector<double> solution(_n-2);
-            solution = rhs / M;   // solve the tridiagonal system of equations
-            for (int i=1; i<=_n-2; i++) {
-                _y2[i] = solution[i-1];
-            }
-#else
+            // For 4 or more points we used to use the TMV symmetric tridiagonal matrix solver.
```
jmeyers314, line 661: Should this say `Eigen`?

jmeyers314 (approved):
LGTM.  Just a couple small questions below.

devel/getACSBandpass.py
```diff
@@ -20,7 +20,6 @@
 this script, which are the files GALSIM_DIR/share/bandpasses/ACS*.dat, are already included in the
 repository.  This script just lets users know where these files came from and how they were altered.
 """
-from __future__ import print_function
```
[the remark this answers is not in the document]
rmjarvis: Probably not.  I'm quite sure some of these devel scripts used GalSim 1.x API, so they won't work.  Some might even predate 1.0 and have additional problems.  Seems like it would be a lot of work to maintain these at even a minimum level of functionality.  I think they are more present just for reference with the expectation that some work might be needed to get them working again if there is a need to run the code again.

src/Table.cpp
rmjarvis: No, I think TMV.  We used to use TMV for this, which implements band matrices.  I didn't check again if Eigen has added them yet, but when I converted to 2.0 and added Eigen then, they didn't have any band matrix functionality.  So I implemented the version here.  But the historical reference is correct to say TMV.

tests/galsim_test_helpers.py
rmjarvis: Interesting. Good point, although the point of that section was to check that our `__eq__` function were checking everything important.  So that means we don't really have a good check of that anymore.  I don't know if it's worth trying to do something clever with `inspect` maybe to try to get that same test functionality?  Probably not I guess.
I can report that the one place left that used `__getinitargs__`, which you pointed out, doesn't need it.  Removing that function keeps `DistDeviate` serializing just fine.  So that was some old workaround, probably from pre-2.0.  I deleted that, which means this test section now never gets run at all.
So what do you think, should I remove it as useless, or leave it with a note that the argument list it loops over would need to be changed to some other technique to make it useful again?

rmjarvis: Side point.  I also got rid of this little amusing DistDeviate method that I assume used to actually do something:
```
@property
    def _function(self):
        return self.__function if isinstance(self.__function, str) else self.__function
```
😆

src/Table.cpp
jmeyers314: Ah, right.  I didn't read the comment closely enough.  I thought it said we were currently using TMV.

tests/galsim_test_helpers.py
jmeyers314: > I don't know if it's worth trying to do something clever with inspect maybe to try to get that same test functionality?

I'm not sure if there's anything in `inspect` that would take an object and reproduce its ctor args.  We might be able to perturb the items returned by `__getstate__`, but I'm not sure it's worth it.

We still have `all_obj_diff`, of course.  The downside is that we have to remember to use it instead of relying on objects being automatically tested.  So I vote to just remove this code.

jmeyers314: Also, I'm not sure I was very clear above, but by "recommend global-search-replace to just make everything look "new-style"" I meant to comb through the code and remove any inheritance from `object`, just to make the code look more modern.

rmjarvis: > remove any inheritance from object
Ah.  Good.  Yes, I'll do that.  I didn't pick up that intent the first time. :)

PR #1131, opened 2021-07-16
Fix rare possibility of SED.sampleWavelength generating wavelengths outside of bandpass limits.

rmjarvis:
@matroxel ran into a rare bug in the Roman simulations where some wavelengths would be barely outside of the allowed range for the PSF (which is based on the bandpass limits):
```
galsim.errors.GalSimRangeError: Shooting photons outside the interpolated wave_list Value [1192.14102378 
972.39138558 1058.912179  ... 1170.73736364 951.20426042 1180.4948833 ] not in range [885.0, 1245.0]
```
First, that error message isn't particularly useful, since most of the wavelengths are fine.  So I changed the messaging to only list the bad wavelengths.  That produced error messages like the following (for a different band than above):
```
galsim.errors.GalSimRangeError: Shooting photons outside the interpolated wave_list Value 
[1604.9999999999998] not in range [1605.0, 2090.0]
```
Clearly this is just a numerical rounding error.  I tracked it down to the SED.sampleWavelength function.  In that function, we sample wavelengths in the SED's native reference frame.  If it is nominally at some other redshift, then the bandpass limits are divided by (1+z).  Then after the waves are generated, we multiply back by (1+z) to get the final waves.  This round trip can lead to errors of order epsilon, which meant you could get the range error that Troxel found.

The solution in this PR is to just clip the values to the bandpass limits at the end (with `np.clip`).  I'm targeting this at releases/2.3, so we can cut 2.3.2 with this fix.

### DIFF
--- a/galsim/chromatic.py
+++ b/galsim/chromatic.py
@@ -1324,8 +1324,9 @@ class InterpolatedChromaticObject(ChromaticObject):
 
         w = photons.wavelength
         if np.any((w < self.waves[0]) | (w > self.waves[-1])):
+            bad_waves = [w for w in photons.wavelength if w < self.waves[0] or w > self.waves[-1]]
             raise GalSimRangeError("Shooting photons outside the interpolated wave_list",
-                                   w, self.waves[0], self.waves[-1])
+                                   bad_waves, self.waves[0], self.waves[-1])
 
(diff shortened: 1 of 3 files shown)


jmeyers314 (approved):
LGTM

PR #1136, opened 2021-09-10
Fix typo in GSFitsWCS

jmeyers314:


rmjarvis (commented):
I guess this is ok. But you can only trigger this as an error if you also neglect to update wcs_type.  In your test, you make wcs2 a self contradictory wcs specification, since `projection` is `'gnomonic'`, but `wcs_type` is still `'ARC'`.

jmeyers314:
Ah, okay.  What initially caught my eye was the duplicated `np.array_equal(self.cd,other.cd)` line.  I guess I figured the duplication indicated something was missing instead of simply being superfluous.  And now I see that yes, equal `wcs_type`s implies equal `projection`s, so long as one isn't directly manipulating attributes.

No need to complicate `__eq__` and unit tests.  I'll simplify this to a more cosmetic fix.

jmeyers314:
Self-review since now trivial.

PR #1137, opened 2021-09-29
Iterate over boundaries linearly in distortion update

jamesp-epcc:
This PR modifies the distortion update to iterate over the pixel boundaries linearly instead of per-pixel as before. This speeds up the tests by around 15% and simplifies the code.

### DIFF
--- a/src/Silicon.cpp
+++ b/src/Silicon.cpp
@@ -490,97 +490,97 @@ namespace galsim {
                 int polyj1 = std::max(j - (_qDist+1), j1);
                 int polyj2 = std::min(j + (_qDist+1), j2);
 
-		// work out indices of update region in horizontal linear array
-		int horizBoundaryX = (((i - i1) - _qDist) * horizontalPixelStride());
-		int horizDistortionX = ((nxCenter - _qDist) * horizontalPixelStride());
-		int horizW = (((_qDist * 2) + 1) * horizontalPixelStride());
-		if (horizBoundaryX < 0) {
-		    // handle case where stencil goes off left side
-		    horizW += horizBoundaryX;
-		    horizDistortionX -= horizBoundaryX;
-		    horizBoundaryX = 0;
-		}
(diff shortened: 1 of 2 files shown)


src/Silicon.cpp
```diff
@@ -561,27 +489,95 @@ namespace galsim {
...
+                    int index = ((horizBoundaryY + y) * horizontalRowStride(nx)) + horizBoundaryX;
+                    int dist_index = ((horizDistortionY + y) * horizontalRowStride(_nx)) + horizDistortionX;
+                    for (int x = 0; x < horizW; ++x, ++index, ++dist_index) {
+#ifdef _OPENMP
+#pragma omp atomic
 #endif
+                        _horizontalBoundaryPoints[index].x += _horizontalDistortions[dist_index].x * charge;
+#ifdef _OPENMP
+#pragma omp atomic
+#endif
```
rmjarvis, line 531: I would think these atomic pragmas are slowing things down in a parallel context.  I think we could flip the order of loops here to have the outer loop be over the boundary points arrays, which would be the ones we would parallelize over.  Then each thread would have a different array, so they couldn't clobber each other on writes. The inner loops would be over the pixels, which could access the same pixels as each other sometimes, but only in a read capacity, so no race conditions.  That strikes me as likely to be faster.

rmjarvis (changes requested):
Hi James.  This is good so far, but I suspect we can do even better by flipping the order of the loops.  I think this would let us get rid of those atomic pragmas, which (I think?) tend to be bad for parallel efficiency.

src/Silicon.cpp
jamesp-epcc: Ah, I see what you mean. Yes, that might well be faster, certainly for the cases where most pixels have some level of charge in them. The cost of the atomics is indeed quite high (I timed running the tests with the atomic pragmas disabled and they ran 2-3x faster, obviously with some errors due to the race condition, but gives a good idea of how much they slow things down). I'll code up a version with the loops flipped and see how fast it goes.

PR #1139, opened 2021-12-06
Add methods to flip and rotate an Image

rmjarvis:
This is a quick PR to add some convenience functions to the Image class.

I recently wanted to make a transpose of an image, and we don't have anything out of the box that does this.  It's not too hard to do by hand, but it still seems like it would be nice to have this natively.  While I was at it, I added a few other similar methods that could be potentially useful:

im.transpose - swaps x,y coordinates
im.flip_ud - flips y values ("up and down")
im.flip_lr - flips x values ("left and right")
im.rot_cw - rotates 90 degrees clockwise
im.rot_ccw - rotates 90 degrees counter-clockwise
im.rot_180 - rotates 180 degrees

In all cases, the returned image has either the same bounds or the transposed bounds.  That feels more natural I think than the other choices.  If the user wants something different, they can shift or recenter by hand.

Also, the wcs of the returned image is None (i.e. undefined).  It seemed too hard to get the wcs to work correctly, and probably anyone who would be using these methods won't care about the wcs anyway.  If they want to set it to something, they'll have to do that by hand as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -202,6 +202,7 @@ jobs:
                 GALSIM_TEST_PY=0 python setup.py test
 
             - name: Upload coverage to codecov
+              if: matrix.py != 'pypy3'
               run: |
                 cd tests
                 pwd -P
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -34,7 +34,7 @@ Config Updates
 New Features
 ------------
 
-- Added Image methods: tranpose, flip_ud, flip_lr, rot_cw, rot_ccw, rot_180.
+- Added Image methods: tranpose, flip_ud, flip_lr, rot_cw, rot_ccw, rot_180. (#1139)
 
 
 Performance Improvements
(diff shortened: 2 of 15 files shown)


rmjarvis:
Update: I also fixed a test failure that started happening in the hsm code on the macos CI run, completely unrelated to this PR.  It was a bear to track down, since I couldn't replicate it locally.  My best guess is that it was due to some obscure compiler bug, where it was over-optimizing a section of code.  If I had it print an intermediate value, then it worked fine, but if not, then it would get a different result for a bounds calculation.  

Usually these kinds of "heisenbugs" are due to some kind of undefined behavior, such as array overruns or using uninitialized data.  I couldn't find anything like that, but I left in some of the infrastructure I added to verify that -- the Image class now keeps track of that maximum allocated pointer to make sure that writes don't go past that point in various places where we use direct pointer arithmetic.

Anyway, the solution was just to simplify that calculation to something not quite so convoluted.  Now it seems to be working fine.

jmeyers314 (approved):
LGTM.

rmjarvis:
Thanks Josh!

PR #1140, opened 2021-12-16
Use single-precision for Silicon internal storage

rmjarvis:
This PR changes the internal storage of the pixel boundaries from double to single precision.  I made sure everything still uses double precision for all calculations, so there isn't too much loss of precision in the various places.  

This is most important when the sides may add 1 to their positions depending on which side of the square they are on.  If these are done in single precision, there is more loss of precision.  Probably still acceptable, TBH, but it seems prudent to do as much as possible in double precision nonetheless.

Anyway, only one test ended up needing to be relaxed in its precision, and I dug into it and convinced myself that it is unavoidable.  Basically, it's when we first apply the tree rings to all the boundaries, then go back and calculate areas.  This necessarily passes through float storage, so it only ends up accurate at rtol=2e-8.  That seems completely fine to me.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,13 +147,16 @@ jobs:
                 # Extra packages needed for testing
                 pip install -U -r test_requirements.txt
                 pip install -U nose codecov coverage
-                pip install -U pandas matplotlib astroplan
+                pip install -U matplotlib astroplan
 
             - name: Install CPython-only dependencies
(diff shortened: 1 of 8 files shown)


jamesp-epcc (approved):
This all looks good for me. I'm glad to see that the approach of ensuring intermediate calculations are still done in double precision has improved the accuracy over my original implementation.

rmjarvis:
Thanks James!

PR #1141, opened 2022-01-04
Move Silicon recalculation logic to python layer

rmjarvis:
This PR moves all the logic of when to do the pixel updates from the C++ accumulate function to the python layer function.

This should make it much simpler for us to adjust how this is done when we are accumulating over the whole image at once rather than just doing one object at a time.  But so far I haven't changed any logic -- just made the appropriate hooks available in python.

### DIFF
--- a/examples/demo11.py
+++ b/examples/demo11.py
@@ -217,7 +217,7 @@ def main(argv):
     dvdx = numpy.sin(theta) * pixel_scale
     dvdy = numpy.cos(theta) * pixel_scale
     image_center = full_image.true_center
-    affine = galsim.AffineTransform(dudx, dudy, dvdx, dvdy, origin=full_image.true_center)
+    affine = galsim.AffineTransform(dudx, dudy, dvdx, dvdy, origin=image_center)
 
     # We can also put it on the celestial sphere to give it a bit more realism.
     # The TAN projection takes a (u,v) coordinate system on a tangent plane and projects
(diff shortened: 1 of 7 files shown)


examples/demo11.py
```diff
@@ -247,9 +247,11 @@ def main(argv):
         # We will need the image position as well, so use the wcs to get that
         image_pos = wcs.toImage(world_pos)
 
-        # We also need this in the tangent plane, which we call "world coordinates" here,
+        # We also need this in the tangent plane, which we call "uv coordinates" here,
         # since the PowerSpectrum class is really defined on that plane, not in (ra,dec).
-        uv_pos = affine.toWorld(image_pos)
+        world_center = wcs.toWorld(image_center)
+        u, v = world_center.project(world_pos)
+        uv_pos = galsim.PositionD(u/galsim.arcsec, v/galsim.arcsec)
```
rmjarvis, line 254: This change isn't related to the rest of this PR, but a couple of the systems in GitHub CI started failing.  I think it must be a change on their end (either architecture or compiler, etc.) since this code doesn't use the Silicon class at all.  
In exact math, these two ways of calculating uv_pos are equivalent, and on all the machines I have access to, they differed only at the 1.e-16 level, which was fine.  But on some of the ubuntu GHA machines, they differed at 1.e-13, which was enough to eventually propagate into some pixels being different at the 1.e-8 level, so the yaml and py versions of the images (dtype=float32) were not identical, as we require.
So I changed the demo11.py code to calculate uv_pos the same way the config layer does.  I don't thin there's any loss in pedagogy from this, so this seems fine to me.

jamesp-epcc (approved):
This looks good to me. As far as I can see it preserves the old behaviour but with the new division between Python and C++.

cwwalter:
Thanks. I hadn't looked closely at this part of GalSim before.  

What is "effective" about recalc now? It looks like the strength was also still used with the old `recalc`.  Is it because you now need to track `._accum_flux_since_update`?

rmjarvis:
> What is "effective" about recalc now?

It was always effective, but before it was a temporary variable in that function.  I named it `self.effective_recalc` to not clash with the original input `self.recalc`.  

It's only relevant when ramping up the BFE strength.  Normally the two values are identical.  But if the strength is high, you need to lower recalc to do the recalculations at the right points still.  That's the "effective" meaning here.

cwwalter (approved):
LGTM

PR #1142, opened 2022-01-10
roman PSF fix

ztq1996:
Change the unit conversion when reading roman files in `roman.roman_psfs._read_aberrations(SCA)`


rmjarvis (changes requested):
The code looks fine to me, given what you showed on Slack.
One thing I'd like in addition though is a unit test that validates the pixel scale and the conventions.  I was thinking on something motivated by these two plots you made:
image (1)
image
It would be nice to loop through the various gaps (at least a few, but maybe even all of them) and verify that the change in the aberration values across the gap is much less than the typical change across a full SCA.

rmjarvis:
By the way, the test failure has nothing to do with your changes.  Apparently distutils.version.LooseVersion is now deprecated.  Since we don't actually need it anymore, I just removed our use of it in some tests.  So if you rebase to the current releases/2.3, that error should go away.  (I also cherry-picked a couple other recent things like this that had been fixed on main.)

ztq1996:
I have added a unit test that assert that the absolute values of the differences between the conjunction points are much less than or equal to the variation within the entire FOV. I removed some conjunction pairs compared to the original choices, because I think adjacent chips that has an offset (for example SCA15 and SCA12) should **not** have perfect continuity because they are not interpolated from the similar reference positions. The new sets of conjunction points looks like this
image
The test will pass for |Z_DIFF|< 0.05(Z_MAX - Z_MIN), although for the conjunction that are a little far away (for example SCA3 and SCA7), I have to set the condition to |Z_DIFF|< 0.10(Z_MAX - Z_MIN).

rmandelb:
sounds promising; have you checked that e.g. an SCA orientation flip or units confusion in the interpolation will cause this test to fail?

ztq1996:
Yes, either the unit confusion and the orientation flip will fail the test.

tests/test_roman.py
```diff
@@ -1052,6 +1052,84 @@ def test_config_sca():
...
+    # Zhang. The conjunction pairs between the first row and the second row are considered 
+    # 'far', because are further separated. For 'far' pairs, z_diff<= 0.1*(z_max - z_min), 
+    # elsewhere, z_diff<= 0.05*(z_max - z_min).
+
+    print("Start continuity test for aberration interpolation")
+    
+    print('Read the SCA# and sca_pos of the conjunction pairs. (SCA#1, SCA1_X, SCA1_Y, SCA#2, SCA2_X, SCA2_Y )')
+    edge_pair_array = np.loadtxt("roman_files/sca_conjunction_points.txt")
+    # Z_min, Z_max across the FPA
+    Z_min_max = np.loadtxt('roman_files/Z_min_max.txt')
```
rmjarvis, line 1073: I think this would be clearer to just calculate it from the aberration file directly after we read them in.

I think something like this should work:
```
Z_min = np.min(aberration_list, axis=(0,1))
Z_max = np.max(aberration_list, axis=(0,1))
```
but really, however you calculated the values to create this file is probably right.

tests/test_roman.py
```diff
@@ -1052,6 +1052,84 @@ def test_config_sca():
...
+
+
+
+    for pair in edge_pair_array:
+
+        sca1 = int(pair[0])
+        sca_pos_1 = galsim.PositionD(x = pair[1] , y = pair[2])
+
+        sca2 = int(pair[3])
+        sca_pos_2 = galsim.PositionD(x = pair[4] , y = pair[5])
```
rmjarvis, line 1094: This might be harder, but if it's possible to get these values programmatically, that would be better.  You probably need to list the sca pairs explicitly, but if the relevant positions could be computed from that, it would be nice to not need the input file.

So,
```
sca_pairs = [ (1,2), (2,3), (4,5), ...., (11,2), (12,3) ]

for sca1, sca2 in sca_pairs:
    # Calculate the ~midpoint positions along the gap between these two scas.
    sca_pos_1 = ...
    sca_pos_2 = ...
```

tests/test_roman.py
```diff
@@ -1052,6 +1052,84 @@ def test_config_sca():
...
+            if i ==22:
+                continue
+
+
+            z_diff = np.abs(point_2_abe[i] - point_1_abe[i])
+
+            if far:
+
+                assert z_diff<= 0.1*(z_max - z_min), 'z_diff > 0.1(z_max - z_min), failed aberration continuity test.\
+                \n \fail for Z{}, z_diff = {}, sca1 = {}, sca2 = {}'.format(i,z_diff,sca1,sca2)
```
rmjarvis, line 1121: I think the "\fail" isn't what you meant.  Should be "Fail" I guess, to match the format below?

rmjarvis (changes requested):
Thank Tianqing!

This is all basically fine I think.  My only request would be to get the number values for the test programmatically, rather than from input files.  I think this is easy for the Z_min/max, but maybe harder for the positions.

If you used code to do the calculations to make that file, then you can probably just port that into the test directly, which would keep the test more self-contained.  Also easier to maintain if anything about the focal plane geometry changes in the future.

tests/test_roman.py
ztq1996: Resolved

tests/test_roman.py
ztq1996: Resolved

galsim/roman/roman_psfs.py
```diff
@@ -382,10 +382,11 @@ def _read_aberrations(SCA):
...
     aberrations[:,1:] = dat[:,5:]
-    # Also get the field position.  The file gives it in arcsec with respect to the center, but we
-    # want it in pixels with respect to the corner.
-    x_sca_pos = dat[:,1]/pixel_scale + n_pix/2
-    y_sca_pos = dat[:,2]/pixel_scale + n_pix/2
+    # Also get the field position.  The file gives it in mm with respect to the center, but we
+    # want it in pixels with respect to the corner. The pixel size of the detector is 0.01 mm/pixel
+    pixel_size = 0.01 
+    x_sca_pos = dat[:,1]/pixel_size + n_pix/2
+    y_sca_pos = n_pix/2 - dat[:,2]/pixel_size
```
rmandelb, line 389: Perhaps it would help future developers to include a comment explaining this sign flip?

tests/test_roman.py
```diff
@@ -1052,6 +1051,173 @@ def test_config_sca():
...
+    sca_xc_mm = -sca_data[3,:]
+    sca_yc_mm = sca_data[4,:]
+    
+    horizontal_adjacent = []
+    # list of horizontal contact (SCA# DOWN, SCA# UP)
+    for i in range(1, 17, 3):
+        horizontal_adjacent.append((i,i+1))
+        horizontal_adjacent.append((i+1,i+2))
+    
+    verticle_adjacent = []
```
rmandelb, line 1082:
```suggestion
    vertical_adjacent = []
```

tests/test_roman.py
```diff
@@ -1052,6 +1051,173 @@ def test_config_sca():
...
+    # list of verticle contact (SCA# LEFT, SCA# RIGHT)
```
rmandelb, line 1083:
```suggestion
    # list of vertical contact (SCA# LEFT, SCA# RIGHT)
```

tests/test_roman.py
```diff
@@ -1052,6 +1051,173 @@ def test_config_sca():
...
+    for i in range(10,13):
+        verticle_adjacent.append((i,i-9))
```
rmandelb, line 1085:
```suggestion
        vertical_adjacent.append((i,i-9))
```

galsim/roman/roman_psfs.py
```diff
@@ -382,10 +382,11 @@ def _read_aberrations(SCA):
     # use entries 1-22).  The units are waves.
     aberrations = np.zeros((5,23))
     aberrations[:,1:] = dat[:,5:]
-    # Also get the field position.  The file gives it in arcsec with respect to the center, but we
-    # want it in pixels with respect to the corner.
-    x_sca_pos = dat[:,1]/pixel_scale + n_pix/2
-    y_sca_pos = dat[:,2]/pixel_scale + n_pix/2
+    # Also get the field position.  The file gives it in mm with respect to the center, but we
+    # want it in pixels with respect to the corner. The pixel size of the detector is 0.01 mm/pixel
+    pixel_size = 0.01
```
rmandelb, line 387: I was wondering if it's worth storing this number in `__init__.py` along with other basic parameters of Roman, so it can be used here and in the test suite without hard coding it as 0.01?  Seems reasonable to keep it with the rest of the basic reference info there.

tests/test_roman.py
```diff
@@ -1052,6 +1051,173 @@ def test_config_sca():
...
+
+        sca_pos_1 = galsim.PositionD(x = npix/2 -x_1_mm/pixel_size , y = y_1_mm/pixel_size + npix/2)
+        sca_pos_2 = galsim.PositionD(x = npix/2 -x_2_mm/pixel_size , y = y_2_mm/pixel_size + npix/2)
+
+
+        edge_pair_list.append([sca1+1,sca_pos_1.x, sca_pos_1.y, sca2+1, sca_pos_2.x, sca_pos_2.y])
+
+
+
+    for pair in range(len(verticle_adjacent)):
```
rmandelb, line 1113:
```suggestion
    for pair in range(len(vertical_adjacent)):
```

tests/test_roman.py
```diff
@@ -1052,6 +1051,173 @@ def test_config_sca():
...
+        sca1 = verticle_adjacent[pair][0]-1
```
rmandelb, line 1114:
```suggestion
        sca1 = vertical_adjacent[pair][0]-1
```

tests/test_roman.py
```diff
@@ -1052,6 +1051,173 @@ def test_config_sca():
...
+        sca2 = verticle_adjacent[pair][1]-1
```
rmandelb, line 1115:
```suggestion
        sca2 = vertical_adjacent[pair][1]-1
```

rmandelb (changes requested):
Thanks for finding and fixing this bug!  I have a few suggestions for your consideration.

rmjarvis (approved):

galsim/roman/roman_psfs.py
ztq1996: I add a line here to state that the y-coordinate is flipped, but the roman file that says this cannot be found online, nor is it in the shared folder.

galsim/roman/roman_psfs.py
ztq1996: Yeah. I have stored the value as `pixel_scale_mm` in the roman `__init__.py`

ztq1996:
Hi @rmandelb!
Thanks for proposing these changes, I have accepted and made corresponding changes.

rmandelb (approved):

rmandelb:
@rmjarvis - is the idea that we should merge this PR into `releases/2.3` as a bug fix, but also get these changes onto master?

rmjarvis:
Yep.  I'll release the bugfix version and cherry-pick this onto main.  Thanks for this fix Tianqing!

rmjarvis:
v2.3.4 has been released.

PR #1143, opened 2022-01-14
Add feature to allow registration of templates

rmjarvis:
This PR adds the option to register template names, so people can use that rather than a file name.

The target use case is imSim, so it can register names like 'dc2_image' or 'flat_field', etc., which could point to a config file in the imSim repo.  This file location is likely to be slightly obscure, so having a comprehensible alias would be both easier for the user and more portable across platforms.

### DIFF
--- a/galsim/config/process.py
+++ b/galsim/config/process.py
@@ -170,7 +170,13 @@ def ProcessTemplate(config, base, logger=None):
         # Copy over the template config into this one.
         new_params = config.copy()  # N.B. Already popped config['template'].
         config.clear()
+
         config.update(template)
+
+        if 'modules' in config:
+            # We want to keep all the modules from either place in the config dict
+            # so things like Eval can import everything that might be needed.
+            config['modules'].extend(new_params.pop('modules', []))
 
         # Update the config with the requested changes
         UpdateConfig(config, new_params)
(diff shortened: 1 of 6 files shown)


galsim/config/process.py
```diff
@@ -221,14 +246,17 @@ def Process(config, logger=None, njobs=1, job=1, new_params=None, except_abort=F
...
     # Process any template specifications in the dict.
     ProcessAllTemplates(config, logger)
 
     # Update using any new_params that are given:
     if new_params is not None:
         UpdateConfig(config, new_params)
 
-    # Import any modules if requested
+    # Do this again in case any new modules were added by the templates or command line params.
     ImportModules(config)
```
cwwalter, line 260: I think I have seen cases where, depending on what people are doing, importing modules twice can cause errors.  Could that be a worry here?

rmjarvis: It shouldn't for any well behaved package.  I'd be very surprised if this caused problems for anyone.  Python normally allows a repeated import statement to basically act as a no op.

The reason for this is that we want both:

1. Named templates in one of the packages in the `modules` list to be available for `template` specifications.  (This is the use case for the imsim user config for instance)
2. Template files that include a `modules` directive should be loaded, so any registered things there are processed properly.  (We have example configs in the great3 and des subdirs that rely on this.)

So importing the modules needs to happen both before and after the template processing.

Issue #1145, opened 2022-04-03
Si(x) and Ci(x) approximations accuracy for x>4


enzymepower:
I've found the current Si(x) and Ci(x) approximations, when x>4, to be quite inaccurate.
I've plotted Si(x), the current approximation and some other approximations for comparison.

'smallsi' is the polynomial for the Si(x) pade approximation 0<=x<=4
'fgApprox' is the current approximation using pade approximated f(x) and g(x) and the smallsi for 0<=0<4
'xx2Approx' uses f(x)=1/max(1, x) and g(x)=1/max(1, x^2)
'approxSisecond' uses a few terms from the series expansion at infinity and the smallsi for 0<=0<4
'Si(x)' is Si(x)

![wide_sinintegrals](https://user-images.githubusercontent.com/6540723/161421218-a5d646ea-470b-4902-8d21-b04d1bdeed84.png)
![wide_sinintegralsError](https://user-images.githubusercontent.com/6540723/161421243-43042bc8-4f38-4372-9482-88c6103aae96.png)

As you can see, the current fg approximation is almost always worse than just using 1/x and 1/(x^2) and even crosses the x axis while x>0.

rmjarvis:
I'm not sure how you're computing these, but I'm not able to reproduce your plot with the GalSim implementation of Si.  I exposed our implementation to the python layer to make it easier to plot and test (branch `expose_si`), and I'm getting extremely good agreement with the scipy.special implementation.  Less than 1.e-15 difference at all 0<x<100.

![si](https://user-images.githubusercontent.com/623887/161550701-1e56e21a-ae33-4128-8c5c-c3a266e4f30a.png)

rmjarvis:
cf #1146.  In short, all tests pass, so I don't think there is anything to fix here.

enzymepower:
Sorry, my mistake. It is Excellent.
![wide_sinintegralsError](https://user-images.githubusercontent.com/6540723/161658591-d2fce073-30d0-4dab-907f-66a6e598c41a.png)

PR #1146, opened 2022-04-04
Expose Si(x) and Ci(x) to python layer

rmjarvis:
In response to Issue #1145 I added code to expose our C++ layer Si(x) implementation to python so I could write unit tests for it.  It agrees with scipy.special's implementation to better than 1.e-15 everywhere, so there doesn't seem to be any problem with the code.

While I was at it, I exposed sinc and gammap (as gammainc in python to match scipy) and wrote tests for them too.  I also went ahead and implemented the formula we have in Rowe et al for Ci(x) and exposed that as well.  In short, everything passes, so our C++ implementations all seem to be completely fine.

For visual comparison, here are Si and Ci from both GalSim and scipy along with their difference for 0 < x < 100.  (Script is in devel directory.)

cisi

### DIFF
--- a/devel/plot_si.py
+++ b/devel/plot_si.py
@@ -39,5 +39,4 @@ ax[1,1].set_xlabel('x')
 ax[1,1].set_ylabel('Ci_galsim(x) - Ci_scipy(x)')
 ax[1,1].set_ylim(-1.e-15, 1.e-15)
 
-fig.savefig('cisi.pdf')
 fig.savefig('cisi.png')


tests/test_bessel.py
```diff
@@ -374,6 +374,133 @@ def test_jv_root():
...
+    print('x = ',x_list)
+    print('vals1 = ',vals1)
+
+    vals2 = [ scipy.special.gammainc(a,x) for a,x in zip(a_list,x_list) ]
+    print('vals2 = ',vals2)
+    np.testing.assert_allclose(
+        vals1, vals2, rtol=1.e-10, err_msg="bessel.gammainc disagrees with scipy.special.gammainc")
+
+    # These values are what scipy returns.
+    vals2 = [   0.05336162361853338,
```
rmandelb, line 392: I do not fully understand the logic in testing twice against scipy, once calling the relevant routine and once using stored values.  It's not *wrong* but it seems redundant.  (Comment applies to some other tests.)  Is there some reason that I am missing?

rmandelb (approved):
Thank you!  Looks good to me.  I had one question but it's not something that should block merging this.

tests/test_bessel.py
rmjarvis: It is redundant.  It used to be useful when scipy was optional (since it used to be a pain to install), but we now require it for tests.  I still find it slightly useful to scan over the values to make sure we're hitting a good dynamic range of possible outputs as well as inputs, so I haven't gotten rid of them.

PR #1147, opened 2022-04-14
Persist pupil samples uv in photon array

jmeyers314:
Abuses the `PhotonArray.dxdz` and `.dydz` attributes to hold onto pupil samples generated by `PhaseScreenPsf._shoot`.

I realized along the way that we may wish/need to propagate these values forwards through photon array convolutions, so the rule is if exactly 1 convolutant has `.dxdz` or `.dydz`, then the result will just get the same values.  If more than one convolutant has these attributes, then an exception is raised.

I went ahead and also implemented similar logic for wavelengths, except in this case, GalSim already had a case where two photon arrays with populated wavelength arrays would be convolved; in `ChromaticObject.applyTo`.  Here, it's actually innocuous, since the two photon arrays will have the _same_ wavelength arrays by construction.  So to allow this code to continue as is, I also decided to permit convolution of two photon arrays both with populated wavelength arrays as long as those arrays are identical (via the python `is` operator).

### DIFF
--- a/galsim/chromatic.py
+++ b/galsim/chromatic.py
@@ -586,7 +586,7 @@ class ChromaticObject:
         if not photon_array.hasAllocatedWavelengths():
             raise GalSimError("Using ChromaticObject as a PhotonOp requires wavelengths be set")
         p1 = PhotonArray(len(photon_array))
-        p1.wavelength = photon_array.wavelength
+        p1._wave = photon_array._wave
         obj = local_wcs.toImage(self) if local_wcs is not None else self
         rng = BaseDeviate(rng)
         obj._shoot(p1, rng)
--- a/galsim/phase_psf.py
+++ b/galsim/phase_psf.py
@@ -1566,8 +1566,9 @@ class PhaseScreenPSF(GSObject):
         ud.generate(pick)
         pick *= len(u)
         pick = pick.astype(int)
-        u = u[pick]
-        v = v[pick]
+        # abuse dxdz and dydz to persist u and v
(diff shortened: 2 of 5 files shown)


rmjarvis (changes requested):
Code looks good Josh.  I'm a little worried about abusing the existing names though to mean something different.  I know that's what we agreed on the call, but I think now that we should maybe add some code to allow these to be accessed as `u`, `v` when that's what they mean.  And maybe not `dxdz`, `dydz`.  Still use the same array behind the scene, which enforces that you can't have both meanings, but I think this would potentially be less confusing, and consequently less error prone.

jmeyers314:
I'm wondering now if it'd be better design to just add new `pupil_u`, `pupil_v` attributes to `PhotonArray` without reusing the `dxdz`, `dydz` storage.  Both cases can be allocated on-the-fly, and we can always reclaim the storage when we're done by explicitly setting these back to `None`.  This way we don't have to handle collisions in meaning.

jmeyers314:
(And in fact, we might not even need to expose `pupil_u`, `pupil_v` to the c++ layer.)

rmjarvis:
That also sounds fine to me.  If any C++ function down the line wants them, we can add that exposure then.

jmeyers314:
Added `pupil_u` and `pupil_v` as `PhotonArray` attrs.  This is ready again Mike.

rmjarvis:
LGTM.  I like this UI better.

rmjarvis (approved):

PR #1152, opened 2022-06-08
Improve checks in GSParams.checks

EiffL:
While working on https://github.com/GalSim-developers/JAX-GalSim/pull/7 @ismael-mendoza found that the following would trigger an exception, not correctly captured:
```python
import galsim
galsim.GSParams.check(None, default="anything not a GSParams")
```
This small PR adds a check that the default is indeed a GSParams instance. 

I would have added a test for this, but I couldn't find a file with dedicated gsparams tests, what would be a good place for that?

### DIFF
--- a/galsim/gsparams.py
+++ b/galsim/gsparams.py
@@ -181,7 +181,13 @@ class GSParams:
         (GSParams.default if no other default specified).
         """
         if gsparams is None:
-            gsparams = default if default is not None else GSParams.default
+            if default is not None:
+                if isinstance(default, GSParams):
+                    gsparams = default 
+                else:
+                    TypeError("Invalid default GSParams: %s"%default)
+            else:
+                gsparams = GSParams.default
         elif not isinstance(gsparams, GSParams):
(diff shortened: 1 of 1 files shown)


rmjarvis:
I'm not sure I see the necessity of this.  This isn't a method that users typically (maybe ever) call, so I'm not sure we need it to be super hardened to bad user input.  So I'm worried that this is merely a performance regression (it's called a lot in relatively low-level functions) without any actual improvement in user interface.

esheldon:
profile :)

EiffL:
Fair, it is true that places that use this function are using an existing `gsobjet.gsparams` for the default argument, which are themselves systematically checked at the creation of a new gsobject. 
Feel free to close this PR if you prefer saving this `isinstance` check. But still interested if you have a pointer to a where GSParams is tested in the test suite :-)

rmjarvis:
I think I'm going to close this as unnecessary.  But please do keep raising issues as you come across things you think might be wrong or non-ideal in the course of your Jax efforts.

PR #1154, opened 2022-06-11
Add option to have InterpolatedImage de-pixelize the image

rmjarvis:
One of the persistent issues with using PSF models such as those from PSFEx or Piff is that they generally model the "effective PSF", including the pixel convolution, rather than the underlying true PSF.  This means that any objects that use this PSF object have to be rendered with `method='no_pixel'`, rather than `'auto'`, or especially `'phot'`.  Using photon shooting would implicitly convolve a second pixel, so lead to significant biases in the resulting images.

This PR adds the capability of fitting for a "depixelized" image to input into `InterpolatedImage`, which would represent the underlying profile, without the pixel convolution.  The expected user API for this is simply to add the parameter `depixelize=True` to the `InterpolatedImage` constructor.  

Then, to use a Piff model, for instance, with photon shooting, one can draw it normally, and then model it as `InterpolatedImage(piff_image, depixelize=True)`.  Then convolve this by the galaxy profile and you can render the combination using `method='phot'`.

It's a big matrix equation, so it scales badly for large images, but for the kinds of images we usually have for PSFs, it takes around 0.1-0.2 seconds, which doesn't seem to bad for many use cases.  Especially if you're in a testing regime where you can treat the PSF as constant for a reasonable number of galaxies.

### DIFF
--- a/galsim/image.py
+++ b/galsim/image.py
@@ -1875,15 +1875,28 @@ class Image:
     def depixelize(self, x_interpolant):
         """Return a depixelized version of the image.
 
-        Specifically, this function creates an image that could be used with InterpolatedImage
+        Specifically, this function creates an image that could be used with `InterpolatedImage`
         with the given x_interpolant, which when drawn with method=auto would produce the
         current image.
 
             >>> alt_image = image.depixelize(x_interpolant)
             >>> ii = galsim.InterpolatedImage(alt_image, x_interpolant=x_interpolant)
-            >>> image2 = ii.drawImage(image.copy())
+            >>> image2 = ii.drawImage(image.copy(), method='auto')
 
         image2 will end up approximately equalt to the original image.
+
+        .. warning::
+
+            This function is fairly expensive, both in memory and CPU time, so it should
+            only be called on fairly small images (~100x100 or smaller typically).
+            The memory requirement scales as Npix^2, and the execution time scales as Npix^3.
+
+        Parameters:
+            x_interpolant:  The `Interpolant` to use in the `InterpolatedImage` to describe
+                            how the profile should be interpolated between the pixel centers.
+
+        Returns:
+            an `Image` representing the underlying profile without the pixel convolution.
         """
         ny, nx = self.array.shape
         npix = nx * ny
--- a/galsim/image.py
+++ b/galsim/image.py
@@ -1902,12 +1902,12 @@ class Image:
         npix = nx * ny
 
         # Each kernel is the integral of the interpolant over 1 pixel.
-        kernels = x_interpolant.unit_integrals(max_len=max(nx,ny))
+        unit_integrals = x_interpolant.unit_integrals(max_len=max(nx,ny))
 
         # The rest of the implementation is done in C++.  cf. src/Image.cpp
         im2 = self.copy()
-        _kernels = kernels.__array_interface__['data'][0]
-        _galsim.depixelizeImage(im2._image, _kernels, kernels.size)
+        _unit_integrals = unit_integrals.__array_interface__['data'][0]
+        _galsim.depixelizeImage(im2._image, _unit_integrals, unit_integrals.size)
         return im2
 
     def __eq__(self, other):
--- a/galsim/image.py
+++ b/galsim/image.py
@@ -1883,7 +1883,7 @@ class Image:
             >>> ii = galsim.InterpolatedImage(alt_image, x_interpolant=x_interpolant)
             >>> image2 = ii.drawImage(image.copy(), method='auto')
 
-        image2 will end up approximately equalt to the original image.
+        image2 will end up approximately equal to the original image.
 
         .. warning::
 
--- a/galsim/image.py
+++ b/galsim/image.py
@@ -1886,17 +1886,8 @@ class Image:
         image2 will end up approximately equalt to the original image.
         """
 
-        nx, ny = self.array.shape
-
-        # Get a list of (h,k) indices of each pixel
-        allh,allk = np.meshgrid(range(nx), range(ny), indexing='ij')
-        allh = allh.ravel()
-        allk = allk.ravel()
-        npix = len(allh)
-
-        # Make a reverse lookup dict saying which (h,k) corresponds to linear index.
-        # I.e. rev[(h,k)] = index of that position in raveled array.
-        rev = dict(zip(zip(allh,allk), range(npix)))
+        ny, nx = self.array.shape
+        npix = nx * ny
 
         # kernel is the integral of the interpolant over 1 pixel.
         kernel = x_interpolant.unit_integrals(max_len=max(nx,ny))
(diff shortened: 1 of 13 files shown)


beckermr (commented):
I think I follow the method from reading the code. You basically compute an equation that is the integral of the interpolant over an image. It's linear so you solve it using a direct solver. 

The tests look complete. I cannot comment on the other random bug fixes. 

I do have one performance suggestion. If I understand the method correctly, for a fixed image size and interpolant type, the matrix A that you run through the LLT decomposition is independent of the underlying image. I am betting most of the compute time is this decomposition and not the final solve step (called back substitution IIRC). I am guessing most folks dealing with psf models will draw with the same size and interpolant always (or would be willing to anyways). Thus if we cache the matrix decomposition, we should be able to have a fair more efficiency in the code. I also could be misunderstanding things. 

Final note is that I did not check the unit integrals of the interpolants in detail. However the tests look like they cover any possible mistakes in the analytic ones.

rmjarvis:
Yes, that's a very good point about the caching possibility.  I'll have to think about how best to implement the cache, but that definitely seems worth doing.

rmjarvis:
Hi Matt, I think I have the cache working.  Thanks for the idea!

Some timings:

32x32: 
* First time: 0.013 s
* Second time: 0.0007 s
48x48: 
* First time: 0.13 s
* Second time: 0.0019 s
51x51:
* First time: 0.18 s
* Second time: 0.0034 s
64x64:
* First time: 0.70 s
* Second time: 0.0045 s
80x80:
* First time: 2.5 s
* Second time: 0.010 s
96x96:
* First time: 7.3 s
* Second time: 0.021 s
128x128:
* First time:  42 s
* Second time: 0.051 s

So this is a huge improvement.  Even 128x128 is pretty acceptable after the initial pass.  The first pass time looks close to the expected scaling of O(Npix^3), and subsequent ones are only O(Npix^2).

beckermr (approved):
Awesome! LGTM!

arunkannawadi:
This seems like a useful feature to have. Thanks Mike and Matt!

rmjarvis:
Thanks Matt.  I'll let this sit another day or two before merging in case anyone else wants to have a look.

galsim/image.py
```diff
@@ -1864,13 +1872,58 @@ def rot_180(self):
...
+
+        Specifically, this function creates an image that could be used with `InterpolatedImage`
+        with the given x_interpolant, which when drawn with method=auto would produce the
+        current image.
+
+            >>> alt_image = image.depixelize(x_interpolant)
+            >>> ii = galsim.InterpolatedImage(alt_image, x_interpolant=x_interpolant)
+            >>> image2 = ii.drawImage(image.copy(), method='auto')
+
+        image2 will end up approximately equalt to the original image.
```
rmandelb, line 1886:
```suggestion
        image2 will end up approximately equal to the original image.
```

tests/test_interpolatedimage.py
```diff
@@ -1647,6 +1713,129 @@ def test_quintic_glagn():
 
         gsobj.drawImage(method='phot', image=image, add_to_image=True)
 
+def test_depixelize():
+    # True, non-II profile.  Something not too symmetric.
```
rmandelb, line 1717: This is not too symmetric, but it is a quite simple profile.  In practice, people will likely want to use this for PSF models, and I wonder if it would be valuable to have a test using say Kolmogorov convolved with Airy or OpticalPSF?  OpticalPSF, especially, has some features that I could imagine pose challenges for the deconvolution.  I realize this could take longer so it could be a specialized test that is not run routinely but is run before new releases, or something like that.  (like the `do_slow_tests` option in `test_optics.py`)

galsim/interpolatedimage.py
```diff
@@ -54,13 +54,20 @@ class InterpolatedImage(GSObject):
...
 
     You can also use images that were drawn with one of the pixel-integrating methods ('auto',
-    'fft', or 'real_space'); however, the resulting profile will not correspond to the one
-    that was used to call `GSObject.drawImage`.  The integration over the pixel is equivalent to
-    convolving the original profile by a `Pixel` and then drawing with ``method='no_pixel'``.  So
-    if you use such an image with InterpolatedImage, the resulting profile will include the `Pixel`
-    convolution already.  As such, if you use it as a PSF for example, then the final objects
-    convolved by this PSF will already include the pixel convolution, so you should draw them
-    using ``method='no_pixel'``.
+    'phot', 'fft', or 'real_space'), or real images where nature has integrated over the pixel for
```
rmandelb, line 57: I feel like one key point that the updated dox omits is that the need to do `no_pixel` prevents the use of photon shooting, so it can have serious impacts on computational cost -- and the existence of the `depixelize` option fixes that issue (at least, assuming the depixelize process is not too costly).  This point is implicitly there but I think it's valuable to mention explicitly.

rmandelb (approved):
I have a few minor comments but they are not really show stoppers in my view, so I am going to hit "Approve" so you can choose what you want to incorporate but not be blocked from merging.  This is a really useful feature -- thanks Mike!

One thought I had, which I didn't put in a comment on the diff because I wasn't sure where it should go: When I read your description of depixelize I thought "this must be some sort of regularized deconvolution" and assumed that in certain regimes this might not work well.  For example: in undersampled data I imagine it could cause serious aliasing; but even if Nyquist sampled it seems there are some assumptions?  Would it be helpful for users to provide some basic guidance about expected accuracy and/or regimes where the code is not meant to be used (like undersampled data, presumably)? If we cannot make a general statement about expected accuracy, perhaps just caution that if they have some really gnarly image they should do some accuracy tests themselves?

tests/test_interpolatedimage.py
rmjarvis: I switched to something more complicated:
```
true_prof = galsim.Convolve(
                    galsim.Kolmogorov(fwhm=1.1, flux=10).shear(g1=0.1, g2=0.2),
                    galsim.OpticalPSF(lam_over_diam=0.6, obscuration=0.4, nstruts=4,
                                      defocus=0.15, astig1=0.13, astig2=-0.14,
                                      coma1=-0.06, trefoil1=-0.08))
```
Screen Shot 2022-06-16 at 5 01 50 PM
And the existing tests still passed.  TBH, I suspect noise is a harder thing for it to deal with, since it doesn't represent a true profile convolved by a pixel.  But I haven't investigated how low noise is required for this to end up sensible.

rmjarvis:
@jmeyers314 I added "We recommend using a Lanczos interpolant with this option for best results.  (Higher order tends to work better here.)" to the discussion of the depixelize option.  I think the maxk issue I identified is independent of depixelize and is instead a problem intrinsic to how we calculate maxk for InterpolatedImages.  So instead of adding something there, I added documentation for the `_force_stepk` and `_force_maxk` parameters:
```
        _force_stepk:       Override the normal stepk calculation (using gsparams.folding_threshold)
                            and force stepk to the given value. [default: 0]
        _force_maxk:        Override the normal maxk calculation (using gsparams.maxk_threshold)
                            and force maxk to the given value.  This option in particular can help
                            reduce FFT artifacts in a manner than is currently unobtainable by
                            lowering maxk_threshold. [default: 0]
```

Issue #1153, opened 2022-06-10
Allow Bounds instances to be expanded directionally.


arunkannawadi:
Currently, `galsim.BoundsI` and `galsim.BoundsD` objects have an `expand` method that takes in a float and expands the bounding box by that factor along both horizontal and vertical directions. There should also be an option for it to take two floats to expand it in different directions. The API could be same as that for `withBorder` that expands the bounding box by specific number of pixels.

Happy to submit a PR if this is a desired feature.

rmjarvis:
Sure.  If someone has a use case where this is desired, feel free to add that feature.  

FWIW, the usual use case for expand is to get a bounding box for finding objects that might overlap an image, including ones whose nominal position is not in the image bounds.  So that's a symmetric application.  I can't think of an obvious application that needs asymmetric expansion, but if you want to add it, please go ahead.

arunkannawadi:
My use case for this was to construct a bounding box that is at least 16x16px and encompasses an object's segment completely. I want the dimensions of the box to be a power of 2 (can be different). So if the box was too small, using `.expand` made more sense than `.withBorder`, and I wanted to grow the box only in one direction if that was sufficient to encompass the object.

PR #1155, opened 2022-06-13
Expand bounds directionally

arunkannawadi:
This is a short PR introducing the feature to expand a bounding box by different factors in different directions. Addresses #1153 .


rmjarvis:
Ah, good point.  If you can add an item in the CHANGELOG, that would be ideal.  Otherwise, I usually try to catch whatever is missing when we cut a release.  But better to remember to add it now.  :)

Issue #1156, opened 2022-06-15
ChromaticSum.drawImage sometimes shoots too many photons


jmeyers314:
If `n_photons` is explicitly set in `ChromaticSum.drawImage` when drawing with `method='phot'`, then that many photons will be used for _each_ summand, instead of that many photons being used amongst all summands.  This can also happen if `ChromaticConvolution` includes a `ChromaticSum` as a convolutant even when `n_photons` is not explicitly specified.  I think we may need some logic similar to [`galsim.Sum.shoot`](https://github.com/GalSim-developers/GalSim/blob/releases/2.3/galsim/sum.py#L317-L337) inside `ChromaticSum` to get this right.

Also, once photons have been drawn from each summand profile, I think it's always okay to pass them through subsequent photons ops together.  I.e., there's no need to repeatedly call the same photon op for each summand, and in fact avoiding this may lead to some small speedup.

Code to reproduce:
```python

import numpy as np
import galsim

sed1 = galsim.SED("CWW_E_ext.sed", wave_type='A', flux_type='flambda')
sed2 = galsim.SED("CWW_Sbc_ext.sed", wave_type='A', flux_type='flambda')
bandpass = galsim.Bandpass("LSST_r.dat", wave_type="nm")

obj1 = (galsim.Gaussian(fwhm=0.6) * sed1).withFlux(990.0, bandpass)
obj2 = (galsim.Gaussian(fwhm=0.3) * sed2).withFlux(10.0, bandpass)
obj = obj1 + obj2

class Counter(galsim.PhotonOp):
    def __init__(self):
        self.nphot = []
        self.meanflux = []

    def applyTo(self, photon_array, local_wcs=None, rng=None):
        self.nphot.append(len(photon_array))
        self.meanflux.append(np.mean(photon_array.flux))

# Looks okay when n_photons unspecified
counter = Counter()
img = obj.drawImage(
    bandpass, nx=24, ny=24, scale=0.2, method='phot',
    photon_ops=[counter]
)
print(f"{counter.nphot = }")
print(f"{counter.meanflux = }")
# RNG-dependent, but I got:
# counter.nphot = [962, 7]
# counter.meanflux = [1.0, 0.9999999999999999]

# When n_photons is explicit, shoots too many photons
counter = Counter()
img = obj.drawImage(
    bandpass, nx=24, ny=24, scale=0.2, method='phot',
    photon_ops=[counter], n_photons=101, poisson_flux=False
)
[...]

PR #1157, opened 2022-06-15
#1156 Fix error in ChromaticSum photon shooting when n_photons is given explicitly

rmjarvis:
See #1156 for description of the problem.

This PR fixes it and adds tests to confirm.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -45,3 +45,4 @@ Performance Improvements
 Bug Fixes
 ---------
 
+- Fix error in ChromaticSum photon shooting when n_photons is explicitly given. (#1156)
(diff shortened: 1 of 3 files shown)


jmeyers314:
Looks good to me.

If there's an easy way to aggregate the photons from the different summands before applying the extra photon_ops, I think that'd be a win (batoid traces 40,000 photons in 1 batch faster than 20000+20000 photons split into two batches).  But I'd understand if you want to leave that to a future issue rather than as part of this bugfix.

rmjarvis:
They don't work on python 3.7, which we still support.  And it's in the tests, so speed isn't relevant.

rmjarvis:
> If there's an easy way to aggregate the photons from the different summands before applying the extra photon_ops, I think that'd be a win.

I think a lot of the chromatic photon shooting could be usefully refactored in a way that would enable this.  But right now I think it would be difficult to implement that.

jmeyers314 (approved):

Issue #1148, opened 2022-04-25
PhotonDCR config bugs


jmeyers314:
Placing here so I don't forget:

Looks like `PhotonDCR._opt_params` is missing an entry for `obj_coord`.

Also, there's a typo: `zenit_coord` -> `zenith_coord`.

Need to add some config usage tests.

rmjarvis:
One comment about this.  The `zenit_coord` is of course a typo-bug, which apparently isn't covered by unit tests.  But `obj_coord` is handled by the `PhotonDCRBuilder` here: https://github.com/GalSim-developers/GalSim/blob/releases/2.3/galsim/config/photon_ops.py#L220

rmjarvis:
Done via #1160

PR #1160, opened 2022-06-21
#1148 Fix PhotonDCR config layer bugs, along with a number of other edge cases it turned up.

rmjarvis:
@jmeyers314 noticed some clear typos in the PhotonDCR `opt_params` class variable, which made it obvious that we don't have any unit tests testing the various ways to make a PhotonDCR object in the config layer.  So this PR adds some.

Along the way, I discovered the following minor bugs and a few features I thought we should have, so I fixed them as well:

* There was no way to specify the dtype of the image made by config processing.  Usually float32 is fine, but we should be able to specify something else if we want.  So now, the image field takes an optional `dtype`, which can be e.g. `np.float64', 'int', 'float', 'numpy.uint32', etc.  Basically all the types that images can be using either np or numpy prefixes where appropriate.
* When testing that feature with int images, I discovered that (completely unrelated to config-based usage) sometimes int images would have values 1 less than what they should be, even when photon shooting, so everything should be integer values.  This turned out to be due to numpy's `astype` not rounding when casting to integers as I think I always assumed.  Some bits of the calculation go through floats, so they could be e.g. 29.99999999999974, which became 29, not 30.  So now GalSim.Image has a `_safe_cast` method that rounds if casting to an integer value before doing `astype`.
* Image type Single had a non-standard usage of world_pos, which I think used to have a rationale a long time ago, but the reason for it has been superseded by `sky_pos`.  This "feature" also wasn't documented anywhere other than in a unit test that confirms it works in that non-standard way.  So I considered it a bug, and I fixed it to work the normal way.
* I wanted to set a flux for an object with an SED, which wasn't (easily) possible in config.  So I added a feature that if you set both flux and sed, it will apply the SED and then call `withFlux(flux, bandpass)` to normalize to that flux value, which seems intuitive.
* I discovered that for image type=Single, sometimes the wcs in the final image would be a Jacobian at the local image pos, rather than the intended wcs.  So I fixed that.
* Finally, there were the typos Josh noticed for PhotonDCR, and also it assumed sky_pos always exists when using it for obj_coord, which isn't true, so now I check for it.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -43,6 +43,8 @@ Config Updates
 - Added option to specify a dtype other than np.float32 for images built by config. (#1148)
 - Added option to specify a dtype other than np.float32 for images built by config. (#1160)
 - Fixed inconsistent behavior of image.world_pos in image type=Single. (#1160)
+- Let a flux item for an object with an SED normalize the SED for the bandpass being
+  simulated. (#1160)
 - Changed the way the internal random number sequence works so that running multiple simulations
   with sequential random seed values doesn't end up with duplicated random values across the
   two (or more) simulations. (#1169)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -72,3 +72,5 @@ Bug Fixes
 - Fixed error in InterpolatedImage.withGSParams not correctly updating stepk and maxk
   if the updated parameters merited it. (#1154)
 - Fix error in ChromaticSum photon shooting when n_photons is explicitly given. (#1156)
+- Fixed some rounding errors that could happen when rendering integer-typed images
+  (e.g. ImageI) that could cause values to be off by 1.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -41,6 +41,8 @@ Config Updates
 - Added galaxy_sample input type with corresponding SampleGalaxy GSObject type. (#731)
 - Added COSMOSValue and SampleValue value types. (#954)
 - Added option to specify a dtype other than np.float32 for images built by config. (#1148)
+- Added option to specify a dtype other than np.float32 for images built by config. (#1160)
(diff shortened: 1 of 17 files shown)


CHANGELOG.rst
```diff
@@ -40,6 +40,13 @@ Config Updates
 - Added Correlated noise type as a generalization of the more specific COSMOS noise type. (#731)
 - Added galaxy_sample input type with corresponding SampleGalaxy GSObject type. (#731)
 - Added COSMOSValue and SampleValue value types. (#954)
+- Added option to specify a dtype other than np.float32 for images built by config. (#1148)
+- Fixed some errors in PhotonDCR usage in the config layer. (#1148)
+- Added option to specify a dtype other than np.float32 for images built by config. (#1160)
```
jmeyers314, line 45: Duplicates line 43 (except for issue #)?

tests/test_config_image.py
```diff
@@ -1359,6 +1404,12 @@ def test_tiled():
         galsim.config.BuildImage(config)
     config['image']['stamp_ysize'] = ysize
     config['image']['yborder'] = xborder
+    with assert_raises(galsim.GalSimConfigError):
+        galsim.config.BuildImage(config)
+    config['image']['yborder'] = -yborder
+
+    # numpy types need np. or numpy. prefix.
+    config['image']['dtype'] = 'float64'
```
jmeyers314, line 1412: Plain `float64` seems like an intuitive use of dtype here too (`str(np.array(3).dtype)` returns just the 'int64', e.g.).  Can we enable this?

jmeyers314 (approved):
LGTM.  I'm especially happy you can now specify a flux and not just a flux density for a chromatic object in the config.  Left a couple questions.

tests/test_config_image.py
rmjarvis: OK.  Done.

Issue #1150, opened 2022-05-28
C++ Interface


JiachuanXu:
Hi all, currently I'm developing a project which depends heavily on GalSim and is very time-consuming. I'm thinking about calling the GalSim functions in C++ to make the code faster. However, after searching for a while, I can not find any instructions on how to call those routines in C++. Does anyone have some notes on that? Thanks!

rmjarvis:
Sorry no, we don't really have much documentation about the C++ interface, since it's not our primary user interface.  For instance, we don't make any semantic versioning guarantees that the C++ function signatures will remain the same through minor version updates like we do for the python layer.  (Though we do guarantee they won't change in bugfix releases.)  

As such, we haven't put any effort into making nice documentation for the C++-layer calls.  The only documentation in our main docs related to the C++ layer is [instructions on making the C++-linkable library](http://galsim-developers.github.io/GalSim/_build/html/install_pip.html#installing-the-c-shared-library), which you can do with a special setup.py call, and which is done for you in the conda release.

Beyond that, the best I can do is point you to the [header files](https://github.com/GalSim-developers/GalSim/tree/releases/2.3/include/galsim), which usually have doxygen-style comments giving basic documentation.  But we don't try to compile those into pretty web pages, and I wouldn't be too surprised if there are errors in them.  But it might be enough to get you started.  Especially if you just have a few functions in mind to use.

jamesp-epcc:
I've used this a bit recently. I have a very simple pure C++ test case here that I've been using to test my GPU port of GalSim:

https://github.com/jamesp-epcc/GalSimGPUTest

Just ignore the GPU related stuff and it might be useful.

rmjarvis:
If you would like to checkout branch `cpp_docs`, I added a first pass at documenting the C++ layer code.  To try it out, do the following:
```
git clone https://github.com/GalSim-developers/GalSim.git
cd GalSim/
git checkout cpp_docs
pip install .
./update_docs
open docs/_build/html/cpp.html
```
See also the PR for this update at #1161.

PR #1161, opened 2022-06-22
Add documentation for the public C++ API in the main Sphinx docs.

rmjarvis:
A long time ago, we used to use Doxygen for our docs, and had it make both C++ and python documentation.  It never looked very pretty, so we switched to Sphinx for the python.  But Sphinx doesn't really do C++, so we just dropped the C++ docs at the time.  

However, in response to issue #1150, I looked into it and discovered that there is a project called Breathe, which can take doxygen output and put it in Sphinx docs.  So I tried it out last night and it seems to work ok.  It's not super pretty, but it's basically functional.  

This PR adds a first pass at this, adding a C++ Layer section to the left side bar in the docs. I think I added all the relevant C++ functions and classes that we expect C++ GalSim users to potentially care about, skipping a lot of the implementation details.  I also didn't try to improve out doxygen markup at all, so whatever was there is all we have right now.  I also added some quick text giving an overview of how people can get the C++ library to link to and such.

I think this is probably fine for now as a straw man for people to improve later if they are so motivated.


CHANGELOG.rst
```diff
@@ -24,6 +24,11 @@ Dependency Changes
 API Changes
 -----------
 
+- Removed CppEllipse in C++ layer, which had been deprecated since the 1.x series, but we forgot
+  to actually get rid of. (#1129)
+- Removed AstronomicalConstants.h in C++ layer, which we never used. (#1129)
+- Removed AttributeDict, which we had stting in utilities, but which we have never used.
```
rmjarvis, line 30:
```suggestion
- Removed AttributeDict, which we had sitting in utilities, but which we have never used.
```

docs/cpp.rst
```diff
@@ -0,0 +1,95 @@
...
+While GalSim is primarily a Python package, much of the implementation is done in C++
+for improved speed for the hard-core numerical calculations.
+
+If you would like to use the C++-layer functions in your own C++ project, you can do
+so with the caveat that these are officially implementation details, so we don't
+strictly enforce semantic versioning for the C++-layer API.  That is, function signatures
+may change on minor version updates (e.g. 2.3.x to 2.4.0).  We don't often make huge
+changes to the C++ API, so most of the time you will be fine upgrading, but you should
+be prepared that you may need to update your code when upgrading GalSim.  (We do
+guarantee that we won't change the C++ API for bugfix updates, e.g. 2.4.1 to 2.4.2.)
```
beckermr, line 13: API or ABI? 

In other words, can people who build their code against 2.4.1 link against 2.4.2 or do they need to recompile? If they need to recompile, then I think it is important to mention this.

rmjarvis: Probably mostly ABI, but I don't know really how to guarantee that.  I.e. I don't know what tests to do to ensure that we keep things ABI compatible.  There is nothing in our test suite that checks this.

rmjarvis: Ah, OK.  That I can do!  :)

beckermr (approved):

rmjarvis:
Thanks Matt!

Issue #1164, opened 2022-06-30
drawReal of specific InterpolatedImage leads to seg fault


jmeyers314:
While running Piff (via Rubin science pipelines) on some DECam data, @leeskelvin encountered the following seg fault.  The relevant line in Piff is https://github.com/rmjarvis/Piff/blob/releases/1.2/piff/pixelgrid.py#L249, but I've distilled the error to what's posted below:

[prof.pkl.gz](https://github.com/GalSim-developers/GalSim/files/9021638/prof.pkl.gz)

Code to reproduce:

```
import pickle
import galsim

with open("prof.pkl", 'rb') as f:
    prof = pickle.load(f)

image = galsim.Image(
    galsim.BoundsI(
        xmin=-12,
        xmax=12,
        ymin=-12,
        ymax=12
    ),
    dtype=float,
    scale=1
)

prof.drawReal(image)
```

Seg-faults in both linux and macos.

rmjarvis:
Yep.  I reproduce this on my laptop.  Which means I can fix it.  I'm on it.

leeskelvin:
Thank you both for looking at this and [diagnosing it so quickly](https://github.com/GalSim-developers/GalSim/pull/1165)!

rmjarvis:
No problem.  And thanks much for the bug report!

PR #1165, opened 2022-06-30
Fix seg fault possibility if InterpolatedImage is fully off the target image

rmjarvis:
Fixes seg fault bug reported in #1164.

Let me know if you need a bugfix release for this on the 2.3 series, or if this can just be included in 2.4.


tests/test_interpolatedimage.py
```diff
@@ -1839,6 +1839,39 @@ def test_depixelize():
...
+            ymax=12
+        ),
+        dtype=float,
+        scale=1
+    )
+
+    image.fill(3)
+    prof.drawReal(image)
+
+    # The prolem was that the object is shifted fully off the target image and that was leading
```
leeskelvin, line 1867: small typo (problem)

rmjarvis, line 1867:
```suggestion
    # The problem was that the object is shifted fully off the target image and that was leading
```

rmjarvis: Thanks.  Fixed.

rmjarvis: Thanks.  Fixed.

jmeyers314 (approved):
LGTM.

jmeyers314:
BTW - I do think this would be good for a backport to 2.3.

rmjarvis:
Released v2.3.5 with this fix.  (Wheels are turning getting it onto pip and conda.)

PR #1167, opened 2022-07-07
BUG return variance in ADU units for CCDNoise

### LINKED ISSUE #1166: Should the `addNoise` method of CCDNoise return `sky / gain` not just `sky`?
I think this line (https://github.com/GalSim-developers/GalSim/blob/releases/2.3/galsim/config/noise.py#L532) doesn't match the expected `sky/gain` units that are returned by say this line (https://github.com/GalSim-developers/GalSim/blob/releases/2.3/galsim/config/noise.py#L544).

The units in the return values of those two methods match for the other noise types AFAICT.

beckermr:
This fixes a bug where the CCDNoise class does not return the variance in ADU units, matching the other methods that deal with variance. 

closes #1166

### DIFF
--- a/galsim/config/noise.py
+++ b/galsim/config/noise.py
@@ -458,6 +458,11 @@ class CCDNoiseBuilder(NoiseBuilder):
         sky = GetSky(base['image'], base, logger, full=True)
         extra_sky = GetSky(config, base, logger, full=True)
         total_sky = sky + extra_sky
+        if isinstance(total_sky, Image):
+            var_adu = np.mean(total_sky.array) / gain + read_noise_var / gain**2
+        else:
+            var_adu = total_sky / gain + read_noise_var / gain**2
+
 
         # If we already have some variance in the image (from whitening), then we try to subtract
         # it from the read noise if possible.  If not, we subtract the rest off of the sky level.
@@ -465,7 +470,7 @@ class CCDNoiseBuilder(NoiseBuilder):
         # Poisson, but it's the best we can do.
         if current_var:
             logger.debug('image %d, obj %d: Target variance is %f, current variance is %f',
-                        base.get('image_num',0),base.get('obj_num',0), var, current_var)
+                        base.get('image_num',0),base.get('obj_num',0), var_adu, current_var)
             read_noise_var_adu = read_noise_var / gain**2
             if isinstance(total_sky, Image):
                 test = np.any(total_sky.array/gain + read_noise_var_adu < current_var)
@@ -524,11 +529,6 @@ class CCDNoiseBuilder(NoiseBuilder):
 
         logger.debug('image %d, obj %d: Added CCD noise with gain = %f, read_noise = %f',
                      base.get('image_num',0),base.get('obj_num',0), gain, read_noise)
-
-        if isinstance(total_sky, Image):
-            var_adu = np.mean(total_sky.array) / gain + read_noise_var / gain**2
-        else:
-            var_adu = total_sky / gain + read_noise_var / gain**2
 
         return var_adu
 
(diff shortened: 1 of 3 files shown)


galsim/config/noise.py
```diff
@@ -457,19 +457,20 @@ def addNoise(self, config, base, im, rng, current_var, draw_method, logger):
...
         sky = GetSky(base['image'], base, logger, full=True)
         extra_sky = GetSky(config, base, logger, full=True)
-        total_sky = sky + extra_sky # for the return value
+        total_sky = sky + extra_sky
         if isinstance(total_sky, Image):
-            var = np.mean(total_sky.array) + read_noise_var
+            var_adu = np.mean(total_sky.array) / gain + read_noise_var / gain**2
         else:
-            var = total_sky + read_noise_var
+            var_adu = total_sky / gain + read_noise_var / gain**2
```
rmjarvis, line 464: This is now the same as `target_var` below, so you can remove that name and use `var_adu` there.

tests/test_config_noise.py
```diff
@@ -776,12 +776,13 @@ def test_whiten():
...
+    gain = 3.7
+    config['image']['noise']['gain'] = gain
     galsim.config.RemoveCurrent(config)
     im5a = im1a.copy()
-    rn = math.sqrt(25-cv1a * 3.7**2)
+    rn = math.sqrt((25/gain-cv1a) * gain**2)
     rng.reset(rng1.duplicate())
-    im5a.addNoise(galsim.CCDNoise(sky_level=25, read_noise=rn, gain=3.7, rng=rng))
+    im5a.addNoise(galsim.CCDNoise(sky_level=25, read_noise=rn, gain=gain, rng=rng))
     im5b, cv5b = galsim.config.BuildStamp(config)
```
rmjarvis, line 786: I suggest adding a test along the lines of
```
    # Check that this is (moderately) close to the actual noise in the image.
    np.testing.assert_allclose(cv5b, np.var(im5b.array - im1a.array), rtol=0.1)
```
This checks that the noise in the image actually is close to the return value.
Then I think the right expected value is now:
```
target_var = 25/3.7 + 5**2/3.7**2
np.testing.assert_almost_equal(cv5b, target_var, decimal=5)
```

tests/test_config_noise.py
```diff
@@ -776,12 +776,13 @@ def test_whiten():
     np.testing.assert_almost_equal(im4b.array, im4a.array, decimal=5)
 
     # Repeat with gain != 1
-    config['image']['noise']['gain'] = 3.7
+    gain = 3.7
+    config['image']['noise']['gain'] = gain
     galsim.config.RemoveCurrent(config)
     im5a = im1a.copy()
-    rn = math.sqrt(25-cv1a * 3.7**2)
+    rn = math.sqrt((25/gain-cv1a) * gain**2)
```
rmjarvis, line 783: I don't think this edit is right.  It's still just 25, which is the nominal read_noise variance in e-.  You want to subtract off the current variance in e-, which is cv1a * gain**2.  Then sqrt to get read_noise sigma, still in e-.

rmjarvis (changes requested):
I think the test that is failing is wrong.

tests/test_config_noise.py
rmjarvis: Actually, adding that check of the cv return value for all the noise types is probably a good idea.  I think the rest of them come out fine already, but it's worth checking the claim that this return value is the current noise variance in ADU as the doc string says.

rmjarvis: Or, actually, I guess the doc string doesn't specify ADU.  Might be worth adding that clarification somewhere.

tests/test_config_noise.py
beckermr: Hmmmm. The config docs say read noise should be in ADU, not e-.

```
read_noise = float_value (default = 0.0) The CCD read noise in ADU.
```

I guess the class expects different and a conversion is done elsewhere?

tests/test_config_noise.py
beckermr: OK. I added all of the tests here. The return value was already tested each time. Now there is a test for the variance.

PR #1169, opened 2022-07-17
Update how the config layer rng sequence is done.

rmjarvis:
There is currently a "gotcha" in the way the config layer does random number sequences.  Internally it takes the initial random_seed value and uses sequential values after this number for each galaxy or each image (according to some other config options).  This lets simulations that use multiprocessing across objects to have deterministic results if the galaxies have (pseudo-)random values for each galaxy.

However, multiple users (@NiallMac @erykoff @danielgruen) have been caught by this behavior when they ran a single config dict with sequential random seeds, expecting totally different simulations.  If you only increment the seed by 1, then all but one of the galaxies will be the same as the previous simulation.  This was documented, but highly non-intuitive.

This PR changes the way we treat the given random_seed value.  Now the first thing that the config layer does is scramble the seed to a new value by essentially doing `random_seed = galsim.BaseDeviate(random_seed).raw()` before doing the sequential pattern we used to do.  Then if the user does sequential seed values, the internal seed sequences will start at very different values, so the two simulations will be completely different.

This change of course broke a lot of the tests.  Lots of them had to switch to follow the new seed scrambling technique, as did the py versions of the demos.  In addition, some were testing very particular results, which had to change to match the particularities of the new prescription.


docs/config_image.rst
```diff
@@ -13,9 +13,15 @@ Some attributes that are allowed for all image types are:
...
 
     * If ``random_seed`` is a list, then multiple random number generators will be available for each object according to the multiple seed specifications.  This is normally used to have one random number repeat with some cadence (e.g. repeat for each image in an exposure to make sure you generate the same PSFs for multiple CCDs in an exposure).  Whenever you want to use an rng other than the first one, add ``rng_num`` to the field and set it to the number of the rng you want to use in this list.
+    .. note::
+
+        Within this ``random_seed`` list, any items that are either an int or str will be
+        evaluated once and converted into a sequence in the normal way.  So if you want an item
+        to be some complex Eval operation, you should make it a dict with an explicit
+        ``type: Eval``, rather than use the ``$`` shorthand notation.
```
beckermr, line 24: There must be a good reason for this gotcha but I don't know it. Is it worth writing down why things must be done this way in the docs here?

examples/demo13.py
```diff
@@ -219,7 +220,10 @@ def main(argv):
         # There are simpler ways to do this in a python script (e.g. probably only need 2
         # rngs, not 3), but this way of setting it up matches the way the config file initializes
         # the random number generators.
-        image_rng = galsim.UniformDeviate(seed + ifilter * nobj)
+        # Also, note that the second seed given in the config file, doesn't get the
+        # BaseDeviate(...).raw() treatment.  Only the first item, which parses as an int.
+        # When a random_seed config item is already a dict, GalSim leaves it as is.
+        image_rng = galsim.UniformDeviate(seed1 + ifilter * nobj)
```
beckermr, line 226: This comment seems related to my comment above.

examples/demo13.yaml
```diff
@@ -119,7 +119,15 @@ image:
         - 12345
 
         # Used for objects.  Repeats sequence for each filter
-        - '$12345 + obj_num % @image.nobjects + 10**6 * (1 + image_num // len(@image.filter.items))'
+        # Note: Don't use $ shorthand here, since that will implicitly be evaluated once and then
+        # treated the same way as an integer (i.e. making a regular sequence starting from that
+        # value).  Using an explicit dict with an Eval type means GalSim will leave it alone and
+        # evaluate it as is for each object.
```
beckermr, line 125: Same gotcha I don't follow.

galsim/config/stamp.py
```diff
@@ -925,7 +925,6 @@ def updateSkip(self, prof, image, method, offset, config, base, logger):
             if image is None:
                 prof = base['wcs'].toImage(prof, image_pos=base['image_pos'])
                 N = prof.getGoodImageSize(1.)
-                N += 2 + int(np.abs(offset.x) + np.abs(offset.y))
```
beckermr, line 928: This change seems unrelated? What is going on here?

galsim/config/util.py
```diff
@@ -379,19 +379,21 @@ def ParseRandomSeed(config, param_name, base, seed_offset):
...
-    # for the first item and go up sequentially from there for each object.
+    # Normally, random_seed parameter is just a number, which really means to use that seed
+    # start a sequential sequence of seeds for each object.
     # However, we allow for random_seed to be a gettable parameter, so for the
     # normal case, we just convert it into a Sequence.
-    if isinstance(config[param_name], int):
-         # The "first" is actually the seed value to use for anything at file or image scope
-         # using the obj_num of the first object in the file or image.  Seeds for objects
-         # will start at 1 more than this.
+    if isinstance(config[param_name], (int, str)):
```
beckermr, line 386: IDK why strings get done here. I think this is related to the gotcha above?

galsim/config/input_powerspectrum.py
```diff
@@ -70,6 +71,7 @@ def getKwargs(self, config, base, logger):
             rs = base['image']['random_seed']
             if not isinstance(rs, list): rs = [rs]
             first_seed += 31415  # An arbitrary offset to avoid the chance of unwanted correlations
+            first_seed = BaseDeviate(first_seed).raw()
```
beckermr, line 74: This hard-coded value seems dangerous. Do we need it now that we redraw things? Maybe rewdraw things twice?

beckermr (commented):
Thanks a bunch for this fix here @rmjarvis! This PR was a lot of work and is much appreciated. 

My only question is why the eval shorthand is not allowed for the random seed list. This seems like a deliberate choice to parse strings as ints for random seeds, and I don't follow why. 

It'd be nice to remove this gotcha if we can.

galsim/config/input_powerspectrum.py
beckermr: By rewdraw twice I mean, `BaseDeviate(BaseDeviate(first_seed).raw()).raw()` or something.

rmjarvis: I think now that we are subsequently applying raw() to get the final value, even +=1 is fine.  Before, I wanted it big enough that there wasn't some weird chance of the shear power spectrum realization coupling to the potentially random locations of galaxies in an image.  Probably was me being overly paranoid to even make a second rng here in the first place.  But now I think any offset value (1 or 31415 or anything but 0) is functionally equivalent.

galsim/config/stamp.py
rmjarvis: This was never necessary, and while not exactly wrong, it's different from what I do everywhere else to calculate image sizes.  It normally doesn't matter at all, since it's just deciding whether to skip an object, so one extra pixel border rarely matters, and if it (because of this extra pixel) decides not to skip here, it just does a few extra calculations that don't end up adding anything to the image (since the eventual stamp is slightly smaller). 

Anyway, it turned out that with the new random seeds, one of the demo scripts had a difference in the output truth files (not the images -- they were still identical), because the yaml one didn't skip an object, so it made it to the truth file, even though it was fully off the image, but the python one skipped it.

galsim/config/util.py
beckermr: I think the issue is that the code conflates two things:

1. the seed being an integer type 
2. the sequence of seeds is setup properly as a sequence type

So my guess is that we simply need to make the if statement more precise with something like:

```
if not (isinstance(config[param_name], dict) and "type" in config[param_name] and config[param_name]["type"] == "Sequence"):
```

beckermr: That might even fail for custom seed specifications where part of the sequence definition also needs to be parsed. We could set a special `__rng_is_setup` variable or something.

rmjarvis: It definitely doesn't have to be explicitly a Sequence.  E.g. your guard would not work right for demo13.  (Either the old or new version).

rmjarvis: FWIW, I think the only potential gotcha left in the version I ended up with in this PR is if someone wanted to make an rng sequence using an Eval and wanted to do it using the shorthand `$` notation (as we used to do in demo13.  That's now doubly documented -- both in the demo script where it would have happened and in the Sphinx docs.  So there's pretty fair warning about it.  
The workaround to avoid it is also pretty simple.  Just don't use the `$` shorthand.

rmjarvis: Currently:
* (int, str) indicate that this is just a single value, which gets converted into a Sequence.
* (dict) indicates that either the above already happened or the user set it up themselves to do whatever they want.

beckermr: Got it. 

I made a PR into this one with another design: https://github.com/GalSim-developers/GalSim/pull/1173

What I was going for was for the code to attempt to parse the seed as an int. If that passes, then use that int to set up the seed sequence. Otherwise, don't do anything. IDK if this might have other problems though.

beckermr: It only does this once using a variable stored in the config to control the state.

beckermr: I think the current design excludes a case where someone uses a short-hand eval string to indicate a value they'd actually like to be a constant.

rmjarvis: No.  The opposite.  Right now, a short-hand eval string (i.e. with `$`) is taken to be a constant, which is parsed as a single int and then made into a Sequence.
The current design excludes someone who wants to use some complicated dict to calculate a single int once, and have that value be treated as a constant and turned into a Sequence.  
That seems unlikely, but if someone really wants to do this, they can, by doing this in an eval_variable rather than directly in the image.random_seed field.  Call this `my_first_seed` or whatever.  Then they can use `image.random_seed = $my_first_seed` to evaluate it that one time and then sequence from there.

beckermr (approved):
I don't follow but don't have the energy right now to understand this more. Lgtm!

rmjarvis:
@beckermr I know you already approved, but it felt like you still had misgivings.  I think this doc edit ([d969b95](https://github.com/GalSim-developers/GalSim/pull/1169/commits/d969b95369715b9110bf1114f63ff0a5c4b76192)) might help clarify things.  I made it much more explicit what all the different options are for the random_seed specification and how they all work.  LMK if this assuages any of your remaining concerns.

beckermr:
That doc edit helps a lot. My qualm is hijacking the shorthand eval syntax for seeding sequences, but really the int syntax was already hijacked. Thus I don't think there is really anything to do about this. LGTM and thank you!

PR #1172, opened 2022-07-19
Add ability to project shear to and from world coordinates

rmjarvis:
This is a quick feature addition requested by @beckermr.  It's actually code that I have copied around a few places when I needed to do this, and he pointed out that it is probably widely useful enough to be put into GalSim.

Basically, `wcs.toWorld(shear)` will now return the equivalent shear in world coordinates when the input is the measured shear in image coordinates (e.g. from HSM).

### DIFF
--- a/galsim/wcs.py
+++ b/galsim/wcs.py
@@ -173,7 +173,7 @@ class BaseWCS:
     def toWorld(self, *args, **kwargs):
         """Convert from image coordinates to world coordinates.
 
-        There are essentially three overloaded versions of this function here.
+        There are essentially four overloaded versions of this function here.
 
         1. The first converts a `Position` from image coordinates to world coordinates.
            It returns the corresponding position in world coordinates as a `PositionD` if the WCS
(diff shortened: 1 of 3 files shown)


beckermr (commented):
My only comment is that I don't know the test suite well enough to understand the distortion level in the WCS mappings being used to test the shear. It'd be good to ensure this works even in high distortion cases. Many celestial WCS classes are close to a pixel scale depending on their use. 

Otherwise, LGTM and thank you!

rmjarvis:
The test suite has some pretty crazy WCS's with very non-square pixels.  One was distorted enough that I had to put the |g| < 0.8 guard in for hsm, since the image was coming back as practically a straight line and hsm couldn't do anything useful with it.  But yeah, a number of the others were quite non-trivial tests of the math.

beckermr (approved):

galsim/wcs.py
```diff
@@ -345,6 +397,25 @@ def profileToImage(self, world_profile, image_pos=None, world_pos=None, color=No
         return self.local(image_pos, world_pos, color=color)._profileToImage(
                     world_profile, flux_ratio, PositionD(offset))
 
+    def shearToImage(self, world_shear, image_pos=None, world_pos=None, color=None):
+        """Convert a shear measured in world coordinates to image coordinates
+
+        This is equivalent to ``wcs.toImage(shear, ...)``.
+
+        This reverses the conventions described in `shearToWorld`.
```
rmandelb, line 405:
```suggestion
        This reverses the process described in `shearToWorld`.
```
I think the shear convention is the same and it's the process (of moving between two coordinate systems) that is reversed.

rmjarvis: Agreed.  That's a better word here.

Issue #731, opened 2016-03-29
name for correlated noise in config


rmandelb:
Hi @rmjarvis - Do I understand correctly that the only way to make correlated noise using config is with the `COSMOS` noise option?  And that the `file_name`, `cosmos_scale`, and `variance` can be reset to any other correlated noise you want?

If so, I wonder if we could perhaps change the name of this from `COSMOS` to something more general?  The name implies that you can only do COSMOS-like noise, but it seems like it's much more powerful than that.  Given those inputs it seems you could make essentially any arbitrary correlated noise field provided that you have a description of its correlation function.

rmjarvis:
I guess so.  But it uses the `correlatednoise.getCOSMOSNoise` function, so if you want to change the name, we should probably change the name of that function too.  

I think @barnabytprowe named it that because we don't have any other examples to use for the file here.  cf. this part of the doc string for `getCOSMOSNoise`:
```
    This function uses a stacked estimate of the correlation function in COSMOS noise fields.
    The correlation function was computed by the GalSim team as described in

        /YOUR/REPO/PATH/GalSim/devel/external/hst/make_cosmos_cfimage.py

    and the resulting file is distributed with GalSim as

        /YOUR/REPO/PATH/GalSim/share/acs_I_unrot_sci_20_cf.fits
```
However, I guess one could run `make_cosmos_cfimage.py` with different input images and produce the correlated noise file for a different telescope.

rmandelb:
Yes.  I'm doing this with an externally-produced noise CF right now, and it's not for COSMOS.

I can think of a few options (other than leaving it as-is):

1. Leave getCOSMOSNoise and the "COSMOS" noise option as-is, but make a more general one as well, like ```correlatednoise.getCorrNoise``` and "Correlated" (for config).  That one would not have defaults.

2. Change the names of the python routine and the config option to be more general, like the options in (1), but let the keywords keep their current COSMOS-specific defaults.

3. Change the names of the python routine and the config option to be more general, like the options in (1).  Make the keywords for filename, pixel scale, and variance not have the COSMOS default, but add a keyword for ```noise_type``` that can be set to COSMOS and if so, use the current COSMOS-specific defaults.

By the way, I think this doesn't have to be resolved before we merge the PR for #732.  May as well get the bug fix quickly onto master, even if we don't quickly converge to an answer on this issue.

rmjarvis:
[...]

PR #1174, opened 2022-07-24
#731 Generalize COSMOSCatalog and COSMOS noise to be more general

### LINKED ISSUE #954: Make it easier to access the zphot record in the COSMOS catalog
@rearmstr asked about accessing the redshift information that is in the COSMOS catalog for use with an NFWHalo simulation (which uses the redshift of each galaxy to compute the correct amount of lensing).

It doesn't seem like there is currently any easy way to do this either in config (his use case) or even using the regular COSMOSCatalog class.  It requires a bit of insider knowledge of the record attribute and such.  

Anyway, here is the custom ValueType code that I proposed to access this value from a config file:

```
def COSMOS_ZPhot(config, base, value_type):
    cosmos_cat = galsim.config.GetInputObj('cosmos_catalog', config, base, 'COSMOS_ZPhot')
    # Requires that galaxy uses an explicit index.
    # This will raise an exception if index is not set.
    index = galsim.config.GetCurrentValue('gal.index', base, value_type=int)
    record = cosmos_cat.getParametricRecord(index)
    return record['zphot']
```
However, it requires that the gal object explicitly set the index to be used.  It can be simply `type : Random`, which will approximate the effect of not using an explicit type, just without the optional weight scheme for selecting an appropriate galaxy.

I propose the following updates in this issue:
* Put something like the above code into GalSim-proper, so as not to require a custom type to do this
* Have the regular COSMOSGalaxy code store the index that gets used somewhere, whether it is set via an explicit `index` parameter or not (including if the weights are used).
* Update the above code to use the stored index value, rather than using `GetCurrentValue`.
* Maybe also add a few other accessors.  E.g. flux, half-light-radius, others?
* Probably also make the config code not require the galaxy type to be COSMOSGalaxy in order to use these accessors.
* Add this to the demo9 example.  It is currently our example for InclinedExponential, but it would be nice if we could take the size, flux, and redshift from the COSMOS catalog using the above new accessors.

rmjarvis:
This issue has been right on the edge of being worth anyone's time to do for a long time.  We (I probably) keep adding it to milestones aspirationally, only to remove it when we finally cut a release.  I didn't want to do that again, so I finally went ahead and did it.

Most of the functionality in our `COSMOSCatalog` class and `getCOSMOSNoise` function would work for things other that COSMOS specifically.  So this PR adds general versions of each as laid out in the issue #731.  Namely:

* `GalaxySample` is the new generic version of `COSMOSCatalog`.  It requires the file name to be specified explicitly rather than figuring it out from a `sample` keyword.  It also takes some extra parameters that are provided automatically in the COSMOS case (e.g. orig_area, orig_exptime).
* `BaseCorrelatedNoise.from_file` is the new generic version of `getCOSMOSNoise`.  The only difference is that the file_name has to be explicitly specified.
* A new config noise type `Correlated` is like `COSMOS`, but requiring an explicit file name.
* A new config input type `galaxy_sample` implements the `GalaxySample` class with corresponding object type `SampleGalaxy`.
* scene.py has been renamed galaxy_sample.py to better match the names of the included functionality.
* There are new value types COSMOSValue and SampleValue, which allow access to the parametric records in the GalaxySample catalog.  (This fixes #954.)

I also wanted to close out #795, but I'm not actually sure from reading that issue what is left to fix.  I know I fixed part of it a while ago, but it looked like @rmandelb still wanted some use case to be enabled that I couldn't quite parse.  Rachel, could you check if there is still anything to do there, or if the current code does everything you need with respect to the name of the sample catalog?  Happy to try to fix anything else relevant as part of this.


rmandelb:
Thanks for doing all this, Mike.  Before reviewing the code I have some comments about your top-level summary.

> GalaxySample is the new generic version of COSMOSCatalog

This is generalized along two axes and made more specific along a third.  What I mean is:

1. Removing COSMOS makes it more general in terms of the origin of the sample.
2. In my mind `Sample` is more general than `Catalog` (e.g., you could imagine defining a sample through draws from parametric distributions, or by using neural network models like in https://github.com/McWilliamsCenter/galsim_hub).
3. `Galaxy` is more specific.  In principle, you could imagine having a class that one can use to specify sky scenes, whether galaxies, stars, or a mix.  But this specifies it to galaxies, which is consistent with our past use but maybe not to potential future uses?  (I'm wondering about studies of chromatic effects, for example, where you might want to have mixes of stars and galaxies with various SEDs.)  Maybe this was OK, I just wanted to check.

This comment re: specificity also applies to

> A new config input type galaxy_sample implements the GalaxySample class with corresponding object type SampleGalaxy.
scene.py has been renamed galaxy_sample.py to better match the names of the included functionality.

Maybe it's fine, but we originally chose `scene` because one could define sky scenes with mixes of objects, rather than only galaxies.

I suppose if we start using it in a more general manner we could always rename again, but with many years of galaxies-only it is probably safe to assume it's staying that way?

From the description of what you did in this PR, it sounds like my request in #795 may actually be obsolete, because it sounds like the class no longer makes an assumption about the filename and requires you to specify it - the issue in #795 came from the filename assumptions for the main file and auxiliary files.

rmjarvis:
We do have an example that mixes stars and galaxies using the MixedScene stamp type (sourced from galsim-extra, but copied locally) in demo13.  So I think if we wanted something more sophisticated, it would be reasonable to have a StarSample class that gave some useful mix of stellar SEDs or something.  But TBH, I suspect it's unlikely such a thing will eventually make it into GalSim proper.  Seems like something a user would make on their own, and if generous, submit to galsim-extra for others to use.  I don't think it would need to necessarily live in GalSim itself.

galsim/correlatednoise.py
```diff
@@ -74,6 +75,9 @@ class BaseCorrelatedNoise:
     is that users instead instantiate derived classes, such as the `CorrelatedNoise`, which are
     able to guarantee the above.
 
+    If you have the correlation function as an image on file, you can use the class  method
+    `from_file`, which does confirm that the file has the appopriate symmetry.
```
rmandelb, line 79:
```suggestion
    `from_file`, which does confirm that the file has the appropriate symmetry.
```

galsim/correlatednoise.py
```diff
@@ -117,6 +121,106 @@ def __init__(self, rng, gsobject, wcs=PixelScale(1.0)):
...
+        2. If the `BaseCorrelatedNoise` instance ``cn`` was instantiated with an image of scale
+           comparable to that in the final output, and ``cn`` has been rotated or otherwise transformed
+           (e.g.  via the `BaseCorrelatedNoise.rotate`, `BaseCorrelatedNoise.shear` methods; see below).
+
+        Conversely, the approximation will work best in the case where the correlated noise used to
+        instantiate the ``cn`` is taken from an input image for which ``image.scale`` is smaller than
+        that in the desired output.  This is the most common use case in the practical treatment of
+        correlated noise when simulating galaxies from space as observed in ground-based surveys.
+
+        Changing from the default bilinear interpolant is made possible, but not recommended.  For more
```
rmandelb, line 163: This paragraph makes it sound like the PR discusses interpolant choices, which I was curious about, so I went there.  I couldn't find anything on it there - did I miss it?  I was actually wondering how much testing we've done with modifying the interpolants for these CorrelatedNoise objects.

galsim/correlatednoise.py
```diff
@@ -117,6 +121,106 @@ def __init__(self, rng, gsobject, wcs=PixelScale(1.0)):
...
+
+        # Check for invalid images:
+        if cfimage.nrow != cfimage.ncol:
+            raise GalSimError("Input image is not square.")
+        if cfimage.nrow % 2 != 1:
+            raise GalSimError("Input image does not have odd size.")
+        if np.max(np.abs((cfimage - cfimage.rot_180()).array)) > 1.e-12:
+            raise GalSimError("Input image does not have 180 degree rotational symmetry.")
+
+        # Also check for invalid negative variance
```
rmandelb, line 202: Should we also check that the cross-covariances obey -1<=C[i,j]/C[0,0]<=1?  Am I right in thinking that this must be mathematically the case or is that not true?

rmandelb: This could help avoid a scenario, for example, that somebody puts the variance at (0,0) instead of in the center.  Such a case might pass your previous test if the center pixel happens to have a positive cross-covariance, no matter how small.

galsim/galaxy_sample.py
```diff
@@ -31,39 +31,23 @@
...
+class GalaxySample:
     """
-    A class representing a random subsample of galaxies from the COSMOS sample with F814W<25.2
-    (default), or alternatively the entire sample with F814W<23.5.
+    A class representing a random subsample of galaxies from an arbitrary deep survey.
 
     Depending on the keyword arguments, particularly ``use_real``, the catalog will either have
-    information about real galaxies, and/or parametric ones.  To use this with either type of
-    galaxies, you need to get the COSMOS datasets in the format that GalSim recognizes; see
+    information about real galaxies, and/or parametric ones.
```
rmandelb, line 39: I realize this isn't new text.  But I think it would be more accurate to say `use_real` tells the software what type of information you want to use for making the sims, right?  I mean, a catalog might have both and `use_real` chooses between them.

galsim/galaxy_sample.py
```diff
@@ -31,39 +31,23 @@
...
+    The original version of this functionality is now in the subclass `COSMOSCatalog`, which
+    specializes to HST observations of the COSMOS field.  `GalaxySample` is a generalization of
+    that, which works for arbitrary data sets, although the user is responsible for building
+    the appropriate input files to use with it.
 
-    Finally, we provide a program that will download the large COSMOS sample for you and
-    put it in the $PREFIX/share/galsim directory of your installation path.  The program is::
+    Unlike `COSMOSCatalog`, which has easy options for picking out one of the two galaxy subsets
+    that are available for download using ``galsim_download_cosmos``, for this class you need to
+    manually specifiy the file name.
```
rmandelb, line 48:
```suggestion
    manually specify the file name.
```

galsim/galaxy_sample.py
```diff
@@ -137,7 +121,9 @@ class COSMOSCatalog:
                             - "marginal": Apply the above cuts, plus ones that eliminate some more
                               marginal cases.
 
-                            [default: "marginal"]
+                            Use of "bad_stamp" or "marginal" requires a ``CATALOG_selection.fits``
+                            file (where CATALOG refers to the regular catalog base name).
```
rmandelb, line 125: Say how/where that base name is specified?

galsim/galaxy_sample.py
```diff
@@ -270,18 +260,12 @@ def _apply_exclusion(self, exclusion_level, min_hlr=0, max_hlr=0, min_flux=0, ma
...
-                cut_ratio = 0.2
-                sn_limit = 20.0
-            else:
-                cut_ratio = 0.8
-                sn_limit = 12.0
             div_val = self.selection_cat['peak_image_pixel_count']
             div_val[div_val == 0.] = 1.e-5
-            mask &= ( (self.selection_cat['sn_ellip_gauss'] >= sn_limit) &
+            mask &= ( (self.selection_cat['sn_ellip_gauss'] >= self.sn_limit) &
                       ((self.selection_cat['min_mask_dist_pixels'] > 11.0) |
```
rmandelb, line 266: This still seems to be hardcoded to 11, whereas the other parameters are now quantities that can be passed in; was this distinction deliberate?

galsim/galaxy_sample.py
```diff
@@ -758,15 +744,185 @@ def getParametricRecord(self, index):
...
                  self.real_cat == other.real_cat and
                  np.array_equal(self.param_cat, other.param_cat) and
                  np.array_equal(self.orig_index, other.orig_index)))
+
+class COSMOSCatalog(GalaxySample):
+    """
+    A class representing a random subsample of galaxies from the COSMOS sample with F814W<25.2
+    (default), or alternatively the entire sample with F814W<23.5.
+
+    Depending on the keyword arguments, particularly ``use_real``, the catalog will either have
```
rmandelb, line 770: Same comment from GalaxySample docs applies here.

rmandelb (approved):
These changes look good to me, thanks Mike! I had minor questions and suggestions. However, I'm leaving for vacation later today so I am clicking "approve" because I don't want to hold you up. 

I was wondering about opportunities to incorporate this into the examples?  There are some that use COSMOSCatalog() so it might be as simple as a comment mentioning the more general GalaxySample option.

galsim/correlatednoise.py
rmjarvis: Well, I tracked it down.  Here is the relevant thread:  https://github.com/GalSim-developers/GalSim/pull/452/files#r5701561.  I'm not sure why it shows up in files, but not conversation...

However, this probably isn't super useful to point users to go looking at.  It's basically me and Barney repeatedly saying we don't really understand what the right thing is here, and that Linear just seems to work out the best.  Pretty unsatisfying, really.  So I'm just going to change this to something more generic: "Changing from the default bilinear interpolant is made possible, but not recommended. In our validation tests, we found that the `Linear` interpolant usually gave the most accurate results."

(BTW, this text is all copied verbatim from the `getCOSMOSNoise` function.)

galsim/correlatednoise.py
rmjarvis: Yep, good idea.  I added that.

galsim/galaxy_sample.py
rmjarvis: Right.  This is talking about the sample, not the source file.  (Catalog is ambiguous here.)  I changed it to "Depending on the keyword arguments, particularly ``use_real``, a `GalaxySample` may be able to generate real galaxies, parametric galaxies, or both."

galsim/galaxy_sample.py
rmjarvis: Changed to "(where CATALOG is ``file_name`` without the ".fits" extension)".

galsim/galaxy_sample.py
rmjarvis: No.  Just an oversight.  Thanks.  I made it a parameter now.

Issue #1067, opened 2019-12-01
Add a GalSim deviates (partial) drop-in replacement for numpy.RandomState.


jmeyers314:
In numpy, if I want a random array uniformly distributed between -0.6 and +0.6 with 100 elements, I'd write:

```
rng = numpy.random.RandomState(seed)
array = rng.uniform(-0.6, 0.6, 100)
```

Using GalSim deviates, I'd do something like:

```
ud = galsim.UniformDeviate(seed)
array = np.empty(100)
ud.generate(array)
array = array*1.2 - 0.6
```

I find the former easier to read.  I don't think it'd be too difficult to create an object based on GalSim deviates with  methods like `.uniform`, `.normal`=`.gaussian`, `.binomial`, `.poisson`, etc that have the same syntax as numpy.  Making a drop-in replacement also means that it's easier for external packages (e.g., imsim, batoid, ...) to switch from numpy random numbers to Galsim random numbers with their stronger guarantee of stability over time.

rmjarvis:
Sounds good.
I wouldn't make a new class for this though.  These can be methods of BaseDeviate I think.

jmeyers314:
I thought about that.  The only awkward part is that then `UniformDeviate` will have an inherited method `.poisson` and so on.  Maybe that's fine though.

rmjarvis:
Doesn't bother me.  :)  But either way.  If you prefer to make a new subclass of BaseDeviate that would enable these various driver methods, that's also fine.

rmjarvis:
Done via #1179

PR #1176, opened 2022-08-01
Add pupil samplers

jmeyers314:
Thanks for starting this @rmjarvis .  I changed `PupilSampler` -> `PupilImageSampler` to leave room for `PupilAnnulusSampler` which I added.

### DIFF
--- a/galsim/phase_psf.py
+++ b/galsim/phase_psf.py
@@ -632,10 +632,23 @@ class Aperture:
         pick = np.empty((n_photons,), dtype=float)
         ud.generate(pick)
         pick *= len(u)
-        pick = pick.astype(int)
-        photons.pupil_u = u[pick]
-        photons.pupil_v = v[pick]
-
+        pickint = pick.astype(int)
+        photons.pupil_u = u[pickint]
+        photons.pupil_v = v[pickint]
+        # Make continuous by adding +/- 0.5 pixels shifts.
+        uscale = self.u[0, 1] - self.u[0, 0]
+        vscale = self.v[1, 0] - self.v[0, 0]
+        # Reuse pick but rename for clarity
+        du = pick
+        ud.generate(du)
+        du -= 0.5
+        du *= uscale
+        photons.pupil_u += du
+        dv = pick
+        ud.generate(dv)
+        dv -= 0.5
+        dv *= vscale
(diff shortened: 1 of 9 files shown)


galsim/phase_psf.py
```diff
@@ -620,6 +620,23 @@ def __getstate__(self):
...
+        n_photons = len(photons)
+        u = self.u_illuminated
+        v = self.v_illuminated
+        ud = UniformDeviate(rng)
+        pick = np.empty((n_photons,), dtype=float)
+        ud.generate(pick)
+        pick *= len(u)
+        pick = pick.astype(int)
+        photons.pupil_u = u[pick]
+        photons.pupil_v = v[pick]
```
rmjarvis, line 637: I thought you might add to this to get a more continuous distribution of u,v values.  Maybe add a random +-0.5 pixel in each direction?
```
duv = pick   # reuse this array, but renamed to not be confusing.
ud.generate(duv)
duv -= 0.5
duv *= self.pupil_plane_scale
photons.pupil_u += duv[:len(photons.pupil_u)]
ud.generate(duv)
duv -= 0.5
duv *= self.pupil_plane_scale
photons.pupil_v += duv[:len(photons.pupil_v)]
```
And all of this boilerplate is making me want to do #1067...   :)

jmeyers314: > Maybe add a random +-0.5 pixel in each direction?

Sure.  I guess that's a bit wrong at the edges, but probably not more wrong than the current strategy.

jmeyers314: > And all of this boilerplate is making me want to do https://github.com/GalSim-developers/GalSim/issues/1067... :)

Not sure, but I think the current version of this might just be to subclass [`numpy.random.BitGenerator`](https://numpy.org/doc/stable/reference/random/bit_generators/generated/numpy.random.BitGenerator.html#numpy.random.BitGenerator) using `BaseDeviate` to generate the bits.

rmjarvis: "just" is not as easy as the docs imply.  Here is the longer explanation of how to subclass this: https://numpy.org/doc/stable/reference/random/extending.html

rmjarvis (approved):
LGTM

PR #1178, opened 2022-08-12
Add TimeSampler

jmeyers314:
Same idea here as #1176 where we factored out pupil sampling from `PhaseScreenPSF._shoot`.  This time, factor out the temporal sampling.  This isn't immediately useful for anything directly inside GalSim, but would likely be important if we want to rotate the diffraction spikes on an Alt-Az mounted telescope in the future, e.g..

### DIFF
--- a/tests/test_chromatic.py
+++ b/tests/test_chromatic.py
@@ -2190,15 +2190,7 @@ def test_phot():
         print('psf = ',psf)
         atol = 4.e-4
         if psf in [psf5, psf6]:
-            atol = 5.e-4  # OpticalPSF doesn't match quite as well as the others.
-                          # geometric_shooting=False does better, even with the larger aberrations,
-                          # but it doesn't reliably pass at 3e-4 under different random seeds,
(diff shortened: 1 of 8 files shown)


tests/test_chromatic.py
```diff
@@ -2199,6 +2199,8 @@ def test_phot():
                           # OpticalPSF.  We effectively rotate the PSF 180 degrees here, but not the
                           # galaxy, so we end up with a slightly different profile.  I (JM) think we
                           # just got lucky before.
+                          # Bumped again.  We were barely failing with 5.0361e-4 > 5e-4 after adding
+                          # the time sampling.
```
rmjarvis, line 2203: I'm not sure we need the complete history of this atol choice.  We rarely have a priori-derived tolerances.  Very often these are just something that seems suitably small and passes.  Honestly, I think just writing "OpitcalPSF doesn't match quite as well as the others." is completely sufficient as justification.

rmjarvis (approved):
Looks good.  Just vv minor text suggestions.

PR #1179, opened 2022-08-16
#1067 Enable galsim.BaseDeviate to be used as a numpy.Generator

rmjarvis:
This PR adds the ability to recast a `galsim.BaseDeviate` object as a `numpy.Generator`.  I allow this via `rng.as_numpy_generator()`, for people who prefer verbose methods for this kind of thing.  And a quick and dirty `rng.np` property for people who prefer an abbreviated syntax.  I think I'll prefer the latter, but they do the same thing, so to each their own.

This took me a while to figure out, since the documentation for making your own numpy.random.BitGenerator is pretty spare.  And the primary source for all this is done in cython, which I didn't want to deal with on our end.  

But it turns out not to be that hard once I figured out how everything works.  You basically just let the numpy base class make all the initial stuff.  Then you overwrite the elements of the `capsule`'s `bitgen_t` struct with the functions you actually want to use.  And there are only three functions returning `uint32_t`, `uint64_t` and `double`.  So those are pretty easy.  I was almost surprised when it actually worked for me.  (To be fair, this was attempt number 30 or so, so I'd gotten used to failing by that point.)

I also updated a few places in the code to use the new numpy syntax, where I thought the numpy syntax was clearly superior.  (Including the recent PR's of @jmeyers314 which prompted this effort.)

### DIFF
--- a/pysrc/Random.cpp
+++ b/pysrc/Random.cpp
@@ -20,7 +20,46 @@
 #include "PyBind11Helper.h"
 #include "Random.h"
 
+// From numpy/random/bitgen.h
+typedef struct bitgen {
+  void *state;
+  uint64_t (*next_uint64)(void *st);
+  uint32_t (*next_uint32)(void *st);
+  double (*next_double)(void *st);
+  uint64_t (*next_raw)(void *st);
+} bitgen_t;
+typedef bitgen bitgen_t;
+
 namespace galsim {
+
+    uint32_t next_uint32(void* st)
+    {
+        BaseDeviate* rng = static_cast<BaseDeviate*>(st);
+        return (uint32_t) rng->raw();
+    }
+
+    uint64_t next_uint64(void* st)
+    {
+        uint32_t i1 = next_uint32(st);
+        uint32_t i2 = next_uint32(st);
+        return ((uint64_t) i1 << 32) + i2;
+    }
+
+    double next_double(void* st)
+    {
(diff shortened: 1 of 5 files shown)


rmjarvis:
Ref: https://stackoverflow.com/a/73367856/1332281

pysrc/Random.cpp
```diff
@@ -19,9 +19,49 @@
...
+    uint64_t next_uint64(void* st)
+    {
+        uint32_t i1 = next_uint32(st);
+        uint32_t i2 = next_uint32(st);
+        return ((uint64_t) i1 << 32) + i2;
+    }
+
+    double next_double(void* st)
+    {
+        return next_uint32(st) / ((double) std::numeric_limits<uint32_t>::max()+1.);
```
jmeyers314, line 52: Could this use `next_uint64` instead to give a larger range of possible doubles?  Not worth the speed hit?

jmeyers314 (approved):
LGTM.  Couple small questions/comments.

pysrc/Random.cpp
rmjarvis: I'd be surprised if any use case would benefit from using 64 bits here rather than 32.  Certainly nothing in GalSim proper needs it, as this version matches the current `ud.generate()` functionality, which is what we're (sometimes) replacing with this.  So I think it's probably not worth cutting the speed in half to get a real 64 bit range of output values.

PR #1183, opened 2022-10-10
Make it possible to delete items from a config list

rmjarvis:
@g-braeunlich found that the current syntax for deleting an item from the config doesn't work to delete things from a list.  This PR remedies this.

The one potentially confusing thing that can result from this is if the user also wants to modify other items in the same list.  Once an item is deleted, these would need to use the new indices, not the original ones.  This seems enough error-prone that I emit a logger warning.

I've targeted this to releases/2.4.  I think this can count as a bug, so could go into a bugfix release.

### DIFF
--- a/galsim/config/util.py
+++ b/galsim/config/util.py
@@ -591,7 +591,7 @@ def GetFromConfig(config, key):
             "Unable to parse extended key %s.  Field %s is invalid."%(key,k)) from None
     return value
 
-def SetInConfig(config, key, value):
+def SetInConfig(config, key, value, logger=None):
     """Set the value of a (possibly extended) key in a config dict.
 
     If key is a simple string, then this is equivalent to config[key] = value.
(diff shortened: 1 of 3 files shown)


galsim/config/util.py
```diff
@@ -608,7 +608,17 @@ def SetInConfig(config, key, value):
...
         # This means remove it, if it is there.
-        d.pop(k,None)
+        try:
+            d.pop(k,None)
+        except TypeError:
+            # Then d was really a list.
+            # This has some potentially counter-intuitive consequences, so let the user
+            # know about them.
+            if logger:
+                logger.warning("Warnign: Removing item %d from %s. "%(k,key[:key.rfind('.')]) +
```
jmeyers314, line 618:
```suggestion
                logger.warning("Warning: Removing item %d from %s. "%(k,key[:key.rfind('.')]) +
```

jmeyers314 (approved):
LGTM.

PR #1184, opened 2022-10-12
Fix bug in how input fields check when they are current

rmjarvis:
This fixes another bug discovered during imSim development.  @jchiang87 found that sometimes the skycatalog input object wasn't being loaded correctly when there were multiple ccds being rendered in the same config run.  It took me a while to figure out the problem, but it stemmed from how the `current` item was being set (and checked), which was different from how I do that for regular (non-input) values.  It mostly worked, but wasn't quite correct, leading to the bug that Jim found.

Anyway, this mostly just switches over the current checks there to more closely mimic the way I do it elsewhere.  And I added a unit test that fails for the old code, but works now.  

TBH, I feel a little silly about not already having written this test, which would have found the error sooner.  It's sort of an obvious extension to a test that we already had, but which skipped running things the normal way using Process, so it managed to not hit the code paths that had the error.

### DIFF
--- a/galsim/config/input.py
+++ b/galsim/config/input.py
@@ -21,7 +21,7 @@ import logging
 
 from .value import RegisterValueType
 from .util import LoggerWrapper, RemoveCurrent, GetRNG, GetLoggerProxy, get_cls_params
-from .util import SafeManager
+from .util import SafeManager, GetIndex, PropagateIndexKeyRNGNum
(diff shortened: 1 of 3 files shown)


jchiang87 (approved):
I confirm that this fixes the problem I was seeing in imSim, so I'm ok with merging, but I don't understand this code well enough to do a real review of these changes.

PR #1186, opened 2022-10-21
Fix hsm assert when step or stride is < 0

rmjarvis:
The asserts in the hsm code are not correct when either step or stride is negative.  For now, just skip these checks when either step or stride is negative.  But possibly down the line we might want to make this check more rigorous to include the case of negative steps or strides.

### DIFF
--- a/include/galsim/Image.h
+++ b/include/galsim/Image.h
@@ -188,6 +188,13 @@ namespace galsim {
         const T* getData() const { return _data; }
         const T* getMaxPtr() const { return _maxptr; }
 
+        /// Check that the ptr is valid, properly accounting for step, stride
+        // Normally (if step, stride > 0), this check is just _data <= p < _maxptr.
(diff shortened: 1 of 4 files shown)


jmeyers314:
Thanks, this looks good to me.

FWIW, I think rigorous here probably just means:
```c++
_minptr = _data + min(_step*(_ncol-1), 0) + min(_stride*(_nrow-1), 0);
_maxptr = _data + max(_step*(_ncol-1), 0) + max(_stride*(_nrow-1), 0) + 1;
```
and then `return _minptr < p && p < maxptr;`.

That doesn't seem too bad to give a go at now, but I admit I didn't try to look for all the possible ways `_maxptr` currently gets set.

rmjarvis:
I think I'll leave the rigorous version for later if it ever seems like it would be useful.  Certainly adding another variable to the class now would cause ABI incompatibility, so we shouldn't knowingly do that on a bugfix release.

PR #1189, opened 2022-10-31
Fix RADec docs capitalization

jmeyers314:
@rmjarvis , this held me up a bit this morning.  Not sure if you want to update on v2.4 docs or not.


rmjarvis (approved):
The title says capitalization.  Were there other doc errors than the one here you wanted to fix?

rmjarvis:
Oh, never mind.  There is a caps issue there too.  I missed that, even when specifically looking for it.  🤣

PR #1190, opened 2022-11-09
Fix Scattered positions if world_pos is in the stamp field

rmjarvis:
Josh discovered that the Scattered type doesn't respect `world_pos` if it's in the stamp field, as it should.  Instead it adds its own `image_pos` specification in the image field, which ends up taking precedence.

It turns out that an `image_pos` in the stamp field did work, kind of by accident, just because when we make the position, it checks stamp first, finds it, and uses it, and just ignores the default one that was made in the image field.  But now the Scattered function explicitly checks the stamp field and only makes the default if there are no other positions anywhere.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,8 @@ API Changes
 Config Updates
 --------------
 
+- Fixed a bug in Scattered type, where it didn't respect world_pos being specified in the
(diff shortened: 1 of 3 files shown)


jmeyers314 (approved):
LGTM.  Thanks!

PR #1191, opened 2022-11-09
Deprecate getter allocation in PhotonArray

rmjarvis:
Currently, accessing any array in PhotonArray will make sure that it gets allocated if it isn't already and then return the relevant numpy array.  This is intended to make it easy for the user, who can just pretend all the arrays are already allocated and initialized to zeros, but only use up that memory if the user actually indicates that they are using the arrays.

However, Josh pointed out that this can be confusing if you think you've assigned values to something, but didn't and then access the arrays, it will happily give back zero arrays.  This is likely almost always an error.  And for the rare cases where the user might have intended this usage (e.g. doing `rng.generate(photon_array.dxdz)` to populate it), it's easy to just explicitly call `allocateWhatever()` first to make sure the appropriate array is allocated.  

The more typical case of using the setter method (e.g. `photon_array.dxdz = dxdz_values`) will still work as before.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,12 @@ Changes from v2.4 to v2.5
 API Changes
 -----------
 
+- Deprecated automatic allocation of `PhotonArray` arrays via "get" access rather than
+  "set" access for the arrays that are not initially allocated.  E.g. writing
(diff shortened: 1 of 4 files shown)


jmeyers314 (approved):
This one looks good too.  Thanks again!

PR #1192, opened 2022-11-15
Try out GHA with python 3.11

jmeyers314:


rmjarvis:
```
tests/run_examples.py::test_psf_wf_movie
tests/run_examples.py::test_psf_wf_movie
  /home/runner/work/GalSim/GalSim/examples/psf_wf_movie.py:34: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
    if LooseVersion(matplotlib.__version__) < LooseVersion('1.2'):

tests/run_examples.py::test_fft_vs_geom_movie
tests/run_examples.py::test_fft_vs_geom_movie
  /home/runner/work/GalSim/GalSim/examples/fft_vs_geom_movie.py:54: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
    if LooseVersion(matplotlib.__version__) < LooseVersion('1.2'):
```
No errors. These are the only warnings.  I think we can probably just remove these checks entirely.  matplotlib v1.2 is really quite old at this point.  Probably fine to let the user get an error if they still have it.

jmeyers314:
I'm wondering if something is wrong with the pip caching.  For the Linux builds, it's 10-30 MB.  For the Macos build, it's about 2.2 GB and consequently rather slow to download.  Should we kick it maybe?

jmeyers314:
Okay, macos cache is still coming in at 2.2GB.  I'll leave it to you, @rmjarvis , if you want to continue to investigate, but I think I'm happy enough now with this.

rmjarvis:
I fought a while with the caches on Travis back when we used that.  GHA seems better, but I don't really have any insights as to why mac would be so different from the linux ones.  Probably just leave it for now.

rmjarvis (approved):

PR #1193, opened 2022-11-17
Fix some problems when using photon_ops with fft drawing.

rmjarvis:
@g-braeunlich discovered that using photon_ops with method='fft' doesn't quite work as intended.

First, if you don't explicitly give a sensor object, then it bails out telling you that this isn't allowed.
Second, there is no way to limit the number of photons used as we do when photon shooting very bright objects, so the memory explodes if you try to do this for very bright objects (as we want to do in imSim).

So this fixes some of the sanity checks to work correctly when the user is applying photon ops on an fft image.  And it implements the maxN parameter for this use case, following the same pattern as we do in drawPhot.

### DIFF
--- a/galsim/gsobject.py
+++ b/galsim/gsobject.py
@@ -1275,7 +1275,7 @@ class GSObject:
                   method='auto', area=1., exptime=1., gain=1., add_to_image=False,
                   center=None, use_true_center=True, offset=None,
                   n_photons=0., rng=None, max_extra_noise=0.,
-                  poisson_flux=None, sensor=None, photon_ops=(), n_subsample=3, maxN=None,
+                  poisson_flux=None, sensor=None, photon_ops=None, n_subsample=3, maxN=None,
                   save_photons=False, bandpass=None, setup_only=False,
                   surface_ops=None):
         """Draws an `Image` of the object.
@@ -1597,7 +1597,7 @@ class GSObject:
                             photons onto the image. [default: None]
             photon_ops:     A list of operators that can modify the photon array that will be
                             applied in order before accumulating the photons on the sensor.
-                            [default: ()]
+                            [default: None]
             n_subsample:    The number of sub-pixels per final pixel to use for fft drawing when
                             using a sensor.  The sensor step needs to know the sub-pixel positions
                             of the photons, which is lost in the fft method.  So using smaller
@@ -1676,8 +1676,11 @@ class GSObject:
                     method=method, sensor=sensor, poisson_flux=poisson_flux)
 
         # If using photon ops with fft, then need a sensor.
-        if photon_ops != () and method != 'phot' and sensor is None:
+        if photon_ops is not None and method != 'phot' and sensor is None:
             sensor = Sensor()
+        if photon_ops is None:
+            # Easier to just make it an empty tuple, rather than deal with None below.
+            photon_ops = ()
 
         # Some parameters are only relevant for either phot or when using a sensor.
         if method != 'phot' and sensor is None:
--- a/galsim/gsobject.py
+++ b/galsim/gsobject.py
@@ -1625,6 +1625,7 @@ class GSObject:
         from .box import Pixel
         from .wcs import PixelScale
         from .photon_array import PhotonArray
+        from .sensor import Sensor
 
         if surface_ops is not None:
             from .deprecated import depr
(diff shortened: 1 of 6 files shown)


rmjarvis:
> I think somewhere in the FFT pipeline, there also should be a call to samplePupil. Or I missed another piece of config maybe.

No, that would be an imsim error, not a galsim one.  Your photon op requires that the pupil positions be set, so you need to include a SamplePupil operator before the diffraction one.  GalSim would have no possible way of knowing what kind of pupil to sample from on its own.

galsim/gsobject.py
```diff
@@ -1677,17 +1674,24 @@ def drawImage(self, image=None, nx=None, ny=None, bounds=None, scale=None, wcs=N
                 raise GalSimIncompatibleValuesError(
                     "poisson_flux is only relevant for method='phot'",
                     method=method, sensor=sensor, poisson_flux=poisson_flux)
-            if photon_ops != ():
+
+        # If using photon ops with fft, then need a sensor.
+        if photon_ops != () and method != 'phot' and sensor is None:
```
jmeyers314, line 1679: Maybe `tuple(photon_ops) != ()`, or some other way of checking for a generic empty Sequence?

tests/test_photon_array.py
```diff
@@ -686,6 +686,35 @@ def test_dcr():
...
+    res3 = im3.FindAdaptiveMom()
+    np.testing.assert_allclose(res3.moments_amp, res2.moments_amp, rtol=1.e-2)
+    np.testing.assert_allclose(res3.moments_sigma, res2.moments_sigma, rtol=1.e-2)
+    np.testing.assert_allclose(res3.observed_shape.e1, res2.observed_shape.e1, atol=1.e-2)
+    np.testing.assert_allclose(res3.observed_shape.e2, res2.observed_shape.e2, atol=1.e-2)
+    np.testing.assert_allclose(res3.moments_centroid.x, res2.moments_centroid.x, rtol=1.e-2)
+    np.testing.assert_allclose(res3.moments_centroid.y, res2.moments_centroid.y, rtol=1.e-2)
+
+    # Repeat with maxN < flux
+    achrom.drawImage(image=im3, method='auto', rng=rng, photon_ops=photon_ops, maxN=10**4)
```
jmeyers314, line 706: Is (or should) the result be independent of the value of `maxN`?  If so, then somewhere an `assert_equal` feels necessary.

jmeyers314 (approved):
LGTM.  A couple quick questions mostly.

galsim/gsobject.py
rmjarvis: I'm not sure.  
`()` is the default value if the user doesn't specify something else, which is why that's what I'm comparing here.  If the user specifies `[]` for instance instead, that feels intentional, like maybe the photon_ops was generated programmatically, and happened not to have any elements after whatever conditions were run through.  In that case, I think the user might hit some of these sanity checks, thinking it was ok to pass `rng`, since some of the potential photon_ops that might have been added (but weren't) need it.  The pure default empty tuple seems most likely to be just the default value, not an accidental user-supplied empty tuple.

tests/test_photon_array.py
rmjarvis: Good point.  I think it should be identical.  Or at least very close to equal.  (There might be some order of operations differences, so maybe rtol=1.e-15 kind of test).  I'll add a better test here.

galsim/gsobject.py
jmeyers314: Sure, it just feels weird to treat `[]` and `()` differently when whether the Sequence is mutable or not really doesn't matter here.  Could make the default `None` I guess.  Or just leave it alone; I'm sure this is a pretty obscure edge case.

rmjarvis: I agree.  I made the default value None.  That's more in line with how we do other defaults, so that's probably the better design here.

tests/test_photon_array.py
rmjarvis: Actually, it turns out they aren't identical, since they run through the random numbers very differently.  They are two equally valid realizations of the photon positions corresponding to the FFT image.  But not the same realization.  So I think the current test is appropriate.

jmeyers314: Sounds reasonable, and glad you added the comment to the test.  I think this is probably also worth mentioning in the drawImage docstring.

rmjarvis: Done.

PR #1195, opened 2022-12-03
Let Input Proxy objects access attributes, not just methods.

rmjarvis:
This PR fixes a long term annoyance about input objects in the config.  We use Proxies to avoid having input objects get cloned into every process, since they are often relatively large.  This works fairly well, but there is a catch -- the standard multiprocessing AutoProxy class doesn't expose attributes and properties of the wrapped object.  Just methods.  So you need to write getters for any attribute you want to access.  This has recently become onerous in ImSim.

Fortunately, @jmeyers314 discovered [this SO answer](https://stackoverflow.com/a/63741184/1332281), which shows how to circumvent this behavior.  I actually used a [slight modification](https://stackoverflow.com/a/65466807/1332281), which doesn't use exec, and then had to fix a [bug](https://stackoverflow.com/questions/65451457/how-can-i-update-class-members-in-processes#comment131796241_65466807) in it.  But basically, that was all that was needed.

No new unit tests.  I just changed some places in the code where we used getters rather than attributes to the more natural attribute syntax.  This failed without the new InputProxy class, and it works with that change.

### DIFF
--- a/galsim/config/input.py
+++ b/galsim/config/input.py
@@ -18,6 +18,9 @@
 
 import os
 import logging
+import types
+
+from multiprocessing.managers import NamespaceProxy
 
 from .value import RegisterValueType
 from .util import LoggerWrapper, RemoveCurrent, GetRNG, GetLoggerProxy, get_cls_params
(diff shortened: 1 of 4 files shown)


jmeyers314 (approved):
LGTM.  Glad this turned out to be so relatively painless!

PR #1198, opened 2022-12-05
Fix the flux scaling of drawReal for objects with non-diagonal jacobian

### LINKED ISSUE #1197: SBProfile::draw : Flux/imscale scaling questionable when full jacobian is provided.
Hi looking at 
```c++
void SBProfile::draw(ImageView<T> image, double imscale,
                         double* jac, double xoff, double yoff, double flux_ratio) const
```
I wander if there is not a typo concerning the relation of `flux_ratio` and `imscale` notably when 
the jacobian is non null and have non-diag elements, as then 
```c++
invdet *= imscale;
flux_ratio *= std::abs(invdet);
```
In fact all the other jacobian use-case leads to a scaling as `imscale*imscale`.

As I have missed something?
Thanks

rmjarvis:
@jecampagne noticed a bug in the C++-layer `SBProfile.draw` function that gets the flux wrong when using `drawReal` on an non-unit-pixel-scale image with an object with a non-diagonal Jacobian.  cf. discussion on #1197

Fortunately, this is a pretty obscure edge case, so I suspect it's unlikely this has ever caused a problem in practice.  But it's definitely better to have the code correct, of course.  :)

Fixes #1197

### DIFF
--- a/src/SBProfile.cpp
+++ b/src/SBProfile.cpp
@@ -357,7 +357,7 @@ namespace galsim {
             double dy = mA * invdet;
             double dyx = -mC * invdet;
             _pimpl->fillXImage(image, new_x0, dx, dxy, new_y0, dy, dyx);
-            flux_ratio *= std::abs(invdet);
+            flux_ratio *= std::abs(invdet * imscale);
         }
 
         // At this point the "flux" ration is really an amplitude ratio, including det(jac)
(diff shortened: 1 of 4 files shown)


rmandelb (approved):
LGTM, and the test addition makes sense as well.  Thanks!

rmjarvis:
Thanks Rachel!

PR #1201, opened 2022-12-13
Add --log_format to GalSim executable

sidneymau:
This pull request adds the `--log_format` option to the GalSim executable. The default behavior is `"%(message)s"` (i.e., consistent with the past behavior of the GalSim logger).

Example usage would be
```
$ galsim --log_format="%(asctime)s - %(name)s - %(levelname)s - %(message)s" demo1.yaml 
2022-12-13 14:12:47,824 - galsim - WARNING - Reading config file demo1.yaml
2022-12-13 14:12:47,834 - galsim - WARNING - Start file 0 = output_yaml/demo1.fits
2022-12-13 14:12:47,858 - galsim - WARNING - File 0 = output_yaml/demo1.fits: time = 0.024030 sec
2022-12-13 14:12:47,858 - galsim - WARNING - Done building files
```

### DIFF
--- a/galsim/download_cosmos.py
+++ b/galsim/download_cosmos.py
@@ -80,6 +80,7 @@ def parse_args(command_args):
         help="Don't link to the alternate directory from share/galsim")
     args = parser.parse_args(command_args)
     args.log_file = None
+    args.log_format = None
 
     # Return the args
     return args
--- a/galsim/main.py
+++ b/galsim/main.py
@@ -52,6 +52,9 @@ def parse_args(command_args):
         '-v', '--verbosity', type=int, action='store', default=1, choices=(0, 1, 2, 3),
         help='integer verbosity level: min=0, max=3 [default=1]')
     parser.add_argument(
+        '--log_format', type=str, action='store', default="%(message)s",
+        help='logger format string [default="%(message)s"]')
+    parser.add_argument(
         '-l', '--log_file', type=str, action='store', default=None,
         help='filename for storing logging output [default is to stream to stdout]')
     parser.add_argument(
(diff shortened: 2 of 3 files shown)


sidneymau:
Should probably add a single character alias for the command line parser, but both `-l` (log) and `-f` (format) are currently in use...

rmjarvis (changes requested):
Thanks Sid.  The code looks fine, but this needs some unit tests.
cf. https://github.com/GalSim-developers/GalSim/blob/releases/2.4/tests/test_main.py
Probably add something relevant in both test_args and test_logger functions.

sidneymau:
Thanks for pointing out the tests. I'm having problems running these locally (`galsim --help` leads to a `KeyError` in argparse on my machine, tested with both python 3.10 and 3.11 -- not sure if the behavior changed at some point), but hopefully it's straightforward enough that I didn't miss aything

rmjarvis:
So, the KeyError looks like a real bug.  I think the galsim --help ends up parsing the format string somewhere and locally at that point (somewhere within the ArgumentParser code) there is no variable `message`.  So probably it needs to be default=None, and then manually make that mean the simple `%(message)s`.
The other errors are that galsim_download_cosmos also calls the `make_logger` function and doesn't have this option.  So probably the best solution is to set the default in that function.
```
if not hasattr(args, 'log_format') or args.log_format is None:
    args.log_format = '%(message)s'
```
I think that should fix most of the test suite errors.  The last one (in test_logger) is a new test that I didn't try to figure out why it's failing.  FWIW, the test looked ok to me.

sidneymau:
I looked into this a bit more. The problem doesn't seem to be the argument itself but rather the `help` string (hence why the code works fine but `galsim --help` runs into an error). Fix by escaping the `%` (found this suggestion here: https://stackoverflow.com/questions/2537929/is-it-possible-to-temporarily-disable-pythons-string-interpolation)

sidneymau:
Fixed the galsim_download_cosmos test error by manually setting the log_format argument to None (in the same way as done for the log_file argument) in `galsim.download_cosmos.parse_args`

rmjarvis (approved):

rmjarvis:
Looks good to me Sid.  Thanks!

PR #1202, opened 2022-12-14
Allow input objects to return an approximate number of objects.

rmjarvis:
The calculation of the number of objects on each image in imsim can take quite a while when doing a lot of CCDs, since it needs to load the WCS and compute the overlap of the CCD with heal-pixels, etc.  This is currently being done in the serial part of the processing, rather than the parallel part, since GalSim needs to know the object numbers to assign to each image before starting the parallelization.  

However, we realized that this doesn't really need to be precise.  An approximage estimate of the number of objects is fine.  Or really an upper bound.  And in fact, if it misses and accidentally under-estimates, it's not the end of the world for most use cases.  It would just end up in some objects having the same random number sequence for whatever they use that for.

So this PR adds the option for an input object, who is slated to inform about the number of objects to render, to return an estimate of this number using the method `getApproxNObjects`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -139,3 +139,9 @@ Changes from v2.4.5 to v2.4.6
 
 - Fixed drawImage to work correctly for method=fft when using photon_ops. (#1193)
 - Fixed the proxies used by config Input items to allow access to attributes. (#1195)
+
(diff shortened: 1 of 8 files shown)


jchiang87 (approved):
LGTM.  I also tried it out, and it works well.

PR #1206, opened 2023-01-04
Make input object using AtmosphericScreen work

rmjarvis:
In imSim, we have an input object that uses an AtmosphericScreen, and it fails when using nproc > 1.  This is basically because the trick that AtmosphericScreen uses for shared memory doesn't mix well with the Proxy objects we use for inputs when multiprocessing.

This PR adds a few extra options to the InputLoader class:

1. `use_proxy=False`.  This takes care of the immediate problem, and actually is sufficient when nimages=1, but nproc>1.  This tells the ProcessInput function not to use a proxy when loading an input object.
2. `worker_init, worker_initargs`.  These enable the workers to use the kind of shared memory that AtmosphericScreen requires.  For any class that uses AtmosphericScreen, these should be `galsim.phase_screeens.initWorker` and `galsim.phase_screens.initWorkerArgs` respectively.  But the functionality is general, so if some user input type wants to mimic this pattern, they can write corresponding functions and pass them in when registering the input type.

I also replaced the `InputProxy` function with a different implementation, which fixes a different problem we had in imSim with some input types not working correctly.

I checked that this works with the imSim AtmosphericPSF class, and it seems to work correctly with nfiles=1, nproc=2, which had been failing previously.  This is implemented in the imSim atm_input branch.


jmeyers314 (approved):
LGTM.  Thanks for figuring this out!

galsim/config/input.py
```diff
@@ -427,12 +442,39 @@ class InputLoader:
...
+
+    worker_init
+                One way to avoid using a proxy is to follow the shared memory model decsribed
+                in the doc string of `AtmosphericScreen`.  This mode uses a global dict, which
+                is initialized for each worker at the start of multi-processing.  To use this
+                method, one needs to provide an initialization function that gets run when
+                each worker starts.  [default: None]
+
+    worker_initargs
+                A function to provude the necessary arguments to worker_init. [default: None]
```
rmjarvis, line 463:
```suggestion
                A function to provide the necessary arguments to worker_init. [default: None]
```

Issue #1208, opened 2023-01-21
Moffat: maxK


jecampagne:
Hi, 
In the SBMoffat.cpp : `SBMoffat::SBMoffatImpl::maxK()` , I have found this comment (in the case of trunc=0, aka no truncation)
```C++
                // f(k) = 4 K(beta-1,k) (k/2)^beta / Gamma(beta-1)
                //
                // The asymptotic formula for K(beta-1,k) is
                //     K(beta-1,k) ~= sqrt(pi/(2k)) exp(-k)
                //
                // So f(k) becomes
                //
                // f(k) ~= 2 sqrt(pi) (k/2)^(beta-1/2) exp(-k) / Gamma(beta-1)
                //
                // Solve for f(k) = maxk_threshold
```
Well, the point is that the code reflect this comment, so it rises me a question
When you perform the 2D Fourier Transform of the Moffat Integrant (1+r^2)^(-\beta) you will find (here the Mathematica default Fourier Parameters)
```
(k/2)^(beta-1) K[beta-1, k] / Gamma[beta]
```
Of course this is exactly  the result of the Hankel Transform (with J_0, aka nu=0) of the Moffat Integrant. 

So as you notice besides some constant what is puzzling to me is the k power scaling in the comment ie k^beta  which is not what I find ie k^(beta-1).

Do I make a mistake/misunderstanding ?

rmjarvis:
Thanks Jean-Eric.  I finally had some time to look into this. 

First, you're quite correct that that bit of code is missing a power of k^-1.  However, even fixing that, this approximation, using the high k asymptotic behavior, seems not to be a very good one for finding maxk.  Especially for larger beta values, where that asymptote doesn't kick in until much higher k than what we need for maxk. 

So I ended up scrapping that and just doing a proper fit for maxk using the real formula with the Bessel K function.  It's plenty fast enough for this purpose, so there's not really a big need to do an approximate calculation.

I'll have a PR for this shortly.

PR #1210, opened 2023-02-02
Fix error in Moffat maxk calculation

rmjarvis:
@jecampagne pointed out an error in our maxk calculation for Moffat.  See details in #1208.  The upshot is that our Moffat FFT renderings were a bit more accurate and a bit slower than they needed to be for the nominal maxk_threshold.  This PR fixes the calculation to be only as accurate as requested.

### DIFF
--- a/include/galsim/SBMoffat.h
+++ b/include/galsim/SBMoffat.h
@@ -79,6 +79,7 @@ namespace galsim {
     protected:
 
         class SBMoffatImpl;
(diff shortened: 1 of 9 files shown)


jmeyers314 (approved):
I didn't really try to follow the math, but changes all look reasonable (and I have pretty high confidence in all the unit testing).

rmjarvis:
Thanks Josh!

PR #1212, opened 2023-02-13
Omp target offload

jamesp-epcc:
This is my GPU acceleration work on the sensor model, rebased against the current main branch. Obviously this is a very extensive change, so I'm open to discussing how it's implemented and making changes to better fit the existing code. There is a single unified code base for both GPU and CPU implementations; when offloading is disabled (which can be done via an environment variable or by building with a non-GPU-aware compiler), the offloaded regions act like OpenMP parallel regions, so the loops still take advantage of multiple CPU cores as before. A version of Clang with offloading support is required to build this for GPU.


jamesp-epcc:
I see the CI is failing. Apparently gcc doesn't like me mapping `this` via OpenMP. Unfortunately this is needed to build a working GPU version with Clang. I had thought it should be harmless to have the offload directives there even for CPU builds, but it might be necessary to wrap them in `#ifdefs` so that the compiler only sees them when actually building for GPU.

jamesp-epcc:
I've wrapped the GPU directives in `#ifdefs` and the checks now pass. The original code already had similar checks in place for the OpenMP parallel directives so it's not too big a departure from how we previously did things. I still need to investigate whether actually building and running the GPU version in CI is feasible.

src/Silicon.cpp
```diff
@@ -874,124 +1135,493 @@ namespace galsim {
...
+	    if (photonsHasAllocatedWavelengths) {
+		double lambda = photonsWavelength[i];
+
+		// perform abs_length_table lookup with linear interpolation
+		int tableIdx = int((lambda - 255.0) / 5.0);
+		double alpha = (lambda - ((float(tableIdx) * 5.0) + 255.0)) / 5.0;
+                if (tableIdx < 0) tableIdx = 0;
+		int tableIdx1 = tableIdx + 1;
+                if (tableIdx > 239) tableIdx = 239;
+                if (tableIdx1 > 239) tableIdx1 = 239;
```
rmjarvis, line 1211: Too many special numbers here.  255.0, 5.0, 239.  These all need to be computed from the input abs_length_table you get from python.  Don't assume these are necessarily the values forever.

src/Silicon.cpp
```diff
@@ -833,26 +781,339 @@ namespace galsim {
...
+	}
+        double* abs_length_table_data = _abs_length_table_GPU.data();
+        
+	int emptypolySize = _emptypoly.size();
+        _emptypolyGPU.resize(emptypolySize);
+	for (int i = 0; i < emptypolySize; i++) {
+	    _emptypolyGPU[i].x = _emptypoly[i].x;
+	    _emptypolyGPU[i].y = _emptypoly[i].y;
+	}
+        Position<double>* emptypolyData = _emptypolyGPU.data();
```
rmjarvis, line 856: Do we need this GPU stuff set when not targeting the GPU?  If so, the names are wrong.  If not, this section should be guarded if ifdefs.

src/Silicon.cpp
```diff
@@ -874,124 +1135,493 @@ namespace galsim {
...
+	// Get everything out of C++ classes and into arrays/structures suitable for GPU
+	// photons
+        // can't get pointers to internal data from a const reference
+        PhotonArray& photonsMutable = const_cast<PhotonArray&>(photons);        
+	double* photonsX = photonsMutable.getXArray();
+	double* photonsY = photonsMutable.getYArray();
+	double* photonsDXDZ = photonsMutable.getDXDZArray();
+	double* photonsDYDZ = photonsMutable.getDYDZArray();
+	double* photonsFlux = photonsMutable.getFluxArray();
+	double* photonsWavelength = photonsMutable.getWavelengthArray();
```
rmjarvis, line 1147: Do you really need pointers to mutable values?  If not, you should just add const versions of these in the PhotonArray class that returns `const double*` pointers rather than using const_cast.

src/Silicon.cpp
```diff
@@ -874,124 +1135,493 @@ namespace galsim {
...
-            double dz = calculateConversionDepth(photons, i, conversionDepthRandom[i-i1]);
-            if (photons.hasAllocatedAngles()) {
-                double dxdz = photons.getDXDZ(i);
-                double dydz = photons.getDYDZ(i);
-                double dz_pixel = dz * invPixelSize;
-                x0 += dxdz * dz_pixel; // dx in pixels
-                y0 += dydz * dz_pixel; // dy in pixels
-            }
-            xdbg<<" => "<<x0<<','<<y0;
-            // This is the reverse of depth. zconv is how far above the substrate the e- converts.
```
rmjarvis, line 897: You got rid of most of these inline comments in your refactoring.  Please add them back.

include/galsim/Silicon.h
```diff
@@ -249,6 +256,12 @@ namespace galsim
 
         void updatePixelBounds(int nx, int ny, size_t k);
 
+        void updatePixelBoundsGPU(int nx, int ny, size_t k,
+                                  Bounds<double>* pixelInnerBoundsData,
+                                  Bounds<double>* pixelOuterBoundsData,
+                                  Position<float>* horizontalBoundaryPointsData,
+                                  Position<float>* verticalBoundaryPointsData);
```
rmjarvis, line 263: Let's not have two ways to do this please.  Rewrite the existing updatePixelBounds to work with the GPU.  Don't repeat everything in a second function.

src/Silicon.cpp
```diff
@@ -874,124 +1135,493 @@ namespace galsim {
...
+#else
+#pragma omp target teams distribute parallel for
 #endif
-        return addedFlux;
-    }
-
-    template <typename T>
-    void Silicon::update(ImageView<T> target)
-    {
-        updatePixelDistortions(_delta.view());
```
rmjarvis, line 990: Why did you move the entirety of `updatePixelDistortions` into `update`?  Can't you just convert that function the way you did with updatePixelBounds to work on a GPU?  It makes it very hard to review this, since I can't actually tell what are the substantive changes you needed to make to work on the GPU.

src/Silicon.cpp
```diff
@@ -874,124 +1135,493 @@ namespace galsim {
...
-        target += _delta;
-        _delta.setZero();
+#endif
+        for (int i = 0; i < imageDataSize; i++) {
+            targetData[i] += deltaData[i];
+            deltaData[i] = 0.0;
+	}
```
rmjarvis, line 1622: I assume this should be a call to `addDelta`, right?  Looks like an oversight in the previous code, but it's more important now that addDelta isn't just a one-liner.

rmjarvis (changes requested):
This PR is pretty hard to review right now.  The src/Silicon.cpp file in particular.

You moved some of the functions around, which means they don't show up as a useful diff currently.  There is a full function removed, and then a very similar function added in a different location.  You also removed some inline comments, which makes git think blocks of code that are otherwise identical are actually different.  There are also some new functions, which are just old functions with some slight changes to data structures, not actually (I think!) any new code.

Please reorganize this so that the Silicon.cpp file show on github as a reasonable diff, so I can review what the real changes are.  Thanks!

jamesp-epcc:
I have addressed Mike's feedback and rebased against the current main branch.
* The function re-ordering, the whitespace changes and the loss of comments were an artefact of me adding GPU versions of the functions initially alongside the originals, then later renaming them and removing the originals, but hopefully all those issues are resolved now.
* The values from `_abs_length_table` are now taken from the table passed in from the Python layer rather than being hardcoded (this was meant to be a temporary change but I forgot about it, apologies).
* Const getter methods have been added to `PhotonArray` instead of `const_cast`ing it.
* The separate CPU version of `updatePixelDistortions` has now been removed. By making a few changes I was able to use the GPU-enabled version for all purposes instead.
* The pixel distortion update has been moved back out of the `update` method, as it was originally. `addDelta` is not being used in `update`, but I have added a comment explaining why.

src/Silicon.cpp
```diff
@@ -959,39 +1322,210 @@ namespace galsim {
...
+                const int xoff[9] = {0,1,1,0,-1,-1,-1,0,1}; // Displacements to neighboring pixels
+                const int yoff[9] = {0,0,1,1,1,0,-1,-1,-1}; // Displacements to neighboring pixels
+                int n = (randomArray[(i-i1) * 4 + 2] > 0.5) ? 0 : step;
                 ix = ix + xoff[n];
                 iy = iy + yoff[n];
             }
 
-            if (b.includes(ix,iy)) {
+            if ((ix >= b.getXMin()) && (ix <= b.getXMax()) &&
+                (iy >= b.getYMin()) && (iy <= b.getYMax())) {
```
rmjarvis, line 1345: Another place where you replaced includes.  Why can't you just use includes here?

rmjarvis (changes requested):

include/galsim/Silicon.h
```diff
@@ -265,6 +278,19 @@ namespace galsim
         Table _abs_length_table;
         bool _transpose;
         ImageAlloc<double> _delta;
+        std::shared_ptr<bool> _changed;
```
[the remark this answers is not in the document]
jamesp-epcc: I moved _changed to be a member variable rather than local to avoid the overhead of allocating it on the GPU every time update is called. The reason it's now a shared_ptr to a bare array instead of a vector is that bool vectors have an optimised implementation that packs multiple bools into each byte, so you can't get a pointer to the raw data, which the GPU requires.

I agree that std::min and std::max would be more readable, but surprisingly the GPU doesn't support them (I was using them initially but they caused weird bugs that took a long time to track down). We could implement custom, GPU-friendly min and max functions rather than writing out the algorithm though.

src/Silicon.cpp
```diff
@@ -310,44 +339,52 @@ namespace galsim {
...
                 }
             }
-
+            
             // update changed array
             if (change) {
-                if (y < ny) changed[(x * ny) + y] = true; // pixel above
-                if (y > 0)  changed[(x * ny) + (y - 1)] = true; // pixel below
+                if (y < ny) changedData[(x * ny) + y] = true; // pixel above
+                if (y > 0)  changedData[(x * ny) + (y - 1)] = true; // pixel below
```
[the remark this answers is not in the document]
jamesp-epcc: _changed.get() will return the raw bool pointer from inside the shared pointer, so this works as intended. See above for explanation of why this is no longer a vector.

src/Silicon.cpp
```diff
@@ -514,41 +565,31 @@ namespace galsim {
         // photon within the pixel, with (0,0) in the lower left
 
         // If test pixel is off the image, return false.  (Avoids seg faults!)
-        if (!target.getBounds().includes(Position<int>(ix,iy))) {
+        if ((ix < targetBounds.getXMin()) || (ix > targetBounds.getXMax()) ||
+            (iy < targetBounds.getYMin()) || (iy > targetBounds.getYMax())) {
```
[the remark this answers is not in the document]
jamesp-epcc: I've got rid of the old CPU version of this and now using the GPU version everywhere. I also tested and found that `includes` does work on GPU, so I've switched back to that as suggested.

src/Silicon.cpp
```diff
@@ -825,89 +954,311 @@ namespace galsim {
...
+        int deltaStep = _delta.getStep();
+        int deltaStride = _delta.getStride();
+
+        int emptypolySize = _emptypoly.size();
+
+        double* deltaData = _delta.getData();
+        Bounds<double>* pixelInnerBoundsData = _pixelInnerBounds.data();
+        Bounds<double>* pixelOuterBoundsData = _pixelOuterBounds.data();
+        Position<float>* horizontalBoundaryPointsData = _horizontalBoundaryPoints.data();
+        Position<float>* verticalBoundaryPointsData = _verticalBoundaryPoints.data();
```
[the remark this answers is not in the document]
jamesp-epcc: That's correct - I tested accessing the vectors directly on GPU but it fails with an invalid memory access. I think this is because mapping the vector to the GPU only maps the vector structure and not the underlying data array (which we can't map because it's private to the vector object). So I've added a comment to explain this.

src/Silicon.cpp
```diff
@@ -959,39 +1322,210 @@ namespace galsim {
...
+                searchNeighbors(*this, ix, iy, x, y, zconv, b, step, emptyPolySize,
+                                pixelInnerBoundsData, pixelOuterBoundsData,
+                                horizontalBoundaryPointsData,
+                                verticalBoundaryPointsData, emptyPolyData);
                 set_verbose(1);
 #endif
-                int n = (pixelNotFoundRandom[i-i1] > 0.5) ? 0 : step;
+                const int xoff[9] = {0,1,1,0,-1,-1,-1,0,1}; // Displacements to neighboring pixels
+                const int yoff[9] = {0,0,1,1,1,0,-1,-1,-1}; // Displacements to neighboring pixels
+                int n = (randomArray[(i-i1) * 4 + 2] > 0.5) ? 0 : step;
```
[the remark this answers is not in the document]
jamesp-epcc: I've added comments to explain these.

src/Silicon.cpp
```diff
@@ -825,89 +954,311 @@ namespace galsim {
...
+        }
+#endif
+    }
+    
     template <typename T>
     void Silicon::subtractDelta(ImageView<T> target)
     {
-        target -= _delta;
+        // perform update on GPU
+        int imageDataSize = (_delta.getXMax() - _delta.getXMin()) * _delta.getStep() + (_delta.getYMax() - _delta.getYMin()) * _delta.getStride() + 1;
```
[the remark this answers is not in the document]
jamesp-epcc: Yes, that is a cleaner way of computing it. Thanks, have updated all the functions that used this.

src/Silicon.cpp
```diff
@@ -825,70 +935,273 @@ namespace galsim {
...
+        int emptypolySize = _emptypoly.size();
+        _emptypolyGPU.resize(emptypolySize);
+        for (int i = 0; i < emptypolySize; i++) {
+            _emptypolyGPU[i].x = _emptypoly[i].x;
+            _emptypolyGPU[i].y = _emptypoly[i].y;
+        }
+        Position<double>* emptypolyData = _emptypolyGPU.data();
+        
+        int nxny = nx * ny;
+        _changed = std::shared_ptr<bool>(new bool[nxny]);
```
rmjarvis, line 984: This only plausibly works with C++17, not C++11, which is our nominal minimum requirement.  And even in C++17, this isn't quite right.  It should be `std::shared_ptr<bool[]>`.  Otherwise it won't delete properly.  cf. https://kezunlin.me/post/b82753fc/

My recommendation is to use `default_delete<bool[]>` as the simplest portable version.  Or possibly switch to `unique_ptr`, since supposedly it has overloads for array deletion.  But you still need to use `bool[]` for the template type.

jamesp-epcc: Good point, I missed that subtlety. I suspect it worked fine in practice because `bool` is a primitive type and doesn't need anything special for array deletion, but I've replaced it with `std::unique_ptr<bool[]>` to be safe.

src/Silicon.cpp
```diff
@@ -959,39 +1287,210 @@ namespace galsim {
...
-#endif
+        
         return addedFlux;
     }
 
+    void Silicon::updatePixelBounds(int nx, int ny, size_t k,
+                                    Bounds<double>* pixelInnerBoundsData,
+                                    Bounds<double>* pixelOuterBoundsData,
+                                    Position<float>* horizontalBoundaryPointsData,
+                                    Position<float>* verticalBoundaryPointsData)
```
rmjarvis, line 1330: Please move this up to where this function used to be so it shows up as a diff, rather than an whole new function (and a deleted one above).

rmjarvis (changes requested):
OK, I think I'm mostly happy with this now.  Just a couple minor comments.  

But I do still want Josh to take a look at some of the OpenMP GPU directives to see if he has any comments there.

src/Silicon.cpp
```diff
@@ -217,51 +218,163 @@ namespace galsim {
...
-                             Position<double> p = pt;
-                             if (rhs) p.x += 1.0;
-                             if (top) p.y += 1.0;
-                             _pixelOuterBounds[k] += p;
-                             });
-
-        Position<double> center = _pixelOuterBounds[k].center();
+        // initialise outer bounds
+        double obxmin = 1000000.0, obxmax = -1000000.0;
+        double obymin = 1000000.0, obymax = -1000000.0;
```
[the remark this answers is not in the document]
jamesp-epcc: I did some more testing and it turns out the += operator does work, so I've switched back to that, which is a big improvement. (The original function wouldn't run on the GPU but I think that was just due to the callback in iteratePixelBoundary not being supported).

rmjarvis (approved):

src/Silicon.cpp
```diff
@@ -275,31 +375,52 @@ namespace galsim {
...
             // Calculate which pixel we are currently below
             int x = p % nx;
             int y = p / nx;
 
             // Loop over rectangle of pixels that could affect this row of points
-            int polyi1 = std::max(x - _qDist, 0);
-            int polyi2 = std::min(x + _qDist, nx - 1);
+            // std::min and std::max are causing this loop to crash, even though
+            // the same functions run fine in the accumulate loop.
+            int polyi1 = x - _qDist;
```
jmeyers314, line 409: I think the versions with min() and max() are clearer.  Would a macro a la https://stackoverflow.com/a/3437484 work?  (And same comment throughout this file...)

src/Silicon.cpp
```diff
@@ -825,70 +1082,252 @@ namespace galsim {
...
+        Position<float>* verticalDistortionsData = _verticalDistortions.data();
+
+        double* abs_length_table_data = _abs_length_table_GPU.data();
+        
+        Bounds<int> b = _delta.getBounds();
+        const int nx = b.getXMax() - b.getXMin() + 1;
+        const int ny = b.getYMax() - b.getYMin() + 1;
+        int nxny = nx * ny;
+
+        int pixelInnerBoundsSize = _pixelInnerBounds.size();
```
jmeyers314, line 1150: Maybe just call this `pixelBoundsSize` since it applies for both `_pixelInnerBounds` and `_pixelOuterBounds`.  (I was concerned for a moment reading `_pixelOuterBounds[0:pixelInnerBoundsSize]` below)

jmeyers314 (approved):
Looks good to me.  I left a couple small comments.

src/Silicon.cpp
jamesp-epcc: Good point, I've renamed it to just `pixelBoundsSize`.

PR #1213, opened 2023-03-22
Fix LoggerWrapper inefficiency

rmjarvis:
We have a class in GalSim config called LoggerWrapper, which is intended to avoid communicating with the main logger via the proxy when we aren't actually going to output anything.  E.g. logger.debug statements when the logging level is only INFO or WARNING.  This is mostly a big improvement.

However, @beckermr and @jchiang87 have each recently found huge inefficiencies when running on Perlmutter with ~128 processes.  I tracked it down to when we make a new LoggerWrapper instance and call `logger.getEffectiveLevel()`.  This had always been going back to the original proxy logger instance even when we already have a wrapped one.  And for lots of processes, this can end up clogging the inter-process communication completely unnecessarily. 

This PR makes the fairly trivial change that when we're re-wrapping something that is already wrapped, it doesn't call `getEffectiveLevel`.  It just gets the already received value stored as an attribute.  This seems to resolve the blockage that Matt (and I presume Jim) was seeing.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -152,7 +152,7 @@ Changes from v2.4.6 to v2.4.7
 - Only use proxies for input objects if not yet in a multiprocessing context. This had been
   causing significant slow downs in some imsim runs. (#1206)
 
-Changes from v2.4.6 to v2.4.8
+Changes from v2.4.7 to v2.4.8
 =============================
 
-- Fix a slow-down in multiprocessing especially when running very many (>10) processes.
+- Fix a slow-down in multiprocessing especially when running very many (>10) processes. (#1213)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -151,3 +151,8 @@ Changes from v2.4.6 to v2.4.7
 - Add options to InputLoader to make inputs with AtmosphericScreens work properly. (#1206)
 - Only use proxies for input objects if not yet in a multiprocessing context. This had been
   causing significant slow downs in some imsim runs. (#1206)
+
+Changes from v2.4.6 to v2.4.8
+=============================
+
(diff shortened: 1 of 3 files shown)


beckermr (approved):
I don't think there's a good test for this so LGTM!

CHANGELOG.rst
```diff
@@ -151,3 +151,8 @@ Changes from v2.4.6 to v2.4.7
 - Add options to InputLoader to make inputs with AtmosphericScreens work properly. (#1206)
 - Only use proxies for input objects if not yet in a multiprocessing context. This had been
   causing significant slow downs in some imsim runs. (#1206)
+
+Changes from v2.4.6 to v2.4.8
```
rmjarvis, line 155:
```suggestion
Changes from v2.4.7 to v2.4.8
```

CHANGELOG.rst
```diff
@@ -151,3 +151,8 @@ Changes from v2.4.6 to v2.4.7
...
+=============================
+
+- Fix a slow-down in multiprocessing especially when running very many (>10) processes.
```
rmjarvis, line 158:
```suggestion
- Fix a slow-down in multiprocessing especially when running very many (>10) processes. (#1213)
```

beckermr:
For posterity, here is the timing of a simple [test](https://github.com/beckermr/misc/tree/main/work/galsim_par_test) on perlmutter:

```
timing 128 processes...
128 5.450411558151245
timing 128 processes...
128 5.5901055335998535
timing 64 processes...
64 5.405612945556641
timing 32 processes...
32 7.624440431594849
timing 16 processes...
16 11.752846002578735
timing 8 processes...
8 20.898335933685303
timing 4 processes...
4 39.02970886230469
timing 2 processes...
2 81.31851077079773
```

PR #1216, opened 2023-05-16
Fix possible crash in PhotonArray.makeFromImage if image has negative pixels.

rmjarvis:
@g-braeunlich ran into a bug in the PhotonArray `makeFromImage` function where it would sometimes seg fault.  It turned out that the calculation for how many photons would be needed was wrong when there were negative-flux pixels, so it wasn't allocating enough space and the arrays were being overrun.  This PR fixes that bug.  

It also happened to be the case that that function shouldn't even have been run for what Geri was doing.  But he had the photon_ops set to [], rather than None.  Those should logically mean the same thing, but GalSim assumed that since photon_ops wasn't None, it needed photons.  Now photon_ops = () or [] means the same thing as None in the code.

### DIFF
--- a/galsim/gsobject.py
+++ b/galsim/gsobject.py
@@ -1700,7 +1700,8 @@ class GSObject:
                     method=method, sensor=sensor, poisson_flux=poisson_flux)
 
         # If using photon ops with fft, then need a sensor.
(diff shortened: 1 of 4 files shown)


jmeyers314 (approved):
LGTM.  Nothing to add.

PR #1219, opened 2023-06-06
Add options to convert HSM moments to sky coordinates

rmjarvis:
This PR involves code that I've been copying around to various places (e.g. it is in the piff.Star.hsm method) knowing each time that I should really put it into GalSim directly.  But it was always more expedient to just copy it.  So today, as I was about to copy it yet again, I decided to do it right and add it here.

The HSM moments are natively measured in pixel coordinates.  Indeed the back end code doesn't have any knowledge of the WCS.  However, it's fairly straightforward to convert these (x,y) moments in to (u,v) moments using the jacobian of the WCS.  This PR adds two ways the user can do this:

`ShapeData.applyWCS(wcs)` converts a ShapeData instance into one where the size and shape values are based on the u,v moments rather than x,y.
`image.FindAdaptiveMom(use_sky_coords=True)` does this for you using the wcs that is in the image.

### DIFF
--- a/galsim/hsm.py
+++ b/galsim/hsm.py
@@ -687,7 +687,8 @@ def EstimateShear(gal_image, PSF_image, weight=None, badpix=None, sky_var=0.0,
             return ShapeData(error_message = str(err))
 
 def FindAdaptiveMom(object_image, weight=None, badpix=None, guess_sig=5.0, precision=1.0e-6,
-                    guess_centroid=None, strict=True, round_moments=False, hsmparams=None):
+                    guess_centroid=None, strict=True, round_moments=False, hsmparams=None,
+                    use_sky_coords=False):
     """Measure adaptive moments of an object.
 
     This method estimates the best-fit elliptical Gaussian to the object (see Hirata & Seljak 2003
@@ -773,6 +774,10 @@ def FindAdaptiveMom(object_image, weight=None, badpix=None, guess_sig=5.0, preci
         hsmparams:          The hsmparams keyword can be used to change the settings used by
                             FindAdaptiveMom when estimating moments; see `HSMParams` documentation
                             for more information. [default: None]
+        use_sky_coords:     Whether to convert the measured moments to sky_coordinates.
+                            Setting this to true is equivalent to running
+                            ``applyWCS(object_image.wcs, image_pos=object_image.true_center)``
+                            on the result.  [default: False]
 
     Returns:
         a `ShapeData` object containing the results of moment measurement.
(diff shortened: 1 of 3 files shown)


galsim/hsm.py
```diff
@@ -220,6 +220,82 @@ def __getstate__(self):
...
+        moments to (u,v) coordinates, where +v is towards North and +u is towards West.
+
+        The values that are different from the original are:
+
+        moments_sigma is changed to be in units of arcsec.
+        observed_shape is changed to be in (u,v) coordinates.
+        moments_centroid is changed to be in (u,v) coordinates in units of arcsec.
+        moments_rho4 is changed to be in units of arcsec^4.
+
+        .. note::
```
jmeyers314, line 238: I think I'd rather invalidate (set to NaN or None?) the untouched parameters.  Having an object where `.moments_sigma` is in arcsec but `.psf_sigma` is in pixels or `.moments_shape` and `.psf_shape` are in different coordinate systems feels error prone.

jmeyers314 (changes requested):
Thanks for this.  I also have frequently gone back to copy-paste the piff solution :)

Just one suggestion.

galsim/hsm.py
rmjarvis: Sure.  Sounds reasonable to me. I changed it to reset any EstimateShear values to defaults.

jmeyers314 (approved):

beckermr (changes requested):
Tests and code look good. There are some unused variables in the tests that should be removed.

rmjarvis:
Thanks Matt.  They were relics of a copy/paste I forgot to clean up.

beckermr (approved):

rmandelb (changes requested):
Hi Mike - thanks for checking, I noticed these reviews going by.  I have a change to suggest but it can't go on the diff because it's in bits of the documentation that haven't changed.  There are a bunch of places in the dox that were written in way that is incompatible with the new code, so they should be updated.  For example (going from RTD on https://galsim-developers.github.io/GalSim/_build/html/hsm.html):

- "This routine assumes that (at least locally) the WCS can be approximated as a [PixelScale](https://galsim-developers.github.io/GalSim/_build/html/wcs.html#galsim.PixelScale), with no distortion or non-trivial remapping. Any non-trivial WCS gets completely ignored."  This has to be modified.
- The "HSM output" section gives units that are correct for the current code but not for if you've used the `use_sky_coords` kwarg, since that changes the units - might need to update here.
- The "HSM parameters" section describes default parameter values, some of which do have units if I recall correctly (like pixels of centroid shift).  If i understand correctly, those should still be in pixels even if using `use_sky_coords` because underneath the hood it's still doing the calculations that way and then converting at the end, but it might be worth saying that explicitly?

I might have missed something, so it might be worth a more careful read.

I also wondered if there are any assumptions being made about the WCS, like it being locally affine or something?  Constant across the object?  If so, either would be fine for nearly all reasonable cases, but it would be good to state explicitly.

rmandelb:
Also, thanks for your work on this feature; it's a nice addition!

rmandelb (approved):

PR #1220, opened 2023-06-09
Add list (and dict, LookupTable) as a registered type for Eval

rmjarvis:
@jmeyers314  discovered a bug in the config parsing that prevented Eval types from being used to generate a list object, when parsed in the normal way.  The oversight was that list (and dict and LookupTable) were not registered as valid types associated with the Eval type.  

The workaround is to use None for the value_type rather than list, but this is inelegant and less type-safe.

### DIFF
--- a/tests/test_config_value.py
+++ b/tests/test_config_value.py
@@ -1608,7 +1608,9 @@ def test_table_value():
                               ]
                   },
         'cur1' : { 'type' : 'Current', 'key' : 'file3' },
-        'cur2' : { 'type' : 'Current', 'key' : 'list1.items.1' }
+        'cur2' : { 'type' : 'Current', 'key' : 'list1.items.1' },
+        'eval1' : '$galsim.LookupTable([0,1,2,3], [0,10,10,0], interpolant="linear")',
+        'eval2' : '$galsim.LookupTable.from_file(@file1.file_name)',
     }
 
     # Test direct values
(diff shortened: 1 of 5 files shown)


tests/test_config_value.py
```diff
@@ -1700,6 +1700,12 @@ def test_eval():
         'bad4' : { 'type' : 'Eval', 'str' : 'np.exp(-0.5 * q**2)', 'fx' : 1.8 },
         'bad5' : { 'type' : 'Eval', 'eval_str' : 'np.exp(-0.5 * x**2)', 'fx' : 1.8 },
 
+        # Check that a list can be made using Eval
+        'list0' : [0,1,2,3,4,5],
+        'list1' : '$np.arange(6)',
+        'list2' : (0,1,2,3,4,5),
+        'list3' : '$(0,1,2,3,4,5)',
+
```
jmeyers314, line 1708: Looks good.  Do you want to test `Eval`-ing a LookupTable and dict too, since you just added those?

jmeyers314 (approved):
LGTM.

tests/test_config_value.py
rmjarvis: Thanks.  This was a good idea.  These were fine as implemented, but it turned up other places where some config code assumed that all dicts would have str keys, which isn't necessarily the case for objects that are intended to be a simple dict, rather than a construction directive.

PR #1221, opened 2023-06-19
Double Zernikes

jmeyers314:
Adds support for Double Zernikes, which are Cartesian products of single Zernikes.  These are useful for modeling, e.g., the combined field and pupil dependence of an optical wavefront.  I added most of the functionality currently available for single Zernikes to the DZs, which is enough that one can quickly analytically produce a map of optical PSF second moments across the field-of-view from DZ coefficients. (see the new `test_dz_to_T` unit test).  

Along the way, I also:
- Changed the technique for converting from Cartesian polynomial coefficients to Noll coefficients.  This had previously been a least-squares fit of precomputed Cartesian coefficients of each Noll coefficient to target Cartesian coefficients.  The new technique is to take advantage of Zernike orthonormality and integrate over an annulus to find each coefficient.  While for small maximum Noll index, the computation is slightly slower now, for large Noll indices there is a dramatic ~10x (or more) improvement.
- To support the above, added Gaussian Quadrature points and weights for integrating over an annulus.
- Added a method `zernikeGradBases` similar to the existing `zernikeBasis` that returns basis vectors for the x and y gradients of Zernike terms.  (I implemented this previously directly in the Rubin wavefront estimation pipeline [here](https://github.com/lsst-ts/ts_wep/blob/develop/python/lsst/ts/wep/cwfs/algorithm.py#L741-L757), but it's nice to have an officially supported API in GalSim for this).
- Silenced a scipy interp2d deprecation warning.


jmeyers314:
Hmm...  Looks like I have a Heisenbug here.  I _was_ getting failures in identically evaluating DoubleZernikes before/after pickling [here](https://github.com/GalSim-developers/GalSim/actions/runs/5314980503/jobs/9622821848#step:15:756), but after adding some additional tests [here](https://github.com/GalSim-developers/GalSim/pull/1221/files#diff-c81764b312924e9e848b9740ed42e0d62f7cb9073b4cc1dac155c31c32f82616R793-R799), the originally failing `do_pickle` line is now passing.  Best guess: something is slightly inconsistent in Horner2d (at ~1e-15 level) depending on whether SSE2 is used or not (which depends on memory alignment)?

galsim/zernike.py
```diff
@@ -18,8 +18,9 @@
 
 import numpy as np
 
-from .utilities import LRU_Cache, binomial, horner2d, nCr, lazy_property
-from .errors import GalSimValueError, GalSimRangeError
+from .utilities import LRU_Cache, binomial, horner2d, horner4d, nCr, lazy_property
+from .integ import _gq_annulus_points
```
rmjarvis, line 22: Why the private function rather than the `gg_annulus` function you made?  If you feel the need to use this undocumented function instead, that strikes me as a sign that the front end function with nice documentation wasn't the right API.  Probably worth rethinking this a bit so that this file can use the documented function.  Or possibly just promote this function to a non-leading_underscore function with better docs.

galsim/zernike.py
```diff
@@ -876,3 +912,611 @@ def zernikeBasis(jmax, x, y, R_outer=1.0, R_inner=0.0):
...
+
+    The double Zernike's therefore satisfy the orthonormality condition:
+
+    .. math::
+        \int_{annuli} DZ_(k, j) DZ_(k', j') = A1 A2 \delta_{k, k'} \delta_{j, j'}
+
+    where A1 and A2 are the areas of the two annuli.
+
+    Double Zernikes series are useful for representing the field and pupil
+    dependent wavefronts of a telescope.  Typically, the first index (the `k`
```
rmjarvis, line 975: Rather than "typically", something like "We adopt the typical convention that..."

galsim/zernike.py
```diff
@@ -499,6 +504,17 @@ def __init__(self, coef, R_outer=1.0, R_inner=0.0):
         self.R_outer = float(R_outer)
         self.R_inner = float(R_inner)
 
+    @staticmethod
```
rmjarvis, line 507: Factory methods are more typically classmethod.  I don't think there is much functional difference between them, at least in the absence of a class hierarchy.  But I believe it's more pythonic to use a classmethod here.  (And below for _from_uvxy)

rmjarvis (approved):
This looks good Josh.  I didn't try to follow all of the math in detail, but I think I get the main points.  test_dz_to_T was helpful for understanding the point of this.  I assume we can use this class in Piff in the Optical model to handle some of the bookkeeping more easily, but I din't try to think about how it would get used there.

galsim/zernike.py
jmeyers314: Decided to make this one public (and add docs).

jmeyers314:
> I assume we can use this class in Piff in the Optical model to handle some of the bookkeeping more easily

Certainly this could be useful for bookkeeping.  I'm not sure if the analytic size and ellipticity predictions will be useful or not; they technically depend on a full pupil annulus (no vignetting).  It's possible they're still good approximations or heuristics though even when there is a small bit of vignetting.

PR #1222, opened 2023-06-19
Support GPU offloading when building with gcc

jamesp-epcc:
This PR adds support for GPU offloading when building with gcc, in addition to the previous clang support. Tested on the Cirrus cluster in Edinburgh with both a version of gcc that supports offloading and one that doesn't.

I also removed the `-stdlib=libc++` stuff from the GPU detection function. I realised that was pointless because neither of the GPU compilers have that flag in the first place.


rmjarvis (approved):
I guess this is fine.  I don't have any real way to evaluate this, so let's go with it, and if there are reported errors, we can try to deal with them then.

PR #1223, opened 2023-06-21
Enable Eval within List in config

jmeyers314:
I was trying to figure out how to simulate all Rubin sensors in ImSim except the guiders and came up with this snippet:

```yaml
  det_num:
    type: List
    items: $[i for i in range(205) if i not in [189, 190, 193, 194, 197, 198, 201, 202]]  # no guiders
```

The `Eval` within the `List` currently produces `galsim.errors.GalSimConfigError: items entry for type=List is not list-like.`, but is fixed on this PR.


rmjarvis (approved):
LGTM

PR #1225, opened 2023-07-14
Chromatic seeing when photon shooting

jmeyers314:
Adds chromatic seeing to atmospheric code for use when using geometric photon shooting (I believe similar behavior is already emergent when drawing with Fourier techniques).

To implement, I added a `wavelength` argument to the various `wavefront_gradient` methods.  In retrospect, I maybe should only have added the extra argument to the private version `_wavefront_gradient`, since as a purely geometric object, the real wavefront gradient isn't wavelength dependent.  Adding wavelength dependence here is kind of a hack to ensure we get the lambda^-beta chromatic dependence (beta ~ 0.2-0.4) expected from the fully diffractive calculation.  But I could lean either way on this one so curious for feedback.

### DIFF
--- a/galsim/phase_psf.py
+++ b/galsim/phase_psf.py
@@ -914,7 +914,7 @@ class PhaseScreenList:
         else:
             return self._layers[0].wavefront(u, v, t, theta)
 
-    def wavefront_gradient(self, u, v, t, theta=(0.0*radians, 0.0*radians)):
+    def wavefront_gradient(self, u, v, t, w=None, theta=(0.0*radians, 0.0*radians)):
         """ Compute cumulative wavefront gradient due to all phase screens in `PhaseScreenList`.
 
         Parameters:
@@ -927,6 +927,7 @@ class PhaseScreenList:
                     will be used for all u, v.  If scalar, then the size will be broadcast up to
                     match that of u and v.  If iterable, then the shape must match the shapes of
                     u and v.
+            w:      Wavelengths (in nanometers) at which to evaluate wavefront gradient.
             theta:  Field angle at which to evaluate wavefront, as a 2-tuple of `galsim.Angle`
                     instances. [default: (0.0*galsim.arcmin, 0.0*galsim.arcmin)]
                     Only a single theta is permitted.
@@ -935,9 +936,9 @@ class PhaseScreenList:
             Arrays dWdu and dWdv of wavefront lag or lead gradient in nm/m.
         """
         if len(self._layers) > 1:
-            return np.sum([layer.wavefront_gradient(u, v, t, theta) for layer in self], axis=0)
+            return np.sum([layer.wavefront_gradient(u, v, t, w, theta) for layer in self], axis=0)
         else:
-            return self._layers[0].wavefront_gradient(u, v, t, theta)
+            return self._layers[0].wavefront_gradient(u, v, t, w, theta)
 
     def _wavefront(self, u, v, t, theta):
         if len(self._layers) > 1:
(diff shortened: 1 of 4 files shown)


galsim/phase_screens.py
```diff
@@ -221,6 +221,11 @@ def work(i, atm):
                             significantly faster than computing PSFs with non-frozen-flow
                             atmospheres.  If ``alpha`` != 1.0, then it is required that a
                             ``time_step`` is also specified.  [default: 1.0]
+        seeing_exp:         Power law index for wavelength dependent seeing when using geometric
```
rmjarvis, line 224: We had talked about deriving this value from L0.  Is that possible?  If so, it would make it easier for users who may be using a VonKarman profile and want to get this to match up correctly.

galsim/phase_psf.py
```diff
@@ -1579,6 +1580,10 @@ def _shoot(self, photons, rng):
         u = photons.pupil_u
         v = photons.pupil_v
         t = photons.time
+        if photons.hasAllocatedWavelengths():
+            w = photons.wavelength
```
rmjarvis, line 1584: Need a test that uses this feature.  (Which is the feature request that triggered this PR...)  Maybe one that compares this implementation to one using ChromaticAtmosphere?

galsim/phase_screens.py
jmeyers314: Yeah, my memory for how these are related was faulty.  There's a whole slide deck [here](https://docs.google.com/presentation/d/1dzB7xHbTjP1Q4Crbw64JSBjJHxzfnBSHlbU9miCT6WY/edit#slide=id.g17ffb075ae_0_59) showing some attempts to infer a relation, but I'm not confident this can or should be used directly.  Safer to just let the user specify a value I think.

galsim/phase_psf.py
jmeyers314: Yeah, I had naïvely thought that adding a `WavelengthSampler` photon op to `drawImage` would trigger this, but I guess those ops only get executed after the base object is drawn.  Looking into how to actually trigger this now...

jmeyers314: It's a little bit of a design conundrum even: `PhaseScreenPSF` is currently a `GSObject` (and not a `ChromaticObject`).  That's perfectly fine for the original usage, and would continue to be fine if `seeing_exp = 0`.  But the profile _is_ chromatic, of course, if `seeing_exp != 0`.  Duck-typing might work here, but may require changing some of the code around https://github.com/GalSim-developers/GalSim/blob/main/galsim/gsobject.py#L419.  For instance, the chromatic version of this profile is not separable.

rmjarvis: Should we just be using ChromaticAtmosphere then?

jmeyers314: >Should we just be using ChromaticAtmosphere then?

Probably not; ChromaticAtmosphere is something quite different.  It just applies some chromatic transformations to a base object (e.g. Moffat, Kolmogorov, etc) consistent with DCR and chromatic seeing, but doesn't do anything with phase screens.

rmjarvis: Can't the base object of ChromaticAtmosphere be a PhaseScreenPSF?

jmeyers314: Hmmm... Maybe?  Let me try.

jmeyers314: Okay, that does appear to be close.  The main difference is that ChromaticAtmosphere applies its shift to the entire base object at once.  So we couldn't specify an exponent per screen.  This is probably good enough for the atmospheric screens, but won't be correct for any OpticalScreens and I don't think will be correct for the second_kick component of the atmosphere (which has a limiting case of an Airy, e.g.).

rmjarvis: Can we do something a little custom in ChromaticAtmosphere to make it work right when the base is a PhaseScreenPSF?

jmeyers314: Yeah.  That or tweak https://github.com/LSSTDESC/imSim/blob/main/imsim/atmPSF.py#L83 to build 3 separate things (atm+opt+2kick) instead of embedding them all inside a single PhaseScreenPSF.  Then we can use ChromaticAtmosphere to only add chromatic seeing to the right part.

rmjarvis: The latter is probably simpler...

rmjarvis:
I think we're abandoning this now, right?  Since it didn't really work the way we expected.

PR #1227, opened 2023-07-16
Don't abort if wavelengths are outside valid range for absorption

rmjarvis:
Bekah Polen ran into an edge case where some photons had bad wavelengths being accumulated via the Silicon sensor.  They also had flux=0, so the other details didn't matter, but it was raising an exception that the wavelengths weren't valid for the absorption depth table.

This PR fixes this by taking the absorption depth to be the value for the minimum or maximum valid wavelength whenever the wavelength is too blue or too red, respectively.

We could probably add an optimization to check for flux=0 sooner as an optimization, but I didn't do that here.


jchiang87 (approved):
lgtm

PR #1230, opened 2023-07-21
Add support for building GPU-enabled GalSim using NVIDIA compiler

jamesp-epcc:
This PR adds support for building with the NVIDIA compiler, nvc++. This compiler doesn't support one of the warnings commonly enabled in the other compilers so we have to remove it. It also formats its version information a little differently from gcc and clang.

This also adds a missing macro definition for building for GPU with gcc.

### DIFF
--- a/src/Silicon.cpp
+++ b/src/Silicon.cpp
@@ -1372,7 +1372,9 @@ namespace galsim {
             // Get the location where the photon strikes the silicon:
             double x0 = photonsX[i]; // in pixels
             double y0 = photonsY[i]; // in pixels
+#ifdef DEBUGLOGGING
             xdbg<<"x0,y0 = "<<x0<<','<<y0;
+#endif
 
             // get uniform random number for conversion depth from randomArray
             // (4th of 4 numbers for this photon)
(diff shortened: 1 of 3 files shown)


src/Silicon.cpp
```diff
@@ -1372,7 +1372,9 @@ namespace galsim {
             // Get the location where the photon strikes the silicon:
             double x0 = photonsX[i]; // in pixels
             double y0 = photonsY[i]; // in pixels
+#ifdef DEBUGLOGGING
```
rmjarvis, line 1375: These ifdef guards are superfluous.  Please remove them.  (xdbg is already a no op when DEBUGLOGGING is not defined.)

jamesp-epcc: Without the guards, the GPU build fails on both gcc and nvc++ (though not clang). My guess would be that the GPU compiler sees std::cerr and throws an error before realising that it's unreachable. I think we either need the guards or we'll need to modify the implementation of xdbg so that disabling it works differently.

rmjarvis: How about changing `std:cerr` to `*dbgout` and have an `extern std::ostream* dbgout` in the no DEBUGLOGGING block.  You might need to define this variable somewhere too in a .cpp file, but maybe not if the compiler always converts it to a no op in the .o files.

jamesp-epcc: That solves the problem.

rmjarvis (approved):

PR #1233, opened 2023-08-02
Recurse ProcessTemplate call in case template file has further templates

rmjarvis:
@cwwalter pointed out that the template processing doesn't work correctly if the template file itself has a template field.  This PR fixes that.

### DIFF
--- a/tests/config_input/template.yaml
+++ b/tests/config_input/template.yaml
@@ -21,6 +21,7 @@ template: multirng
 
 modules: 
   - template_register
+  - astropy.time
 
 output:
   file_name: test_template.fits
@@ -40,7 +41,7 @@ gal.ellip:
   template: multirng:psf.ellip
 
 gal.half_light_radius: ""
-gal.scale_radius: 1.6
+gal.scale_radius: "$astropy.time.Time(1.6, format='jd').to_value('jd')"
 
 psf:
   type: List
--- a/tests/config_input/template.yaml
+++ b/tests/config_input/template.yaml
@@ -1,0 +1,57 @@
+# Copyright (c) 2012-2022 by the GalSim developers team on GitHub
+# https://github.com/GalSim-developers
+#
+# This file is part of GalSim: The modular galaxy image simulation toolkit.
+# https://github.com/GalSim-developers/GalSim
+#
+# GalSim is free software: redistribution and use in source and binary forms,
+# with or without modification, are permitted provided that the following
+# conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this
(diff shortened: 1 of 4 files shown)


cwwalter:
Looks good.  Thanks!  So this can go into 2.4 too?

rmjarvis:
Yes, I realized we can put this on the 2.4 series.  And can you turn your "looks good" into an actual review please?

tests/config_input/template.yaml
```diff
@@ -0,0 +1,58 @@
...
+  e2: 0
+
+input.power_spectrum.1.grid_spacing: 2
+
+gal.magnify:
+  type: PowerSpectrumMagnification
+  num: 1
+
+gal.ellip:
+  template: multirng:psf.ellip
```
cwwalter, line 41: This is good, but just a note that it doesn't explicitly test the thing I noted:

```
file1.yaml
  template: mytemplate.yaml

file2.yaml
  template: file1.yaml
```

Although from reading the code I think it is handled the same.

tests/test_config_image.py
```diff
@@ -2455,6 +2460,29 @@ def test_template():
...
+    config8 = galsim.config.ReadConfig('config_input/template.yaml')[0]
+    galsim.config.ProcessAllTemplates(config8)
+    for key in config3:
+        print(key)
+        print(config4[key])
+        print(config8[key])
+        assert config8[key] == config4[key]
+    assert config8 == config4
+
+    # Make sure nested templating works
```
cwwalter, line 2473: I think the section above that tests the config_input/template.yaml file also tests nesting.. might want to note it above also.

tests/config_input/template.yaml
cwwalter, line 41: This is good, but just a note that it doesn't explicitly test the thing I noted that didn't work:

```
file1.yaml
  template: mytemplate.yaml

file2.yaml
  template: file1.yaml
```

Although from reading the code I think it is handled the same.

cwwalter (approved):
LGTM. 

I left just a few comments about things that were a bit unclear to me.

tests/test_config_image.py
rmjarvis: No.  The previous test just confirms that template.yaml is equivalent to the thing we already tested above in python code.  config9 is the one that tests your use case of nesting templates.

tests/config_input/template.yaml
rmjarvis: That's tested by config9, not this template.yaml.  I thought it gratuitous to make another file that does this, but config9 does it in code.

cwwalter: OK got it.  Thanks.  LGTM.

Issue #1229, opened 2023-07-19
Can't photon shoot simple chromatic transformation


jmeyers314:
`ChromaticTransformation._shoot` attempts to evaluate wavelength-dependent jacobian, offset, and flux_ratios [here](https://github.com/GalSim-developers/GalSim/blob/releases/2.4/galsim/chromatic.py#L1933), but some of the underlying methods (for example in `ChromaticObject.expand` [here](https://github.com/GalSim-developers/GalSim/blob/releases/2.4/galsim/chromatic.py#L840-L842)) don't work correctly when handed an array of wavelengths.  Here's a short snippet that fails due to the above:

```python
import galsim
bandpass = galsim.Bandpass('LSST_r.dat', 'nm')
psf = galsim.ChromaticObject(
    galsim.Gaussian(fwhm=1)
).dilate(lambda w: (w/500)**-0.3)
star = galsim.DeltaFunction()*galsim.SED('vega.txt', 'nm', 'flambda')
obj = galsim.Convolve(psf, star)
img = obj.drawImage(bandpass, nx=15, ny=15, scale=1, method='phot')
```

rmjarvis:
Done #1236

PR #1236, opened 2023-08-14
#1229 Fix a few errors in photon shooting chromatic transformations.

rmjarvis:
Josh pointed out in #1229 that chromatic objects with wavelength-dependent transformations don't work with photon shooting.  This fixes that.

I also fixed two other issues in this part of the code:

* GalSim automatically adds a WavelengthSampler to the photon_ops for photon shooting chromatic objects.  But if the user already did that, then it (1) redundantly samples the wavelengths twice, which is inefficient and (2) warns the user that this is likely an error.  Well, IMO, the error is in the GalSim code, not the user code.  So now it skips adding a redundant WavelengthSampler and consequently doesn't emit the warning.
* I realized when writing the test that the handling of the poisson_flux option in ChromaticConvolution was incorrect.  It neglected to apply the g factor returned by `_calculate_nphotons`, which is required to get the rendered poisson flux correct.  (Basically because the way we implement this, shooting always tries to get the right total flux with whatever n_photons is, but then multiplying by g gets them back to the poisson value.)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -44,3 +44,5 @@ Bug Fixes
 
 - Fixed a bug that prevented Eval types from generating lists in config files in some contexts.
 - Changed the SED class to correctly broadcast over waves when the SED is constant. (#1228)
+- Fixed some errors when drawing ChromaticTransformation objects with photon shooting. (#1229)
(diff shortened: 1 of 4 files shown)


jmeyers314 (approved):
LGTM

PR #1237, opened 2023-08-15
Add initial_image input type

rmjarvis:
This PR adds the ability for the config processing to read in an initial image from a file and draw additional things onto that.  The implementation is pretty simple:
```
input:
    initial_image:
        file_name: inital_image_file.fits
```
My test for this is to simulate a supernova time series.  It draws a single reference image, and then draws a point source with variable flux 12 times.  The SN appears between the 3rd and 4th image.  And then it decays exponentially.
Screen Shot 2023-08-15 at 4 11 04 PM

### DIFF
--- a/tests/config_input/sn.yaml
+++ b/tests/config_input/sn.yaml
@@ -1,0 +1,86 @@
+# Copyright (c) 2012-2022 by the GalSim developers team on GitHub
+# https://github.com/GalSim-developers
+#
+# This file is part of GalSim: The modular galaxy image simulation toolkit.
+# https://github.com/GalSim-developers/GalSim
+#
+# GalSim is free software: redistribution and use in source and binary forms,
+# with or without modification, are permitted provided that the following
+# conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this
+#    list of conditions, and the disclaimer given in the accompanying LICENSE
+#    file.
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions, and the disclaimer given in the documentation
+#    and/or other materials provided with the distribution.
+#
+
+
+modules:
+    - numpy
+
+psf:
+    type: Moffat
+    beta: 3
+    fwhm: 0.6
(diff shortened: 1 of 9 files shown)


tests/config_input/sn.yaml
```diff
@@ -0,0 +1,86 @@
...
+#    list of conditions, and the disclaimer given in the accompanying LICENSE
+#    file.
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions, and the disclaimer given in the documentation
+#    and/or other materials provided with the distribution.
+#
+
+
+modules:
+    - numpy
```
jmeyers314, line 21: Is this needed for the `np.exp(-time/50)` bit down below?  I know `import numpy as np` is canonical, but I'm a little surprised `np.exp` works instead of `numpy.exp` given this line.  Just curious how this works.

jmeyers314 (approved):
LGTM.  Just a have few points around my own curiosity; nothing for you to address in code.

tests/config_input/sn.yaml
```diff
@@ -0,0 +1,86 @@
...
+image:
+    type: Single
+    size: 128
+    dtype: float
+    random_seed:
+        - 12345
+    pixel_scale: 0.2
+    # No noise, which is unrealistic, but makes the test easier.
+
+---
```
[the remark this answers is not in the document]
rmjarvis: Yes.  This is kind of a different way of doing templates where everything is self-contained in one file.  The first document is the base for everything else.  Then each other document starts with that as a template and adds modifications.  It's described in some detail in demo6.yaml.  Then also used by demos 8, 9, 12.

tests/config_input/sn.yaml
rmjarvis: No.  I'm not sure why this is here.  I think it must be a relic from something that since changed.  I removed it.

But to answer your question about np.exp, we automatically import numpy, both as numpy and as np, so those are always available without anything special on the part of the user.

PR #1238, opened 2023-08-15
Add option to skip failures in stamp processing.

rmjarvis:
This PR adds a new option in stamp processing, skip_failures, which will skip an object if some exception is raised during processing.  This is similar in spirit to the existing retry_failures, but more appropriate when the potential cause of failure is from an input catalog rather than a random value, so no amount of retrying would resolve the problem.


jmeyers314 (approved):
LGTM

PR #1240, opened 2023-08-18
Move some of our test utilities to galsim.utilities

rmjarvis:
Some of the things that are in our galsim_test_helpers, I've used in other projects and just copied them over.  It would be easier if these lived in galsim.utilities instead, so I could just use them directly.  This PR puts the ones that I think are potentially useful for other project in our installed code base:

* do_pickle
* all_objs_diff
* timer
* profile
* CaptureLog


galsim/utilities.py
```diff
@@ -1913,3 +1913,244 @@ def __exit__(self, type, value, traceback):
...
+        return self
+
+    def __exit__(self, type, value, traceback):
+        self.handler.flush()
+        self.output = self.stream.getvalue().strip()
+        self.handler.close()
+
+
+# Context to make it easier to profile bits of the code
+class profile:
```
jmeyers314, line 2120: FWIW, I'd actually love to rename this one so it doesn't conflict with using `@profile` via [line_profiler](https://github.com/pyutils/line_profiler#line_profiler).

galsim/utilities.py
```diff
@@ -1913,3 +1913,244 @@ def __exit__(self, type, value, traceback):
...
+    4. copy.copy(obj) returns an equivalent copy unless random=True.
+    5. copy.deepcopy(obj) returns an equivalent copy.
+    6. eval(repr(obj)) returns an equivalent copy unless random or irreprable=True.
+    7. repr(obj) makes something if irreprable=False.
+
+    Parameters:
+        obj:            The object to test
+        func:           A function of obj to use to test for equivalence.  [default: lambda x: x]
+        irreprable:     Whether to skip the eval(repr(obj)) test.  [default: False]
+        random:         Whether the obj has some intrinsic randomness. [default: False, unless
```
jmeyers314, line 1939: Looks like default is `None`?

jmeyers314 (approved):
LGTM. Minor comments.

galsim/utilities.py
rmjarvis: Right.  I was describing what that default means, which is False normally, but tries to notice randomness.

galsim/utilities.py
rmjarvis: OK.  Suggestions?  Maybe just capitalize?  `with Profile():`  That would fit in with most of our GalSim style for classes.

rmjarvis: I was also thinking of renaming a couple others.  `do_pickle` and `all_diff` aren't great names either, tbh.  I was thinking 'check_pickle` and `check_all_diff` for those.

jmeyers314: Yeah, these all sound better to me.

PR #1242, opened 2023-08-22
Don't error if two photon_arrays being convolved both have allocated ancillary arrays.

rmjarvis:
When convolving two PhotonArrays, we had a feature where GalSim would check if both of them had various ancillary arrays set (and not identical), and if they were both set, it would raise an exception.  The thought was that this probably indicated an error.  But in practice it meant that when two PhaseScreenPSFs were convolved, they couldn't be drawn with photon shooting, since each of them would have set the pupil coordinates and the time independently.

If you want to be really careful about these ancillary arrays, then you can put the PSFs in photon_ops, in which case they will be processed sequentially and PhaseScreenPSF won't overwrite e.g. pupil coordinates that are already there.  But if you don't much care about these ancillary arrays, it should be allowed to just convolve two PSF objects and draw them any old way.  So now you can.

This bug showed up in imSim when doOpt=True.  The workaround was to put the psfs in photon_ops, as I mentioned above.  cf. https://github.com/LSSTDESC/imSim/pull/410

### DIFF
--- a/tests/test_photon_array.py
+++ b/tests/test_photon_array.py
@@ -350,18 +350,26 @@ def test_convolve():
 
     # Check propagation of dxdz, dydz, wavelength, pupil_u, pupil_v
     for attr, checkFn in zip(
-        ['dxdz', 'wavelength', 'pupil_u', 'time'],
-        ['hasAllocatedAngles', 'hasAllocatedWavelengths', 'hasAllocatedPupil',
+        ['dxdz', 'dydz', 'wavelength', 'pupil_u', 'pupil_v', 'time'],
+        ['hasAllocatedAngles', 'hasAllocatedAngles',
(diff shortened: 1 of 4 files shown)


tests/test_photon_array.py
jmeyers314, line 1: Not a change for this PR, but I'm noticing that the comment on line 351 mentions `dxdz, dydz, wavelength, pupil_u, pupil_v` but the code doesn't actually look at `dydz` or `pupil_v`.  Should probably make this consistent.

jmeyers314 (approved):
Looks good.  One minor comment.

tests/test_photon_array.py
rmjarvis: I added dydz and pupil_v to the list.

PR #1246, opened 2023-09-01
Let vectors in SiliconSensor release their previous allocated memory when getting a smaller target image.

rmjarvis:
The SiliconSensor class uses a rather large amount of memory when the postage stamp is large.  Proportional to the number of pixels, as you might imagine.  However, once it's done with the large image and goes on to smaller postage stamps, it doesn't release that memory.  This PR lets those vectors release the allocated memory whenever the target image size decreases.  It also does the same with the `_delta` image -- releasing the memory when it gets smaller.

### DIFF
--- a/include/galsim/Image.h
+++ b/include/galsim/Image.h
@@ -736,10 +736,16 @@ namespace galsim {
          *  share data with each other, but the tie to this ImageAlloc is severed.
          *
          *  This typically allocates new memory for the array.  The only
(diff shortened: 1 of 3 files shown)


jamesp-epcc (approved):
This looks good to me. The resize happens during a phase of the code where no data is mapped to the GPU so it won't affect the GPU acceleration at all.

PR #1249, opened 2023-10-03
Add galsim.EmissionLine

jmeyers314:


galsim/sed.py
```diff
@@ -1095,5 +1097,131 @@ def __setstate__(self, d):
...
+
+class EmissionLine(SED):
+    """Emission line SED.
+
+    Creates a simple triangle-shaped emission line with a given wavelength and
+    FWHM.
+
+    Parameters:
+        wavelength:    The wavelength of the line.
+        fwhm:          The full-width-half-max of the line.
```
rmjarvis, line 1109: This should probably have  default too, I think.  1 nm or something.

galsim/sed.py
```diff
@@ -1095,5 +1097,131 @@ def __setstate__(self, d):
...
+        flux:          The integrated flux of the line.  [default: 1.0]
+        wave_type:     The units of wavelength used for this SED.  See SED
+                       constructor for options.  [default: 'nm']
```
rmjarvis, line 1112: I guess this is just for fwhm.  Maybe mention that here.

rmjarvis (changes requested):
A few suggestions.  Also, I think we had a few places in the code where we hand-make an SED that is basically an EmissionLine.  It would be nice to replace these with the new class.

galsim/sed.py
rmjarvis: Of course.  Yes, wavelength too.  :)

rmjarvis (approved):

Issue #1254, opened 2023-10-19
SED flux type units lost in translation


sidneymau:
When constructing an SED with a custom flux type via astropy units, the flux type appears to get improperly converted to `fnu` such that multiplying this SED mangles the units (e.g., when calling any `SED.__mul__` [method](https://github.com/GalSim-developers/GalSim/blob/releases/2.5/galsim/sed.py#L546-L548)).
```
>>> disk_sed = galsim.SED(disk_sed_table, u.Angstrom, u.Lsun / u.Hz / u.Mpc**2, redshift=mock["redshift"][igal]))
>>> disk_sed.flux_type
'fnu'
```
This results in wrong fluxes
```
lsst_g = galsim.Bandpass("LSST_g.dat", wave_type="nm")
>>> (disk_sed).calculateFlux(lsst_g)
0.04530669166489878
>>> (1 * disk_sed).calculateFlux(lsst_g)
112691591447658.53
```
We can see that the flux type of the remade SED (even without applying any multiplicative factors) is the cause of this
```
>>> galsim.SED(galsim._LookupTable(disk_sed._spec.getArgs(), np.array(disk_sed._spec.getVals())), disk_sed.wave_type, disk_sed.flux_type, disk_sed.redshift).calculateFlux(lsst_g)
112691591447658.53
```
whereas forcing the original flux type produces the correct value
```
>>> galsim.SED(galsim._LookupTable(disk_sed._spec.getArgs(), np.array(disk_sed._spec.getVals())), disk_sed.wave_type, u.Lsun / u.Hz / u.Mpc**2, disk_sed.redshift).calculateFlux(lsst_g)
0.04530669166489878
```

Tested in GalSim Version 2.5.0

sidneymau:
I noticed that the original flux type is preserved as `SED._flux_type. Using this to reconstruct the SED in the above methods works:
```
>>> galsim.SED(galsim._LookupTable(disk_sed._spec.getArgs(), np.array(disk_sed._spec.getVals())), disk_sed.wave_type, disk_sed._flux_type, disk_sed.redshift).calculateFlux(lsst_g)
0.04530669166489878
```

sidneymau:
Per the above, changing https://github.com/GalSim-developers/GalSim/blob/releases/2.5/galsim/sed.py#L530 from
```
flux_type = self.flux_type
```
to
```
flux_type = self._flux_type
```
fixes the particular problem I'm seeing. With the change,
```
[...]

PR #1256, opened 2023-10-19
Add test to check fluxes calculated for different flux types and propose fix

sidneymau:
This first commit introduces a new test in `test_sed.py` that fails with the current implementation of `SED._mul_scalar` to address #1254.

### DIFF
--- a/tests/test_sed.py
+++ b/tests/test_sed.py
@@ -1209,26 +1209,21 @@ def test_flux_type_calculateFlux():
         wave_type="nm", flux_type="1"
     )
     bp = galsim.Bandpass(galsim.LookupTable([2,4], [1,2]), wave_type='nm')
-    # bp = galsim.Bandpass('1', 'nm', blue_limit=1, red_limit=5)
+    exp_gal = galsim.Exponential(half_light_radius=0.5, flux=1)
 
     for sed in [sed1, sed2, sed3, sed4]:
         flux1 = sed.calculateFlux(bp)
         flux2 = (1 * sed).calculateFlux(bp)
-        flux3 = (galsim.Exponential(half_light_radius=0.5, flux=1) * sed).calculateFlux(bp)
+        flux3 = (exp_gal * sed).calculateFlux(bp)
         flux4 = (flat_sed * sed).calculateFlux(bp)
         print('flux = {}, {}, {}'.format(flux1, flux2, flux3))
         wave = np.linspace(bp.blue_limit, bp.red_limit, 10000)
         f = sed(wave) * bp(wave)
         flux5 = np.trapz(f, wave)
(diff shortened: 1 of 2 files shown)


tests/test_sed.py
```diff
@@ -1186,6 +1186,37 @@ def test_emission_line():
...
+    )
+    sed4 = galsim.SED(
+        galsim.LookupTable([1,2,3,4,5], [1.1, 1.9, 1.4, 1.8, 2.0]),
+        wave_type="nm", flux_type=units.Lsun / units.Hz / units.Mpc**2
+    )
+    bp = galsim.Bandpass(galsim.LookupTable([2,4], [1,2]), wave_type='nm')
+
+    for sed in [sed1, sed2, sed3, sed4]:
+        flux1 = sed.calculateFlux(bp)
+        flux2 = (1 * sed).calculateFlux(bp)
```
rmjarvis, line 1211: Probably worth also adding a test of multiplying a gsobject by sed, rather than just 1.  Especially since that was the original way you found this bug.  I think currently that hits the same code as `1*sed`, but I can imagine them being separate code paths for some reason in the future.

rmjarvis (changes requested):
Looks good.  Thanks Sid!  Just one suggestion in the test function.

tests/test_sed.py
sidneymau: Would a flat SED be sufficient to test this? It's degenerate with multiplying by 1 but should hit `SED._mul_sed` instead of `SED._mul_scalar`

tests/test_sed.py
```diff
@@ -1204,17 +1204,31 @@ def test_flux_type_calculateFlux():
...
+        flux5 = np.trapz(f, wave)
         np.testing.assert_allclose(flux1, flux3)
         np.testing.assert_allclose(flux2, flux3)
+        np.testing.assert_allclose(flux1, flux4)
+        np.testing.assert_allclose(flux2, flux4)
+        np.testing.assert_allclose(flux3, flux4)
+        np.testing.assert_allclose(flux1, flux5)
+        np.testing.assert_allclose(flux2, flux5)
+        np.testing.assert_allclose(flux3, flux5)
+        np.testing.assert_allclose(flux4, flux5)
```
rmjarvis, line 1231: I think it's probably fine to just compare everything to one of the values.  E.g. flux1 say.  Transitivity should handle the rest.

rmjarvis (approved):

tests/test_sed.py
sidneymau: I realized I didn't compare `flux1` and `flux2` directly, so I used the next commit as an opportunity to address this as well.

Issue #1187, opened 2022-10-22
Revisit use of spline LookupTable in SED class


rmjarvis:
Currently SEDs are much faster if their underlying function (`spec`) uses a LookupTable with `interpolation='linear'`.  When this is the case, various manipulations preserve the function tabulation (e.g. multiplying by a bandpass), but when the function is anything else, the result is a lambda function.

This can be a gotcha for users who think they are providing a LookupTable, but use the default `interpolation='spline'`.  Then it's very easy to do pretty normal things to an SED and end up with very slow integration when sampling wavelengths for chromatic photon shooting (for instance).

I *think* the reason for some of the choices here was that linear interpolated tables were the only functions that were quick to integrate.  But now `LookupTable.integrate` can handle spline interpolation equally quickly.  So I suspect we should revisit our choices here and try expand the use cases where we keep a tabulated function.  And hopefully avoid the kind of gratuitous inefficiency trap that is easy to fall into currently.

rmjarvis:
Done (a long time ago) in #1257

PR #1257, opened 2023-10-27
Improve behavior of SED class, especially with respect to pickling and spline interpolation.

rmjarvis:
This PR addresses a number of things that have turned up in various imsim/roman_sim runs related to the SED classes.  The initial impetus for this work was to let the SEDs we get from skyCatalogs to be picklable, so image.nproc would work correctly, and other use cases as well would not be quite so finicky about pickling.

* The main solution for this problem was to have mul_sed try to return a tabulated sed if possible.  I.e. when either (or both) of them uses a lookup table.
* Another thing that helps for some cases is to not try to pickle the derived SEDs in chromatic objects that have the SED as a lazy_property.  It's not very inefficient to remove them from the dictionary before pickling and remake them if needed.  (Indeed it might be usually more efficient, since there is less serialization and communication involved.)
* While I was at it I addressed issue #1187, making it generally fine to use spline interpolation for Bandpass and SED objects.  The default in these classes is still linear (e.g. when reading from a file), but now it's not inefficient to build an SED with a regular LookupTable using the default interpolant='spline'.
* Given the above, I added an `interpolant` option to these classes so if you want to read from a file using spline interpolation, it is now possible.
* I added a clip_neg option to DistDeviate to clip any negative values in the probability distribution to 0.  This is now used in sampleWavelengths, since spline interpolations can lead to slightly negative values in some cases.  Rather than try to avoid it in the SED class, it seems more sensible to just allow DistDeviate to gracefully ignore those negative values.  The default behavior of DistDeviate is unchanged -- with the default clip_neg=False, it raises an error if there are any negative probabilities.
* The hardest thing for getting spline to work well was to get the thin() algorithm to work correctly with a spline table.  I couldn't find an O(N) algorithm for that, so it uses the slower algorithm, but I think it's basically working now.
* Along the way I found two bugs in the new EmissionLine class, which I fixed.  First, the `atRedshift` function was wrong.  It both set the redshift value and changed the wavelength of the line.  The nominal wavelength shouldn't change, since the given spectrum is supposed to define the rest-frame SED, not the observed SED.  
* Also, in some use cases, the 0 at the edge of the emission line could turn into a 1.e-16 kind of number.  When integrating from there to "infinity", that could lead to crazy bolometric fluxes.  My fix for this was to add a second 0 another fwhm away on either size to really force it to zero when not near the line.
* Related to that, I changed out definition of infinity in bolometric fluxes from 1.e100 to 1.e30 (nm).  That's still 1.e21 meters, which is way past any conceivable radio wave wavelength.  But it allows functions of wave with some modest power like wave**3 to not blow up.
* Finally, I added an interpolant option to our trapz function to make it easier do the integral in the thin_tabulated function using the specified interpolant.  I didn't end up using it in as many places as I originally expected to, but still, I think it's a useful enough feature, so I kept it.

### DIFF
--- a/galsim/sed.py
+++ b/galsim/sed.py
@@ -891,12 +891,13 @@ class SED:
             spec_native = self._spec(rest_wave_native)
 
             # Note that this is thinning in native units, not nm and photons/nm.
-            newx, newf = utilities.thin_tabulated_values(
-                    rest_wave_native, spec_native, rel_err=rel_err,
-                    trim_zeros=trim_zeros, preserve_range=preserve_range, fast_search=fast_search)
-
             interpolant = (self.interpolant if not isinstance(self._spec, LookupTable)
                            else self._spec.interpolant)
+            newx, newf = utilities.thin_tabulated_values(
+                    rest_wave_native, spec_native, rel_err=rel_err,
+                    trim_zeros=trim_zeros, preserve_range=preserve_range,
+                    fast_search=fast_search, interpolant=interpolant)
+
             newspec = _LookupTable(newx, newf, interpolant=interpolant)
             return SED(newspec, self.wave_type, self.flux_type, redshift=self.redshift,
                        fast=self.fast)
--- a/galsim/sed.py
+++ b/galsim/sed.py
@@ -1139,9 +1139,13 @@ class EmissionLine(SED):
         _, wave_factor = SED._parse_wave_type(wave_type)
         assert max_wave > wavelength + fwhm
 
+        w = wavelength
+        # Some operations can turn a 0 into 1.e-15, which can lead to large errors
+        # when integrating from w+fwhm to max_wave.  So add a second set of 0's at
+        # w +- 2*fwhm to make sure it's exactly 0 for most of the range.
         spec = LookupTable(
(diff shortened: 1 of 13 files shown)


galsim/sed.py
```diff
@@ -413,7 +421,8 @@ def _fast_spec(self):
                     f = self._rest_nm_to_photons(x)
                 else:
                     f = self._rest_nm_to_dimensionless(x)
-                return _LookupTable(x, f, interpolant='linear')
+                interp = self._spec.interpolant if isinstance(self._spec, LookupTable) else 'spline'
```
jmeyers314, line 424: Most of the other default interps look to be 'linear', why 'spline' here?

jmeyers314 (approved):
LGTM.  Thanks for catching the EmissionLine bugs.  Just one small question.

galsim/sed.py
rmjarvis: Good catch.  I was originally a little aggressive about trying to get most things to use spline by default, but then I backpedaled on that idea, since it felt a little too API-changing. I just missed this one.

PR #1258, opened 2023-10-29
Don't err if SiliconSensor accumulates on image outside range of treering function

rmjarvis:
In ImSim, we sometimes make a stamp image that is large enough to extend past the allowed range of the treering function.  The treatment outside that range is irrelevant, since we just ignore the part that is outside the bounds of the main image, but the current code was raising an exception.

This PR changes it to just ignore tree rings on any pixels that are outside the range of the treering function.


jmeyers314 (approved):
LGTM

PR #1259, opened 2023-10-31
Implement ChromaticSum._shoot

rmjarvis:
In PR #1100, I implemented the applyTo method for chromatic objects, so they can be used as photon_ops.  This is now the canonical way to do photon shooting with chromatic objects, since it can use each photon's specific wavelength to do wavelength-dependent things.

One that I skipped in that PR was ChromaticSum.  I suspect I thought that class would only be used for things like bulge + disk, where each object has a spectral SED.  These can't be used as photon_ops, so I didn't implement that functionality for them.

However, @matroxel wants to use ChromaticSum for the RomanPSF to smoothly interpolate across a sensor using a weighted sum of 4 PSFs, corresponding to the 4 corners of the sensor.  So this PR rectifies my oversight, and implements ChromaticSum._shoot, which is the back end function for all PSF applyTo calls.

Note: when implementing this, I added PhotonArray._copyFrom, which can copy portions of the data from one PhotonArray to another.  It copies all the allocated arrays, using an arbitrary mask in both the original and target PhotonArray.  Similar to `assignAt` but more flexible.  (And `assignAt` now uses `_copyFrom` for its implementation.)

I now use this all the places where we were doing this.  I think this is safer than what we were doing, since some PSFs may need things like the pupil positions or time and when generating the positions/fluxes, they may have generated such things.  So it's a little bit less efficient in most cases (but not much) in the interest of avoiding potentially subtle bugs that may turn up in various edge cases.  (We don't currently have any unit tests that care about such things.)

### DIFF
--- a/galsim/photon_array.py
+++ b/galsim/photon_array.py
@@ -440,28 +440,51 @@ class PhotonArray:
 
     def assignAt(self, istart, rhs):
         """Assign the contents of another `PhotonArray` to this one starting at istart.
+
+        Parameters:
+            istart:     The first index at which to insert new values.
+            rhs:        The other `PhotonArray` from which to get the values.
         """
         if istart + rhs.size() > self.size():
             raise GalSimValueError(
                 "The given rhs does not fit into this array starting at %d"%istart, rhs)
         s = slice(istart, istart + rhs.size())
-        self.x[s] = rhs.x
-        self.y[s] = rhs.y
-        self.flux[s] = rhs.flux
+        self._copyFrom(rhs, s, slice(None))
+
+    def _copyFrom(self, rhs, s1, s2):
+        """Copy some contents of another `PhotonArray` to some elements of this one.
+
+        Specifically each element of rhs[s2] is mapped to self[s1].  The values s1 and s2
+        may be slices, list of indices or any other type that is a valid key for a
+        numpy array.
+
+        .. warning::
+
+            There are no checks that the given s1 and s2 are valid keys for the two
+            photon arrays.  It is up to the user to ensure these are valid.
+
+        Parameters:
+            rhs:        The `PhotonArray` from which to get values.
+            s1:         The indices at which to assign values in the current PhotonArray (self).
+            s2:         The indices from which to get values from the PhotonArray, ``rhs``.
+        """
+        self.x[s1] = rhs.x[s2]
+        self.y[s1] = rhs.y[s2]
+        self.flux[s1] = rhs.flux[s2]
         if rhs.hasAllocatedAngles():
             self.allocateAngles()
-            self.dxdz[s] = rhs.dxdz
-            self.dydz[s] = rhs.dydz
+            self.dxdz[s1] = rhs.dxdz[s2]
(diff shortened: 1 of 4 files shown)


galsim/photon_array.py
```diff
@@ -205,7 +205,13 @@ def _fromArrays(
         ret._pupil_u = pupil_u
         ret._pupil_v = pupil_v
         ret._time = time
-        ret._is_corr = is_corr
+        ret._is_corr = False
+        if is_corr:
+            from .deprecated import depr
+            depr('is_corr=True', 2.5, '',
```
jmeyers314, line 211: What was the former use case and why is it no longer necessary?

tests/test_photon_array.py
```diff
@@ -243,6 +233,69 @@ def test_photon_array():
...
+    assert pa2.pupil_u[47] != pa1.pupil_u[18]
+    assert pa2.pupil_v[47] != pa1.pupil_v[18]
+    assert pa2.time[47] != pa1.time[18]
+
+    # Error if indices are invalid
+    assert_raises(ValueError, pa2.copyFrom, pa1, slice(50,None), slice(50,None))
+    assert_raises(ValueError, pa2.copyFrom, pa1, 100, 0)
+    assert_raises(ValueError, pa2.copyFrom, pa1, 0, slice(None))
+    assert_raises(ValueError, pa2.copyFrom, pa1)
+
```
jmeyers314, line 298: Should probably add some tests for numpy fancy indexing too.
I.e., `pa1.copyFrom(pa2, pa2.flux>10)` and `pa1.copyFrom(pa2, np.where(pa2.flux>10))`

jmeyers314: And maybe even discontiguous slices.  `pa1.copyFrom(pa2, slice(0, 100, 2))`

jmeyers314 (approved):
Looks good to me.  I have one question and one additional testing suggestion.

galsim/photon_array.py
rmjarvis: It was for the Sum._shoot implementation.  That had done all the photons in one component first, then all the ones in the other component.  I switched that to do the same algorithm I used for ChromaticSum.  This is potentially important if the PSF is time dependent for instance (e.g. includes a PhasePSF), so the order in the PhotonArray matters.  It seems like there are subtle implications to letting them be correlated like that, and it was very easy to switch it to not ever be correlated in the first place.

tests/test_photon_array.py
rmjarvis: Yep.  Good ideas.  Will do.

rmjarvis:
Question for you.  Do you think we should deprecate assignAt as well in favor of the new copyFrom?  
I think the new name is clearer (and matches a similar method in Image).  The only place we had used assignAt was in the old Sum._shoot method, so that's now gone with the new algorithm.  And I doubt any users probably use assignAt for anything.

jmeyers314:
> Do you think we should deprecate assignAt as well in favor of the new copyFrom?

Yep, this sounds good to me.

PR #1261, opened 2023-11-07
Add max_sun_angle to the roman module global variables

rmjarvis:
For the upcoming Rubin-Roman sims, the time delay group (the TDS PIT) wants to be able to pretend that the ELAIS field is in the Roman continuous viewing zone so the transient light curves don't end up getting split up into discrete seasons.  

I think the least-intrusive way to allow this, without overly encouraging people to just ignore the sun position in Roman observations, is to make the maximum angle between the solar panels and the sun be a galsim.roman module variable, rather than hard-coded.  This is good programming practice anyway (it's currently written as just 36, rather than even named in the function).  And it would allow the roman_imsim package to add an option to edit it to something larger, which would let GalSim still calculate reasonable roll angles and just pretend that the solar panels could handle larger incidence angles.

I.e. `galsim.roman.max_sun_angle = 36`, but is potentially settable by user code.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -105,3 +105,9 @@ Changes from v2.5.0 to v2.5.1
   necessary anymore. (#1259)
 - Deprecated `PhotonArray.assignAt` in favor of the more flexible `PhotonArray.copyFrom`
   method. (#1259)
+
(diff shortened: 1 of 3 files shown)


matroxel (approved):
Looks good to me.

Issue #1272, opened 2024-01-24
Image array throws error about having no setter but still performs operation


sidneymau:
I guess this only works in an interactive session / debug shell, but modifying the underlying numpy array of an image _can_ be done despite an AttributeError being thrown. For example, the following throws an `AttributeError` but the operation still has taken place. This is somewhat confusing behavior
```
(Pdb) psf_image.array
array([[-4.3576780e-13, -4.2518140e-13, -4.0360353e-13, ...,
         1.5450338e-13, -1.7318126e-14, -2.6135360e-13],
       [-4.2518140e-13, -5.3816332e-13, -5.9797236e-13, ...,
         4.2668004e-13,  1.6594023e-13, -1.3191374e-13],
       [-4.0360353e-13, -5.9797236e-13, -7.3459999e-13, ...,
         6.1851040e-13,  3.0497444e-13, -7.1951275e-14],
       ...,
       [ 1.5450338e-13,  4.2668004e-13,  6.1851040e-13, ...,
        -3.1695967e-13, -1.8824241e-13, -9.3841832e-14],
       [-1.7318127e-14,  1.6594022e-13,  3.0497444e-13, ...,
        -1.8824241e-13, -2.0354703e-13, -2.3963579e-13],
       [-2.6135357e-13, -1.3191373e-13, -7.1951275e-14, ...,
        -9.3841818e-14, -2.3963579e-13, -4.1648504e-13]], dtype=float32)
(Pdb) psf_image.array /= 10
*** AttributeError: property 'array' of 'Image' object has no setter
(Pdb) psf_image.array
array([[-4.3576779e-14, -4.2518140e-14, -4.0360354e-14, ...,
         1.5450338e-14, -1.7318125e-15, -2.6135361e-14],
       [-4.2518140e-14, -5.3816333e-14, -5.9797233e-14, ...,
         4.2668004e-14,  1.6594023e-14, -1.3191374e-14],
       [-4.0360354e-14, -5.9797233e-14, -7.3459996e-14, ...,
         6.1851037e-14,  3.0497445e-14, -7.1951273e-15],
       ...,
       [ 1.5450338e-14,  4.2668004e-14,  6.1851037e-14, ...,
        -3.1695966e-14, -1.8824240e-14, -9.3841832e-15],
       [-1.7318128e-15,  1.6594023e-14,  3.0497445e-14, ...,
        -1.8824240e-14, -2.0354703e-14, -2.3963579e-14],
[...]

PR #1273, opened 2024-02-23
Define setter for image.array to allow inplace operations

sidneymau:
This pull request addresses #1272 by defining a setter for the `Image.array` attribute. This protects against any case where a user does an inplace operation on the image array attribute in an interactive environment: without the setter, this would raise an error while still performing the operation, leading to some confusion. This change also obviates the need to operate on a view of the underling array; e.g., instead of
```
image.array[:] += 1
```
we can now do
```
image.array += 1
```


rmjarvis (changes requested):

rmjarvis:
Also, this shouldn't be directed at releases/2.5.  It's not a bugfit.  Rebase it to main and direct the PR to main.

sidneymau:
Evidently I don't know how to tell GitHub how to rebase, will try this in a new PR...

Issue #1262, opened 2023-11-07
CelestialWCS method docstring erratum?


sidneymau:
The docstring from `CelestialWCS.radecToxy`  [says](https://github.com/GalSim-developers/GalSim/blob/19291306afc43d860399fe8ce1749e43b821dd11/galsim/wcs.py#L1307)
>This is equivalent to wcs.toWorld(ra,dec, units=units)

but I think this is meant to say it is equivalent to `wcs.toImage`?

PR #1275, opened 2024-02-23
Fix CelestialWCS.radecToxy docstring erratum

sidneymau:
This addresses https://github.com/GalSim-developers/GalSim/issues/1262


rmjarvis (approved):
Thanks Sid!

PR #1276, opened 2024-02-26
Define setter for Image.array to allow inplace operations

sidneymau:
(Duplicate of https://github.com/GalSim-developers/GalSim/pull/1273 due to rebasing/etc.)
This pull request addresses https://github.com/GalSim-developers/GalSim/issues/1272 by defining a setter for the Image.array attribute. This protects against any case where a user does an inplace operation on the image array attribute in an interactive environment: without the setter, this would raise an error while still performing the operation, leading to some confusion. This change also obviates the need to operate on a view of the underling array; e.g., instead of
```
image.array[:] += 1
```
we can now do
```
image.array += 1
```

### DIFF
--- a/tests/test_image.py
+++ b/tests/test_image.py
@@ -1742,7 +1742,7 @@ def test_Image_inplace_setter():
             type_index = (i + j) % ntypes
             if type_index != i:
                 with assert_raises(galsim.GalSimError):
-                    image5 = galsim.Image(zeros_array)
+                    image5 = galsim.Image(ref_array.astype(types[i]))
                     image5.array = ref_array.astype(types[type_index])
 
 
--- a/tests/test_image.py
+++ b/tests/test_image.py
@@ -1719,7 +1719,11 @@ def test_Image_inplace_setter():
         # operations on the Image array
         internal_array = np.zeros_like(ref_array_view)
         image3 = galsim.Image(internal_array)
+        image3_array_id = id(image3.array)
         image3.array += ref_array_view
+
+        # Check that the id of the Image array attribute is unchanged
+        assert id(image3.array) == image3_array_id
 
         # NumPy arrays will return views with new IDs, so we must check if the
         # underlying data buffers are equal rather than just the IDs
@@ -1729,16 +1733,17 @@ def test_Image_inplace_setter():
 
         # Test that attempting to set an Image's array to an array of different
         # shape raises a ValueError
-        with assert_raises(ValueError):
+        with assert_raises(galsim.GalSimError):
             image4 = galsim.Image(zeros_array)
             image4.array = large_array
 
-        # Test that attempting to set an Image array from that of a real dtype
-        # to that of a complex dtype raises a TypeError
-        if np.isreal(types[i]):
-            with assert_raises(TypeError):
-                image4 = galsim.Image(zeros_array)
-                image4.array =  1j * ref_array_view
+        # Test that setting an Image array to a different dtype raises an error
+        for j in range(ntypes):
+            type_index = (i + j) % ntypes
+            if type_index != i:
+                with assert_raises(galsim.GalSimError):
+                    image5 = galsim.Image(zeros_array)
+                    image5.array = ref_array.astype(types[type_index])
 
 
 @timer
--- a/tests/test_image.py
+++ b/tests/test_image.py
@@ -1711,6 +1711,12 @@ def test_Image_inplace_add():
         image3 += (image2 / 17) * 17
         np.testing.assert_allclose(image3.array, image1.array)
 
+        # Check inplace operations on the underlying array
+        image1 = galsim.Image(ref_array.astype(types[i]))
+        image1.array += 1
+        image2 = galsim.Image(ref_array.astype(types[i]) + 1)
+        np.testing.assert_allclose(image1.array, image2.array)
+
         # Then try using the eval command to mimic use via ImageD, ImageF etc.
         image_init_func = eval("galsim.Image"+tchar[i])
         slice_array = large_array.copy().astype(types[i])[::3,::2]
--- a/tests/test_image.py
+++ b/tests/test_image.py
@@ -1758,11 +1758,11 @@ def test_Image_inplace_setter():
 def test_Image_setter_alignment():
     # Make an array that is _not_ memory aligned
     # c.f. https://stackoverflow.com/a/9895986
-    m = n = 1000
+    m = n = 2
     dtype = np.dtype(np.float64)
     nbytes = m * n * dtype.itemsize
     buf = np.empty(nbytes + 16, dtype=np.uint8)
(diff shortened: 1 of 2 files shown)


tests/test_image.py
sidneymau, line 1: Copying comment from Mike:

> I'd like a few more tests that guarantee that the setter doesn't allocate new memory or change fundamental things about the array. I think the implementation is ok, but there are other ways to have written the setter which would pass this test, but would break the image object.
e.g.
> 
> ```image1.array = 1  # should keep array shape and dtype
> image1.array = 1+2j  # should only work if array is already complex
> image1.array = np.zeros((3,3))  # should only work if current shape is (3,3)
> ```
> Also, all of those should fail if the image is a cont view.
> In other words, try to come up with as many stress tests as possible and make sure it behaves as we would want it to behave.

sidneymau: An alternative implementation of the setter is the following:
```
    @array.setter
    def array(self, other):
        """Set the numpy array in-place.
        """
        self_array = self.array
        other_array = np.asarray(other)
        if self_array.shape != other_array.shape:
            raise GalSimError("Other array shape is not equal to current array shape")
        if self_array.dtype != other_array.dtype:
            raise GalSimError("Other array dtype is not equal to current array dtype")
        self._array = other_array
```
This has an advantage in explicitly guarding against mutations to shape and/or dtype. I believe this is more robust and closer to what we want

rmjarvis: This looks worse to me.  I think numpy already guarded against these things.  And you lost the feature that writing to a const view would raise an exception.

rmjarvis: Also, the dtype check is wrong.  For instance, if dtype is float, you should be allowed to assign integers to it and let it remain a float array.  (Which I believe the previous implementation did correctly.)

rmjarvis: Oh, another problem with this implementation, which might merit a test is that it loses the careful 64 bit alignment that we do for image arrays.  The array should always satisfy:
```
assert image.array.ctypes.data % 16 == 0
```
Looks like we don't currently have tests for this, but it makes the SSE stuff in C++ work faster.

sidneymau: Hmm, both implementations currently pass all of the current tests, so I guess I need to write a test that (correctly) breaks one of them.

I do believe this second version does allow for manipulations with integers while remaining a float array (unless I am misunderstanding your comment); e.g., given an Image initialized with a float array, `image.array += 1`, `image.array += np.ones((2, 2), dtype=int)` etc. work fine. The implementations do differ in how they handle broadcasting when an Image is initialized with an int array.

e.g., `self._array = other[:]` version
```
>>> image = galsim.Image(np.zeros((2, 2), dtype=int))
>>> image.array = np.array([[1.2, 3.8], [5.6, 6.7]])
>>> image.array
array([[1, 3],
       [5, 6]], dtype=int32)
```

and `self._array = np.asarray(other)` version
```
>>> image = galsim.Image(np.zeros((2, 2), dtype=int))
>>> image.array = np.array([[1.2, 3.8], [5.6, 6.7]])
  File "/home/smau/Documents/software/GalSim/galsim/image.py", line 460, in array
    raise GalSimIncompatibleValuesError(
galsim.errors.GalSimIncompatibleValuesError: Image array dtypes are inconsistent Values {'arr1': array([[0, 0],
       [0, 0]], dtype=int32), 'arr2': array([[1.2, 3.8],
       [5.6, 6.7]])}
```

I'm not sure which of these is preferred in GalSim—the second is certainly less confusing to me

sidneymau: I suppose there's a third version: `self._array = np.asarray(other)` _without_ the type checking:
```
>>> image = galsim.Image(np.zeros((2, 2), dtype=int))
>>> image.array = np.array([[1.2, 3.8], [5.6, 6.7]])
>>> image.array
array([[1.2, 3.8],
       [5.6, 6.7]])
```
This is probably preferred to each of the above? Unless we want to enforce that the dtype of an Image array doesn't mutate

sidneymau: > Oh, another problem with this implementation, which might merit a test is that it loses the careful 128 bit alignment that we do for image arrays. The array should always satisfy:
> 
> ```
> assert image.array.ctypes.data % 16 == 0
> ```
> 
> Looks like we don't currently have tests for this, but it makes the SSE stuff in C++ work faster.

There's currently a commented assert checking this elsewhere in `Image`—is this worth raising an error/warning for in the class, or just writing a test?

rmjarvis: > I have not succeeded in making a numpy array that does not satisfy array.ctypes.data % 16 == 0, so will have to figure out how to do that to write this test

Maybe numpy added that as a feature.  They didn't used to guarantee that.

tests/test_image.py
```diff
@@ -1692,6 +1692,112 @@ def test_Image_binary_scalar_pow():
...
+    start_index = -buf.ctypes.data % 16 + 1  # offset by 1 to unalign
+    unaligned_array = buf[start_index:start_index + nbytes].view(dtype).reshape(m, n)
+    unaligned_array[:] = 1
+    assert unaligned_array.ctypes.data % 16 != 0
+
+    # Make arrays that are memory-aligned by default
+    zeros_array = np.zeros((m, n))
+    ones_array = np.ones((m, n))
+    assert zeros_array.ctypes.data % 16 == 0
+    assert ones_array.ctypes.data % 16 == 0
```
sidneymau, line 1774: These will break if numpy is _not_ guaranteeing that arrays are memory-aligned by default, which I gather was not previously the case. We can get around this by writing a function to explicitly make memory-aligned arrays, but doing so may not be necessary

tests/test_image.py
```diff
@@ -1692,6 +1692,112 @@ def test_Image_binary_scalar_pow():
...
+@timer
+def test_Image_inplace_setter():
+    """Test that all six types of supported Images set correctly
+    """
+    # Test that setting an Image initialized with an int array to a float array
+    # doesn't downcast and truncate values
+    farray = np.array([[1.2, 3.8], [5.6, 6.7]])
+    image = galsim.Image(np.zeros((2, 2), dtype=int))
+    image.array = farray
+    np.testing.assert_array_equal(image.array, farray)
```
sidneymau, line 1704: I'm not sure what the preference for broadcasting Image arrays is—this tests that an int array gets successfully cast to a float array, which may not be the desired behavior

rmjarvis: It is not desired behavior.

rmjarvis (changes requested):

rmjarvis:
@sidneymau This has been dormant for a couple months now.  Did you want to try to finish it, or should we abandon this effort?

sidneymau:
Thanks for bumping this—I'm happy to consider this abandoned for now and have the PR closed and the branch pruned. I may pick it up again when I have more time, but I don't expect that to be any time soon.

PR #1277, opened 2024-03-06
Spiral profile

esheldon:
Adds a new profile `Spiral` that creates a logarithmic spiral object, represented by point sources.

The points are created in three dimensions in a disk, with possible inclination along the line of sight and rotations in the plane of the sky.  The number of spiral arms is also configurable.

Some examples

two spiral arms
-----------------------

Points in 3d as well as a rendered image

Edge on

More arms
----------------

Full scenes
-----------------
rgb-1021620208
rgb-119650250

### DIFF
--- a/tests/test_spiral.py
+++ b/tests/test_spiral.py
@@ -1,0 +1,435 @@
+import numpy as np
+import galsim
+from galsim_test_helpers import (
+    check_basic,
+    do_shoot,
+    do_kvalue,
+    check_pickle,
+)
+from galsim.errors import (
+    GalSimValueError,
+    GalSimRangeError,
+    # GalSimIncompatibleValuesError,
+)
+import pytest
+
+
+def _get_example_kw():
+    return {
+        'npoints': 100,
+        'half_light_radius': 1.0,
+        'flux': 3.5,
+        'narms': 3,
+        'angle_fuzz': 0.07,
+        'xy_fuzz': 0.09,
+        'rel_height': 0.13,
+        'inclination': 45 * galsim.degrees,
+        'rotation': -20 * galsim.degrees,
+    }
+
+
+def test_spiral_defaults():
+    """
+    Create a spiral galaxy and test that the getters work for
+    default inputs
+    """
+
+    seed = 1234
+    npoints = 100
+    hlr = 1.0
+    rng = galsim.BaseDeviate(seed)
+    sp = galsim.Spiral(npoints, half_light_radius=hlr, rng=rng)
+
+    assert sp.npoints == npoints, (
+        'expected npoints==%d, got %d' % (npoints, sp.npoints)
+    )
+    assert sp.input_half_light_radius == hlr, (
+        'expected hlr==%g, got %g' % (hlr, sp.input_half_light_radius)
+    )
+
+    nobj = sp.points.shape[0]
+    assert nobj == npoints, (
+        'expected %d objects, got %d' % (npoints, nobj)
+    )
+
+    pts = sp.points
+    assert pts.shape == (npoints, 3), (
+        'expected (%d,2) shape for points, got %s' % (npoints, pts.shape)
+    )
+    np.testing.assert_almost_equal(sp.centroid.x, pts[:, 0].mean())
+    np.testing.assert_almost_equal(sp.centroid.y, pts[:, 1].mean())
+
+    sp.calculateHLR()
+    sp.calculate_e1e2()
+
+    gsp = galsim.GSParams(xvalue_accuracy=1.e-8, kvalue_accuracy=1.e-8)
+    rng2 = galsim.BaseDeviate(seed)
+    sp2 = galsim.Spiral(npoints, half_light_radius=hlr, rng=rng2, gsparams=gsp)
+
+    assert sp2 != sp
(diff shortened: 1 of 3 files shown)


cwwalter:
Cool!  Would you still add a bulge component separately?

esheldon:
Yes, that can be added separately.

In the color images above, the galaxies have up to three parts
- spiral using the `Spiral` class
- knots, using the exact same `Spiral` profile parameters but with fewer points, more flux per point, and bluer color
- bulge as a DeVauc profile

rmjarvis:
This is quite cool, Erin!  The images look great.

esheldon:
Some things for discussion
* Mike mentioned that nowadays we should be able to replace the `_sbp` property with something better.
* The use of points in 3d to represent the profile makes the code simple, but a lot of points are needed to make it look right when the spiral is well resolved.  This can make it slow to render, even with photon shooting.  Is there some way to hack up an analytic form?

rmandelb:
Thanks Erin - this is super cool.  I was going to ask about computational expense and then you commented on it yourself... it feels to me like there must be some trick one could use, relying on the fact that each point is fundamentally the same exact light profile after PSF convolution, modulo sub-pixel shifts.  (I assume we are still making the approximation that there is a single well-defined PSF for the entire `Spiral` object... i.e. no PSF spatial variation across the `Spiral` object and no color gradients within the `Spiral`)

tests/test_spiral.py
```diff
@@ -0,0 +1,512 @@
...
+
+    pts = sp.points
+    assert pts.shape == (npoints, 3), (
+        'expected (%d,2) shape for points, got %s' % (npoints, pts.shape)
+    )
+    np.testing.assert_almost_equal(sp.centroid.x, pts[:, 0].mean())
+    np.testing.assert_almost_equal(sp.centroid.y, pts[:, 1].mean())
+
+    sp.calculateHLR()
+    sp.calculate_e1e2()
```
rmjarvis, line 63: Should probably assert something about the return values of these two, rather than just that they run without error.

tests/test_spiral.py
```diff
@@ -0,0 +1,512 @@
...
+    assert sp2 == sp.withGSParams(gsp)
+    assert sp2 == sp.withGSParams(xvalue_accuracy=1.e-8, kvalue_accuracy=1.e-8)
+
+    # Check that they produce identical images.
+    psf = galsim.Gaussian(sigma=0.8)
+    conv1 = galsim.Convolve(sp.withGSParams(gsp), psf)
+    conv2 = galsim.Convolve(sp2, psf)
+    im1 = conv1.drawImage()
+    im2 = conv2.drawImage()
+    assert im1 == im2
```
rmjarvis, line 79: I guess this is asserting that xvalue_accuracy and kvalue_accuracy are not used by the Spiral class anywhere.  Seems more useful to test that something that does matter causes the image to be different.

tests/test_spiral.py
```diff
@@ -0,0 +1,512 @@
...
+
+    # Check that image is not sensitive to use of rng by other objects.
+    rng3 = galsim.BaseDeviate(seed)
+    sp3 = galsim.Spiral(npoints, half_light_radius=hlr, rng=rng3)
+    rng3.discard(523)
```
rmjarvis, line 84: I'm not sure this is an interesting test.  I think the rng is only used by the constructor.  Once the points are made, even if the rng wasn't duplicated, I think this would still work.  Perhaps a better test that you are sequestering the rng would be that copying it still produces the same image.
```
rng3.discard(523)
sp4 = sp3.withFlux(sp3.flux)
sp5 = sp3.copy()
sp6 = sp3.shift(0,0)
...
```
These should all be equivalent to sp3, and I think exercise the step of duplicating the rng.

tests/test_spiral.py
```diff
@@ -0,0 +1,512 @@
...
+    check_basic(conv1, 'Spiral')
+
+    im = galsim.ImageD(256, 256, scale=0.2)
+
+    do_shoot(conv1, im, 'Spiral')
+    do_kvalue(conv1, im, 'Spiral')
+
+    check_pickle(sp, irreprable=True)
+    check_pickle(conv1, irreprable=True)
+    check_pickle(conv1, lambda x: x.drawImage(scale=1), irreprable=True)
```
rmjarvis, line 101: I'd like these to all work with irrerable=False.

rmjarvis (changes requested):
Thanks Erin.  I have some requests in detail, but the overall code looks very good.  This will be a nice addition!

rmandelb:
Hi @esheldon - thanks have been quiet on this PR for a while, so I wanted to check in and let you know that we've been discussing tagging GalSim v2.6 in about a month from now.  Do you want this feature in the next tagged version, and if so, do you anticipate that timeline working for you to address the review?

esheldon:
I started this as a fun project with my son while on vacation. I haven't had any time since to work on it.

If I don't get to Mike's comments in time, please don't wait for me.

rmandelb:
Ah, the vacation project; I can relate to that phenomenon.  Thanks for letting us know... we won't wait on this, but if you do find time for it, we can definitely include it.

Issue #1264, opened 2023-12-05
Construct images without sanity checks in `_convertMask` function in the HSM module


arunkannawadi:
When measuring the moments using `FindAdaptiveMom`, I find that a significant chunk of time is taken by the `_convertMask` function, which arises from various Image instantiations. Given that this is an internal function, I expect that the inputs provided to create an Image instance would always pass these checks and can directly use the `_Image`.

For reference, 38 out of 186 seconds is spend in `_convertMask` when calling `FindAdaptiveMom`.

rmjarvis:
I think the main thing is probably that there are a couple of checks currently that the mask is sane (no negative values and not all zero).  So probably adding an option to skip those will speed this up a lot.

rmjarvis:
Probably simplest would be to add a leading underscore version of this `_FindAdaptiveMom` that skips all the sanity checks and requires that `object_image` and `weight` are already the right dtype and everything.

arunkannawadi:
That sounds like the least disruptive approach to me (not saying that modifying `_convertMask` would have been), but I could see a user potentially calling `FindAdaptiveMom` with inconsistent inputs. I'm happy to code this up, quantify the performance improvements and raise a PR here.

erykoff:
An additional public interface to `observed_e1` and `observed_e2` would also be nice.  And the current return for `observed_shape` should probably use the optimized `_Shape` rather than `Shape`.

arunkannawadi:
Closing this issue as completed.

PR #1279, opened 2024-03-08
Optimize bits of HSM code

arunkannawadi:
This PR addresses a lot of the concerns raised in #1264. I also separated conversion code from code that checks things as much as possible. I can add some additional unit tests if this all looks sensible.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,6 +1,6 @@
-.. image:: https://github.com/GalSim-developers/GalSim/workflows/GalSim%20CI/badge.svg
+.. image:: https://github.com/GalSim-developers/GalSim/workflows/GalSim%20CI/badge.svg?branch=main
         :target: https://github.com/GalSim-developers/GalSim
-.. image:: https://codecov.io/gh/GalSim-developers/GalSim/branch/master/graph/badge.svg
+.. image:: https://codecov.io/gh/GalSim-developers/GalSim/branch/master/graph/badge.svg?branch=main
         :target: https://codecov.io/gh/GalSim-developers/GalSim
 .. image:: https://img.shields.io/badge/astro--ph.IM-1407.7676-B31B1B.svg
         :target: https://arxiv.org/abs/1407.7676
--- a/galsim/hsm.py
+++ b/galsim/hsm.py
@@ -563,7 +563,7 @@ def _convertImage(image):
 def EstimateShear(gal_image, PSF_image, weight=None, badpix=None, sky_var=0.0,
                   shear_est="REGAUSS", recompute_flux="FIT", guess_sig_gal=5.0,
                   guess_sig_PSF=3.0, precision=1.0e-6, guess_centroid=None,
-                  strict=True, hsmparams=None):
+                  strict=True, check=True, hsmparams=None):
     """Carry out moments-based PSF correction routines.
 
     Carry out PSF correction using one of the methods of the HSM package (see references in
@@ -667,6 +667,9 @@ def EstimateShear(gal_image, PSF_image, weight=None, badpix=None, sky_var=0.0,
                         ``False``, then information about failures will be silently stored in
                         the output ShapeData object. [default: True]
 
+        check:          Check if the object_image, weight are in the correct format and valid.
+                        [default: True]
+
         hsmparams:      The hsmparams keyword can be used to change the settings used by
                         `EstimateShear` when estimating shears; see `HSMParams` documentation
                         for more information. [default: None]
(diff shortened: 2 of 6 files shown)


galsim/hsm.py
```diff
@@ -127,9 +130,17 @@ def image_bounds(self): return BoundsI(self._data.image_bounds)
...
+        return self._data.observed_e1
+
+    @property
+    def observed_e2(self):
+        return self._data.observed_e2
+
     @property
     def observed_shape(self):
-        return Shear(e1=self._data.observed_e1, e2=self._data.observed_e2)
+        return Shear(e1=self.observed_e1, e2=self.observed_e2)
```
arunkannawadi, line 143: @erykoff wanted this to use `galsim._Shear` here, but that requires conversion from `e` to `g` before we invoke `_Shear` and might not be all the worthwhile.

rmjarvis: If you want to optimize this too, you could try:
```
e = self._data.observed_e1 + 1j*self._data.observed_e2
g = e / (1 + np.sqrt(1.-np.abs(e)**2))
return _Shear(g)
```
I suspect that's a little faster, but might just be nibbling at the edges.

galsim/hsm.py
```diff
@@ -661,8 +680,10 @@ def EstimateShear(gal_image, PSF_image, weight=None, badpix=None, sky_var=0.0,
     # prepare inputs to C++ routines: ImageF or ImageD for galaxy, PSF, and ImageI for weight map
     gal_image = _convertImage(gal_image)
     PSF_image = _convertImage(PSF_image)
-    weight = _convertMask(gal_image, weight=weight, badpix=badpix)
     hsmparams = HSMParams.check(hsmparams)
+    if check:
+        _checkWeightAndBadpix(gal_image, weight=weight, badpix=badpix)
```
rmjarvis, line 685: You need to add tests of `check=False`.  (For both functions.)

rmjarvis (approved):
LGTM, module a small coverage drop.  And thanks for the typo/doc fixes.

galsim/hsm.py
arunkannawadi: Of course! This was not quite what we discussed in #1264 (we talked about going for a `_FindAdaptiveMomView` there) which is why I first wanted to run this by you before I wrote the tests for the two cases.

galsim/hsm.py
arunkannawadi: In the LSST DM code base, the reason to get `observed_shape` was because there was no way to get `observed_e1` directly up until now. Now that we can access them directly, I honestly don't think using `_Shear` is necessary.

galsim/hsm.py
```diff
@@ -779,6 +800,8 @@ def FindAdaptiveMom(object_image, weight=None, badpix=None, guess_sig=5.0, preci
                             ``GalSimHSMError`` exception if shear estimation fails.  If set to
                             ``False``, then information about failures will be silently stored in
                             the output ShapeData object. [default: True]
+        check:              Check if the object_image, weight are in the correct format and valid.
```
arunkannawadi, line 803: I missed out on `badpix` here and will include that as well.

galsim/hsm.py
rmjarvis: Sure.  This seems like a reasonable API to me.

arunkannawadi:
OK, something completely unrelated seems to be failing. `test_interpolateImage.py` should set up some global variables but that don't seem to be recognized in the test cases. All tests pass locally on my machine.

README.rst
```diff
@@ -1,6 +1,6 @@
-.. image:: https://github.com/GalSim-developers/GalSim/workflows/GalSim%20CI/badge.svg
+.. image:: https://github.com/GalSim-developers/GalSim/workflows/GalSim%20CI/badge.svg?branch=main
```
arunkannawadi, line 1: I also updated this link to show the badge corresponding to the run on the `main` branch, typically after merging a feature branch. I don't suppose you want to show case that tests on a development branch failed in the repo's README.

rmjarvis:
For now, just pin pytest to <8.  They changed something with pytest v8 about how they run (or don't!) setup() before the tests functions.  I haven't dug in to figure out what to change to make it work with v8 yet.

galsim/hsm.py
```diff
@@ -478,6 +489,27 @@ def __setstate__(self, d):
...
+        if weight.bounds != image.bounds:
+            raise GalSimIncompatibleValuesError(
+                "Weight image does not have same bounds as the input Image.",
+                weight=weight, image=image)
+                # also make sure there are no negative values
+
+        if np.any(weight.array < 0):
+            raise GalSimValueError("Weight image cannot contain negative values.", weight)
+
+    if badpix and badpix.bounds != image.bounds:
```
rmjarvis, line 507:
```suggestion
    if badpix is not None and badpix.bounds != image.bounds:
```

rmjarvis:
Test suite looks good.  Thanks.

PR #1281, opened 2024-04-04
Fix bug in using eval with templates not at top level

rmjarvis:
I was trying to craft a graceful solution to ImSim Issue [#452](https://github.com/LSSTDESC/imSim/issues/452), and I think I did, but I ran into a bug in GalSim, which this PR fixes.

Basically, if there is a template embedded in one of the fields of the config dict, and that template uses an eval, then Python doesn't like it, since you end up changing the size of the overall config dict due to the eval_gdict item.  Using `list(config.items())` fixes this issue.


jmeyers314 (approved):
LGTM!

PR #1282, opened 2024-04-05
Fix CelestialWCS._local computation of local jacobian when position is near ra=0

rmjarvis:
@esheldon ran into a bug in the computation of the local Jacobian near ra=0 when using a CelestialWCS.  It's the normal thing about 0 and 2pi not being close to each other, so the finite difference estimate of the jacobian ends up badly wrong.

This bug doesn't affect GSFitsWCS, which has its own implementation of this function.  So that's probably why it hasn't surfaced earlier, but Erin's use case involve a wrapper of the LSST DM wcs class, where this finite difference estimate is used.  The fix is (as usual) to wrap the values to be near one particular value, so the finite differences do what they are intended to do.


esheldon (approved):
This fixes an issue I was seeing in lsst psf tests

Issue #1284, opened 2024-04-29
Only photons from the final object in a `ChromaticSum` are saved


welucas2:
When `save_photons = True`, drawing a `ChromaticObject` returns all the photons requested. However, if those objects are combined into a `ChromaticSum` only the photons drawn for the final object are returned:

```python
import galsim

rng = galsim.BaseDeviate(215324)
sed = galsim.SED('vega.txt', wave_type="nm", flux_type="fphotons")
bandpass = galsim.Bandpass("LSST_r.dat", wave_type="nm")

# Create a bulge and disk and draw with 1000 photons each.
bulge = galsim.Sersic(n=3, half_light_radius=0.8) * sed
disk = galsim.Exponential(half_light_radius=1.4) * sed
bulge_image = bulge.drawImage(method="phot", n_photons=1000, save_photons=True, bandpass=bandpass, rng=rng)
disk_image = disk.drawImage(method="phot", n_photons=1000, save_photons=True, bandpass=bandpass, rng=rng)

# All 1000 photons are returned for each of the separate bulge and disk.
print(bulge_image.photons, disk_image.photons)

# Create a ChromaticSum object combining the bulge and disk then draw it with 1000 photons.
gal = bulge + disk
gal_image = gal.drawImage(method="phot", n_photons=1000, save_photons=True, bandpass=bandpass, rng=rng)

# The ChromaticSum's photon array only contains 511 photons.
print(gal_image.photons)
```

`ChromaticSum.drawImage` loops through the individual components, [drawing them in sequence](https://github.com/GalSim-developers/GalSim/blob/9f17d317e7d082f153497762f964a5c646a980ce/galsim/chromatic.py#L2400C1-L2401C94). The photons are actually saved in `GSObject.drawImage` at the [very end of the function](https://github.com/GalSim-developers/GalSim/blob/9f17d317e7d082f153497762f964a5c646a980ce/galsim/gsobject.py#L1836C9-L1837C36) but simply does `image.photons = photons`, replacing any previously saved object's photon array with the current one.

I think either a full-sized photon array should be allocated in `ChromaticSum.drawImage` and then filled in `GSObject.drawImage`, or else the latter should check if a photon array has already been stored and if so extend rather than overwrite it.

rmjarvis:
Fixed in #1289.

PR #1285, opened 2024-04-30
Save photons from the component objects of a ChromaticSum

welucas2:
This commit addresses #1284 by checking whether `image` already has a `photons` attribute at the end of `GSObject.drawImage`. If not, it simply assigns it the newly created photon array as previously. If it does exist, a new array with the correct length is created, the extant `image.photons` copied in at the beginning and the new `photons` after, and the new array assigned to `image.photons`.

This fix corrects the example in #1284 and also fixes previously incorrect output from the imSim photon pooling pipeline noted in LSSTDESC/imSim#424.

### DIFF
--- a/galsim/gsobject.py
+++ b/galsim/gsobject.py
@@ -1834,7 +1834,21 @@ class GSObject:
 
         image.added_flux = added_photons / flux_scale
         if save_photons:
-            image.photons = photons
+            # We need to check if image already contains photons from a previous call to drawImage,
+            # for example if doing a ChromaticSum.
+            if not hasattr(image, 'photons'):
+                # If not, then simply put the photons into the image.
+                image.photons = photons
+            else:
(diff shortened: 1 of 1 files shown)


rmjarvis:
@welucas2 You'll need to rebase this to the current main to get the fix to CI in #1288.

welucas2:
Thanks for the heads-up @rmjarvis, just rebased.

welucas2:
I'm certainly not familiar with all the use cases so I'll take that as the correct approach! I guess the imSim photon pooling is only concerned that the photon array coming back out again from `ChromaticSum` contains the complete set of photons drawn for the object as a whole. If doing it in `ChromaticSum`, maybe the full-sized array can be created and then photons coming back from `drawImage` can be copied in analogously to `added_flux`..?

rmjarvis:
See #1289 instead.

PR #1288, opened 2024-05-01
Fix fftw on new Mac GHA machines

rmjarvis:
The new MacOS runners on GitHub actions seem to have recently turned on "System Integrity Protection".  This means the way we had been setting the FFTW include and link directories didn't work right anymore.  We now need to explicitly set FFTW_DIR when doing pip install.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,7 @@ jobs:
                 py: [ 3.7, 3.8, 3.9, "3.10", 3.11, 3.12 ]
                 CC: [ gcc ]
                 CXX: [ g++ ]
+                FFTW_DIR: [ "/usr/local/lib" ]
 
                 # Add some other particular combinations to test
                 include:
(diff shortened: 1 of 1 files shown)


jmeyers314:
Yep.  I think the location homebrew installs to is different for MacOS x86 and MacOS ARM too if we ever want to enable both in CI.

jmeyers314 (approved):

rmjarvis:
Thanks Josh!

jmeyers314:
Also note that the system clang on the GHA runners (both x86 and ARM) don't include OpenMP (https://github.com/GalSim-developers/GalSim/actions/runs/8911049524/job/24471556168#step:13:513).  

I don't know if there's a "pro" side to disabling omp for CI, but to compile with it you can grab llvm from homebrew (which is the right thing to do for pypi uploads I think).  See bits around https://github.com/jmeyers314/batoid/blob/releases/0.6/.github/workflows/wheels.yml#L90-L97.

rmjarvis:
Yeah, we should include the llvm download when we do the arm wheels.

jmeyers314:
> Yeah, we should include the llvm download when we do the arm wheels.

I can't actually tell from the GHA logs, but I suspect we should do this for the x86 wheels too.  I.e., I suspect the MacOS wheels currently on PyPI are single-thread only.

Issue #1271, opened 2024-01-17
`save_photons` does not work for `ChromaticAiry`


jfcrenshaw:
When drawing a chromatic PSF using photon shooting, the `save_photons` parameter does not seem to work.

For example, the following achromatic code works fine:
```python
psf = galsim.Airy(lam=500, diam=8)
image = psf.drawImage(method="phot", n_photons=1000, save_photons=True)
photons = image.photons
```

However, the following chromatic code raises `AttributeError: 'Image' object has no attribute 'photons'`:
```python
sed = galsim.SED('vega.txt', wave_type="nm", flux_type="fphotons")
bandpass = galsim.Bandpass("LSST_r.dat", wave_type="nm")

psf_chromatic = galsim.ChromaticAiry(lam=500, diam=8) * sed
image = psf_chromatic.drawImage(
    method="phot",
    n_photons=1000,
    save_photons=True,
    bandpass=bandpass,
)
photons = image.photons
```

rmjarvis:
Fixed in #1289

PR #1289, opened 2024-05-06
Fix chromatic save photons

rmjarvis:
Fixes a few issues with Chromatic objects' use of the save_photons=True option.

Specifically, issues #1271 and #1284.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,7 +72,7 @@ jobs:
                   echo 'event.after: ${{ github.event.after }}'
 
             - name: Set up Python ${{ matrix.py }}
-              uses: actions/setup-python@v4
+              uses: actions/setup-python@v5
               with:
                 python-version: ${{ matrix.py }}
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,9 +130,7 @@ jobs:
                 brew link --overwrite fftw gcc wget || true
                 brew install eigen || true
                 brew link --overwrite eigen || true
-                brew cask install gfortran || true
-                brew install libav || true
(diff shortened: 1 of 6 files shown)


rmjarvis:
@jmeyers314 When working up the test suite for this, I had trouble with chromatic objects that use atRedshift.  In particular ones that use the base class's implementation of atRedshift to just use `ChromaticTransformation(obj, redshift=redshift)`.  It's not even clear what this *should* do in the general case.

When the object is simply `gal * sed`, it's straightforward.  Basically just apply atRedshift to the sed.  And some slightly more complicated objects also work fine.  But for instance what should atRedshift do to this object:
`galsim.Convolve(disk * disk_SED, optical_psf).atRedshift(1.2)`?  Only apply the redshift to disk_SED?  Or also shift the wavelength dependence of the PSF?  We have a test_atredshift in test_chromatic.py that enforces the latter, which seems weird to me.  And I couldn't find an elegant way to implement the same thing for photon shooting.  (That test uses FFT rendering.)

Anyway, my eventual takeaway was that maybe we should get rid of the whole option to call atRedshift on objects rather than just on SEDs.  I think that would still cover all the cases anyone actually needs, and it would avoid the ambiguity about what convolutions at a redshift mean, since the user would be explicit about it.  But do you know if that would get rid of some use case that needs this feature?

jmeyers314:
LGTM.  I agree that we should only permit `.atRedshift` on SEDs.  Redshifting the wavelength dependence of a PSF feels like an error.

jmeyers314 (approved):

rmjarvis:
Thanks Josh.  I'll do the `atRedshift` deprecation on a separate PR to keep things clean.

PR #1291, opened 2024-05-16
Deprecate ChromaticObject.atRedshift

rmjarvis:
When working on #1289, I realized that the ChromaticObject.atRedshift feature doesn't always make much sense.  

If the ChormaticObject is fairly simple, like `(gal * sed).atRedshift(z)` then `atRedshift` just applies the redshift to the SED and everything is fine.  But in that case, it's just as easy for the user to call `atRedshift` on the SED directly: `gal * sed.atRedshift(z)`.

However, if object is a convolution of a chormatic galaxy with a chromatic PSF, then it's unclear what `atRedshift` should even mean?  Probably you just want to shift the galaxy part, but the implementation also changed the wavelength dependence of the PSF part, which seems likely an error.  Rather than try to find an implementation that does this sensibly, it seems more straightforward to just require the user to apply the redshift to the galaxy SED instead.

So this PR deprecates the use of `atRedshift` on Chromatic objects as well as putting a `redshift` parameter in the `gal` field of a config dict, unless the galaxy in achromatic (in which case redshift is just an attribute, so it's not problematic).


jmeyers314 (approved):
LGTM!

PR #1297, opened 2024-06-05
Updates to be compatible with numpy 2.0

rmjarvis:
Numpy made a few changes to their API with version 2.0.  This PR addresses the places where they impact GalSim code.

1. np.array(a, copy=False) is now an error if a copy is required for the dtype.  Using `copy=None` now does the equivalent to what `copy=False` used to do.  But in most cases, it's better to just use `np.asarray(a)`, which is equivalent in both versions.
2. There is one case where we have a more complicated decision about whether to copy, so in that case we need to use an explicit version check to get the right kwarg.
3. They changed the implicit dtype conversion for values in e.g. += and similar operations.  For GalSim purposes, this mostly just required changing test expectations.  E.g. using `Image(..., dtype=float)` when we want to test accuracy at double precision.
4. One other gotcha is that Python int cannot be implicitly assigned to np.uint types.  This feels like a bug to me, since np.int types do work.  Just not a regular Python int.  cf. numpy issue https://github.com/numpy/numpy/issues/26483  However, it is likely that in most cases where this would show up in user code, it is likely some kind of user error.  So I think it's best to not work around it, but rather just let the exception be raised.  We explicitly avoid it by using np.int types in tests where it would otherwise be a problem.
5. Related to the above, `Image.__neg__` now uses np.int64(-1), rather than just -1, so it will continue to work the same way it used to for unsigned dtypes.
6. Finally, np.trapz is deprecated.  The new name is `np.trapezoid`. We mostly don't use it, since our own implementation is faster, but there were a few places in the tests that needed to be updated.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,7 @@ jobs:
                       #py: pypy-3.7
                       #CC: gcc
                       #CXX: g++
+
         env:
             FFTW_DIR: ${{ matrix.FFTW_DIR }}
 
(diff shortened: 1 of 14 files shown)


beckermr (approved):
Thanks for the hard work!

PR #1298, opened 2024-06-06
Update wheel.yml to build ARM wheels and switch to trusted PyPi upload rather than using a token.

rmjarvis:
In preparation for releasing version 2.5.3, I added arm wheels to our wheels.yaml script.
Also, PyPi now prefers using trusted publishing.  (cf. https://docs.pypi.org/trusted-publishers/adding-a-publisher/) So I switched to that as well.

This is hard to confirm before actually running the script, which will happen when we make the release tag v2.5.3.  But @jmeyers314 if you can look through it once to see if there are any errors that you can notice in advance, that would be appreciated.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -13,19 +13,20 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: actions/setup-python@v3
+      - uses: actions/setup-python@v5
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.16.2
+        uses: pypa/cibuildwheel@v2.16.5
         env:
           CIBW_BUILD: "*manylinux*"
           CIBW_ARCHS: auto64
-          CIBW_SKIP: cp36* pp*
+          CIBW_SKIP: cp36* cp37* pp*
           # I think yum might always work here.  But leave all options available.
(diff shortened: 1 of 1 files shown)


rmjarvis:
Note also that I followed your lead in removing python 3.7 from our list of wheel builds.

.github/workflows/wheels.yml
```diff
@@ -75,10 +106,16 @@ jobs:
     # Just need to build sdist on a single machine
```
jmeyers314, line 106: I think you'll need to update the above to
`needs: [build_linux_wheels, build_musl_wheels, build_macosx_intel_wheels, build_macosx_arm_wheels]`

.github/workflows/wheels.yml
```diff
@@ -34,39 +35,69 @@ jobs:
...
+      - uses: actions/upload-artifact@v5
         with:
+          name: whl-musl
           path: ./wheelhouse/*.whl
 
-  build_macosx_wheels:
-    name: Build wheels on macosx
+  build_macosx_intel_wheels:
+    name: Build wheels on MacOS Intel
     runs-on: macos-latest
```
jmeyers314, line 56: You _might_ need to change this to macos-13 to get an intel runner.  It's possible an ARM runner will cross-compile though, just need to try and see I guess.

jmeyers314 (approved):
I think there's at least one bug.  But mostly just have to try this and see.

rmjarvis:
Thanks!  I think you're right on both counts.

Issue #1299, opened 2024-07-09
comma/space delimited PositionD string parsed incorrectly in config layer


jmeyers314:
According to https://galsim-developers.github.io/GalSim/_build/html/config_values.html#pos-value, I ought to be able to use:

```yaml
...
image:
  image_pos: 123.4, 567.8
```

to set the image position for an object.  But this gets parsed by [_GetPositionValue](https://github.com/GalSim-developers/GalSim/blob/releases/2.5/galsim/config/value.py#L450) to (1.0, 2.0) because strings are indexable.

Reproducer:
```python
import galsim
config = {'image_pos': '123.4, 567.8'}
print(galsim.config.ParseValue(config, 'image_pos', config, galsim.PositionD)[0])
assert galsim.config.ParseValue(config, 'image_pos', config, galsim.PositionD)[0] == galsim.PositionD(123.4, 567.8)
```

rmjarvis:
Interesting.  It seems to only be a problem if the first number is >= 10.  Our unit test for this uses '0.1, 0.2'.  But like you said, it fails if the first number is e.g. 123.4, since trying to parse the first two elements of the string succeeds.

FWIW, a workaround for now is to add $ at the front.
```
config = {'image_pos': '$123.4, 567.8'}
```

PR #1300, opened 2024-07-09
Fix PositionD parsing in config from strings like '123.1, 234.2', which had been doing the wrong thing.

rmjarvis:
@jmeyers314 noticed in #1299 that a string like '123.4, 567.8' doesn't get parsed correctly.  This PR fixes the implementation to work correctly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,13 +141,6 @@ jobs:
                 if ! test -d $HOME/des_data || ! test -f $HOME/des_data/DECam_00154912_01.fits.fz; then wget --no-check-certificate http://www.sas.upenn.edu/~mjarvis/des_data.tar.gz; tar xfz des_data.tar.gz -C $HOME; fi
                 ln -s $HOME/des_data examples/des/
 
-            - name: Install numpy 2.0.0rc2
-              # Just on python 3.12, use the rc2 branch of numpy 2.0.0.
-              # XXX: Once the default numpy is >= 2.0, remove this.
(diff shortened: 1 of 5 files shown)


jmeyers314 (approved):
LGTM.  Thanks!

jmeyers314:
Although, it just occurred to me this will still fail if given:

```yaml
pos : 12
```

=> `PositionD(1.0, 2.0)`.

Maybe that's bizarre enough to not bother defending against though.

rmjarvis:
```
pos : 12
```
That will give an appropriate error:
```
  File "/Users/Mike/GalSim/galsim/config/value.py", line 206, in ParseValue
    val = _GetPositionValue(param)
  File "/Users/Mike/GalSim/galsim/config/value.py", line 463, in _GetPositionValue
    raise GalSimConfigError("Unable to parse %s as a PositionD. Caught %s"%(param,e))
galsim.errors.GalSimConfigError: Unable to parse 12 as a PositionD. Caught 'int' object has no attribute 'split'
```
The one that will work inappropriately is
```
pos : "12"
```
which I think is probably bizarre enough not to bother defending.

jmeyers314:
Ah, right.  Yaml sees that as an int, not a str, so all is well.

rmjarvis:
Is there any urgency in getting this into a bugfix release?  Or should we let this sit until we have more bugs fixed?

jmeyers314:
No urgency I think.  The original workaround you proposed seems fine.

Issue #1253, opened 2023-10-16
Remove hidden scipy dependency


rmjarvis:
GalSim does not have a formal dependency on scipy. For good reason -- that brings in a pretty large dependency tree.  (Historically for a better reason, that scipy was notoriously hard to install for many years -- but that reason doesn't apply so much after scipy v1.0.)

However, we have a hidden scipy dependency in FittedSIPWCS.  I think we can rewrite sections of that to use numpy.linalg.lstsq rather than scipy.optimize.least_squares.  

If we choose not to, we should instead add scipy as a formal requirement I think, since this functionality isn't something that would obviously pull in an additional dependency requirement, so I don't think users should expect this behavior.

PR #1305, opened 2024-07-17
Remove implicit scipy dependency.

jmeyers314:
Addresses #1253 .

Adds a basic Levenberg-Marquardt optimizer that can fill in for `scipy.linalg.least_squares` in simple cases (such as used in `FittedSIPWCS`).  It's a small bit slower on average, so we might want to consider making it a fallback option when scipy is unavailable though I didn't bother for now.  (The fitted sip unit tests takes 3.0 s now, vs. 2.25 s using scipy).

Along the way I also refactored our `if __name__ == __main__` test running code so you can use `-k testname` (similar to pytest) when running directly with `python`.  There are command line options for running the profiler too.

### DIFF
--- a/galsim/utilities.py
+++ b/galsim/utilities.py
@@ -1832,3 +1832,72 @@ class Profile:
         if self.reverse:  # pragma: no cover
             ps = ps.reverse_order()
         ps.print_stats(self.nlines)
+
+
+def least_squares(fun, x0, args=(), kwargs={}, max_iter=1000, tol=1e-9, lambda_init=1.0):
+    """Perform a non-linear least squares fit using the Levenberg-Marquardt algorithm.
+
+    Drop in replacement for scipy.optimize.least_squares when using default options,
+    though many fewer options available in general.
+
+    Parameters:
+        fun: Function which computes vector of residuals, with the signature
+             fun(params, *args, **kwargs) -> np.ndarray.
+        x0: Initial guess for the parameters.
+        args: Additional arguments to pass to the function.
+        kwargs: Additional keyword arguments to pass to the function.
+        max_iter: Maximum number of iterations.  [default: 1000]
+        tol: Tolerance for convergence.  [default: 1e-9]
+        lambda_init: Initial damping factor for Levenberg-Marquardt.  [default: 1.0]
+
+    Returns:
+        A named tuple with fields:
+            x: The final parameter values.
+            cost: The final cost (sum of squared residuals).
+    """
+    # JM: This is a tweaked version of a ChatGPT-generated implementation of
+    # Levenberg-Marquardt (cross-checked against the wikipedia page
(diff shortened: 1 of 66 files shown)


galsim/utilities.py
```diff
@@ -1821,3 +1821,68 @@ def __exit__(self, type, value, traceback):
...
+
+        if np.linalg.norm(new_residuals) < np.linalg.norm(residuals):
+            params = new_params
+            residuals = new_residuals
+            lambda_ /= 3  # reduce damping
+        else:
+            lambda_ *= 3  # increase damping
+
+        if np.linalg.norm(delta_params) < tol:
+            break
```
rmjarvis, line 1880: I assume you copied this algorithm from somewhere.  Should give the source.

tests/test_vonkarman.py
```diff
@@ -25,19 +25,13 @@
...
-        do_deltas = [False, True]
-    else:
-        lams = [500.0]
-        r0_500s = [0.2]
-        L0s = [25.0]
-        do_deltas = [False]
+    lams = [300.0, 500.0, 1100.0]
+    r0_500s = [0.05, 0.15, 0.3]
+    L0s = [1e10, 25.0, 10.0]
+    do_deltas = [False, True]
```
rmjarvis, line 34: How slow were the slow tests?  I'm assuming you checked this and found the "slow" ones were still fast enough to run all the time.

galsim/utilities.py
```diff
@@ -1821,3 +1821,68 @@ def __exit__(self, type, value, traceback):
...
+    Returns:
+        A named tuple with fields:
+            x: The final parameter values.
+            cost: The final cost (sum of squared residuals).
+    """
+    from collections import namedtuple
+    params = np.array(x0)
+    lambda_ = lambda_init
+
+    for _ in range(max_iter):
```
rmjarvis, line 1851: The code coverage miss is this loop never ending.  Should either add a test with a low max_iter, (e.g. 1 probably), or add `#pragma: no branch` to this line.

rmjarvis (approved):
A couple comments/questions, but this is essentially fine.

tests/test_vonkarman.py
jmeyers314: On my lap this this is 2.07 sec for all tests vs 0.07 sec for just the fast tests.  I can wrap this in a `if __name__ == '__main__'` block if you prefer.

rmjarvis: Sure.  I usually try to keep each test under 1 sec if possible for they pytest run.  There are some exceptions, but this seems like one where we don't need the slow tests to spot regressions.

galsim/utilities.py
jmeyers314: It's ChatGPT + wikipedia.  I can add a comment to that effect.

PR #1306, opened 2024-07-24
Fix #1302

### LINKED ISSUE #1302: Drawing error for ChromaticConvolution with pixel response
Noticed that Galsim throws an error when trying to draw a convolution between a chromatic PSF and the pixel response function.  Here is a code snippet showing the issue:

```
import galsim
import galsim.roman as roman

filt = 'F184'
bandpass = roman.getBandpasses()[filt]
vega_sed = galsim.SED('vega.txt', 'nm', 'flambda')
scale = roman.pixel_scale/4
psfInt = roman.getPSF(8, filt, n_waves=10,  pupil_bin=8)
pixel_response = galsim.Pixel(roman.pixel_scale)
star = galsim.DeltaFunction()*vega_sed
galaxy = galsim.Gaussian(sigma=0.2)*vega_sed
eff_psf = galsim.Convolve(psfInt, pixel_response) 
star_convolved = galsim.Convolve(star, eff_psf)
galaxy_convolved = galsim.Convolve(galaxy, eff_psf)
star_image = star_convolved.drawImage(bandpass, nx=128, ny=128, scale=scale, method = 'no_pixel')
galaxy_image = galaxy_convolved.drawImage(bandpass, nx=128, ny=128, scale=scale, method = 'no_pixel')
```

The vega SED here is simply used as an example SED, but running this will throw the same error for the bottom two lines: 
`AttributeError: 'Pixel' object has no attribute '_fiducial_profile'`. 

Curiously there is a practical workaround for to avoid this error if you replace the PSF as a sum of the PSF at discrete wavelengths:
```
waves = np.linspace(bandpass.blue_limit, bandpass.red_limit, 10)
psfInt = roman.getPSF(8, filt, wavelength=waves[0],  pupil_bin=8)
for wave in waves[1:]:
    psf = roman.getPSF(8, filt, wavelength=wave,  pupil_bin=8)
    psfInt += psf
```
 which does not result in an error. However, I don't think this implementation is correct.

Any comments or suggestions are greatly appreciated!

rmjarvis:
Fix an error found by @FedericoBerlfein that amounted to GSObjects missing a method needed to properly duck type as ChromaticObjects in some cases.  

Specifically, if an inseparable chromatic object is convolved with a GSObject (in Federico's case, a Pixel), then a bit of code tries to call `_fiducial_profile` on the separable components, which didn't work, since GSObject didn't have this method.  The fix is simply to add a trivial version of that method so the GSObject duck types properly in this context.

### DIFF
--- a/galsim/gsobject.py
+++ b/galsim/gsobject.py
@@ -450,6 +450,7 @@ class GSObject:
     def dimensionless(self): return True
     @property
     def wave_list(self): return np.array([], dtype=float)
+    def _fiducial_profile(self, bandpass): return bandpass.effective_wavelength, self
 
     # Also need these methods to duck-type as a ChromaticObject
     def evaluateAtWavelength(self, wave):
--- a/tests/test_chromatic.py
+++ b/tests/test_chromatic.py
@@ -2518,6 +2518,40 @@ def test_chromatic_fiducial_wavelength():
     with assert_raises(galsim.GalSimError):
         gal3.drawImage(bp)
 
+@timer
+def test_convolve_pixel():
+    """Check that convolving a chromatic object with a Pixel works
+    """
+    # In response to issue #1302
+    # The problem here had been that if some of the components of the convolution were
+    # inseparable, then when separating out the separable and inseparable components,
+    # the code called `obj._fiducual_profile(bandpass)` on the separable ones.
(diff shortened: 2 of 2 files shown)


jmeyers314 (approved):
LGTM

rmandelb (approved):
It's amazing the difference one line makes.

Issue #1018, opened 2019-03-21
Simulate prism observations


rmjarvis:
Rebekah Hounsell wants to use GalSim to simulate prism observations from Roman.  This mostly involves implementing an old (and closed because part of it was done) [issue of 1-d degenerate objects](#533).  In this case, the 1-d function would be the SED in the direction of the prism dispersion.

matroxel:
Working on this with SN PIT. First prototype prism model implemented as photon op in https://github.com/matroxel/roman_imsim/blob/prism/roman_imsim/photonOps.py

Need to update some roman module things to enable prism elements that we had previously never implemented, which will happen in this branch.

arunkannawadi:
I think I'll take this issue over from Troxel. I have computed sky backgrounds corresponding to prism and grism, but noticing that I don't get the current values for the other filters exactly. I know they are accurate to only 10% or so, but I was hoping to get the exact values reproduced by existing filters as a cross check. I'll raise a PR later today so we can see if they seem reasonable.

arunkannawadi:
Troxel tells me that @rmandelb is integrating the code to compute the sky backgrounds on the fly instead of using a lookup table. That's great!

But in the mean time, could we not use the sky backgrounds for `SNPrism` that are already here? I see that it was added in #1017 however, the sky level for prism and grism bands continue to be skipped here: https://github.com/GalSim-developers/GalSim/blob/releases/2.5/galsim/roman/roman_bandpass.py#L138-L139

Can we remove those lines?

rmandelb:
Hi @arunkannawadi - For now, for various practical reasons, I have integrated the code to compute sky backgrounds on the fly (based on Troxel's port of Chris's ETC code, with both of their permission) into a separate repo in which we are developing tools for Euclid-like simulations.  However, the idea is to make that public soon.  If we then want to move that out of the Euclid-like code base and into GalSim, that would be fine.

rmjarvis:
I believe some people have been implementing prism and grism action using photon operators.  Regardless, I think any effort in this direction will happen outside of GalSim proper, so we can close this issue.  (Of course, feel free to reopen if you think there is something useful to be done in GalSim iteself.)

arunkannawadi:
Yes, this is indeed happening outside of GalSim, but I found that some calls to the bandpasses should return the non-imaging Roman bands too to do that consistently. It's fine to close the issue as broadly stated, and I will raise a PR soon.

rmjarvis:
[...]

PR #1307, opened 2024-07-31
Include prism bandpasses in the roman module

arunkannawadi:
Addresses #1018

### DIFF
--- a/galsim/roman/roman_bandpass.py
+++ b/galsim/roman/roman_bandpass.py
@@ -101,7 +101,7 @@ def getBandpasses(AB_zeropoint=True, default_thin_trunc=True, include_all_bands=
 
     @returns A dictionary containing bandpasses for all Roman imaging filters.
     """
-    from . import collecting_area
+    from . import collecting_area, non_imaging_bands
 
     # Begin by reading in the file containing the info.
     datafile = os.path.join(meta_data.share_dir, "roman", "Roman_effarea_20210614.txt")
@@ -139,7 +139,7 @@ def getBandpasses(AB_zeropoint=True, default_thin_trunc=True, include_all_bands=
     bandpass_dict = {}
     # Loop over the bands.
     for index, bp_name in enumerate(data.dtype.names[1:]):
-        if include_all_bands is False and bp_name in ('Grism_0thOrder', 'Grism_1stOrder', 'SNPrism'):
+        if include_all_bands is False and bp_name in non_imaging_bands:
             continue
 
         # Initialize the bandpass object.
--- a/galsim/roman/roman_bandpass.py
+++ b/galsim/roman/roman_bandpass.py
@@ -135,7 +135,7 @@ def getBandpasses(AB_zeropoint=True, default_thin_trunc=True, **kwargs):
     # Loop over the bands.
     for index, bp_name in enumerate(data.dtype.names[1:]):
         # Need to skip the prism and grism (not used for weak lensing imaging).
-        if bp_name=='SNPrism' or bp_name=='Grism_1stOrder' or bp_name=='Grism_0thOrder':
+        if bp_name=='Grism_1stOrder' or bp_name=='Grism_0thOrder':
             continue
 
         # Initialize the bandpass object.
@@ -154,9 +154,10 @@ def getBandpasses(AB_zeropoint=True, default_thin_trunc=True, **kwargs):
             bp = bp.withZeropoint('AB')
 
         # Store the sky level information as an attribute.
-        bp._ecliptic_lat = ecliptic_lat
-        bp._ecliptic_lon = ecliptic_lon
-        bp._sky_level = sky_data[2+index, :]
+        if bp_name!='SNPrism':
+            bp._ecliptic_lat = ecliptic_lat
+            bp._ecliptic_lon = ecliptic_lon
+            bp._sky_level = sky_data[2+index, :]
 
         # Add it to the dictionary.
         bp.name = bp_name if bp_name != 'W149' else 'W146'
--- a/galsim/roman/roman_bandpass.py
+++ b/galsim/roman/roman_bandpass.py
@@ -29,7 +29,7 @@ from .. import meta_data
 from ..errors import galsim_warn
 from .. import Bandpass, LookupTable
 
-def getBandpasses(AB_zeropoint=True, default_thin_trunc=True, **kwargs):
+def getBandpasses(AB_zeropoint=True, default_thin_trunc=True, include_all_bands=False, **kwargs):
     """Utility to get a dictionary containing the Roman ST bandpasses used for imaging.
 
     This routine reads in a file containing a list of wavelengths and throughput for all Roman
@@ -91,6 +91,9 @@ def getBandpasses(AB_zeropoint=True, default_thin_trunc=True, **kwargs):
                             use no thinning and truncation of bandpasses, or who want control over
                             the level of thinning and truncation, should have this be False.
                             [default: True]
+        include_all_bands:  Should the routine include the non-imaging bands (e.g., grisms)?
+                            This does not implement any dispersion physics by itself.
+                            [default: False]
         **kwargs:           Other kwargs are passed to either `Bandpass.thin` or
                             `Bandpass.truncate` as appropriate.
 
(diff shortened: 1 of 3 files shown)


arunkannawadi:
There's a lot of chatter about setuptools removing the test command. How do we want to fix the breakage?

rmjarvis:
Eugh.  I guess pin setuptools to <72 for now.  We use it to run the C++ tests.  I guess we'll need to come up with some other builder/runner for them.

rmjarvis:
BTW, are you planning to actually do something to implement prism observations?  This PR doesn't do anything except allow those bands to be selected.  I think that's a bad idea if we don't also do something to make them simulate the prism dispersion physics.

arunkannawadi:
I figured that somebody would be confused because the changes in the PR doesn't do justice to the title :-)

Troxel and I have generated prism images along with Roman SN PIT with these changes here, some in `skyCatalogs` [here](https://github.com/LSSTDESC/skyCatalogs/pull/104) using the dispersive photon operator [here](https://github.com/Roman-Supernova-PIT/galsim-slitless-spec/blob/main/slitlessspec.py). All of these come together in the `roman_imsim` package [here](https://github.com/matroxel/roman_imsim/tree/prism/roman_imsim). 

image


I'd still like to have this merged by v2.6 since someone at STScI is going to take this over from us and I'd like them to be able to use GalSim+skyCatalogs without needing any modifications. Apart from having a similar photon operator here in the unit tests, do you have anything in mind? I am not sure if this would be meaningful at all.

rmjarvis:
I feel like some kind of sanity check is warranted here.  I'm worried that someone who isn't using the roman_imsim code will just try to make and image with one of these as the filter, and then be upset that the image doesn't actually have any dispersion.  I don't know what the best place for it is though.

Maybe just add a logger warning along the lines of "Warning: selecting a dispersive bandpass.  This should be accompanied by a suitable photon operator (e.g. from roman_imsim) to actually implement the dispersive effect.  This compatibility is not checked."  

@rmandelb @matroxel What do you think?

matroxel:
We need to at least get the right thermal backgrounds for these bandpasses and an implementation of the zodiacal light that works before merging into main I'd think?

A warning would be good, and we should probably include an example dispersion photon op and some updated documentation in demo13 or a new demo?

arunkannawadi:
Do we have the thermal backgrounds from the project? For other short wavelength bands, they are unavailable and set to 0 here, so I followed the same. We have nominal zodical backgrounds through those bandpasses [here](https://github.com/GalSim-developers/GalSim/blob/releases/2.5/share/roman/roman_sky_backgrounds.txt) although they may not be precise enough for prism.

Warning + demo with dispersion photon operator sounds good. If we have to switch the backgrounds, then I'd suggest we keep the issue open but merge this PR with warning+demo.

matroxel:
The blue bands average to zero at some decimal place. The wide prism/grism bandpasses won't. This is something we'll have to ask about, its not on the Roman page that I found.

matroxel:
I guess we should be properly accounting for it vs wavelength like we will for the zodiacal light for the W filter and prism/grism. W filter is also not well represented right now.

arunkannawadi:
For some context, Roman SN PIT folks just want to be able to test their spectral extraction code from a reasonable-looking image, not necessarily realistic. Depending on when v2.6 is aimed for, getting realistic backgrounds may be out of scope. For this initial PR, my goal is to be able to get the bandpasses by calling `galsim.roman.getBandpasses()`, which is not possible at present.

.github/workflows/ci.yml
```diff
@@ -148,7 +148,7 @@ jobs:
 
                 # Do these three first to clarify potential conflicts
                 pip install -U numpy
-                pip install -U setuptools
+                pip install -U setuptools<72
```
rmjarvis, line 151:
```suggestion
                pip install -U "setuptools<72"
```

.github/workflows/wheels.yml
```diff
@@ -125,7 +125,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install -U pip
-          pip install -U numpy setuptools
+          pip install -U numpy setuptools<72
```
rmjarvis, line 128:
```suggestion
          pip install -U numpy "setuptools<72"
```

arunkannawadi: I didn't quite follow why these had to be installed first when they are anyway specified in the `requirements.txt`. Something is still trying to download the more recent `setuptools`.

rmjarvis: pyproject.toml.  Probably the pre-install isn't necessary anymore, but older versions of pip were pretty flaky when it came to installing things that were used for the build.  So it was safer to install numpy and setuptools by hand first to be sure.

rmjarvis (changes requested):
I think this PR is currently waiting on some kind of warning to let users know that selecting the Prism doesn't automatically give them Prism physics.  Probably pointing to the implementation in roman_imsim that does.

AFAIU, the other items mentioned above (thermal backgrounds zodiacal light, etc.) for the Prism/Grism are considered out of scope for this PR and would be implemented in a later one.  But please correct me if we want something better than just 0.0 for those now.

arunkannawadi:
I changed the PR title to reflect the code changes.

For the warnings, I could think of two ways - either include an optional argument to `getBandpasses` that will decide whether to fetch non-imaging bands and warn the user if they are fetched or modify the `__getitem__` method for the `bandpass_dict` that will warn if the keys corresponding to non-imaging bands are passed. I kinda like the second one, but I don't know if this would annoy users who might be used to just iterating over all the bands.

rmjarvis:
Here's an idea, which might be better anyway.  Add an option to `getBandpasses` like `include_nonimaging_bands=True`, which would include the Prism and Grism, which defaults to False.  And maybe then it would be sufficient to just put a warning in the doc string about how that isn't sufficient to get prism physics and where the user might go to get it.  At least then, it requires the user to be more intentional about selecting those bands, so maybe they would see the documentation about it.

tests/test_roman.py
```diff
@@ -322,7 +322,7 @@ def test_roman_backgrounds():
     # The routine should not allow us to look directly at the sun since the background there is high
     # (to understate the problem).  If no date is supplied, then the routine assumes RA=dec=0 means
     # we are looking at the sun.
-    bp_dict = galsim.roman.getBandpasses()
+    bp_dict = galsim.roman.getBandpasses(include_all_bands=True)
```
arunkannawadi, line 325: I know you normally like to keep the old tests as they were, but I _think_ this should be fine?

arunkannawadi (commented):
Do we have at least a ball park value for the backgrounds in these bands? I can add a note in the docstring appropriately.

tests/test_roman.py
rmjarvis: This change shouldn't be necessary.  However, you should add a test that the Prism/Grism bands aren't included yet.  Then redo with `include_all_bands=True` and check that they are included.

tests/test_roman.py
```diff
@@ -485,6 +485,20 @@ def test_roman_bandpass():
...
+    bp_imaging = galsim.roman.getBandpasses(AB_zeropoint=True)
+    bp_all = galsim.roman.getBandpasses(AB_zeropoint=True, include_all_bands=True)
+
+    # Check that the imaging bandpasses arein the all bandpasses
+    for key in bp_imaging:
+        assert key in bp_all
+
+    assert 'Grism_0thOrder' in bp_all
+    assert 'Grism_1stOrder' in bp_all
+    assert 'SNPrism' in bp_all
```
rmjarvis, line 501: Or do it here, adding the converse asserts that these are *not* in bp_imaging.

rmjarvis (approved):

rmjarvis:
I'm happy with this now.  @matroxel or @rmandelb did either of you want to take a look?

rmjarvis:
@arunkannawadi Looks like you might need to rebase it to the current main.  GitHub is complaining.

galsim/roman/roman_bandpass.py
```diff
@@ -134,8 +139,7 @@ def getBandpasses(AB_zeropoint=True, default_thin_trunc=True, **kwargs):
     bandpass_dict = {}
     # Loop over the bands.
     for index, bp_name in enumerate(data.dtype.names[1:]):
-        # Need to skip the prism and grism (not used for weak lensing imaging).
-        if bp_name=='SNPrism' or bp_name=='Grism_1stOrder' or bp_name=='Grism_0thOrder':
+        if include_all_bands is False and bp_name in ('Grism_0thOrder', 'Grism_1stOrder', 'SNPrism'):
```
arunkannawadi, line 142: Do we want to save this list somewhere (and reuse them in the tests may be) instead of hardcoding them here?

arunkannawadi: I could imagine this going in `galsim/roman/__init__.py` as `nonimaging_bands` or something.

rmjarvis: Sure.

PR #1308, opened 2024-08-02
Fix Python 3 compatibility, remove some unused modules

AlexKurek:
Tests:
`732 passed, 1 skipped in 225.68s (0:03:45)`
`Test_photon_array.py ......s...........`

### DIFF
--- a/devel/brighter-fatter/bf_plots_07aug18.py
+++ b/devel/brighter-fatter/bf_plots_07aug18.py
@@ -34,7 +34,6 @@ latest 'hole20' branch of the Poisson_CCD22 code
 import sys
 import os
 import math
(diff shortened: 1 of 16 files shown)


AlexKurek:
These are different results when Im running it locally. Maybe I should back out of trying to commit to GalSim; Im not advanced enough to fix this. Sorry for the confusion.

rmjarvis:
These changes are all in the devel directory.  That is (as the name implies) a sandbox area for developers.  Not code that needs to be maintained.  The files you are updating are at best of historical interest at this point to document how various things were developed a long time ago.  I'm inclined to reject this PR as unnecessary.

PR #1309, opened 2024-08-08
Update random_seed behavior in config to be less confusing.

rmjarvis:
@kailicao discovered an error in the recent Roman image simulations where the random seed was not updating for each galaxy as expected.  This only affected Poisson noise on the flux values (the "realized flux"), since most values were from SkyCatalog, not generated from an rng.  But the statistics of the fluxes ended up messed up.

@matroxel and I tracked down the problem to the random_seed specification in the config file, which used a dict to evaluate a value based on the input file.  This was intended to be the initial seed, to be converted into a sequence as usual.  But the ([documented](http://galsim-developers.github.io/GalSim/_build/html/config_image.html#image-field-attributes)) behavior in that case is to take the user's dict specification as is, and not convert it.  This is confusing behavior, since most of the time evaluation like this is equivalent to setting a literal value.  

Troxel also noticed that imsim's example config has the same problem, although the config files used by @jchiang87 for the recent simulation runs didn't use that.  (It used a string eval value for the seed.)  So imsim users may also be getting wrong random values currently.

So I changed the behavior to always convert the first random_seed in the list (or only one if not a list) into the sequence in the usual way.  Any other items in the list are evaluated as the user specifies.  This is technically an API change, since it's possible that some users have been relying on the old behavior to leave dict specifications as is and convert int or str values.  However, (a) I think it's unlikely that anyone was using this feature other than to have the first item be normal and later items be dicts, and (b) we know of two accidental erroneous usages mentioned above, and it wouldn't surprise me if there were others.  So this is more like a bug fix than a feature change.  Anyway, I document it as an API change nonetheless to warn users who might be relying on the old behavior.

### DIFF
--- a/docs/config_image.rst
+++ b/docs/config_image.rst
@@ -17,17 +17,7 @@ Some attributes that are allowed for all image types are:
 * ``random_seed`` = *int_value* or *list* (optional) The random seed to use for random numbers in the simulation.
 
     * The typical use case is that this is a simple integer value. Then, internally we scramble this value in a deterministic way and use a sequence of seeds based on the scrambled value so that the output is deterministic even when using multiple processes to build each image.
-    * If ``random_seed`` is a string, then it is parsed as a integer and treated in the same way as above.  This allows you to do a simple calculation using the ``$`` Eval shorthand to e.g. compute the initial seed from some other kind of ID number you might have in the config file.  It also means you don't have to be careful in your YAML writing to not accidentally write e.g. ``'1234'`` rather than ``1234``.
-    * If ``random_seed`` is a dict, it should parse as an *int_value*, but in this case, GalSim will not convert it into a sequence based on a single value.  Rather, it will respect your specification and evaluate the seed for each object as you specifiy.  This let's you do advanced things like use a different cadence for your random numbers (e.g. potentially to repeat the same seed for multiple objects; cf. `Demo 7`).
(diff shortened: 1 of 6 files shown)


docs/config_image.rst
```diff
@@ -17,17 +17,7 @@ Some attributes that are allowed for all image types are:
...
-
-    .. warning::
-
-        Within this ``random_seed`` list, any items that are either an int or str will be
-        evaluated once and converted into a sequence in the normal way.  So if you want an item
-        to be some complex Eval operation, you should make it a dict with an explicit
-        ``type: Eval``, rather than use the ``$`` shorthand notation.
-
+    * If ``random_seed`` is a list, then the first one will be converted as described above, but the later ones will not. Rather, it will respect your specification and evaluate the seed for each object as you specifiy. This will create multiple random number generators, according to the multiple seed specifications. This is normally used to have one random number behave normally for noise and such, and another one (or more) repeat with some other cadence (e.g. repeat for each image in an exposure to make sure you generate the same PSFs for multiple CCDs in an exposure). See `Demo 7` and `Demo 13` for examples of this. Whenever you want to use an rng other than the first one, add ``rng_num`` to the field and set it to the number of the rng you want to use in this list.
```
jmeyers314, line 20: Maybe specify that `rng_num` starts at `0`.

galsim/config/util.py
```diff
@@ -392,18 +392,36 @@ def ParseRandomSeed(config, param_name, base, seed_offset):
...
+            if (isinstance(config[param_name], dict)
+                and config[param_name].get('type',None) == 'Sequence'):
+                    # Not really a deprecation, but similar.
+                    # Warn users of the change in behavior.
+                    galsim_warn('You have a custom Sequence as the (first) random_seed. '
+                                'As of GalSim version 2.6, this will be parsed as an integer '
+                                'and converted to a new sequence indexed by obj_num, '
+                                'rather than whatever sequence you have specified. '
+                                'You probably want to put your custom random_seed sequence '
+                                'as the second item in the radom_seed list and use rng_num=1.')
```
jmeyers314, line 408:
```suggestion
                                'as the second item in the random_seed list and use rng_num=1.')
```

jmeyers314 (approved):
LGTM!

Issue #1194, opened 2022-11-17
Add Quantity as a valid value_type in config


rmjarvis:
There are a few places in GalSim where we have unit-ful values as parameters to various things, which are documented to assume a particular unit.  (E.g. our wavelengths are all in nm.)  It would be nice if the user could pass an astropy.Quantity for such things and specify their units without worrying about whether their assumption matches ours.  This is even more true for some imsim values where both m and mm are plausible units for various distances in the telescope.

Proposal:  Let such parameters be specified in a config dict as:
1. A float value as currently with the documented assumed units.
2. An astropy.Quantity with whatever units the user wants (so long as dimensionally compatible ofc).
3. A string that can eval into an astropy.Quantity
4. A "Quantity" type that has two items: `value` and `unit`.  This lets the `value` (a float) be further processed by the normal config processing, if that is desired.

rmandelb:
As I understand it, astropy is already a dependency for GalSim, so this makes sense to me - it'll help some users and not complicate the installation.

Does astropy have convenience routines to check dimensional compatibility?  At a quick glance, it looks like it does (since for example they've defined this error: https://docs.astropy.org/en/stable/api/astropy.units.UnitConversionError.html#astropy.units.UnitConversionError) so we wouldn't have to do any of that checking ourselves.

PR #1311, opened 2024-09-13
Add astropy.units support to config layer.

jmeyers314:
Starting to add astropy Quantity/Unit support to GalSim config layer...  WIP, but could be worth a quick look at what's here so far.


jmeyers314:
I think this is ready now.  I added `Quantity` and `Unit` as config-layer value types.  These can be instantiated as described in #1194 .  I then added support for a number of concrete config classes, including:

* `Airy`, `Kolmogorov`, `VonKarman`, `OpticalPSF` for wavelengths, Fried parameters, telescope diameters.
* Bandpass units
* SED units and flux densities
* DCR conditions (pressure/temperature/H2O_pressure)

I think there are a lot of places where this will particularly benefit ImSim.

cwwalter:
This is great Josh.  

FYI, here is an old imSim issue about this:

https://github.com/LSSTDESC/imSim/issues/48

docs/config_values.rst
```diff
@@ -571,6 +573,49 @@ Options are:
...
+
+These represent `astropy.units.Quantity` values, which are a combination of a float and a unit (specifically, an `astropy.units.Unit`).
+
+Options are:
+
+* An `astropy.units.Quantity` value (e.g. '8.7*u.m', where 'u' is the `astropy.units` module)
+* A string interpretable by `astropy.units.Quantity` (e.g. '8.7 m')
+* A dict with:
+
+    * ``type`` = *str* (required)  There is only one valid option:
```
rmjarvis, line 587: There seem to be two valid options.

docs/config_values.rst
```diff
@@ -571,6 +573,49 @@ Options are:
...
+
+These represent `astropy.units.Unit` values.
+
+Options are:
+
+* An `astropy.units.Unit` value (e.g. 'u.m', where 'u' is the `astropy.units` module)
+* A string interpretable by `astropy.units.Unit` (e.g. 'm')
+* A dict with:
+
+    * ``type`` = *str* (required)  There is only one valid option:
```
rmjarvis, line 609: Here too.

rmjarvis (approved):
Looks great Josh!  Just a couple minor doc errors.

docs/config_values.rst
jmeyers314: Looks like I mimicked the `sky_value` section, which also technically has two valid options.  I'll fix that one too.

PR #1316, opened 2024-10-24
Use builtin methods on lists

arunkannawadi:
I have been spending some time this week hacking GalSim and noticed a few places where builtins are known to be faster than the `numpy` equivalents generally. This is not likely to move the needle in terms of the overall efficiency, but this is something I had to do anyway during the hack and I thought I might as well try to get it merged here.


rmjarvis (approved):
This is fine I guess.  But the length of these lists is typically 2 or 3. So the time difference between any implementation of these operations is basically zero.  It's completely irrelevant to any efficiency concern.

arunkannawadi:
Oh, I don't doubt that one bit. These are some changes that I had to do anyway to get my hacky version running. Btw, I'm not in any hurry and this need not result in a patch release immediately.

Issue #1322, opened 2024-12-07
gso.drawImage shifts the object


zhouconghao:
Hi! 

@beckermr and I were testing some source injection code, and we found that if we call gso.drawImage, the returned image is shifted by 1 pixel when the stamp is even. If we draw into images directly,y it wouldn't be a problem. Here is a minimal reproducible test written by Matt:

```
import numpy as np
import galsim
import pytest

def _compute_cen(img):
    x, y = img.get_pixel_centers()
    flux = img.array
    xcen = np.sum(x * flux) / np.sum(flux)
    ycen = np.sum(y * flux) / np.sum(flux)
    return galsim.PositionD(xcen, ycen)


@pytest.mark.parametrize(
    "stamp_size", [50, 51],
)
@pytest.mark.parametrize(
    "x_center_offset", [-1.0, -0.5, 0.0, 0.5, 1.0],
)
@pytest.mark.parametrize(
    "y_center_offset", [-1.0, -0.5, 0.0, 0.5, 1.0],
)
def test_injection_simple(stamp_size, x_center_offset, y_center_offset):
    cen = stamp_size // 2
    center = galsim.PositionD(
        cen + x_center_offset,
        cen + y_center_offset,
    )
    wcs = galsim.PixelScale(0.2)
    galsim_im = galsim.Gaussian(fwhm=0.9).drawImage(
        nx=stamp_size,
        ny=stamp_size,
        wcs=wcs.local(center),
        center=center,
    )
    centroid = _compute_cen(galsim_im)
    np.testing.assert_allclose(centroid.x, center.x, atol=1e-4, rtol=0)
    np.testing.assert_allclose(centroid.y, center.y, atol=1e-4, rtol=0)
```

Could you take a look at this? Let us know if more info or more tests are needed. Thank you!

beckermr:
cc @rmjarvis @esheldon

beckermr:
Another clue to this bug is that we do NOT see the error if you draw into an image with the bounds set.

rmjarvis:
Yeah, I see where the bug is.  There is an inconsistency in how it sets the bounds in two different places.  Thanks for the report.  I'll work on getting a fix in soon.

zhouconghao:
Thank you for the fix, Mike!

PR #1323, opened 2024-12-07
Fix centering bug reported in #1322

rmjarvis:
cf. #1322
@beckermr @zhouconghao

### DIFF
--- a/galsim/gsobject.py
+++ b/galsim/gsobject.py
@@ -1225,8 +1225,8 @@ class GSObject:
         elif nx is not None and ny is not None:
             b = BoundsI(1,nx,1,ny)
             if center is not None:
-                b = b.shift(_PositionI(np.floor(center.x+0.5)-b.center.x,
-                                       np.floor(center.y+0.5)-b.center.y))
+                b = b.shift(_PositionI(np.floor(center.x+0.5-b.true_center.x),
+                                       np.floor(center.y+0.5-b.true_center.y)))
             return b
         elif bounds is not None and bounds.isDefined():
             return bounds
--- a/galsim/gsobject.py
+++ b/galsim/gsobject.py
@@ -1164,6 +1164,7 @@ class GSObject:
                         "Must set either both or neither of nx, ny", nx=nx, ny=ny)
                 image = Image(nx, ny, dtype=dtype)
                 if center is not None:
+                    # Note: this needs to match the corresponding calculation in _get_new_bounds
                     image.shift(_PositionI(np.floor(center.x+0.5-image.true_center.x),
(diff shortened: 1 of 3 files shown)


galsim/gsobject.py
```diff
@@ -1225,8 +1225,8 @@ def _get_new_bounds(self, image, nx, ny, bounds, center):
         elif nx is not None and ny is not None:
             b = BoundsI(1,nx,1,ny)
             if center is not None:
-                b = b.shift(_PositionI(np.floor(center.x+0.5)-b.center.x,
-                                       np.floor(center.y+0.5)-b.center.y))
+                b = b.shift(_PositionI(np.floor(center.x+0.5-b.true_center.x),
+                                       np.floor(center.y+0.5-b.true_center.y)))
```
rmjarvis, line 1229: Worth mentioning that [here](https://github.com/GalSim-developers/GalSim/blob/releases/2.6/galsim/gsobject.py#L1167) is the other line, which this should be consistent with (and wasn't).

beckermr (approved):
Thanks for the fix!

I have two related questions, none of them should block this PR but might be good fodder for future work. 

1. Is it worth trying to systematically add tests that exercise all of the drawing options in each combination? I was surprised to find this bug and it reminded me that even if the coverage is near 100%, we can still miss bugs because we have not hit all possible paths through the code. 

2. Is it worth trying to do some internal refactoring on some of the private methods to avoid the need to have certain lines of code be kept in sync? For example, I wonder if it is possible to refactor things such that the wcs, bounds, etc is all made up front and then the image is generated from those. This might help different bits of the code be more robust to changes maybe?

In any case, LGTM and thank you!

beckermr (changes requested):
Ack one other thing. Let's add comments to the two relevant code lines that need to be the same stating that and referring to the other line so future folks understand when reading the code.

rmjarvis:
1. We do try.  We apparently missed these combinations that you discovered.  There were a bunch of other tests that test various combinations of offset and center and such.  But I think mostly with given bounds, rather than creating new ones.  And the ones that tested that combination apparently wasn't comprehensive enough.  If you have further ideas of a test that would be more comprehensive, I'd be happy to add it.
2. Possibly.  I thought about refactoring the `_setup_image` to use `new_bounds` rather than recapitulate the same calculation when necessary.  But (a) I didn't want to gratuitously break any code that uses this private method.  At least not in a bugfix release.  And (b) I wasn't positive that there wasn't a good reason that I didn't already do it that way.

beckermr:
Sounds good. Thank you and LGTM!

beckermr (approved):

PR #1325, opened 2024-12-20
Fix #1324 and add more regression tests of maxk, stepk, etc.

### LINKED ISSUE #1324: `Spergel` `stepk` has two powers of the scale radius
It appears that galsim has two factors of `r0` being applied when computing the `stepk` for the `Spergel` profile.

The first power is here:

https://github.com/GalSim-developers/GalSim/blob/120eab3b413aa7bc1447e72b8e3df4e5f4d6ac43/src/SBSpergel.cpp#L101-L102

but then the scale radius is applied again here:

https://github.com/GalSim-developers/GalSim/blob/120eab3b413aa7bc1447e72b8e3df4e5f4d6ac43/galsim/spergel.py#L206-L211

I compared to the implementation in jax-galsim, and I can only get them to agree if I use a second factor of `r0` within jax-galsim like this

```python
    @property
    def _stepk(self):
        R = calculateFluxRadius(1.0 - self.gsparams.folding_threshold, self.nu)
        R *= self._r0**2
        # Go to at least 5*hlr
        R = jnp.maximum(R, self.gsparams.stepk_minimum_hlr * self.half_light_radius)
        return jnp.pi / R
```

Note that the jax-galsim `calculateFluxRadius` function is in normalized units, and so should have only one factor of `r0`.

rmjarvis:
Fix a bug in the Spergel stepk calculation noticed by @beckermr.  See #1324 for details.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
             matrix:
                 # First all python versions in basic linux
                 os: [ ubuntu-latest ]
-                py: [ 3.7, 3.8, 3.9, "3.10", 3.11, 3.12 ]
+                py: [ 3.8, 3.9, "3.10", 3.11, 3.12, 3.13 ]
                 CC: [ gcc ]
                 CXX: [ g++ ]
                 FFTW_DIR: [ "/usr/local/lib" ]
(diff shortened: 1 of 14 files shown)


beckermr (approved):
LGTM and nice additions to the tests!

rmjarvis:
Thanks Matt.  Would you like this in a bugfix release soon?  Or let it simmer a bit until we get another bug report?  (I'm assuming this one isn't particularly mission critical.)

PR #1328, opened 2025-02-18
Add recalc=0 option to SiliconSensor

rmjarvis:
For some work on imsim that @welucas2 is doing, we realized that there is not really a good API for the user to fully control when pixel boundary recalculations happen in the SiliconSensor class.  This PR adds a mechanism to do that more easily.

Specifically, now nrecalc=0 means don't do any recalculations during a call to accumulate.  And a new recalc=True option to accumulate means do a manual update right at the start.

If resume=False then the sensor will be initialized the normal way, but then (aside from the tree rings) work equivalently to the simple sensor (no pixel boundary adjustments).

If resume=True and recalc=True, then the pixel boundaries will be updates at the start of the accumulate call, but then not afterwards.

If resume=True and recalc=False, then it continues with the current image, and no recalculation is done.  This option is for when we have more batches than we want recalculations.

I think these two options should effectively allow complete control over when the recalculations happen.  At least sufficient for what we need in the imsim refactorization.

### DIFF
--- a/galsim/sensor.py
+++ b/galsim/sensor.py
@@ -129,7 +129,7 @@ class SiliconSensor(Sensor):
 
         lsst_e2v_50_8
                     The E2V sensor being used for LSST, using 8 points along each side of the
-                    pixel boundaries. 
+                    pixel boundaries.
 
         lsst_e2v_50_32
                     The E2V sensor being used for LSST, using 32 points along each side of the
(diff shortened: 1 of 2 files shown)


welucas2 (approved):
Thanks, this looks great to me. Tests pass and I can confirm that the PR code matches the imSim performance I saw with my hacky version.

galsim/sensor.py
```diff
@@ -310,6 +313,8 @@ def accumulate(self, photons, image, orig_center=PositionI(0,0), resume=False):
                             accumulation to see what flux is already on the image, which can
                             be more efficient, especially when the number of pixels is large.
                             [default: False]
+            recalc:         Whether to force a recalculation at the pixel boundaries at the
```
rmjarvis, line 316:
```suggestion
            recalc:         Whether to force a recalculation of the pixel boundaries at the
```

PR #1330, opened 2025-03-18
Make PhotonDCR work correctly with both sky_pos and zenith_angle

rmjarvis:
@vfbraga ran into a bug in the dcr angle parsing where both zenith_angle and obj_coord are set.  This can happen if zenith_angle is given explicitly, and there is a sky_pos being set in the config (which is relatively common).

The fix here is just to allow that combination.  So if zenith_angle is given, the dcr parsing function will pop off obj_coord from the rest of the kwargs that get returned.


jmeyers314 (approved):
LGTM!

PR #1337, opened 2025-10-03
Update calculateFWHM to be insensitive to np.argsort algorithmic details 

### LINKED ISSUE #1336: Result of Image.calculateFWHM() depends on NumPy version
`Image.calculateFWHM()` can return different results depending on the version of NumPy when calling `np.argsort()` at line https://github.com/GalSim-developers/GalSim/blob/4c04a1ea4aa4c652c25019c0c7b5881f7571035f/galsim/image.py#L1587

NumPy 2.0 and more recent has an argument `stable` to `argsort()` which guarantees that it will return in order the indices of equal-valued elements of the array. This is by default not enabled, and before 2.0 the option doesn't exist -- in this case, it seems the indices of equal valued elements can be given in any order that seems to be determined by the version of NumPy.

GalSim's FWHM calculation orders the elements by distance from the image center and finds the first element with a surface brightness < half the max. It then interpolates with the brightness of the previous element, which can have been sorted in any order with the other elements the same distance from the center. From tests it does seem that a given NumPy version will always sort the array the same way, but it can and does change when using different versions. Here's a very small reproducer showing different returned values:
```
import galsim
gal = galsim.Gaussian(flux=1.e5, sigma=5.)
rng = galsim.BaseDeviate(1234)
img = gal.drawImage(method='phot', rng=rng, nx=50, ny=50)
print(img.calculateFWHM())
```
With NumPy 1.26.4 this returns a FWHM of 13.870431838154461.
With NumPy 2.0.2 it instead returns 13.828419784303065.
With NumPy 2.1.0 it returns 13.870431838154461 once more.

This is likely behind an [odd failing test in imSim](https://github.com/LSSTDESC/imSim/pull/493).

rmjarvis:
@welucas2 reported on #1336 that our `calculateFWHM` calculation could return different values for different numpy versions.  Specifically with different argsort algorithmic details.

The problem occurs when the profile is centered directly on a pixel center (or corner or some other special locations) such that there are repeated rsq values.  Argsort will report one of these to be the "first" one to be farther than the half max value, but different numpy versions could report different pixels to be this location.

If the profile is symmetric and drawn directly or with fft, then this still wasn't a problem, since all data at that rsq would have the same value, so it didn't matter which one was the "first" one.  But with photon shooting or asymmetric profiles, that's not the case.

This PR updates our FWHM algorithm to compute the mean data value of all pixels with the desired rsq if there are more than one.  This should be insensitive to the details of how argsort does its thing, and also probably produce a slightly more robust answer for photon shooting, since it will average over the noise somewhat.  When the profile is slightly off-center, this will produce the same result as GalSIm has calculated in the past.

### DIFF
--- a/galsim/image.py
+++ b/galsim/image.py
@@ -1590,13 +1590,18 @@ class Image:
 
         # Find the first value with I < 0.5 * Imax
         k = np.argmax(data < 0.5 * Imax)
-        Ik = data[k] / Imax
+        # If there are duplicate rsq values, argmax won't be deterministic across numpy versions.
+        # To achieve consistent results, we take the average of all data values with the same rsq.
+        k2 = np.searchsorted(rsqf, rsqf[k], side='right')
+        Ik = np.mean(data[k:k2]) / Imax
 
         # Interpolate (linearly) between this and the previous value.
         if k == 0:
             rsqhm = rsqf[0] * (0.5 / Ik)
         else:
-            Ikm1 = data[k-1] / Imax
+            # Similarly, use the mean data for all rsq equal to rsqf[k-1].
+            k1 = np.searchsorted(rsqf, rsqf[k-1], side='left')
+            Ikm1 = np.mean(data[k1:k]) / Imax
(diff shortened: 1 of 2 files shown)


welucas2 (approved):
All good!

Issue #1333, opened 2025-06-28
Curious Out Of Memory during looping on Cosmos Catalogue


jecampagne:
Hello,

I was trying to extract the galaxy&psf couples from the COSMOS_23.5_training_sample catalog
to make some datasets

Please find below the code.

The problem I have is that after having processed 28_000 galaxies (and not yet 29_000) the 
python code crashes due to "Out of Memory" detected by the my local SLURM system 

Is there a "purge" that I have missed somewhere?

```python
    real_galaxy_catalog = galsim.RealGalaxyCatalog(
            dir=path_catalog, sample=sample
        )
    n_total = real_galaxy_catalog.nobjects  # - 56062
    sequence = np.arange(0, n_total) 
    np.random.shuffle(sequence)

    # Loop....
    t0 = time.time()
    for i,gal_idx in enumerate(sequence):
        if i==0 or i%1000==0:
            print(f"Process {i}/{gal_idx} at t:{time.time()-t0:.1f}sec")
        # read HST galaxy and PSF
        gal_ori = galsim.RealGalaxy(real_galaxy_catalog, index=gal_idx)
        psf_hst = real_galaxy_catalog.getPSF(gal_idx)

        # get the corresponding galsim.Image
        gal_ori_image = real_galaxy_catalog.getGalImage(gal_idx)
        psf_ori_image = real_galaxy_catalog.getPSFImage(gal_idx)

        #save into fits files
        fname =  out_catalog_dir +"/"+"gal_"+str(i)+".fits"
        gal_ori_image.write(fname)
        
        fname =  out_catalog_dir +"/"+"psf_"+str(i)+".fits"
        psf_ori_image.write(fname)

```

Thanks for your attention.
Jean_Eric

rmjarvis:
Better documentation about this added in #1340.

PR #1340, opened 2025-12-08
Add documentation to RealGalaxyCatalog.close()

rmjarvis:
This mostly just adds a bit of documentation about `RealGalazyCatalog.close`, which hadn't been documented.  I suspect having it documented would have helped @jecampagne with the problem he reported in #1333.

I also changed the implementation very slightly to also clear out the cached noise images, even though they are usually negligible.  If people want to release memory manually, I figure we might as well release all of the cached stuff.

### DIFF
--- a/galsim/real.py
+++ b/galsim/real.py
@@ -508,6 +508,14 @@ class RealGalaxyCatalog:
        GalSim knows the location of the installation share directory, so it will automatically
        look for it there.
 
+    .. note::
+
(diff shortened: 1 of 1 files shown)


rmandelb (approved):
LGTM.  Optional suggestion: I think it could be helpful to have a one-line note in the `RealGalaxyCatalog` class documentation noting that if memory management is an issue with the various caching steps then the `close` function could be helpful.

PR #1341, opened 2025-12-09
#1332 Switch large FFTs to emitting warnings rather than raising exceptions

### LINKED ISSUE #1332: Change FFT memory Error to Warning?
Currently, GalSim will throw an error if the requested FFT is too large (cf. https://github.com/GalSim-developers/GalSim/blob/4c04a1ea4aa4c652c25019c0c7b5881f7571035f/galsim/gsobject.py#L1970-L1971).

The default value for this is quite low (something like 2GB) and this often results in many jobs crashing and hanging. Every time I configure this to be larger, I still end up finding some object that needs more memory resulting in another crash. It would be helpful for this to be changed into a warning rather than an error.

rmjarvis:
@sidneymau made a good suggestion in #1332 to change our large FFT behavior from being errors to just emitting warnings.

A long time ago, it was rare for machines to have more than 2GB of memory, and it was relatively easy for users to accidentally try to do things that needed FFTs much larger than this.  So rather than crash their computers (and on some machines crashing due to excess memory can be quite annoying), we raised an error before that and told them exactly how much memory they were implicitly asking GalSim to allocate.  So they could override that if they really knew what they were doing, but probably redesigned their code instead.

Anyway, I suspect it is now not as common for people to make this mistake by accident. And many people need to make large FFTs very much on purpose on machines that can handle it. So this guard has probably outlived its usefulness.

This PR changes the behavior to emit a warning instead of an exception, so the code will go on, and if someone's machine crashes from using too much memory, the warning will potentially explain why.  But it won't require people to anticipate how large their FFTs will be when memory isn't really an issue.

### DIFF
--- a/galsim/errors.py
+++ b/galsim/errors.py
@@ -425,7 +425,7 @@ class GalSimFFTSizeWarning(GalSimWarning):
         self.mem = size * size * 24. / 1024**3
         message += "\nThe required FFT size would be {0} x {0}, which requires ".format(size)
         message += "{0:.2f} GB of memory.\n".format(self.mem)
-        message += "If you can handle the large FFT and want to suppress this warning, "
+        message += "If you can handle the large FFT and want to suppress this warning,\n"
         message += "you may update gsparams.maximum_fft_size."
         super(GalSimFFTSizeWarning, self).__init__(message)
     def __repr__(self):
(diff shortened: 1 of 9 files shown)


sidneymau (approved):
Everything looks good to me!

PR #1287, opened 2024-05-01
ENH add critical logger method to logging wrapper

beckermr:
This method is missing and I happened to need it. Adding it here.

PR #1345, opened 2026-02-15
Update all logging to use native Python loggers

sidneymau:
Alright, this is a really big one (albeit more "wide" than "deep").

The core updates are in `galsim.config.util.py`, where the `log_listener` and `log_queue` are introduced to handle the multiprocessed logging.

In the past, the `LoggerWrapper` has presented some challenges to performance in multiprocessing environments (#1213 ); while this was addressed with a simple fix, there is still going to be some inherent cost associated with the `LoggerWrapper` serializing and deserializing each message through the multiprocessing queue.

Beyond the performance aspect, these logging modifications simplify many APIs (and, unfortunately, introduce many breaking changes for developers of GalSim modules) and bring GalSim logging more in-line with idiomatic Python logging.
This will make it much easier to add logging to GalSim library functions and to have those logs propagate through the main script.

The biggest downside here are the changes to the APIs -- in summary, any instance of `logger` being used as a function argument have been removed. This does not impact most casual users but would necessitate changes by any module developers. `imSim`, for example, would need a comparable change; I am happy to follow-up with task. In any case, I based this PR against `main` as I don't think it makes sense to propose such a big change to a release branch right now.

---

Notes:

- When multiprocessing, it appears that only the main process is logging when stamps finish, making it hard to track what's happening in each child process. Why is this? In `galsim.config.util.py`, the `MultiProcess` function passes a `done_func`, which is run at the completion of each process. At present, the logging messages are emitted by these `done_func`s, which are therefore run on the parent process. The process index is passed to the `done_func`, allowing us to still track which process resulted in which message. With the new logging implementation, it would be easier to include logging within the _worker_ processes, which can also emit messages formatted with the process ID.
- When writing logs to a _file_, the GalSim script will open the log file in `w` mode (not the default `a` mode); this was done to pass `test_main.py::test_logger`, but it may be prudent to change the behavior and the test.

---

To compare the emitted logs, I ran demo9 on both the `main` and `logging` branches:
```
galsim demo9.yaml  --log_format '%(process)d - %(name)s - %(message)s'
```
The logs are attached ([main.log](https://github.com/user-attachments/files/25320504/main.log) and 
[logging.log](https://github.com/user-attachments/files/25320505/logging.log)).
To compare,

- The new logging interface can report which _module_ is emitting the messages
- The new logging interface correctly identifies from which _process_ each message is being emitted

---

To test the potential impact of this PR on performance, I ran Matt's [test](https://github.com/beckermr/misc/tree/main/work/galsim_par_test) on a CPU node on Perlmutter.
To create the environment, I ran
```
mamba create -n galsim python galsim
mamba remove --force galsim
pip install -e .  # in GalSim (main | logging)
```

`main` (`aab609d04`)
```
timing 128 processes...
128 7.334399223327637
timing 128 processes...
128 7.499938011169434
timing 64 processes...
64 7.893015384674072
timing 32 processes...
32 10.4623703956604
timing 16 processes...
16 15.256845712661743
timing 8 processes...
8 23.149372100830078
timing 4 processes...
4 41.14827919006348
timing 2 processes...
2 76.40376877784729
```

`logging` (`8c75e6f7b`)
```
timing 128 processes...
128 6.943336009979248
timing 128 processes...
128 6.935553789138794
timing 64 processes...
64 7.929414510726929
timing 32 processes...
32 10.205114126205444
timing 16 processes...
16 14.73991870880127
timing 8 processes...
8 23.33072829246521
timing 4 processes...
4 41.20093846321106
timing 2 processes...
2 75.99361968040466
```
There is no significant difference introduced with the updated logging according to this test.

### DIFF
--- a/galsim/config/util.py
+++ b/galsim/config/util.py
@@ -699,10 +699,6 @@ def MultiProcess(nproc, config, job_func, tasks, item, timeout=900,
         Process = ctx.Process
         Queue = ctx.Queue
 
-        log_queue = Queue()
-        listener = ctx.Process(target=log_listener, args=(log_queue,))
-        listener.start()
-
         # Temporarily mark that we are multiprocessing, so we know not to start another
         # round of multiprocessing later.
         config['current_nproc'] = nproc
@@ -713,6 +709,11 @@ def MultiProcess(nproc, config, job_func, tasks, item, timeout=900,
         # We need to set the number of OpenMP threads to 1 during multiprocessing, otherwise
         # it may spawn e.g. 64 threads in each of 64 processes, which tends to be bad.
         with single_threaded():
+
+            # Initialize the log queue and start the listener for logging from multiple processes.
+            log_queue = Queue()
+            listener = ctx.Process(target=log_listener, args=(log_queue,))
+            listener.start()
 
             # Send the tasks to the task_queue.
             ntasks = len(tasks)
@@ -777,6 +778,7 @@ def MultiProcess(nproc, config, job_func, tasks, item, timeout=900,
                         if except_abort:
                             for j in range(nproc):
                                 p_list[j].terminate()
+                            listener.terminate()
                             raise_error = res
                             break
                     else:
@@ -803,6 +805,13 @@ def MultiProcess(nproc, config, job_func, tasks, item, timeout=900,
                 # And terminate any jobs that might still be running.
                 for j in range(nproc):
                     p_list[j].terminate()
+                # Same for logging
+                while not log_queue.empty():
+                    try:
+                        log_queue.get_nowait()
+                    except Exception:  # pragma: no cover
+                        pass
+                    listener.terminate()
                 raise_error = e
 
             finally:
@@ -823,9 +832,13 @@ def MultiProcess(nproc, config, job_func, tasks, item, timeout=900,
                     p_list[j].join()
                 task_queue.close()
 
-                # Also make sure to stop the log listener with `STOP`
-                log_queue.put('STOP')
+                # Also make sure to stop the log listener with `STOP` and do cleanup.
+                try:
+                    log_queue.put_nowait('STOP')
+                except Exception:  # pragma: no cover
+                    pass
                 listener.join()
+                log_queue.close()
 
         del config['current_nproc']
 


sidneymau:
Looks like there's still some weirdness going on with the `fork`ing causing a process to hang forever that needs to be sorted out... Having a hard time reproducing this locally, though :thinking:

sidneymau:
> This is far too backwards incompatible as it currently stands. At the very least, you would need to deprecate the old syntax to let users know that they need to make changes to their code and how to do so before completely breaking everything. This is a pretty huge deal for something that doesn't seem to have that much benefit. Although maybe I'm not fully appreciating what the benefit of all this is.

Yeah, I definitely agree that this PR is not far too much as it is, I meant to mark it as a draft to test out the updated logging. The main benefit, as I see it, would be to make it easier to keep track of which processes are emitting logs (right now there's a hack in GalSim that passes relative process IDs around, but those logs are ultimately emitted by the main process _after_ the process has finished).
I think there's also a benefit to keeping to native logger objects to reduce the need to maintain the wrapper (e.g., #1287) to maintain consistency with what users expect.

> I'm also not particularly fond of APIs that rely on global variables. Especially global variables that live in another module. They can have their place for sure, and maybe this is a case where it does work better than a more functional style. But I'd like to have more evidence of its benefits before accepting it here.

I think my main goal looking into this was more to follow how Python intends for logging to be done between multiple modules -- I don't necessarily disagree about the global/local logging point you brought up, but it is the case that extra steps need to be taken to integrate GalSim logging with downstream code that is using GalSim as a library.
I also do wonder if there's still some extra overhead with the `LoggerWrapper` (the test I ran from Matt didn't indicate a significant difference, but I was planning to rerun with more verbose logging enabled to check this)

> In any case, I suspect a slightly different implementation would be feasible that I think would give you the new functionality you want without breaking old code and without forcing people to use the global variable API. That would be to allow the logger parameter to be a string in addition to a logger object. Then the first line in functions that use a logger could be
> 
> _logger = logging.getLogger(logger) if isinstance(logger, str) else logger
> 
> Then the lines that actually use the logger currently would use the underscore version.
> 
> The initial call to any config function taking a logger would use __name__ for the logger, rather than making one as is normally done now.
> 
> This would have the additional advantage that you wouldn't need to figure out the problem with fork you are currently running into. And we could more straightforwardly try both systems in the wild in various contexts to see if there are any tradeoffs with the new one vs the old.

Thanks for the suggestion, I'll give it a try when I can. I had another idea to try that might also help support both modes (though I really wouldn't expect this to be out of draft state for a while)

PR #1355, opened 2026-03-27
Add bandpass option to SED.thin

rmjarvis:
For some Piff stuff, I wanted to be able to thin an SED targeting a particular bandpass, rather than generically.  Especially to have the output limits truncated to the bandpass limits, but also targeting the integral for thinning to be specifically the one you get with this bandpass.

This PR adds that functionality.


arunkannawadi (approved):
LGTM! I think this is the case, but can confirm that trying to thin an already thinned SED (for the given bandpass) is trivial the second time?

rmjarvis:
Thanks Arun!

PR #1356, opened 2026-04-29
fix: bug in maxk algorithm for interpolated images

beckermr:
This PR fixes a small algorithmic bug in computing maxk for interpolated images. 

I verified that the test failed without the bug fix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -201,12 +201,11 @@ jobs:
 
             - name: Run unit tests
               run: |
-                # cd tests
-                pytest -vvsx -k test_interpolatedimage_maxk_python
-                # coverage run -m pytest -v
-                # pytest -v run_examples.py
-                # cd ..  # N.B. This seems to happen automatically if omitted.
-                #        # Less confusing to include it explicitly.
+                cd tests
+                coverage run -m pytest -v
+                pytest -v run_examples.py
+                cd ..  # N.B. This seems to happen automatically if omitted.
+                       # Less confusing to include it explicitly.
 
             - name: Check shared library
               if: matrix.py == 3.12
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -201,11 +201,12 @@ jobs:
 
             - name: Run unit tests
               run: |
-                cd tests
-                coverage run -m pytest -v
-                pytest -v run_examples.py
-                cd ..  # N.B. This seems to happen automatically if omitted.
-                       # Less confusing to include it explicitly.
+                # cd tests
+                pytest -vvsx -k test_interpolatedimage_maxk_python
+                # coverage run -m pytest -v
+                # pytest -v run_examples.py
+                # cd ..  # N.B. This seems to happen automatically if omitted.
+                #        # Less confusing to include it explicitly.
 
             - name: Check shared library
               if: matrix.py == 3.12
--- a/tests/test_interpolatedimage.py
+++ b/tests/test_interpolatedimage.py
@@ -1890,98 +1890,55 @@ def test_drawreal_seg_fault():
     np.testing.assert_array_equal(image.array, 0)
 
 
-def _find_maxk(kim, thresh, count_thresh=True):
-    No2 = kim.xmax
-    dk = np.pi / No2
-
-    thresh *= thresh
-    max_ix = No2
-    n_below_thresh = 0
-    maxk_ix = 0
-
-    ix = 0
-    while ix <= max_ix:
-
-        iy = 0
-        while iy <= ix:
-            # The bottom side of the square in the lower-right quadrant.
-            norm_kval = kim(iy, -ix)
-            norm_kval = (norm_kval * norm_kval.conjugate()).real
-
-            if norm_kval <= thresh and iy != ix and ix != No2:
-                # The top side of the square in the upper-right quadrant.
-                norm_kval = kim(iy, ix)
-                norm_kval = (norm_kval * norm_kval.conjugate()).real
-
-            if norm_kval <= thresh and iy > 0:
-                # The right side of the square in the lower-right quadrant.
-                norm_kval = kim(ix, -iy)
-                norm_kval = (norm_kval * norm_kval.conjugate()).real
-
-            if norm_kval <= thresh and ix > 0 and iy != No2:
-                # The right side of the square in the upper-right quadrant.
-                norm_kval = kim(ix, iy)
-                norm_kval = (norm_kval * norm_kval.conjugate()).real
-
-            if norm_kval > thresh:
-                maxk_ix = ix
-                n_below_thresh = 0
-                break
-
-            iy += 1
-
-        if count_thresh:
-            if norm_kval <= thresh:
-                n_below_thresh += 1
-        else:
-            n_below_thresh += 1
-        if n_below_thresh == 5:
-            break
-
-        ix += 1
-
-    maxk_ix += 1
-    return maxk_ix * dk
-
-
 @timer
 def test_interpolatedimage_maxk_python():
     # this code makes an image where there is a gap in the fourier
-    # space image of 4 pixels where pixels go above and below the
-    # maxk threshold. Four pixels is exactly the gap needed to trigger
-    # the bug in the maxk code we are testing for.
+    # space image of a certain number pixels where pixels go above
+    # and below the maxk threshold. At five pixels, galsim should
+    # ignore the gap, but less than that it should increase maxk.
+
+    def _compute_maxk_cpp(xim, iim):
+        # this little function exists only to invoke the C++
+        # maxk code...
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -201,11 +201,12 @@ jobs:
...
-                coverage run -m pytest -v
-                pytest -v run_examples.py
-                cd ..  # N.B. This seems to happen automatically if omitted.
-                       # Less confusing to include it explicitly.
+                # cd tests
+                pytest -vvsx -k test_interpolatedimage_maxk_python
+                # coverage run -m pytest -v
+                # pytest -v run_examples.py
+                # cd ..  # N.B. This seems to happen automatically if omitted.
+                #        # Less confusing to include it explicitly.
```
beckermr:
```suggestion
                cd tests
                coverage run -m pytest -v
                pytest -v run_examples.py
                cd ..  # N.B. This seems to happen automatically if omitted.
                       # Less confusing to include it explicitly.
```

tests/test_interpolatedimage.py
```diff
@@ -1890,5 +1890,58 @@ def test_drawreal_seg_fault():
     np.testing.assert_array_equal(image.array, 0)
 
 
+@timer
+def test_interpolatedimage_maxk_python():
```
beckermr:
```suggestion
def test_interpolatedimage_maxk_kspace_pixel_gap():
```

tests/test_interpolatedimage.py
```diff
@@ -1890,5 +1890,58 @@ def test_drawreal_seg_fault():
     np.testing.assert_array_equal(image.array, 0)
 
 
+@timer
+def test_interpolatedimage_maxk_kspace_pixel_gap():
+    # this code makes an image where there is a gap in the fourier
+    # space image of a certain number pixels where pixels go above
+    # and below the maxk threshold. At five pixels, galsim should
```
beckermr:
```suggestion
    # and below the maxk threshold. At >five pixels, galsim should
```

beckermr:
@rmjarvis This one is ready for review! Getting the tests right was a pain since I don't have a great understand of how to call into the C++ layer properly. Hopefully the comments help explain what I am trying to do. I tried a bunch of versions where I remade new python-level interpolated images, tried comparing maxk for different image rotations/flips/etc. and none of those worked. The test here is a direct test of the details of the algorithm.

tests/test_interpolatedimage.py
```diff
@@ -1890,5 +1890,58 @@ def test_drawreal_seg_fault():
...
+            iim.gsparams._gsp,
+        )
+        sbii.calculateMaxK(0)  # this call is needed to invoke the C++ code
+        return sbii.maxK()
+
+    print(" ")
+    for offset in [3, 4, 5, 6, 7]:
+        im = galsim.Gaussian(fwhm=0.9 / 0.2).drawImage(scale=1)
+        iim = galsim.InterpolatedImage(im, scale=1)
+        orig_maxk = _compute_maxk_cpp(iim._xim, iim)
```
rmjarvis: Why not just `iim.maxk`?  I would have thought that would call the C++ code for you.

beckermr: That works for the initial maxk but doesn't come out right for the manipulated image. When I feed the new image back in, the maxk changes, but some how the fourier grid is different and so things don't match as I would expect.

beckermr: Here is what you get if you try and feed the new image into a new interpolated image object compared to the test I coded up:

| offset | orig | new via direct C++ | new via galsim II |
| ------ | ---- | ------------------ | ----------------- |
|    3 |   1.963495 |   2.159845 |   3.141593 |
|    4 |   1.963495 |   2.225295 |   3.141593 |
|    5 |   1.963495 |   2.290745 |   3.141593 |
|    6 |   1.963495 |   1.963495 |   3.141593 |
|    7 |   1.963495 |   1.963495 |   3.141593 |

The columns are:

- offset: the pixel offset
- orig: the original maxk value (same no matter how it is computed)
- new via direct C++: new maxk value computed via direct call into the C++ layer
- new via galsim II: new maxk value by feeding the new image back into an interpolated image

You can see it always returns `pi` which is really odd. :/

beckermr: I've pushed an update to the test suite that makes this table.

beckermr: I've also pushed some code that shows how the FFTs of the images relate given galsim's storage conventions.

tests/test_interpolatedimage.py
```diff
@@ -1890,5 +1890,73 @@ def test_drawreal_seg_fault():
...
+        new_im = new_im[iim._image.bounds]
+
+        if offset == 0:
+            np.testing.assert_allclose(im.array, new_im.array, atol=1e-6, rtol=1e-6)
+
+        print("| % 6d | %10.6f | %18.6f | %17.6f |" % (
+            offset,
+            orig_maxk,
+            new_maxk,
+            galsim.InterpolatedImage(new_im, scale=1).maxk),
```
rmjarvis: If this uses the full new_im, not the clipped one (i.e. not `new_im = new_im[iim._image.bounds]`) and has `pad_factor=1`, then it matches your code.  If you don't mind, I'd rather use that, rather than a bunch of non-user facing code for this test.

beckermr: Ah great! I'll try that. I tried a bunch of ways of doing this and didn't know the right incantation.

rmjarvis: Not sure if you looked at what the modified xim is when you add a single hot pixel in the outskirts of the kimage, but it's pretty wonky.  So trimming to the original bounds is a pretty massive change.  The resulting round trip is not close to what you wanted.  So you need to keep the full new_im.  pad_factor=1 then keeps it to that size, rather than padding it further, which would again change the effective image from what you are testing.

beckermr:
This is ready for another look @rmjarvis!

tests/test_interpolatedimage.py
```diff
@@ -1890,5 +1890,49 @@ def test_drawreal_seg_fault():
...
+def test_interpolatedimage_maxk_kspace_pixel_gap():
+    # this code makes an image where there is a gap in the fourier
+    # space image of a certain number pixels where pixels go above
+    # and below the maxk threshold. At >five pixels, galsim should
+    # ignore the gap, but less than that it should increase maxk.
+
+    print("\n| offset | orig       | new                |")
+    print("|--------|------------|--------------------|")
+    for offset in [0, 3, 4, 5, 6, 7]:
+        im = galsim.Gaussian(fwhm=0.9 / 0.2).drawImage(scale=1)
```
rmjarvis:
```suggestion
        im = galsim.Gaussian(fwhm=4.5).drawImage(scale=1)
```

tests/test_interpolatedimage.py
```diff
@@ -1890,5 +1890,49 @@ def test_drawreal_seg_fault():
...
+        kx, ky = kim.get_pixel_centers()
+        kx *= kim.scale
+        ky *= kim.scale
+        # this is the last pixel above threshold. galsim adds 1
+        # to the last pixel it finds above threshold to compute orig_maxk
+        # and so we subtract 1
+        maxk_ix = np.floor(orig_maxk / kim.scale).astype(int) - 1
+        if offset > 0:
+            val = kim[maxk_ix + offset, maxk_ix]
+            kim[maxk_ix + offset, maxk_ix] = val / np.abs(val) * kim[0, 0].real
```
rmjarvis: This is fine.  But why preserve the phase here?  Your previous version just set it to kim[0,0].  (Which should always be real, so the `.real` is superfluous here.)

rmjarvis (approved):

tests/test_interpolatedimage.py
beckermr: I was hacking trying to figure out why I could not call interpolate images directly for maxk. You knew how to fix that so we can remove the phase thing. Will do!

tests/test_interpolatedimage.py
```diff
@@ -1890,5 +1890,49 @@ def test_drawreal_seg_fault():
...
+        new_im = kim.calculate_inverse_fft()
+        new_maxk = galsim.InterpolatedImage(new_im, scale=1, pad_factor=1).maxk
```
beckermr:
```suggestion
            kim[maxk_ix + offset, maxk_ix] = kim[0, 0].real
        new_im = kim.calculate_inverse_fft()
        new_maxk = galsim.InterpolatedImage(new_im, scale=1, pad_factor=1).maxk
```

beckermr:
Ok! Assuming the tests still pass this is ready for hopefully one last look @rmjarvis!  I can't merge here so you'll need to merge if it looks good!

rmjarvis:
Thanks Matt!

beckermr:
Thank you!

Issue #1357, opened 2026-05-05
describe_zernike(11) is wrong


jmeyers314:
Prints 

`sqrt(5) * ( - 6y^2 + 6y^4 - 6x^2 + 12x^2y^2 + 6x^4)` 

instead of 

`sqrt(5) * (1 - 6y^2 + 6y^4 - 6x^2 + 12x^2y^2 + 6x^4)`.

jmeyers314:
See #1358 .

PR #1358, opened 2026-05-05
#1357: fix describe_zernike(11).

jmeyers314:
Needed a check for a bare `1` term instead of just a bare `1` polynomial.


rmjarvis (approved):
lgtm

PR #1252, opened 2023-10-15
TST fix tests for JAX-Galsim

beckermr:
This PR has two main features:

1. It has a generalization of the concept of a "galsim backend" with associated functions for the test suite only. In the test suite only, the backend is used to adjust the tests as needed for the jax-based galsim implementation.

2. It has adjustments of the tests for jax-galsim. Most of them are related to cases where jax cannot raise the same errors (e.g., for checking argument types) or when the fact that jax arrays cannot be changed in place causes the APIs to differ.

PR #1360, opened 2026-05-13
test: fixes for precision and fragile tests from jax-galsim PR

beckermr:
This PR has suggested changes to the main galsim test suite from the JAX-GalSim changes PR (https://github.com/GalSim-developers/GalSim/pull/1252) that @rmjarvis reviewed for us a while back.

### DIFF
--- a/tests/test_draw.py
+++ b/tests/test_draw.py
@@ -1097,22 +1097,22 @@ def test_shoot():
     # in exact arithmetic.  We had an assert there which blew up in a not very nice way.
     obj = galsim.Gaussian(sigma=0.2398318) + 0.1*galsim.Gaussian(sigma=0.47966352)
     obj = obj.withFlux(100001)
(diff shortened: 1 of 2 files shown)


rmjarvis:
You'll need to remind me why these changes are necessary.  They don't look particularly "fragile" to me.  Why did we think these changes are appropriate?

beckermr:
These changes came from your suggestions on the JAX-GalSim tests suite PR. Basically, we found that when things were implemented in JAX, the tests had to be changed. This was due to them demanding more precision than needed, sensitivity to the RNG seed, etc.

rmjarvis:
I thought we were moving those test changes to a different repo though...  Am I misremembering the plan here?

beckermr:
That was our plan, but the original comments you made during the code review indicated to me that these changes were needed in galsim as well.

PR #1364, opened 2026-09-15
Keep signed photon weights in Sum._shoot (#1352)

shaneraphel:
Fixes https://github.com/GalSim-developers/GalSim/issues/1352.

`Sum._shoot` assigns each photon to a component, calls that component's `_shoot`, then overwrites `temp.flux` with the unsigned weight `(positive_flux + negative_flux) / n`. `ChromaticSum._shoot` already copies the component array without that assignment.

That overwrite is an identity for profiles with `negative_flux = 0` (Gaussian, Moffat). `OpticalPSF` is an `InterpolatedImage` with ringing, so `positive_flux + negative_flux` is larger than `flux` (here 1.584 on GalSim 2.8.5, Roman J129). Used as a `photon_ops` PSF, the unsigned overwrite inflates the drawn flux.

Labeled check on GalSim 2.8.5, Roman J129, `n_photons=1000`, `nx=51`, `poisson_flux=False`, seed 42, `Add([OpticalPSF, OpticalPSF])/2`:

| path | image sum / true flux |
|---|---|
| achromatic `Sum` (current) | 1.565 |
| same this change with the overwrite removed | 0.983 |
| `ChromaticSum` of `ChromaticOpticalPSF` | 1.057 |

A Gaussian `Sum` drawn as the object (`method='phot'`) stays flux-faithful with or without the line (0.955 vs 0.956 on a 33² stamp). Existing Gaussian `do_shoot` coverage should keep passing.

This is a software routing identity with `ChromaticSum`, not a claim about telescope optics.

AI: a compiled language model drafted the one-line this change (0 gradient-descent training steps). I reviewed the source sibling and the flux lattice before opening this PR.


rmjarvis:
Two issues need addressing:

First, just removing that line breaks other use cases. One of the existing tests failed when removing it. I suspect the right thing is to set it to either +/- this value depending on the sign of the input, but it's been a while since I tracked through all the ways the photon fluxes are set, so I'm not 100% sure of that.

Second, you need a new test showcasing what was wrong with the current code. Something that fails without the code change but passes when you change the code.

